PR #1, opened 2022-08-30
Added scripts and updated the README

YohannParis:
* Added some scripts to start the whole webapp with one script.
* Updated the README to match.
* Added `dist` into the `.gitignore`


.yarnrc.yml
```diff
@@ -1,3 +1,9 @@
-yarnPath: ".yarn/releases/yarn-berry.cjs"
-nodeLinker: "node-modules"
 enableGlobalCache: true
+
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
+    spec: "@yarnpkg/plugin-workspace-tools"
```
mwdchang, line 7: What does the plugin do? Is it to support the `yarn workspace foreach` directive? 

Can you add a comment/URL-link back tot he plugin documentation?

package.json
```diff
@@ -4,5 +4,9 @@
   "private": true,
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "scripts": {
+    "build": "yarn workspaces foreach --parallel --interlaced --verbose --topological run build",
```
mwdchang, line 9: So this relates to a somewhat related thread I started with @bencodrington, do we actually want to build (e.g. transcompile) the serve code? Or are we happy running the server through ts-node/ts-node-esm executables? 

Any thoughts?

README.md
```diff
@@ -17,9 +17,11 @@ yarn install
 Start local dev server for Vue client, with Hot Module Replacement.
 ```
 # Start client
-yarn workspace client run dev
-
+yarn workspace client dev
```
bencodrington, line 20:
```suggestion
yarn workspace client run dev
```

Need that `run` in there.

README.md
```diff
@@ -17,9 +17,11 @@ yarn install
...
 
 # Start server
-yarn workspace server run serve
+yarn workspace server serve
```
bencodrington, line 23:
```suggestion
yarn workspace server run serve
```

Same here.

README.md
```diff
@@ -17,9 +17,11 @@ yarn install
...
+
+# Start client and server
+yarn dev
```
bencodrington, line 26:
```suggestion
yarn run dev
```

package.json
YohannParis: For now I just added main scripts to the project, I think their names should match into which environment we want to run.

README.md
bencodrington: Oh you're totally right, I didn't realize that. Feel free to take it out again if you prefer.

package.json
bencodrington: If that's the case we'll need to do some further configuration. Currently `tsc` fails in the `server` package because it's not able to follow the link to the `shared` package.

mwdchang: So currently the trans-compilation is not working. The `shared` module is not packaged correctly server-side.

mwdchang (approved):

PR #2, opened 2022-08-30
Docker build

mwdchang:
### Summary
Added koa-static for serving client code, added basic Dockerfile, added dockerignore.

Mostly behind the scene for now, so no updates to README.

### Testing
- `docker build . -t terarium-dev`
- `docker run -p 3000:3000 terarium-dev`

Application will be available at http://localhost:3000/, API endpoints available at http://localhost:3000/api

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+**/node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 .pnp.*
 
 node_modules
+dist
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,19 @@
 FROM node:18.8.0-slim
 
 # Install git and source code
-RUN apt-get update
-RUN apt-get -y install git
-RUN git clone https://github.com/DARPA-ASKEM/TERArium.git
+# RUN apt-get update
+# RUN apt-get -y install git
+# RUN git clone https://github.com/DARPA-ASKEM/TERArium.git
+
+COPY . TERArium
 
 WORKDIR TERArium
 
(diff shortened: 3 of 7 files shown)


bencodrington:
I got this error:
```
$ docker built -t . terarium-dev
unknown shorthand flag: 't' in -t
```
Could it have something to do with the docker version? What do you think?

mwdchang:
> docker built -t . terarium-dev

should be `docker build -t . terarium-dev`

YohannParis (commented):
I cannot do the first part, here the message I got:
```
$ git status
On branch docker-build
Your branch is up to date with 'origin/docker-build'.

nothing to commit, working tree clean
$ docker build -t . terarium-dev
invalid argument "." for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
```

bencodrington:
> > docker built -t . terarium-dev
> 
> should be `docker build -t . terarium-dev`

Looks like there have been several layers of typos 😆 including in the original PR description.

The final command that worked for me was:

`docker build . -t terarium-dev`

Note the `.` is before the `-t`


But it did work! Thanks.

YohannParis (approved):
Works great now that the command is fixed.
It will be nice if `docker run -p 3000:3000 terarium-dev` was not blocking the terminal session, and or that `ctrl+c` worked to stop it.

mwdchang:
Oops, sorry about the bad command instructions.

@YohannParis you can use the detached `-d` flag to put the docker process into the background. As for using "ctrl+c" that is a little tricky as the SIGTERM may not be forwarded to the app at all. There are several solutions we can try.

PR #6, opened 2022-09-08
feat: add router

mwdchang:
### Summary
- Adding router setup
- Remove template/boiler-plates from starter kit

## Test
- yarn 
- yarn dev
- http://localhost:5173/ should work with a placeholder message


packages/client/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/vite.svg" />
```
YohannParis, line 5: We can remove the icon as well.
Don't we have an Uncharted one while waiting to make a TERArium icon?

packages/client/src/views/Home.vue
```diff
@@ -0,0 +1,7 @@
+i<script>
```
YohannParis, line 1: random `i`.
you can remove the whole `<script>` tag as well.

YohannParis (approved):

packages/client/src/views/Home.vue
mwdchang: Random `i` is a vim artifact, sometimes you will also see random Js and random Ks 😆

jryu01 (changes requested):

mwdchang:
@jryu01 I have linter warnings but I do not have transpile/compile errors.

`vue/multi-word-component-names` is a little bit annoying.

jryu01 (approved):

PR #13, opened 2022-09-13
test(init): initialize vitest for client and shared

Nithos:


packages/client/tests/unit/__snapshots__/basic.spec.ts.snap
```diff
@@ -0,0 +1,7 @@
+// Vitest Snapshot v1
```
YohannParis, line 1: Is this suppose to be on git? it is an automatically generated file?

YohannParis (approved):

packages/client/tests/unit/__snapshots__/basic.spec.ts.snap
Nithos: the basic tests does a dummy snapshot that is generated yes, now that its there any changes to the code/test will compare to that snapshot. if it does not match it will fail the test (so you would need to update the snapshot if its a valid change)

TLDR; yes it is generated but it should be there as it is used for test comparison
NOTE: this is a dummy test so has no bearing on the app so can be removed, was meant to showcase the usage at this point

YohannParis (approved):

PR #16, opened 2022-09-13
chore(ci): initial CI configuration for GitHub

### LINKED ISSUE #14: CI integration

Nithos:
Basic GitHub CI that is triggered on PR and direct pushes to main. Will be augmented or replaced with GitLab configuration once those runners are available.

resolves #14

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,7 +1,6 @@
-# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests using node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
-name: Node.js CI
+name: Build & Test
 on:
   push:
     branches: ['main']
@@ -10,13 +9,13 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 18.4.0
+      - name: Use Node.js 18.8.0
         uses: actions/setup-node@v3
         with:
-          node-version: 18.4.0
+          node-version: 18.8.0
           cache: 'yarn'
 
       - name: Install packages
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -4,7 +4,7 @@
 name: Node.js CI
 on:
   push:
-    branches: ['main', 'ci-integration']
+    branches: ['main']
   pull_request:
     branches: ['main']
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,29 @@
...
+name: Node.js CI
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['main']
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
YohannParis, line 13: Our DockerFile uses `node:18.8.0-slim` which I think is based on alpine. Should we use the same environment for CI and container?

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js 18.4.0
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.4.0
+          cache: 'yarn'
```
YohannParis, line 20: Same thing for Node.js and Yarn version with DockerFile

YohannParis (approved):

package.json
```diff
@@ -1,31 +1,32 @@
...
+		"postinstall": "husky install"
+	},
+	"lint-staged": {
+		"*.{ts,vue}": [
+			"eslint --fix",
+			"prettier --write"
+		]
+	},
+	"devDependencies": {
+		"@vue/eslint-config-airbnb-with-typescript": "^7.0.0",
```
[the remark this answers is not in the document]
Nithos: we did, this is an existing dependency that was present. I have not gone through and pinned these yet, will be part of the renovate job.

.github/workflows/ci-test.yml
Nithos: this is not the same, in GitHub workflows this defines the underlying system VM its running on and not the docker image. That being said just noticed that they are deprecating 18.x and we should probably be specific on the version so will upgrade to use v22

YohannParis: I understand that they are not the same, I was just wondering if we should have the same configuration between DockerFile and GitHub Actions, just for saving ourselves headaches in the future.

Nithos: don't have an option for alpine on GitHub as far as I can tell, hopefully once we move to gitlab runners we can make this exact

package.json
```diff
@@ -1,31 +1,32 @@
...
+	"name": "terarium",
+	"packageManager": "yarn@3.2.2",
+	"private": true,
+	"workspaces": [
+		"packages/*"
+	],
+	"scripts": {
+		"build": "yarn workspaces foreach --parallel --interlaced --verbose --topological run build",
+		"dev": "yarn workspaces foreach --parallel --interlaced --verbose run dev",
+		"lint": "eslint . --ext .ts,.vue --fix",
```
jryu01, line 11: I wonder if we should have something like `lint:ci` or `prettier:ci` script that doesn't auto fix the linting/formatting errors for you to be used in the CI pipeline.

Nithos: good catch, will add them

Nithos: done

PR #19, opened 2022-09-20
feat: switch to use quarkus server with keycloak on kubernetes

dgauldie:
There is still a lot of work to do setting this project up. But I wanted to get it in soon so that we don't have too many conflicts


kubernetes/local/gateway-httpd-deployment.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    software.uncharted.terarium: gateway-httpd
+  name: gateway-httpd
+spec:
+  containers:
+    - image: docker-hub.uncharted.software/auth/httpd-openidc:dev-1.0.0
```
Nithos, line 9: our proxy should not be used in this project as it is OSS and meant for others to use as well

kubernetes/local/gateway-httpd-htdocs.yaml
```diff
@@ -0,0 +1,36 @@
...
+data:
+  index.html: |
+    <!DOCTYPE html>
+    <html lang="en">
+      <head>
+        <meta charset="utf-8">
+        <title>Landing Page</title>
+        <link rel="stylesheet" href="style.css">
+      <body>
+        <h2>Uncharted Auth</h2>
```
Nithos, line 15: change to Terarium AUth?

packages/client/terarium/package.json
```diff
@@ -4,24 +4,24 @@
 	"version": "0.0.0",
 	"type": "module",
 	"scripts": {
-		"dev": "vite",
+		"dev": "PORT=8080 vite",
```
Nithos, line 7: extract into an .env file for use in both server and client

packages/services/server/src/main/docker/Dockerfile.jvm
```diff
@@ -0,0 +1,93 @@
...
+# - GC_CONTAINER_OPTIONS: Specify Java GC to use. The value of this variable should
+#   contain the necessary JRE command-line options to specify the required GC, which
+#   will override the default of `-XX:+UseParallelGC` (example: -XX:+UseG1GC).
+# - HTTPS_PROXY: The location of the https proxy. (example: "myuser@127.0.0.1:8080")
+# - HTTP_PROXY: The location of the http proxy. (example: "myuser@127.0.0.1:8080")
+# - NO_PROXY: A comma separated lists of hosts, IP addresses or domains that can be
+#   accessed directly. (example: "foo.example.com,bar.example.com")
+#
+###
+FROM registry.access.redhat.com/ubi8/openjdk-17:1.11
```
Nithos, line 78: should we use temurin builds instead?

kubernetes/local/gateway-httpd-deployment.yaml
bigglesandginger: I have lost all knowledge of which docker repo names mean what.  Should we use `docker.uncharted.software` instead?  I will change to that unless someone says otherwise.

kubernetes/local/gateway-httpd-htdocs.yaml
bigglesandginger: done

YohannParis (commented):
This is what I got for running `quarkus dev`:

```Bash
yparis@yparis-uncharted TERArium % quarkus dev     
The operation couldn’t be completed. Unable to locate a Java Runtime.
Please visit http://www.java.com for information on installing Java.

Downloading https://services.gradle.org/distributions/gradle-7.5-bin.zip
...........10%............20%...........30%............40%...........50%............60%...........70%............80%...........90%............100%

Welcome to Gradle 7.5!

Here are the highlights of this release:
 - Support for Java 18
 - Support for building with Groovy 4
 - Much more responsive continuous builds
 - Improved diagnostics for dependency resolution

For more details see https://docs.gradle.org/7.5/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* What went wrong:
Could not open cp_settings generic class cache for settings file '/Users/yparis/dev/TERArium/settings.gradle' (/Users/yparis/.gradle/caches/7.5/scripts/f2lxffi5r1zlou0jfzvhn2ur5).
> BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 63

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 8s
```

packages/client/terarium/package.json
dgauldie: I don't think we need a .env file for this, but I did move specifying the port to the vite config and I set strictPort to true so that vite ensures we run on 8080 and will exit if it can't run on 8080.

Nithos: details... still feel that it may be good to have a local config that is changeable that can affect both as they need to be in sync... but for now fine... still need to settle on the project setup for all this then see what makes the most sense

YohannParis: Yes, I agree, once this is merge and we create a **TERArium** repo that includes scripts to install and run the whole project at different level, the port numbers should be define then

package.json
```diff
@@ -3,12 +3,12 @@
...
 	"private": true,
 	"workspaces": [
-		"packages/*"
+		"packages/client/*"
 	],
 	"scripts": {
 		"build": "yarn workspaces foreach --parallel --interlaced --verbose --topological run build",
 		"dev": "yarn workspaces foreach --parallel --interlaced --verbose run dev",
-		"serve": "yarn workspace server run serve && yarn workspace client run serve && cp -r packages/client/dist/* packages/server/dist/web/",
+		"serve": "yarn workspace terarium run serve && cp -r packages/client/terarium/dist/*",
```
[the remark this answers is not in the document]
Nithos: renamed and removed copy

YohannParis:
So what is missing to get this PR merged?

bigglesandginger:
> So what is missing to get this PR merged?

I think mostly, can you run this branch successfully?  It is "done" as a preliminary workflow idea.

packages/client/webapp/src/utils/uncloak.ts
```diff
@@ -1,17 +1,11 @@
...
-export const getAccessToken = async () => {
-	const response = await fetch('/silent-check-sso.html');
-
-	if (response.ok) {
-		accessToken = response.headers.get('OIDC_access_token');
-	}
-};
+import { useAuthStore } from '../stores/auth';
 
+// eslint-disable-next-line import/prefer-default-export
```
YohannParis, line 3: We should remove this eslint rule instead of adding this `eslint-disable-next-line`

packages/client/webapp/src/stores/auth.ts
```diff
@@ -0,0 +1,55 @@
+import { defineStore } from 'pinia';
```
YohannParis, line 1: Why `pinia` instead of the web api LocaleStorage?

YohannParis (dismissed):

packages/client/webapp/src/utils/uncloak.ts
mwdchang: I have a set of linter rules changes in my branch for the graph and scaffolding utilities (https://github.com/DARPA-ASKEM/TERArium/tree/porting-graph-scaffold) some of which have already discussed with various folks, I think let's take the linter discussion there rather than doing this exercise multiple times.

packages/client/webapp/src/stores/auth.ts
Nithos: localstorage could be used as well but figure the app will need state management so might as well use it for this as well. works well with vue and wont accidentally forget to clear the tokens

packages/services/server/src/main/docker/Dockerfile.jvm
Nithos: @dgauldie any input on this? 
i know the above is from their demo but just wondering if we should be consistent where possible

Nithos (commented):
I think this is probably at a decent stage to merge and unblock others waiting on this

packages/client/webapp/src/stores/auth.ts
YohannParis: I see, I wasn't really keen on using a JS framework specific state management dependency, because people will start to use to bypass simple event/prop communication path, and do weird thing.

But let's move on, and I will keep an eye on it.

YohannParis (approved):

PR #21, opened 2022-09-23
refactor: add placeholder login button and api calls

shawnyama:
Added a Header component above route components. The Header just has a navbar with a logout button. 
Added buttons to the homepage that will test api calls later.


packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
+<script setup lang="ts">
+function logout() {
+	console.log('logout');
+}
+</script>
+
+<template>
+	<nav id="navbar">
```
YohannParis, line 8: No need for ID, just use the type selector

packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
...
+		<div class="navbar-left">
```
YohannParis, line 9: No need for the div `navbar-left/right`

packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
...
+			<h2>TERArium</h2>
```
YohannParis, line 10: Maybe a `header` is more fitting

packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
...
+#navbar {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	border-bottom: 1px solid lightgrey;
+}
+
+.navbar-left,
+.navbar-right {
+	margin: 0.5rem 1rem;
```
YohannParis, line 28: This should be a `padding` within the `#navbar` instead of those margins.

packages/client/src/views/Home.vue
```diff
@@ -1,3 +1,43 @@
...
+		<div class="test-api-calls">
+			<button type="button" @click="apiCall">New Model</button>
+			<button type="button" @click="apiCall">Update Model</button>
+			<button type="button" @click="apiCall">Get Model</button>
+			<button type="button" @click="apiCall">Delete Model</button>
+		</div>
+	</div>
 </template>
+
+<style>
```
YohannParis, line 25: Add `scoped` attribute
https://vuejs.org/api/sfc-css-features.html#scoped-css

packages/client/src/views/Home.vue
```diff
@@ -1,3 +1,43 @@
...
+	padding: 2rem;
+	text-align: center;
+}
+
+.test-api-calls {
+	display: flex;
+	margin: 2rem;
+	justify-content: center;
+}
+.test-api-calls button {
```
YohannParis, line 39:
```suggestion
button [type=button] {
```

YohannParis (commented):
Great Jobs for a first PR !

packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
+<script setup lang="ts">
+function logout() {
+	console.log('logout');
```
Nithos, line 3: eslint rules prevent console in CI (we can enable for local builds just be warned)

packages/client/src/components/Header.vue
Nithos: also make sure your styles are scoped

packages/client/src/components/Header.vue
Nithos: @YohannParis why? I agree that may not be needed for all but some may be useful for e2e testing

packages/client/src/components/Header.vue
```diff
@@ -0,0 +1,34 @@
+<script setup lang="ts">
```
Nithos, line 1: component should have a corresponding test (at the very least a stub for one) in the `tests` folder
please reach out to me if you have question regarding e2e testing using playwright

bencodrington: I think the larger team (or at least I) was unaware of the expectation to have a test for each component.
We'd discussed adding unit tests for business logic in utils/services, but it may be worth having a quick chat with all the devs to make sure we're all on the same page for where and when e2e tests should be included.

packages/client/src/components/Header.vue
YohannParis: True, but then we can add the ID if needed, not in case of its needed.

packages/client/src/components/Header.vue
Nithos: The expectation is to have applications well tested. This in some (but maybe not all) cases should include e2e/component tests as well. This is something new and would start of slowly but want to get into the habit of at least thinking about writing them and the possible things to test there. Nav bar may seem simple but its a good sanity check for things even like auth where some menus are shown/hidden depending on the user etc.. In any case yes it is something to be discussed further with the team.

Nithos: added small reference test

packages/client/src/App.vue
```diff
@@ -1,3 +1,8 @@
+<script setup lang="ts">
+import Header from './components/Header.vue';
```
Nithos, line 2: should use the alias `@` for consistency
may add other aliases in the future if needed

YohannParis:
```suggestion
import Header from '@/components/Header.vue';
```

Nithos (approved):

PR #63, opened 2022-10-06
feat: add a playground and basic plumbing code for TA2/Catlab related experiments

mwdchang:
Note this is a work in progress. But if people want a preview to try things out

### Summary
This adds basic rendering and plumbing code to interact with Catlab.
Mostly for demoing purposes, but much of the rendering and interaction code, I think, will be able to repurpose in the actual implementation.

- The buttons adds places/transitions respectively
- To add an edge, shift click `(place, transition)` or `(transition, place)` . Note Catlab will not allow you to create invalid relations but the UI is not prevent this currently.

This also address a few configuration errors in graph-scaffolder that prevented it from being imported properly.

### Testing
You will need to have the model-transform service running (see https://github.com/DARPA-ASKEM/model-transform). 
Either build it as a docker and run the image
```
docker build . -t aksem-model-transform 
docker run -p 8888:8888 askem-model-transform
```
Or directly run the service
```
julia src/service.jl
```

Once that is done bring up terarium application, the playground is available on 
```
http://localhost:8078/app/#/ta2-playground
```


packages/client/webapp/src/views/TA2Playground.vue
```diff
@@ -0,0 +1,343 @@
...
+				})
+			});
+			this.jsonOutput();
+		}
+	}
+});
+</script>
+
+<template>
+	<div style="margin: 10px">
```
YohannParis, line 334: I'll prefer a `<section>` here.
And also to put all the CSS within a `<script scoped>`

YohannParis (approved):

YohannParis (dismissed):

bencodrington:
Didn't test, but code looks totally fine for experiments.

May want to add the instructions about the model transform service to a README for people trying it out in the future.

PR #65, opened 2022-10-07
feat: collapsible sidebar

shawnyama:
Sidebar with nav options 
- chose icons that made the most sense to me
- can be collapsed if the current nav option is clicked again 
- the only feature within the sidebar is to change the position of the sidebar (this is available when the profile option is chosen)

If you look at `Header.vue` and `App.vue` you can tell I built this around the fact that the top navbar has a height of `3.5rem`. Not sure if this is the best approach.

image


Nithos:
Link relevant ticket when possible, through the sidebar or keywords such as `resolves #x` or `closes #x`

packages/client/webapp/src/App.vue
```diff
@@ -23,5 +27,22 @@ const isAuthenticated = computed(() => auth.isAuthenticated);
...
+		<Sidebar
+			:sidebar-position="sidebarPosition"
+			@updateSidebarPosition="(newPosition) => (sidebarPosition = newPosition)"
+		/>
+		<router-view />
+	</div>
 </template>
+
+<style scoped>
+div {
```
YohannParis, line 40: Sorry, you can do that, you need to name that div something more specific, as it will affect all the other `div` from within the `<router-view>`

packages/client/webapp/src/App.vue
```diff
@@ -1,19 +1,23 @@
...
  */
 onBeforeMount(() => {
 	auth.fetchSSO();
+	if (localStorage.sidebarPosition === undefined) {
+		localStorage.setItem('sidebarPosition', 'left');
+	}
 });
 
 const isAuthenticated = computed(() => auth.isAuthenticated);
+const sidebarPosition = ref(localStorage.sidebarPosition);
```
bencodrington: Nitpick: You may wish to get used to using the syntax:
```suggestion
const sidebarPosition = ref(localStorage.getItem('sidebarPosition'));
```

The only differences are:
- `getItem` returns `null` instead of `undefined` when the item is missing, and
- `getItem` will behave consistently for any item name, whereas the `localStorage.x` syntax will be unpredictable if `x` is any of the 6 existing properties of localStorage, most notably `length` or `key`.

See [this stackoverflow response (specifically the EDIT)](https://stackoverflow.com/a/15439059) for a more detailed answer.

packages/client/webapp/src/App.vue
bencodrington: I don't think that's true, scoped files don't affect descendent elements.

For example, the `div` surrounding the buttons on the home page:

image

It's inside the `router-view`, but it doesn't inherit any of these styles from App.vue:

image

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,134 @@
...
+import MachineLearningModel from '@carbon/icons-vue/es/machine-learning-model/32';
+import Dataset from '@carbon/icons-vue/es/table--split/32';
+import ProvenanceGraph from '@carbon/icons-vue/es/flow/32';
+import Search from '@carbon/icons-vue/es/search--locate/32';
+import User from '@carbon/icons-vue/es/user/32';
+
+const props = defineProps<{ sidebarPosition: String }>();
+const emit = defineEmits(['updateSidebarPosition']);
+
+const mode = ref('Simulation Plan');
```
bencodrington: I would recommend extracting 'Simulation Plan' and the other hardcoded strings to a constant enum at the top of the file. This enables auto-complete and static checking and avoids typos when repeating these strings in multiple places.

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,134 @@
...
+const collapsed = ref(false);
```
bencodrington: It's a useful convention to start boolean variables with `is` or `has`, so this would be `isCollapsed`.

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,134 @@
...
+
+function updateMode(modeOption: string) {
+	if (modeOption === mode.value) {
+		collapsed.value = !collapsed.value;
```
bencodrington: Am I correct in thinking that this can be simplified to the following?

```suggestion
		collapsed.value = true;
```

Seems like if `modeOption === mode.value`, then collapsed must be false.

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,134 @@
...
+import Search from '@carbon/icons-vue/es/search--locate/32';
+import User from '@carbon/icons-vue/es/user/32';
+
+const props = defineProps<{ sidebarPosition: String }>();
+const emit = defineEmits(['updateSidebarPosition']);
+
+const mode = ref('Simulation Plan');
+const collapsed = ref(false);
+
+function updateMode(modeOption: string) {
```
bencodrington: Nitpick: This might be slightly more readable by using a more meaningful variable name for `modeOption`.

Perhaps `selectedMode` or `toggledMode`?

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,134 @@
...
+	if (modeOption === mode.value) {
+		collapsed.value = !collapsed.value;
+		return;
+	}
+	collapsed.value = false;
+	mode.value = modeOption;
+}
+
+function moveSidebar() {
+	if (localStorage.sidebarPosition === 'left') {
```
bencodrington: These `left` and `right` strings should also be extracted to avoid repeating hardcoded strings all over the place.

Consider:
- If the only options we want to support for the time being are "left" and "right", we can use a boolean like `isSidebarPositionedRight` instead of a string. This if/else chain can then be simplified to `const currentPosition = localStorage.getItem('sidebarPosition'); localStorage.setItem(!currentPosition)`
- If we expect to support other positions, or prefer to be explicit about what the "non-right" position is, this is a good place to use a constant enum.

bencodrington (dismissed):
This is looking really good!

I left several comments, some nit-pickier than other, to get a sense of some small programming habits you can get into to improve readability and maintainability longterm.

But overall this looks good and works well! Thanks.

packages/client/webapp/src/App.vue
YohannParis: I agree, I was working on an old version of this branch

packages/client/webapp/src/components/Sidebar.vue
Nithos: Great point Ben, definitely want to avoid repeating yourself. Stay DRY everyone ;)

One followup question. Does/Will the app need language support? If so might be good to get started on localization of all these strings as well.

YohannParis: https://www.typescriptlang.org/docs/handbook/enums.html

packages/client/webapp/src/components/Sidebar.vue
YohannParis: No, it needs to be a toggle.

https://user-images.githubusercontent.com/636801/194637009-d66c2373-165c-4c40-9fe9-ddc97ff8cd92.mov

packages/client/webapp/src/components/Sidebar.vue
YohannParis: we talked about using `mode`, as it is what being passed to the function, and expected based on its name

packages/client/webapp/src/components/Sidebar.vue
bencodrington: @Nithos My guess would be not to worry about localization any time soon. In world modelers it never once became a requirement, and setting up a language support system in the meantime would add overhead when writing and understanding code.

packages/client/webapp/src/components/Sidebar.vue
bencodrington: I see. To me it's unintuitive that there would still be an active/selected tab even though the side panel is collapsed. Wouldn't it be more intuitive if mode.value is null or `''` by when the panel is collapsed?

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -0,0 +1,137 @@
...
+nav.mode-selection ul li {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	cursor: pointer;
+	height: 3rem;
+	width: 3rem;
+	margin: 0.5rem auto;
+	border-radius: 5px;
+	background-color: whitesmoke;
```
YohannParis, line 119: Change all those hardcoded colour for the Uncharted Design tokesn

packages/client/webapp/src/components/Sidebar.vue
shawnyama: Yeah I see what you mean - vscode actually works like that. In this case though (from what I understand) what is displayed in the `<router-view/>` corresponds to what's selected in the sidebar. So it might make sense.

YohannParis: Sorry @shawnyama but this meaning has changed right after our conversation...
So @bencodrington original opinion is now valid, and the one to go trough with.
The `<router-view/>` will not change based on the sidebar main buttons.

shawnyama: Okay I'll change that then

YohannParis (approved):

PR #66, opened 2022-10-07
chore: add task request template

Nithos:
Adding template for general development tasks based on designs or requirements that do not require a full feature

### DIFF
--- a/.github/ISSUE_TEMPLATE/task-request.md
+++ b/.github/ISSUE_TEMPLATE/task-request.md
@@ -1,0 +1,24 @@
+---
+name: Task request
+about: Template for development tasks
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the task**
+A clear and concise description of what the task is.
+
+**Task Items**
+- [ ] list of items for completion
+- [ ] another item
+
+**Goal**
+A clear and concise description of what the end result must have for it to be approved as complete.
+
+**Screenshots**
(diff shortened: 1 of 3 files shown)


YohannParis (approved):

.github/ISSUE_TEMPLATE/task-request.md
```diff
@@ -0,0 +1,24 @@
+---
+name: Task request
+about: Template for development tasks
+title: ''
+labels: ''
```
bencodrington, line 5: You removed these quotes

```suggestion
labels:
```

In the changes above. Do we need them here? What's the correct syntax?

bencodrington (approved):
Looks good to me! One syntax clarification question, otherwise merge away.

Never used the code-owners feature before, seems really handy.

.github/ISSUE_TEMPLATE/task-request.md
Nithos: hmm looking into it... it was done through github generator so not sure why they are inconsistent

Nithos: its just YAML underneath so can be either or unless you need to escape something... made them all consistent now

YohannParis (approved):
I think we should move on with it, and if they fix that Task list issue, then great.
I will make an extra effort to follow the template when creating issues.

PR #69, opened 2022-10-07
fix: small changes for the demo

YohannParis:


packages/client/webapp/src/assets/css/style.css
```diff
@@ -15,6 +15,11 @@
 	--un-color-accent: #2D8E2Dff;
 	--un-color-accent-light: #92E192ff;
 	--un-color-accent-lighter: #E4FAE4ff;
+
+	--un-z-index-sidebar: 50;
+	--un-z-index-navbar: 100;
+	--un-z-index-notification: 300;
+	--un-z-index-modal: 500;
```
mwdchang, line 22: Looks fine to me, but I know @bencodrington has some opinion and/or suggestions about z-indices in WM project because they are relative to the context and not meant as an absolute measure.

YohannParis: Yes, I would love to have opinions on the matter.
In this case, we can just add them if you have a notification or sidebar within a modal with
```css
.modal-notification {
    z-index: calc(var(--un-z-index-modal) + var(--un-z-index-notification));
}
```

packages/client/webapp/src/components/Header.vue
```diff
@@ -1,33 +1,69 @@
...
+		<Button class="dataExplorer" @click="router.push('/explorer')"><IconSearchLocate16 /></Button>
+		<Button v-if="!auth.isAuthenticated" @click="login">Login <IconLogin16 /></Button>
 	</nav>
 </template>
 
 <style scoped>
 nav {
 	align-items: center;
+	background-color: var(--un-color-body-surface-primary);
 	border-bottom: 1px solid var(--un-color-black-20);
```
shawnyama, line 36: If we are using box shadows I think we should get rid of the solid borders. The header border makes the mode config almost look like its 1px higher than the sidebar nav. Since you changed the sidebar nav to a darker green I think it looks cleaner without its solid border as well.

Current version:
image

Without solid borders:
image

YohannParis: Thank you! I think they got back from a merge from `main`.

packages/client/webapp/src/assets/css/style.css
YohannParis: Sold!

packages/client/webapp/src/App.vue
```diff
@@ -33,5 +33,29 @@ const isAuthenticated = computed(() => auth.isAuthenticated);
...
+main > section {
+	z-index: 1;
+}
+</style>
+<style>
+body {
+	isolation: isolate;
+}
+
+header {
```
bencodrington, line 54: Do we want these unscoped styles here? Maybe it would be better to apply a class to the `Header` component and style that within the `scoped` tag. Similarly, can we move the `main` styles below up to the scoped section?

packages/client/webapp/src/components/Header.vue
```diff
@@ -1,33 +1,71 @@
...
+import IconSearchLocate16 from '@carbon/icons-vue/es/search--locate/16';
+import { useRouter } from 'vue-router';
 import { useAuthStore } from '../stores/auth';
 
 const auth = useAuthStore();
-const isAuthenticated = computed(() => auth.isAuthenticated);
+const router = useRouter();
 
 const login = () => window.location.assign('http://localhost:8078');
+const homepage = () => router.push('/');
```
bencodrington, line 12: Nitpick: I think it's a good convention to name functions/methods starting with a verb, e.g. `goToHomePage`

packages/client/webapp/src/components/Header.vue
```diff
@@ -1,33 +1,71 @@
...
-		{{ auth.name }}
-		<Button v-if="!isAuthenticated" @click="login">Login <IconLogin16 /></Button>
-	</nav>
+		<p>
+			<a @click="homepage">Projects</a>
+			<span>{{ projectName }}</span>
+		</p>
+
+		<aside>
+			<Button class="dataExplorer" @click="router.push('/explorer')"><IconSearchLocate16 /></Button>
```
bencodrington, line 26: Not sure if we established this convention team-wide, how do folks feel about consistently using `kebab-case` CSS classes?

packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -50,49 +49,42 @@ function moveSidebar() {
...
+			</li>
+			<li :active="selectedMode === Mode.Profile" @click="updateMode(Mode.Profile)">
+				<IconUser32 />
+			</li>
+		</ul>
+	</nav>
+	<aside v-if="!isCollapsed" class="mode-configuration" :class="{ right: isSidebarPositionRight }">
+		<header>{{ selectedMode }}</header>
+		<main v-if="selectedMode === Mode.Profile">
+			<Button @click="moveSidebar"> Move sidebar </Button>
```
bencodrington, line 76: The move sidebar button is broken now. Clicking it toggles the class, but the class has no effect.

bencodrington (changes requested):
This is looking really nice, thanks for incorporating the z-index stuff! Left a couple of comments to address/discuss.

As a general note, the data-explorer tests seem unrelated to this PR and should be split into their own PR next time we run into this situation.

Thanks!

packages/client/webapp/src/App.vue
YohannParis: I am not a fan of adding extra attributes when unnecessary, but I guess that's part of the single web app design.

bencodrington (dismissed):
Looks good! Thanks.

shawnyama (approved):

bencodrington (approved):

PR #92, opened 2022-10-12
feat: add theia view and model sidebar panel

YohannParis:
# Description

* Added a _Theia_ view to demonstrate the tool for the October 13 demo
* Created a `model-sidebar-panel.vue` component to host content

https://user-images.githubusercontent.com/636801/195395129-f08b7b61-2a74-4ba0-be86-b2ab2683179e.mov

## Type of change

# How Has This Been Tested?

- [x] Added sidepanel _component_ test for the header and the button

# Checklist:


packages/client/webapp/src/components/Sidebar.vue
```diff
@@ -158,8 +162,7 @@ aside {
 	display: flex;
 	flex-direction: column;
 	justify-content: space-between;
-	padding: 0.5rem;
-	width: 15rem;
+	width: max(15em, 20vw);
```
mwdchang, line 165: I don't know if we had discussed this, but are we using `rem` or we are using `em`? Would prefer if we stick to one system rather than mixing both. Slight personal preference for `rem`.

YohannParis: They are not the same, I don't see how can this be an issue?

mwdchang: Exactly, there's nothing technically wrong here, they are both relative measurements but with different basis. 

So when you mix and match it is like you are working with two different coordinate systems. Potentially it can be confusing, especially if components are meant to embed into others. Imaging the inner-most `div` is a reusable component.

```
<html>
  <head>
  </head>
  <body>
    <div style="font-size: 1em"> Hi
      <div style="font-size: 2em"> Hi
        <div style="font-size: 2em"> Hi
        </div>
      </div>
    <div>
    <hr>
    <div style="font-size: 1rem"> Hi
      <div style="font-size: 2rem"> Hi
        <div style="font-size: 2rem"> Hi
        </div>
      </div>
    <div>
    <hr>
    <div style="font-size: 1em"> Hi
      <div style="font-size: 2em"> Hi
        <div style="font-size: 2rem"> Hi
        </div>
      </div>
    <div>
  </body>
</html>
```

image

bencodrington (changes requested):
Looks good! Can you add some styles to stop the iframe from overflowing the page? I think it would be good for the parent component to control the maximum iframe dimensions.

image

packages/client/webapp/src/components/Sidebar.vue
YohannParis: I agree with you, but we are using the font-system from the Uncharted design tokens, and we scope our css as much as possible.

After it's important to let the design be responsibe to a change a font-size within a component or it's parent.
This is why I use `em` when its a behaviour I'm expecting.

YohannParis:
> Looks good! Can you add some styles to stop the iframe from overflowing the page? I think it would be good for the parent component to control the maximum iframe dimensions.

This is an issue because the sidebar is pushing the content of the page instead of being an overlay.
The iframe cannot be resized easily, and because this is not a permanent solution but just an easy way for Powell to demonstrate the tool tomorrow. It will be fine for him to close the sidebar after clicking on the button.

I will create issues to remove the iframe, make the sidebar overlaying the content, etc.

packages/client/webapp/src/components/Sidebar.vue
mwdchang: But we are using `rem` in other places.
The font variables are also specified in `rem` so they are not responsive per se, and  the scoping will not help in this case I believe, because it searches the nearest parent-node regardless.

YohannParis: Fine, I'll stop using `em` and choose a unit tied to the parent component as source of responsiveness instead of font

bencodrington (approved):

PR #128, opened 2022-10-19
fix: broken quarkus dependencies

mwdchang:
# Description
Fix bad dependency where `quarkus dev` wouldn't start properly.

## Type of change


Nithos:
@dgauldie @Tom-Szendrey can this be reviewed and merged please. Think this is blocking and causing some issues.

gradle.properties
```diff
@@ -1,7 +1,7 @@
 #Gradle properties
 quarkusPluginId=io.quarkus
-quarkusPluginVersion=2.12.1.Final
+quarkusPluginVersion=2.13.2.Final
```
bigglesandginger, line 3: should this not be the same version as `quarkusPlatformVersion` - ie `2.13.3.FInal`?

Nithos: updated to match the bom

Nithos:
So this I believe stems from the fact that Renovate Bot updated the BOM but not the plugin, should these always be in sync? If so we need to tell renovate to group these for future reference.

dgauldie (approved):

PR #131, opened 2022-10-19
feat: hide sidebar on home view

### LINKED ISSUE #118: Toggle Sidebar visiblity
- On the homepage, the _Sidebar_ should not be visible.

- Hide/Show the Sidebar based depending on which _Views_ is visible. Hide for `Home`, Show for the others.

echl:
# Description

Enumerated route paths and add a condition to render sidebar if the current route is not home. 

Resolves #118 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/client/webapp/src/App.vue
+++ b/packages/client/webapp/src/App.vue
@@ -4,8 +4,12 @@ import Header from '@/components/Header.vue';
 import Sidebar from '@/components/Sidebar.vue';
 import Overlay from '@/components/Overlay.vue';
 import DataExplorer from '@/views/DataExplorer.vue';
-
+import { useRouter } from 'vue-router';
 import useAuthStore from './stores/auth';
+import { HOME_PATH } from './router/index';
+
+const router = useRouter();
+const sidebarHidden = computed(() => router.currentRoute.value.path === HOME_PATH);
 
 const auth = useAuthStore();
 
@@ -52,7 +56,7 @@ const dataExplorerActivated = ref(false);
 	/>
 	<Header class="header" @show-data-explorer="dataExplorerActivated = true" />
 	<main v-if="isAuthenticated">
-		<Sidebar class="sidebar" />
+		<Sidebar v-if="!sidebarHidden" class="sidebar" />
 		<router-view class="page" />
 	</main>
 </template>
@@ -71,6 +75,10 @@ main {
 
 .sidebar {
 	z-index: 2;
+}
+
+.hidden {
+	visibility: hidden;
 }
 
 .page {
--- a/packages/client/webapp/src/App.vue
+++ b/packages/client/webapp/src/App.vue
@@ -4,12 +4,10 @@ import Header from '@/components/Header.vue';
 import Sidebar from '@/components/Sidebar.vue';
 import Overlay from '@/components/Overlay.vue';
 import DataExplorer from '@/views/DataExplorer.vue';
-import { useRouter } from 'vue-router';
 import useAuthStore from './stores/auth';
-import { HOME_PATH } from './router/index';
+import { useIsCurrentRouteHome } from './router/index';
 
-const router = useRouter();
-const sidebarHidden = computed(() => router.currentRoute.value.path === HOME_PATH);
+const isSidebarVisible = computed(() => !useIsCurrentRouteHome());
 
 const auth = useAuthStore();
 
@@ -56,7 +54,7 @@ const dataExplorerActivated = ref(false);
 	/>
 	<Header class="header" @show-data-explorer="dataExplorerActivated = true" />
 	<main v-if="isAuthenticated">
-		<Sidebar v-if="!sidebarHidden" class="sidebar" />
+		<Sidebar v-if="isSidebarVisible" class="sidebar" />
 		<router-view class="page" />
 	</main>
 </template>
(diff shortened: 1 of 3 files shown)


packages/client/webapp/src/router/index.ts
```diff
@@ -4,8 +4,10 @@ import ResponsiveMatrixCells from '@/components/ResponsiveMatrixCells.vue';
 import TA2Playground from '@/views/TA2Playground.vue';
 import Theia from '@/views/theia.vue';
 
+export const HOME_PATH = '/';
```
YohannParis, line 7: This should be nice as [enum](https://www.typescriptlang.org/docs/handbook/enums.html) of _ALL_ the routes.

packages/client/webapp/src/App.vue
```diff
@@ -4,8 +4,12 @@ import Header from '@/components/Header.vue';
 import Sidebar from '@/components/Sidebar.vue';
 import Overlay from '@/components/Overlay.vue';
 import DataExplorer from '@/views/DataExplorer.vue';
-
+import { useRouter } from 'vue-router';
 import useAuthStore from './stores/auth';
+import { HOME_PATH } from './router/index';
+
+const router = useRouter();
+const sidebarHidden = computed(() => router.currentRoute.value.path === HOME_PATH);
```
YohannParis, line 12:
```suggestion
const isSidebarVisible = computed(() => router.currentRoute.value.path !== HOME_PATH);
```

packages/client/webapp/src/router/index.ts
echl: Done!

packages/client/webapp/src/router/index.ts
```diff
@@ -19,4 +26,8 @@ const router = createRouter({
 	routes
 });
 
+export function useIsCurrentRouteHome() {
```
bencodrington, line 29: This isn't exactly the typical composable pattern, so it might confuse future developers.
- This function depends on router.currentRoute.value.path, which is a reactive dependency changes over time. Just looking at the function signature, though, a developer calling this function might not expect it to change since it's not a computed property or reactive state, and it doesn't have any parameters. I'd recommend either making this a pure function where you pass in `router.currentRoute.value` or make it a computed property directly.
- Composables typically return an object whose properties are all, or almost all, reactive, whereas this function just returns a boolean. If this were a composable its return statement would look something like:
`return { isCurrentRouteHome: computed(() => router.currentRoute.value.path === RoutePath.Home;) };` which seems like overkill to me, a simple computed property would suffice.
- I'm genuinely not sure how using composition API reactivity functions like `computed` would work outside of a component, may need to wrap it in a composable or just move it directly to the App.vue component where it's being used.

Happy to discuss further or have a quick call if you disagree with or are confused by any of that ☝️

bencodrington (changes requested):
Good stuff!

Left a pretty rambly clean up comment but other than that this looks good.

Thanks.

packages/client/webapp/src/router/index.ts
echl: Hi Ben, 

Good points. You're right, it doesn't really need to be a composable, though I disagree on passing in the value as the caller then needs a reference to router first. It seems simpler and more straight forward to have a function you can call without passing in any parameters. Plus, if you already have the name of the current route then evaluating it is trivial enough to not need separate function anyway.

bencodrington: Totally agreed, making it a pure function doesn't really make sense in this case.
My point (buried somewhere in the paragraphs above) is that in general in the app we should have pure functions OR functions with reactive dependencies that live in components or composables.

In the interest of avoiding hard-to-debug reactivity chains as our app grows in size, functions that have reactive dependencies should be distinct. Since we agree pure function isn't the way to go here, I'd recommend creating a composable that exports the `isCurrentRouteHome` computed property.

Then any time this code is used it'll look like
```
const { isCurrentRouteHome } = useEdwinsFancyNewComposable();
const isSidebarVisible = computed(() => isCurrentRouteHome.value === false);
```

and it's impossible to use without that `.value` that signifies it's a reactive dependency that can and will change.

echl: @bencodrington Clarification needed: as I understand it so far, `.value` would imply it's a ref. Are you suggesting I wrap `isCurrentRouteHome` in a ref?

packages/client/webapp/tests/e2e/base.spec.ts
```diff
@@ -19,4 +19,9 @@ test.describe('main landing page test', () => {
 		// Expect an attribute "to be strictly equal" to the value.
 		await expect(header).toBeVisible();
 	});
+
+	test('should not display sidebar on home page', async ({ page }) => {
+		const sidebar = page.locator('data-test-id=sidebar');
```
YohannParis, line 24: The classname `sidebar` did not work for the test.

echl: Selecting based on the class `sidebar` works but I figured class implies it may not be unique.

packages/client/webapp/src/router/index.ts
bencodrington: This has been discussed in a call and resolved.

Thanks for addressing that 👍

YohannParis (approved):

bencodrington (approved):

PR #135, opened 2022-10-21
feat: redirect unauthenticated user to the login page

### LINKED ISSUE #113: [FEAT]: Redirect unauthenticated user to the login page
- When an user is unauthenticated, they can still access the `webapp` with a login button.
- This behaviour makes it difficult and inconvenient to run just the `webapp` locally and for testing.

- Redirect unauthenticated user to the login page and keep the authentication process out of the `webapp`

shawnyama:
# Description

I moved the login check to `main.ts`. The link redirection is now done in `auth.ts`. Not sure if this is the best way to do it - I was just focused on getting auth operations outside of the Vue components.

In this version you won't be automatically redirected to see passwords for adam and ursala (localhost:8078) it will go straight to the keycloak login. I assume that's fine.

Resolves #113 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


packages/client/webapp/src/stores/auth.ts
```diff
@@ -52,6 +52,7 @@ const useAuthStore = defineStore('auth', {
 					: await fetch('/silent-check-sso.html');
 
 			if (!response.ok) {
+				window.location.assign('http://localhost:8078/app/');
```
Nithos, line 55: my only concern here is the presence of `localhost`, does a relative path not work? i.e `/app`

shawnyama: The relative path doesn't work because the localhost port for the webapp defaults to 8080 but keycloak is 8078.

YohannParis: I think this need to be an `.env` variable to know what is the origin of the webapp.
Because on _staging_ and _production_ those will `terarium.ai/` and _local_ might stayt `localhost:8078/app`.

bigglesandginger:
Redirecting to `/app` and bypassing the landing page is a potential issue.  The current landing page is for Uncharted developers to remember what users they have available and have a link to the only UI at the time the landing page was written, however, in the future (now?) that could/should change.  The landing page should be a menu of UI/Tools that can be selected to use, or simply a info dump as to what TERArium is, this might be pointless to developers but could be instrumental to actual users.

packages/client/webapp/src/stores/auth.ts
bigglesandginger: Have to agree, hardcoding something like `localhost` is not great, and can be a source of errors to come.  Ideally, this should be relative, but I get that Yarn insists on a very specific developer environment, a shame as deployment can/will be much freer

bigglesandginger: `/logout`, whilst effectively a no-op, is probably a nice "compromise" since the by product of logging out is being redirected to the front door.  As for the `http://localhost:8078` thing, it should be in the README for devs to find, and if someone doesn't like that specific port, then that can be changed I chose a non 8080 port with no rhyme or reason.

bigglesandginger: How logout works and those "gateway" details are defined within the kubernetes yaml files and are editable.

YohannParis:
Thanks @Nithos 
I believe once we removed the authentication from the front-end we will not need this. Or we could use the staging keycloak with a test user specifically for this purpose. But this means using secrets within test. Is this possible?

Nithos:
> Thanks @Nithos I believe once we removed the authentication from the front-end we will not need this. Or we could use the staging keycloak with a test user specifically for this purpose. But this means using secrets within test. Is this possible?

Authentication is getting removed? How?
Currently this is just an intercept that `mocks` a response from keycloak. This response can be anything really it does not matter, however currently 'auth' is extracting some meta information from the token (name, email) and eventually it could extract the roles that user has access to (analyst, admin, etc..) So if we want to mimick that we just need to encode whatever we need within that token used for tests. 

This was meant to use the dummy user we have but can be changed, and it can also be loaded from a local env if needed. But I think for test purposes we just need to generate ones that meet our needs for the tests only, for now at least. Again this may need to change in the future.

YohannParis (dismissed):

YohannParis (approved):

PR #174, opened 2022-10-31
feat: tom playground

Tom-Szendrey:
# Description
Adding new features to TA2-Playground:
-a petrinet drawing function
-stratify models - Provided 2 models + a type model stratify them, create a new model and draw the output
-Create Models Button (add a few hard coded JSON models)
-load Model -> Provided a model ID draw it on playground

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.
https://github.com/DARPA-ASKEM/model-transform/issues/6

Resolves #(issue)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Manually - This page is meant as a playground and will be shortly overwritten so i have not added any test cases.

- [X] fetchStratificationResult - Error correctly thrown
- Manual testing such as: 
     Create Models 
     Fill in model IDs + Stratify
     Load models based off of ids

# Checklist:


packages/client/webapp/src/utils/TA2/stratify.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { json } from 'd3';
+
+// TODO: Better error handling (models not found for example)
+//       Override function to also accept 2 typed models (need to chat with TA2 to fully understand)
+export async function stratify(modelA, modelB, typeModel) {
+	console.log('Start Stratify');
+
+	if (!modelA || !modelB || !typeModel) {
+		return json('An ID must be provided for each model');
```
bencodrington, line 12: Since this function is running on the frontend, we don't need to wrap the response in JS Object Notation, we could just return the string or an object directly.

Also, since d3 is just the library we're using for rendering graphs we shouldn't need to use its general purpose JSON utility functions at any point.

I could see an argument for either throwing an error:

```suggestion
		throw new Error('An ID must be provided for each model');
```

or logging the error directly to the console and returning an empty object `{}` or something.

Not sure if there's an established precedent in the app for where and how we want to handle errors so either of those would be fine.

packages/client/webapp/src/utils/TA2/stratify.ts
```diff
@@ -0,0 +1,25 @@
...
+	}
+	console.log(`Stratifying: ${modelA} ${modelB} ${typeModel}`);
```
bencodrington, line 14: I think it's a good habit to get into to clear out these console.log()s before raising your PR, it'll slow down debugging in the future if we end up needing to dig through the logs from when each feature was being developed.

packages/client/webapp/src/utils/TA2/stratify.ts
```diff
@@ -0,0 +1,25 @@
+// Providing the ID of 3 Models (model A, model B, and the type Model)
```
bencodrington, line 1: Can you move this file to `/services/` instead of `/utils/` please? Dan mentioned he'd like to split it so that "services" is anything specific to this project or anything that depends on our APIs, and "utils" are general purpose math or DOM or data structure functions that could be copy-pasted to future Uncharted projects.

Also would be good to establish the naming convention of calling it `stratification-service.ts` or `model-transform-service.ts`

packages/client/webapp/src/utils/TA2/stratify.ts
```diff
@@ -0,0 +1,25 @@
...
+// Create a new model of based off of the stratification
+
+import { json } from 'd3';
+
+// TODO: Better error handling (models not found for example)
+//       Override function to also accept 2 typed models (need to chat with TA2 to fully understand)
+export async function stratify(modelA, modelB, typeModel) {
```
bencodrington, line 8: Nitpick: this might be more accurately titled `fetchStratificationResult(...)` which would serve the double purpose of making it easier to avoid name clashes when importing later, e.g.
```
import { fetchStratificationResult } from '@/utils/TA2/stratify';
fetchStratificationResult()
// instead of
import * as strat from '@/utils/TA2/stratify';
strat.stratify()
```

bencodrington (commented):
Left some notes, some nitpickier than others, on the two files you actually changed.

Once all this 40-file gitstorm is resolved I'll actually test the code and do another round of feedback.

Looking really good though! Thanks.

packages/client/webapp/src/utils/TA2/stratify.ts
YohannParis: I'm for throwing Errors in JS, but we have to make sure they are properly handle!
So for each `throw new Error` I would like to see a test handling it.

packages/client/webapp/src/utils/TA2/stratify.ts
Tom-Szendrey: Would you prefer `/services/TA2` or` /services`?

YohannParis: Maybe `/services/models/`

YohannParis:
> I will be removing this draft tomorrow as Im replacing it with #190
> Im not sure why this branch is showing the 38 file changes so updates will be done there

It was simply an issue of `main` going forward compared to your branch.
I updated it and now it's all good.

Tom-Szendrey:
> > I will be removing this draft tomorrow as Im replacing it with #190
> > Im not sure why this branch is showing the 38 file changes so updates will be done there
> 
> It was simply an issue of `main` going forward compared to your branch. I updated it and now it's all good.

Weird that it detected these as file changes from my end. Thanks! pulling from main was causing issues for me for some reason

packages/client/webapp/src/utils/TA2/stratify.ts
Tom-Szendrey: Good point - done

packages/client/webapp/src/utils/TA2/stratify.ts
Tom-Szendrey: Good points. I have written this test case down. Currently just testing the fail case

packages/client/hmi-client/src/views/TA2Playground.vue
```diff
@@ -527,6 +639,155 @@ export default defineComponent({
...
+				this.createModel(outputModel, true);
+			} catch (e: any) {
+				console.error(e.message);
+			}
+		},
+		// Used to create sample models for stratifying tests
+		// Will not be requried in the long run as we will be moving to storing these in DB
+		async createSampleModels() {
+			// TODO: Add Petri Net type to this when merged with other PR
+			const SIRDModel = {
```
bencodrington, line 742: These can be typed correctly now that Shawn's PR is merged.

```suggestion
			const SIRDModel: PetriNet = {
```

packages/client/hmi-client/src/views/TA2Playground.vue
```diff
@@ -73,6 +74,10 @@ interface NodeData {
 interface EdgeData {
 	val: number;
 }
+enum NodeType {
+	species = 'S',
```
bencodrington, line 78: Nitpick: it's a convention to capitalize enum constants, like
```suggestion
	Species = 'S',
```

packages/client/hmi-client/src/views/TA2Playground.vue
```diff
@@ -121,8 +126,10 @@ class SampleRenderer extends graphScaffolder.BasicRenderer<NodeData, EdgeData> {
 	}
 
 	renderNodes(selection: D3SelectionINode<NodeData>) {
-		const species = selection.filter((d) => d.data.type === 'species');
-		const transitions = selection.filter((d) => d.data.type === 'transition');
+		const species = selection.filter((d) => d.data.type === 'species' || d.data.type === 'S');
```
bencodrington, line 129: We should use `NodeType.Species` here instead of the hardcoded string `'S'`.

When would we be using 'species' instead of 'S'? Can we consolidate these?

packages/client/hmi-client/src/views/TA2Playground.vue
```diff
@@ -536,9 +797,29 @@ export default defineComponent({
 		<p>A playground for testing TA2 API integrations.</p>
 		<button type="button" @click="addPlace">Add place</button>
 		<button type="button" @click="addTransition">Add transition</button>
+		<button type="button" @click="createSampleModels">Create Models</button>
 		&nbsp;
 		<button type="button" @click="LotkaVolterra">LotkaVolterra</button>
 		<button type="button" @click="simulate">Simulate</button>
+		<form>
+			<label for="loadModel">
+				<input v-model="loadModelID" type="text" id="loadModelID" placeholder="Model ID" />
```
bencodrington, line 806: We should remove these `id="XYZ"` attributes now that we're accessing the input values through v-model directly.

bencodrington (commented):
This is almost there, just left a couple of notes to clean up.

packages/client/hmi-client/src/views/TA2Playground.vue
```diff
@@ -527,6 +639,155 @@ export default defineComponent({
 					.style('stroke', null)
 					.style('fill', 'blue');
 			}
+		},
+		// Pulls model ID from form and sends model to createModel function for the actual work
+		async drawModel() {
+			console.log(this.loadModelID);
```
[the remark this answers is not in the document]
Tom-Szendrey: Without this log I will not be able to create models and then use those IDs to stratify are you sure?

packages/client/hmi-client/src/views/TA2Playground.vue
Tom-Szendrey: 'species' is used in some of the LotkaVolterra example - so is 'S'
Will definitely update to use the enum for this

packages/client/hmi-client/src/views/TA2Playground.vue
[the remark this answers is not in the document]
bencodrington: There's another one on line 662:
```
console.log(`Model ID: ${modelId}`); // currently required for testing
```
I thought that's what that one was?

packages/client/hmi-client/src/views/TA2Playground.vue
bencodrington: Sounds good. Hopefully we can consolidate going forward.

bencodrington (approved):

PR #182, opened 2022-11-02
feat: petrinet validator utility function

### LINKED ISSUE #122: [FEAT]: Add PetriNet validator
- PetriNet follows a strict set of representation rules.
- Implement the rules into the renderer or as a _utils_ to control user inputs.
- Bonus: make an agnostic validator that accepts a config file as a set of rules for different model framework (i.e. `petrinet.json`, `ode.json`, `mira.json`, etc.)

**Additional context**

shawnyama:
# Description

Added a utility function that checks if the petrinet in the TA2 playground is valid. This works by checking the node IDs linked by `I` and `O` in the petrinet object. ~The Lotka Volterra example will be recognized as valid as long as you make the function accept unbounded petrinets. For unbounded petrinets we should probably add a toggle to the playground for that in the case we want to accept those. You'd just have to put that toggle value as a parameter in the validator function.~

There are 4 checks for a valid petrinet that I made:
- At least one edge
- Every node should be at least either a source or a target
- ~In the case the petrinet is bounded (it is by default) every transition must be recognized as a source and a target~
- Make sure there is only one petrinet body

Cancelled check: It's a bipartite graph (place -> transition -> place...)
This will be handled another way since false edges aren't recorded in the petrinet object

Resolves #122 

## Type of change

# How Has This Been Tested?

A new unit test has been written called `petrinet-validator.spec.ts`. This tests various valid and invalid petrinets. I made some examples in the playground and copied the data representation of those to be tested. The petrinets I made for testing test every requirement needed for validity (list above).

- [ ] Test A
- [ ] Test B

# Checklist:


packages/client/webapp/src/views/TA2Playground.vue
```diff
@@ -323,6 +324,7 @@ export default defineComponent({
 				method: 'GET'
 			});
 			const output = await resp.json();
+			console.log(petrinetValidator(g) ? 'Valid Petrinet' : 'Invalid Petrinet');
```
YohannParis, line 327: Maybe make it part of the JSON output by adding a boolean attribute `isValidPetriNet` or something equivalient?

packages/client/webapp/src/utils/petrinet-validator.ts
```diff
@@ -0,0 +1,121 @@
...
+export const petrinetValidator = (graph: IGraph, isBoundedPetrinet: boolean = true): boolean => {
+	const { nodes, edges }: { nodes: INode[]; edges: IEdge[] } = graph;
+
+	// console.log(graph);
+	// console.table(nodes);
+	// console.table(edges);
+
+	/* ----- 1. Requires at least one edge ----- */
+	if (edges.length < 1) {
+		console.log('#1');
```
YohannParis, line 24: Maybe, instead of doing a `console.log()` you can return a message explaining why it is not valid.

mwdchang:
Can you briefly explain what you mean by a bounded and unbounded petri-net here?

packages/client/webapp/src/utils/petrinet-validator.ts
```diff
@@ -0,0 +1,92 @@
+export interface Petrinet {
```
bencodrington, line 1: Since "Petri net" is two words, this should be `PetriNet` and the other uses of the term in the other functions and variables in this file.
This seems nitpicky but it'll be much easier to be consistent later and avoid typos when they can't be automatically caught if we use the established spelling.

packages/client/webapp/src/utils/petrinet-validator.ts
bencodrington: Agreed, wherever this validator is called will need to be able to access the reason why it failed to help gracefully handle the error.

bencodrington (commented):
Very nice! Thorough and readable.

A couple of minor cleanup notes remaining but once those are addressed this will be good to merge.

packages/client/hmi-client/src/utils/petri-net-validator.ts
```diff
@@ -22,35 +22,30 @@ interface Output {
...
-	if (!checkIfSourceOrTarget(linkedTransitionIDs, T.length + 1)) return false;
 	const linkedStateIDs: number[] = [
 		...new Set([...I.map((input) => input.is), ...O.map((output) => output.os)])
 	];
-	if (!checkIfSourceOrTarget(linkedStateIDs, S.length + 1)) return false;
+	if (
+		!checkIfSourceOrTarget(linkedTransitionIDs, T.length + 1) ||
+		!checkIfSourceOrTarget(linkedStateIDs, S.length + 1)
+	)
+		return 'Invalid petri net: #2. Every transition node should be at least either a source or a target';
```
bencodrington, line 48: I think it'll be a good idea to remove these number codes from the return strings, `#2`, `#3`, etc.

If they're to be displayed to the user, the number will be meaningless and perhaps confusing.
If they're for the programmer, they can just Cmd+F the error message.
And any automatic conditionals we do that rely on this validator don't care about the actual content of the string.

packages/client/hmi-client/src/utils/petri-net-validator.ts
```diff
@@ -82,11 +77,9 @@ export const petrinetValidator = (petrinet: Petrinet): boolean => {
...
-		if (statesToMerge.length === potentialConnections.length) {
-			console.log('#5');
-			return false;
-		}
+		if (statesToMerge.length === potentialConnections.length)
+			return 'Invalid petri net: #3. There are multiple petri net bodies';
 	} while (potentialConnections.length > 0);
 
-	return true; // All checks have been successfully passed
+	return 'Valid petri net'; // All checks have been successfully passed
```
bencodrington, line 84: I think it'll be a good idea to more easily programmatically distinguish successful output from a failure. As it stands currently, the place where we call this validator function would need to know about this magic `'Valid petri net'` string and get the capitalization and whitespace exactly correct to be able to check if the result was successful.

Some options:
- Return an array of errors. If the array is empty, there were no errors so it passed. This might be overkill and confusing to the developer since we're exiting early after one error.
- Return `true` if the petri net is valid. You could make the return type `string | true` and leave a comment to explain that string is for errors. This feels messy to me but would work fine.
- Return an empty string `''` if the petri net is valid. This is also a magic string but would be easier to check for in the place that calls it.

I'm leaning toward the middle option but go with whatever you think is best.

Either way, adding a brief comment to the function header to explain the significance of each return value type is a good idea.

bencodrington (commented):
A couple more minor tweaks about the return value of the validator function, then we can finally put this to rest 😄

bencodrington (approved):
Very nice! Let's put this bad boy to bed.

PR #204, opened 2022-11-08
chore: add missing label for feature

Nithos:
The feature template was using an old `label` type that no longer exists. Updating it to use `feature` now.


dgauldie (approved):
LGTM

PR #218, opened 2022-11-10
ci: switch to use triggers rather than wait on workflow completion

Nithos:
# Description

The workflow_run call is less than desired. Switching to manually trigger tests on publish instead and not have a long dependency chain.

## Type of change

# How Has This Been Tested?

Increased timeout to 10 min on main build phase

# Checklist:

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -6,10 +6,10 @@
 name: Playwright Tests
 on:
   workflow_call:
-  workflow_run:
-    workflows: ['Lint & Test']
-    types:
-      - completed
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['main']
 jobs:
   e2e-tests:
     timeout-minutes: 30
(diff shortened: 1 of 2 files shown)


Nithos:
@dgauldie @YohannParis Okay, should be working now, when this gets merged to 'main' it should automatically build and publish the `latest` tag version.

NOTE: I had to destroy the old `hmi-server` package as it would not let me upload, think it was something to do with permissions as it was manually pushed to. Thus currently there is a version `0.0.0` and `0.0.0-native` version of the server and a `0.0.0` for the client. Users should be made aware of the change so they can update (hopefully its only a few people).

FUTURE TODO: go back and remove the `webapp` and old `theme` package that should no longer be used

TODO: @dgauldie please test and verify the native image to see if we need to make adjustments to the bake file for that

YohannParis (approved):
LGTM
To bad, the idea of workflows sounds nice; we will have to wait for a better support from GitHub.

Nithos:
> 

Workflows are still present and they still have dependencies, its just that they are sometimes doing more work. Its more of a pull rather than push now.

PR #223, opened 2022-11-11
fix: fix bad variables in merge

mwdchang:
## Summary
Hotfix for bad merge


bencodrington (approved):
That did the trick. Thanks!

PR #238, opened 2022-11-16
fix: linter issue

mwdchang:
# Description
Various linting related issues

### DIFF
--- a/.github/workflows/lint-client.yml
+++ b/.github/workflows/lint-client.yml
@@ -42,6 +42,4 @@ jobs:
         run: yarn prettier:ci
 
       - name: Run Typechecks
-        run: |
-          # yarn workspace graph-scaffolder tsc --build
-          yarn typecheck
+        run: yarn typecheck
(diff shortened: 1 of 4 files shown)


YohannParis (approved):
LGTM

.github/workflows/lint-client.yml
```diff
@@ -42,4 +42,6 @@ jobs:
         run: yarn prettier:ci
 
       - name: Run Typechecks
-        run: yarn typecheck
+        run: |
+          yarn workspace graph-scaffolder tsc --build
```
Nithos, line 46: @dgauldie issue is that the types need to be built otherwise it does not properly reference them... not sure what the best option for this is really as the above is not really maintainable

Nithos: moved it to the script in package.json makes more sense there

YohannParis (approved):
Moving it to the script was a good idea, the more code agnostic the GitHub actions are the better.

PR #243, opened 2022-11-18
feat: add axios as dependency for API requests

bencodrington:
# Description

Adds axios dependency for API calls to hmi-server.

## Type of change

- [X] No new feature, just adding a dependency

# How Has This Been Tested?

```
await API.get('models');
```
returns the list of models as expected.

# Checklist:

- [X] I have made corresponding changes to the documentation


packages/client/hmi-client/src/api/api.ts
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios';
+
+const API = axios.create({
+	baseURL: '/api'
+});
```
YohannParis, line 5: Could we add the Keycloack Token as well?

packages/client/hmi-client/README.md
```diff
@@ -9,6 +9,19 @@
...
+## HTTP Requests
+
+You can import and use the singleton [axios](https://axios-http.com/) instance to make requests to the `hmi-server` package like so:
+
+```ts
+import API from '@/api/api';
+
+// ... in an async function:
+const response = await API.get('models');
+```
```
YohannParis, line 21: Instead of importing it on every page we need, can we make it a global in our single Vue app instance?

YohannParis (approved):
LGTM

packages/client/hmi-client/src/api/api.ts
bencodrington: I tried to do this but I think it might require some very roundabout code.

We want to create a new axios instance when we create the Vue app instance which means getting `auth.token` then. But `auth.token` can change or expire which makes me think we'd actually want to get `auth.token` from `useAuthStore` each time we make a request.

That would involve wrapping/replacing axios's request functions... which seems like a much more complicated of just doing something like what we currently have in `uncloak()`. Am I missing something?

packages/client/hmi-client/README.md
bencodrington: I'm concerned that we'll regret getting in the habit of adding global variables. This way we can statically detect which files are making API calls, and any developer that sees an API call being made can easily trace back to where the axios instance was created, rather than just having to know that there's a magic global keyword floating around that you can use.

If you feel strongly about this Im happy to add the global, just wanted to throw in my 2 cents. Let me know what you think.

packages/client/hmi-client/src/api/api.ts
YohannParis: Good questions!
@bigglesandginger do you have any insights?

Nithos: How did this work without the token?
I'm still confused as to why we can't just use the `uncloak` util, or at least use one or the other (I don't think we need both).

If using axios have you tried adding the
`headers: {'Authorization': 'Bearer '+token}` where `token` is retrieved from the store? I doubt it will dynamically evaluate it then but worth a try.

The other option (although probably not the best) is to use an interceptor and add it then.

Again I may be missing something key here though.

bencodrington: I'd also love to know how this worked without the token 😆 I just know that it did. Happy to jump on a call to walk through it together if you'd be able to help me understand that.

I'm also not clear on why we need axios for our current usage, it does feel like overkill at the moment. @YohannParis mentioned that @dgauldie and @mwdchang both wanted to add it so maybe they can chime in.

I tried exactly what you suggested and got errors where the store hadn't been created yet. I figured it wasn't worth spending a bunch of time troubleshooting given that we're going to have the dynamic evaluation issue anyway.

If I understand what you're suggesting with the "interceptor" idea, I had the same thought. At that point seems like a lot of potentially confusing code to do what we're already doing with `uncloak`.

mwdchang: I haven't looked at Ahmed PR so I am not up-to-date with the cloak/uncloak stuff.

Nithos: its just a wrapper around a `fetch` call that adds the token to the request nothing complicated there.
looks like he modified it slightly to allow for more than just GET requests

mwdchang: Yeah I took a look and got the context, so basically if we want axios, `uncloak` should be like a helper wrapper  function that uses axios under the hood.

bencodrington: @mwdchang do you see any benefit to using axios under the hood of `uncloak`? It's working fine now without.

mwdchang: I think we have agreed to leverage axios, so I will close this thread.

packages/client/hmi-client/src/api/api.ts
```diff
@@ -0,0 +1,22 @@
+import axios from 'axios';
+import useAuthStore from '../stores/auth';
+
+const API = axios.create({
```
Nithos, line 4: are all endpoints going to be `/api`? if not is the expectations that we will have multiple axios instances present?

packages/client/hmi-client/src/api/api.ts
```diff
@@ -0,0 +1,22 @@
...
+// - https://medium.com/swlh/handling-access-and-refresh-tokens-using-axios-interceptors-3970b601a5da
+API.interceptors.request.use((config) => {
+	const auth = useAuthStore();
+
+	if (config.headers) {
+		config.headers.Authorization = `Bearer ${auth.token}`;
+	} else {
+		config.headers = { Authorization: `Bearer ${auth.token}` };
+	}
+	return config;
```
Nithos, line 19: Do we need an error handler here? or passing that along to whoever uses it?

packages/client/hmi-client/src/api/api.ts
```diff
@@ -0,0 +1,22 @@
...
+import useAuthStore from '../stores/auth';
+
+const API = axios.create({
+	baseURL: '/api'
+});
+
+// Hook in bearer tokens
+// See
+// - https://medium.com/swlh/handling-access-and-refresh-tokens-using-axios-interceptors-3970b601a5da
+API.interceptors.request.use((config) => {
```
Nithos, line 11: do we foresee any need to eject this interceptor in the future? if so then this should be saved in a const that is added later and can be removed as needed

packages/client/hmi-client/src/api/api.ts
```diff
@@ -0,0 +1,22 @@
+import axios from 'axios';
```
Nithos, line 1: If this is the way forward, can probably remove the `uncloak` file if not used for anything

mwdchang: There are a few parallel threads along the same veins, let's get the foundations in first, before cleaning up other spots, otherwise this PR is just going to drag on and on.

packages/client/hmi-client/src/api/api.ts
mwdchang: I think as far as the client is concerned, it will use the `/api` prefix. We can have debugging/administrative routes with other names that are out of scope of the client application.

packages/client/hmi-client/src/api/api.ts
mwdchang: I don't know what you mean by `eject this interceptor`

Nithos: sorry `axios` terminology, `eject` simply means remove or unsubscribe
ie.
```
const myInterceptor = axios.interceptors.request.use(function () {/*...*/});
axios.interceptors.request.eject(myInterceptor);
```

I don't think we will need to but just covering the bases.

mwdchang: Got it! Yeah I don't think we will need it, but will keep it mind, thanks!

packages/client/hmi-client/src/api/api.ts
mwdchang: I added a stubbed error handler so people know where it is, we can extend it if necessary

mwdchang (approved):

Issue #59, opened 2022-10-05
[FEAT]: First draft of the responsive matrix cells


dgauldie:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #246, opened 2022-11-18
feat: add responsive matrix core functionality (#59)

mj3cheun:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Resolves #(issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


packages/client/hmi-client/src/components/responsive-matrix/pixi-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+				'CustomBufferResource width, height, internalFormat, format, or type invalid'
+			);
+		}
+
+		super(width, height);
+
+		(this as any).data = source;
+		(this as any).internalFormat = internalFormat;
+		(this as any).format = format;
+		(this as any).type = type;
```
mwdchang, line 19: If you are adding/extending class attributes to `Resource`, maybe we can just desclare these class variables so you don't need to do `this as any` all over the place?

e.g.

```
export class X extends Resource {
  internalFormat: string
  type: string 
  ...

  constructor() { ... }
}
```

packages/client/hmi-client/src/components/responsive-matrix/matrix.fs.glsl.ts
```diff
@@ -0,0 +1,30 @@
...
+	return texelFetch(tex, ivec2(col, row), 0);
+}
+
+void main() {
+	float microColLen = float(textureSize(uMicroCol, 0).x);
+	float microRowLen = float(textureSize(uMicroRow, 0).x);
+	uint colIndex = texelFetch(uMicroCol, ivec2(int(vUvs.x * microColLen), 0), 0).r;
+	uint rowIndex = texelFetch(uMicroRow, ivec2(int(vUvs.y * microRowLen), 0), 0).r;
+
+	fragColor = texelFetch(uColor, ivec2(colIndex, rowIndex), 0);
```
mwdchang, line 29: It feels like you are using sampler2D as if it is a sampler1D - is sampler1D not supported in webGL?

Also it seems like `getColorFromRowCol` is not being used - is that in your later PRs or can we remove?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vs.glsl.ts
```diff
@@ -0,0 +1,18 @@
...
+in vec2 aUvs;
+
+uniform mat3 translationMatrix;
+uniform mat3 projectionMatrix;
+
+out vec2 vUvs;
+
+void main() {
+	vUvs = aUvs;
+	gl_Position = vec4((projectionMatrix * translationMatrix * vec3(aVertexPosition, 1.0)).xy, 0.0, 1.0);
```
mwdchang, line 17: This is just a vanilla M*V*P right?

packages/client/hmi-client/src/components/responsive-matrix/matrix.fs.glsl.ts
mwdchang: A bit of comment describing what the main method would be good to have.

packages/client/hmi-client/src/views/ResponsivePlayground.vue
```diff
@@ -0,0 +1,62 @@
...
+			]
+		};
+	},
+
+	methods: {
+		fillColorFn(datum: any, paramsMin: any, paramsMax: any): String {
+			// get value as fraction of min clamped to 0
+			const x = Math.max(datum.I / paramsMax.I, 0);
+			// apply gamma
+			// x **= 0.1;
```
YohannParis, line 46: Explain what this `gamma` does and why it is here.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
+<template>
+	<div class="matrix-container" ref="matrix_container">
```
mwdchang, line 2: nitpick: prefer `matrixContainer`

packages/client/hmi-client/src/components/responsive-matrix/matrix.vs.glsl.ts
mj3cheun: yup, the two matrices come from pixi

packages/client/hmi-client/src/views/ResponsivePlayground.vue
```diff
@@ -0,0 +1,62 @@
...
+
+			return `#${Math.round(x * 255)
+				.toString(16)
+				.padStart(2, '0')}0000`;
```
YohannParis, line 50:
```suggestion
                         // Return a gradient of Red with a #hex colour
			return `#${Math.round(x * 255)
				.toString(16)
				.padStart(2, '0')}0000`;
```

packages/client/hmi-client/src/views/ResponsivePlayground.vue
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+};
+</script>
+
+<style scoped>
+.responsive-matrix-demo {
```
YohannParis, line 57:
```suggestion
section {
```

You don't need a class if you are scoped.

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
```diff
@@ -0,0 +1,50 @@
...
+	BAR = 'ResponsiveCellBarContainer',
+	LINE = 'ResponsiveCellLineContainer'
+}
+
+export enum CellStatus {
+	NONE,
+	SELECTED
+}
+
+export type Uniforms = {
```
YohannParis, line 22: Could you add a comment explaining (or a link that explains) what are `Uniforms`.

packages/client/hmi-client/src/components/responsive-matrix/matrix.fs.glsl.ts
mj3cheun: sampler1d would be ideal for this implementation but in order to get around the texture size limit (around 8192) which is much too low a limit for microrow/col, i planned to convert it to using 2d texture.

`getColorFromRowCol` is a function i stole from grafer that should help with this, but again, not implemented yet.

i should probably do a commit implementing 2d tex

packages/client/hmi-client/src/components/responsive-matrix/pixi-utils.ts
mj3cheun: sounds good will do

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: for ref or class or both?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+import matrixVS from './matrix.vs.glsl';
+import matrixFS from './matrix.fs.glsl';
+// import matrixGridFS from './matrix-grid.fs.glsl';
+
+export default {
+	// ---------------------------------------------------------------------------- //
+	// components                                                                   //
+	// ---------------------------------------------------------------------------- //
+
+	// ///////////////////////////////////////////////////////////////////////////////
```
YohannParis, line 61: Do we need those massive lines, when the comment block are big enough?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+			screenHeight: 0,
+			screenWidth: 0,
+
+			selectedCellList: [] as SelectedCell[], // e.g. [[startrow, startcol, endrow, endcol]]
+			selectedRows: [] as CellStatus[], // e.g. [0, 0, 1, 0]
+			selectedCols: [] as CellStatus[], // e.g. [0, 0, 1, 0]
+			selectedCell: Array(4) as SelectedCell, // e.g. [startrow, startcol, endrow, endcol]
+
+			update: 0,
+			move: 0,
```
mwdchang, line 162: Not too clear what move/update are doing, can you add a comment to clarify?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+		cellLabelCol: {
+			type: Array as PropType<number[] | string[]>,
+			default() {
+				return [];
+			}
+		},
+		fillColorFn: {
+			type: Function,
+			default() {
+				return '#000000';
```
YohannParis, line 98: We are using [Uncharted Design Tokens](https://github.com/unchartedsoftware/design-tokens/blob/main/build/js/es6/uncharted-tokens.js) they are already installed, so you can just import the colours.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: yup, they exist basically to trigger updates in children components when either is required, this should probably have been commented out here and uncommented in the drilldown pr. will add comment

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+		window.removeEventListener('keyup', this.handleKey);
+	},
+
+	// ---------------------------------------------------------------------------- //
+	// methods                                                                      //
+	// ---------------------------------------------------------------------------- //
+
+	// ///////////////////////////////////////////////////////////////////////////////
+
+	methods: {
```
YohannParis, line 469: All those methods could be put in a `services/responsive-matrix.ts` file

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+				:parametersMin="dataParametersMin"
+				:parametersMax="dataParametersMax"
+				:colorFn="getSelectedGraphColorFn(selectedCell)"
+				@click="selectedCellClick(idx)"
+			/> -->
+		</div>
+	</div>
+</template>
+
+<script lang="ts">
```
YohannParis, line 26: Can you add a JSDoc comment to this component explaining what it does and the utilities its importing.

```suggestion
<script lang="ts">
/**
 * Matrix
 *
 * Create a matrix/grid to display results in cells, ....
 * Rendered with WebGL for performance ...
 * Uses: 
 *  - Pixi: ..... (url)
 *  - Choma: .... (url)
 **/
```

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+			uNumRow: this.numRows,
+			uNumCol: this.numCols,
+			uMicroRow: this.uniforms.uMicroRow,
+			uMicroCol: this.uniforms.uMicroCol,
+
+			// cell element color data
+			uColor: Texture.fromBuffer(color, this.numCols, this.numRows),
+
+			uTime: 0
+		};
```
mwdchang, line 390: Your uniforms don't seem to match the shader, e.g. what is `uTime`, `uGridRowDisplayLim` ...etc used for?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+	CellType,
+	Uniforms
+} from '@/types/ResponsiveMatrix';
+import { uint32ArrayToRedIntTex } from './pixi-utils';
+
+// import ResponsiveCellBarContainer from './cell-bar-container.vue';
+// import ResponsiveCellLineContainer from './cell-line-container.vue';
+
+import matrixVS from './matrix.vs.glsl';
+import matrixFS from './matrix.fs.glsl';
```
YohannParis, line 53: Add a comment explaining what those file are doing.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: some of the uniforms are only applicable for the grid, like in the case of `uGridRowDisplayLim`.

`uTime` isnt being used i should remove it.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+		this.worldHeight = vertexPosition[5] - vertexPosition[3];
+		this.worldWidth = vertexPosition[2] - vertexPosition[0];
+		(this as any).$viewport = new Viewport({
+			screenWidth: this.screenWidth,
+			screenHeight: this.screenHeight,
+			worldWidth: this.worldWidth,
+			worldHeight: this.worldHeight,
+			divWheel: matrix as HTMLElement,
+			interaction: app.renderer.plugins.interaction // the interaction module is important for wheel to work properly when renderer.view is placed or scaled
+		});
```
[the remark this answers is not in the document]
mj3cheun: i dont like this but i did this specifically to avoid reactivity (which breaks everything). lemme know if there are any other solutions you can think of im open to suggestions.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+	// unmounted                                                                    //
+	// ---------------------------------------------------------------------------- //
+
+	// ///////////////////////////////////////////////////////////////////////////////
+
+	unmounted() {
+		this.resizeObserver.disconnect();
+		window.removeEventListener('keydown', this.handleKey);
+		window.removeEventListener('keyup', this.handleKey);
+	},
```
mwdchang, line 461: Are there any PIXI-related cleanups we need to do here?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mwdchang: Just for the `ref`.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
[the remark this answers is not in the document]
mwdchang: Hmmm, so your code is kind of in a quirky phase here because it is written in a Vue-2 style, but the rest of the application is using Vue-3.

I think you can declare non-reactive variables using setup()
```
setup() {
  const $viewport = ...

  return { $viewport }
}
```

Going to ping @bencodrington here because he knows a bit more about this than I ...

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
mwdchang: Well if you want to go down that path, there are:
- attribute
- varying
- uniform
- ... maybe something else ???

They are basically all variables that are passed to the shader, with slightly different behaviours and context.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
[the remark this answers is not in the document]
bencodrington: Yep, @mwdchang's suggestion above checks out 👍 variables returned from `setup()` are non-reactive unless you explicitly specify they should be using `ref()`, `reactive()`, or `computed()`.

Assuming you want all this to still run after the component has mounted, you'd do something like:
```ts
setup() {
  let $viewPort: ViewPort | null = null;
  onMounted(() => {
    $viewPort = new ViewPort(...)
  }
  return { $viewPort };
}
```

The [docs page](https://vuejs.org/api/composition-api-setup.html#setup-context) is here, let me know if there's anything else I can clear up for ya.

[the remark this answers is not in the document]
mj3cheun: thanks! will do it this way

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
mj3cheun: do we want to do this, these are all universally understood gl terms, i am not sure we want to embed a gl tutorial in the responsive matrix code. if we still want to do something like this i am more open to putting links in.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: similar to another comment above, vertex and fragment shaders are well understood components of gl, not sure we need a gl tutorial imbedded in the code

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: i personally think its easier to navigate this big file with these things in but i can remove them if people dont like them

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+		},
+
+		numSelectedCols(): number {
+			return this.selectedCols.reduce(
+				(acc: number, val: CellStatus) => acc + Number(val !== CellStatus.NONE),
+				0
+			);
+		},
+
+		microRowSettings(): number[] {
```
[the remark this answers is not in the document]
mj3cheun: i can add a quick blurb about the micro grid/col thing somewhere, either here or down in the methods

packages/client/hmi-client/src/views/ResponsivePlayground.vue
mj3cheun: will put in a link

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
mwdchang: I tend to agree with Manfred here, let's leave it alone and maybe add a note to the technical-specification that we use WebGL and maybe link people here (https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API).

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+				0
+			);
+		},
+
+		numSelectedCols(): number {
+			return this.selectedCols.reduce(
+				(acc: number, val: CellStatus) => acc + Number(val !== CellStatus.NONE),
+				0
+			);
+		},
```
mwdchang, line 195: I think it is more performant, and maybe more readable if you don't use accumulator and the hardcast to Number...maybe something like.

```
return this.selectedRows.filter(d => d !== CellStatus.NONE).length
```

Didn't try on browser, but the reduce is slower running in node

```
const list = [];
for (let i = 0; i< 300; i++) {
  if (Math.random() < 0.5) {
    list.push(0);
  } else {
    list.push(1);
  }
}


function getOnes() {
  return list.reduce((acc, val) => acc + Number(val !== 0), 0)
}

function getOnes_2() {
  return list.filter(d => d !== 0).length;
}


let start = 0;
let end = 0;

start = performance.now();
for (let i = 0; i < 100; i++) {
  getOnes();
}
end = performance.now();
console.log(end - start);

start = performance.now();
for (let i = 0; i < 100; i++) {
  getOnes_2();
}
end = performance.now();
console.log(end - start);
```

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -0,0 +1,902 @@
...
+					1,
+					0,
+					1,
+					1,
+					0,
+					1
+				], // u, v
+				2
+			) // the size of the attribute
+			.addIndex([0, 1, 2, 0, 2, 3]);
```
mwdchang, line 356: Out of curiosity, what does addIndex do here?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
YohannParis: Perfect, a simple comment saying `// Vertex and Fragment Shaders` will be perfect!

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
YohannParis: I understand the need for the big one:
```js
	// ---------------------------------------------------------------------------- //
        // components                                                                   //
	// ---------------------------------------------------------------------------- //
```
But not those ones:
```js
        // ///////////////////////////////////////////////////////////////////////////////
```

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
YohannParis: Okay, we need to add this in the README.
I don't like reading code, and I have to be the one figuring out what things are. Either a comment block at the top should announce that this is using WebGL, and refer people to that.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: hmm i just pasted your benchmark into the browser console a few times and the accumulator is slightly (0.2ms) faster the majority of the time. that performance difference is negligable tho (runtime roughly 1ms for both) and i do like the readability of the filter better. we will use that and change it back if it ever becomes an issue (probably not).

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: its specifying the points making up the 2 triangles forming the quad. so what this is saying here is the first triangle are attributes of index 0, 1, 2, and the second triangle is formed by attributes of index 0, 2, 3

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mwdchang: Ah right, probably because it creates a new array - I have comparable result as yours when I increase the number of iterations.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mwdchang: I see, I thought it would've been a triangle-strip or something, e.g. not size=6.

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mwdchang: Let's do these as a separate round after things have settled in a bit and has demonstrated the utility

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mwdchang: I don't think you can use CSS-variables in a pixi/shader context.

mwdchang (approved):

PR #263, opened 2022-11-23
build: switch to using multi-stage docker builds

Nithos:
# Description

This is an attempt to switch the Dockerfiles to be multi-staged. The reason for this is that during the image build it should handle the compiling of the code internally. This way there is no need for a separate script to be run (in our case a Makefile) to build the source, move it to a directory so that its accessible by the Dockerfile.

Drawbacks:
 - this does complicate the build a bit more
 - some shinanigans needed with the dockerignore file so that we can still pass around what we need (can there be multiple ignore files?)
 - larger context being moved around since we are moving the whole packages section (optimizations can be made here perhaps, also if multiple ignores are possible this could help)

Benefits:
- local and CI builds are self contained, simply call the bake file
- can reuse the reusable bake ci workflow directly now, should now be consistent across repos

## Type of change

# How Has This Been Tested?

Local builds generated, used to standup the whole K8 stack. Seems to work correctly. 
Would want a second opinion test on another machine for those willing.
CI trigger tested

# Checklist:


packages/services/mock-data-service/docker/jvm/Dockerfile.jvm
```diff
@@ -1,11 +1,7 @@
...
 #
-# Before building the container image run:
-#
-# ./gradlew build
-#
-# Then, build the image with:
+# Build the image with:
 #
 # docker build -f src/main/docker/Dockerfile.jvm -t quarkus/server-scaffold-jvm .
 #
```
Nithos, line 7: @dgauldie All this below, is any of that useful? Guessing this is just from a quarkus template just wondering if we can clean it up

.github/workflows/publish.yaml
```diff
@@ -121,36 +44,25 @@ jobs:
...
-            *.cache-from=type=gha,scope=build-${{ matrix.target }}
-            *.cache-to=type=gha,scope=build-${{ matrix.target }},mode=max
+  # Build and Publish all targets associated with specified group
+  bake:
+    needs:
+      - lint
+      - lint-client
+      - test-client
+      - tag-generator
+    uses: darpa-askem/.github/.github/workflows/bake-publish.yml@main
```
YohannParis, line 59: Do we need to save the workflow under `/.github` folder, when they are already in the `.github` repo?

packages/services/hmi-server/docker/native/Dockerfile.native
```diff
@@ -13,15 +9,30 @@
...
+
+RUN ./gradlew build -Dquarkus.package.type=native
+
+## Stage 2 : create the docker final image
+FROM quay.io/quarkus/quarkus-micro-image:2.0
+WORKDIR /work/
+COPY --from=builder /code/build/*-runner /work/application
+RUN chmod 775 /work
+EXPOSE 8080
 CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]
```
YohannParis, line 38: Is this because Quarkus by default runs on port `8080`? then we port the all income connection from `localhost` to it?

packages/services/mock-data-service/docker/jvm/Dockerfile.jvm
YohannParis: I was wondering the same thing about all those comments.
Also @Nithos I don't think we need to bother with mock-data-service anymore, so I'm not sure if it's worth keeping all of those scripts.

packages/services/hmi-server/src/main/resources/application.properties
```diff
@@ -29,5 +29,5 @@ xdd-extraction-service/mp-rest/url=https://xdddev.chtc.io/askem
 ########################################################################################################################
 # Native container configurations
 ########################################################################################################################
-quarkus.native.container-build=true
+quarkus.native.container-build=false
 quarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:22.3.0-java17
```
YohannParis, line 33: Why this needs to be set here, but we do not use it in the `Dockerfile`.
Isn't it confusing to have this builder reference in two places?

packages/client/hmi-client/docker/Dockerfile
```diff
@@ -1,14 +1,29 @@
+## Stage 1 : build with gradle builder image with native capabilities
```
YohannParis, line 1: Comment is mentionning `gradle` but it's using node as builder

packages/client/hmi-client/docker/Dockerfile
```diff
@@ -1,14 +1,29 @@
...
+WORKDIR /code
+COPY package.json /code
+COPY yarn.lock /code
+COPY packages/client/graph-scaffolder/package.json /code/packages/client/graph-scaffolder/
+COPY packages/client/hmi-client/package.json /code/packages/client/hmi-client/
+RUN yarn install
+
+COPY . /code
+RUN yarn workspace graph-scaffolder tsc --build
+RUN yarn workspace hmi-client build
```
YohannParis, line 12: Isn't this step done by `makefile build-hmi-client`?

YohannParis (dismissed):

.github/workflows/publish.yaml
Nithos: Unfortunately yes, we tried without it and it did not work.

packages/services/hmi-server/src/main/resources/application.properties
Nithos: This can be removed completely. Left it in for now just in case. @dgauldie thoughts

packages/services/hmi-server/docker/native/Dockerfile.native
Nithos: @dgauldie ?

packages/client/hmi-client/docker/Dockerfile
Nithos: oops, fixed

packages/client/hmi-client/docker/Dockerfile
Nithos: it is/was, with this change Make file can be removed as this is now fully self contained
Makefile is left for convenience for local testing at the moment

.dockerignore
```diff
@@ -1,7 +1,13 @@
 *
-!build/*-runner
-!build/*-runner.jar
-!build/lib/*
-!build/quarkus-app/*
-
```
YohannParis, line 6: Is it because the build step is done in GitHub actions now?

packages/services/hmi-server/src/main/resources/application.properties
YohannParis: @dgauldie ????

YohannParis (approved):

packages/services/hmi-server/docker/native/Dockerfile.native
bigglesandginger: That should be an error, a hold over from "normal" Java deployments.  That `EXPOSE` should specify `3000` not 8080

packages/services/hmi-server/src/main/resources/application.properties
bigglesandginger: That is for when you run `quarkus` at the command-line at dev time.  When set to true it allows running Quarkus code without needing Java - useful for building native containers and the like, but can be false to have the host use the resident Java/GraalVM - at least that is my brief understanding from `https://quarkus.io/guides/building-native-image`

bigglesandginger (dismissed):

packages/services/hmi-server/src/main/resources/application.properties
Nithos: I believe this is only relevant when building images, but will need @dgauldie to confirrm.

.dockerignore
Nithos: no, b/c of multi stage builds we need to make sure that these are not ignored as they are needed for the secondary stage

Makefile
```diff
@@ -12,7 +12,9 @@ all:
 	@echo ""
 	@echo "Commands:"
 	@echo "  clean                    - remove bin, dist and node_modules directories"
-	@echo "  images                   - build docker images of all targets"
+	@echo "  build-all                - compile all sources"
```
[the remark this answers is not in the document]
Nithos: removed

packages/services/hmi-server/src/main/resources/application.properties
dgauldie: we can remove entirely

dgauldie (approved):
LGTM

PR #264, opened 2022-11-25
feat: 165 display selected model

### LINKED ISSUE #165: [FEAT]: Display user selected model within `Model Renderer` component
**Goals**
* Create a new component named `ModelRendered` to visualise models in the _Models_ view.
* Base code from the TA2 playground.

bencodrington:
# Description

image

Model renderer now displays the model and metadata for the model whose ID is specified in the URL.

I also removed the debugging pane that displayed the model JSON and made the graph pane bigger.

Resolves #165

## Type of change

# Testing instructions

1. Make sure gateway, hmi-server, hmi-client and data-service are running
2. Navigate to http://localhost:8078/app/#/projects/1/model/2 . You should see the graph and metadata for the model whose name starts with "Okuonghae2020"
3. Changing the `2` in the URL will display other models, e.g. `3` is for "Giordano2020"

# Checklist:

### DIFF
--- a/packages/client/hmi-client/src/views/Model.vue
+++ b/packages/client/hmi-client/src/views/Model.vue
@@ -1,7 +1,6 @@
 <script setup lang="ts">
 import { IGraph } from '@graph-scaffolder/index';
-import { onMounted, ref, watch } from 'vue';
-import { PetriNet } from '@/utils/petri-net-validator';
+import { ref, watch } from 'vue';
 import {
 	runDagreLayout,
 	D3SelectionINode,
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
LGTM

packages/client/hmi-client/src/views/Model.vue
```diff
@@ -69,55 +55,43 @@ class ModelPlanRenderer extends BaseComputionGraph<NodeData, EdgeData> {
...
 		model.value = result.data as Model;
 	},
 	{ immediate: true }
 );
+
+const graphElement = ref<HTMLDivElement | null>(null);
+// Render graph whenever a new model is fetched or whenever the HTML element
+//	that we render the graph to changes.
+watch([model, graphElement], async () => {
+	if (model.value === null || graphElement.value === null) return;
```
mwdchang, line 75: I find it a little weird to watch a `div` - can we just watch the model but retain the same conditional check?

bencodrington: This is to handle the case where model.value is set before the graphElement is created on the page.

In that case we get

1) model is set
2) conditional is true, return early
3) graph element is created
4) graph never gets rendered since we're not watching the graph element

The same thing happens if the graph element is rerendered for any reason.

Both of these possibilities are unlikely but I think it's nice to be extra robust.

Is there a downside I'm missing? If not I'd recommend we'd leave it.

bencodrington: Discussed externally.

PR #265, opened 2022-11-25
feat: header fix

YohannParis:
# Description

Change where and how we fetch the information about a project:
 - Created the `sevice/projects.ts` to handle API call and everything related to Project manipulation.
 - Now the call is made in `App.vue`, as this is only made once in the app.
 - The Project is passed as a prop to all views
 - All handled via the URL
 - Updated the Header to display the proper name
 - Make the sidebar closed by default

https://user-images.githubusercontent.com/636801/204052860-48e76b75-3d6a-4950-ae96-339c06b22332.mov

## Type of change

# Checklist:

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -7,12 +7,14 @@ import DataExplorer from '@/views/DataExplorer.vue';
 import Sidebar from '@/components/Sidebar.vue';
 import { Project } from '@/types/Project';
 import * as ProjectService from '@/services/project';
-
 import { useCurrentRouter } from './router/index';
 
+/**
+ * Router
+ */
 const route = useRoute();
 const router = useRouter();
-
+const goToHomepage = () => router.push('/');
 const { isCurrentRouteHome } = useCurrentRouter();
 const isSidebarVisible = computed(() => !isCurrentRouteHome.value);
 
@@ -50,10 +52,13 @@ watch(
 			// Set the Project to null,
 			// and send the user to the homepage.
 			project.value = null;
-			router.push('/');
+			goToHomepage();
 		} else {
 			const id = projectId as string;
 			project.value = await ProjectService.get(id);
+			if (!project.value) {
+				goToHomepage();
+			}
 		}
 	}
 );
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { computed, ref, watch } from 'vue';
-import { useRoute } from 'vue-router';
+import { useRoute, useRouter } from 'vue-router';
 import Header from '@/components/Header.vue';
 import Overlay from '@/components/Overlay.vue';
 import DataExplorer from '@/views/DataExplorer.vue';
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/App.vue
```diff
@@ -36,10 +63,14 @@ const dataExplorerActivated = ref(false);
 		@show-overlay="enableOverlay"
 		@hide-overlay="disableOverlay"
 	/>
-	<Header class="header" @show-data-explorer="dataExplorerActivated = true" />
+	<Header
+		class="header"
+		:projectName="project?.name"
```
bencodrington, line 68: You can take out this `?` since project will always be an instance of `Project`.

```suggestion
		:projectName="project.name"
```

Though, what do you think about using `null` when no project is selected rather than `ProjectService.empty`?
That way when components need to know if a project has been selected they can just check `project === null` rather than having to know the arbitrary ID of the empty project constant.

packages/client/hmi-client/src/components/Header.vue
```diff
@@ -15,7 +16,9 @@ const goToHomepage = () => {
 };
 const goToDataExplorer = () => emit('show-data-explorer');
 
-const projectName = 'Name of the project that can be long for clarity and precision';
+defineProps<{
+	projectName?: Project['name'];
```
bencodrington, line 20: Depending on your response to my other comment you may be able to take this `?` out as well.
```suggestion
	projectName: Project['name'];
```

bencodrington (commented):
Looks good, works well.

Just one comment about the case where no project is selected, then this is good to go.

packages/client/hmi-client/src/services/project.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Get a project per id
+ * @return Project - the appropriate project, or an empty one if none returned by API
+ */
+async function getProject(projectId: string): Promise<Project> {
+	const response = await API.get(`/projects/${projectId}`);
+	return response?.data ?? emptyProject;
+}
+
+export { getProject as get, emptyProject as empty };
```
mwdchang, line 25: Why do we need to rename the exported functions? If you want to call it `get` and `empty`, why not just name them get/empty respectively?

packages/client/hmi-client/src/App.vue
YohannParis: You are correct, I was lazy and didn't wanted to handle `null`, as `emptyProject` was already existing.

packages/client/hmi-client/src/App.vue
```diff
@@ -25,6 +34,29 @@ const disableOverlay = () => {
...
 
 const dataExplorerActivated = ref(false);
+
+/**
+ * Project
+ *
+ * As we use only one Project per application instance.
+ * It is loaded at the root and passed to all views as prop.
+ */
+const project = ref<Project | null>();
```
bencodrington, line 44: Genuinely curious, don't we need to explicitly initialize this to `null`?

```suggestion
const project = ref<Project | null>(null);
```

I'd assume that otherwise it's initialized to `undefined` which would unexpectedly fail any checks for `project.value === null`.

But I also assume TypeScript would yell at you if it was initialized to `undefined`...? Curious about the actual behaviour.

YohannParis: I can't find documentation about it, so I added the `null`

bencodrington: Yeah looks like it gets initialized to `undefined` and the type gets (silently) extended to `Project | null | undefined`.

Adding `null` is the way to go. 👍

bencodrington (approved):
Looks good to me! Thanks.

PR #268, opened 2022-11-28
fix: project setup on start

YohannParis:
# Description

* Fix an issue where the Project API was not call on start
* Added a basic error handling on the Project Service API call

## Type of change

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -45,29 +45,23 @@ const dataExplorerActivated = ref(false);
  */
 const project = ref<Project | null>(null);
 
-async function setProject(id: Project['id']) {
-	// If the Id or the Project are null,
-	// set the Project to null, and send the user to the homepage.
-	if (!id) {
-		project.value = null;
-		goToHomepage();
-	} else {
-		project.value = await ProjectService.get(id);
-		if (!project.value) {
-			goToHomepage();
-		}
-	}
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/App.vue
```diff
@@ -45,21 +45,28 @@ const dataExplorerActivated = ref(false);
...
-		} else {
-			const id = projectId as string;
-			project.value = await ProjectService.get(id);
-			if (!project.value) {
-				goToHomepage();
-			}
-		}
+		const id = projectId as string;
+		setProject(id);
 	}
```
mwdchang, line 70: I was thinking something like below would be an easier fix, in the sense that if we need to track more variables like modelId or simulationId...this provides a cleaner way to read the logic, and potentially less chance s we'd clobber our watched variables together. 

Basically we just rig the original watch's third option to have `{immediate: true}`.


```
watch(
	() => route.params.projectId,
	async (projectId) => {
		if (!projectId) {
			// Set the Project to null,
			// and send the user to the homepage.
			project.value = null;
			goToHomepage();
		} else {
			const id = projectId as string;
			project.value = await ProjectService.get(id);
			if (!project.value) {
				goToHomepage();
			}
		}
	},
	{ immediate: true }
);
```

mwdchang (approved):

bencodrington (approved):

PR #275, opened 2022-11-29
feat: simulation results plumbing

mwdchang:
# Description
First round of displaying simulation results and API plumbing.
- side panel content with existing simulation runs given a project
- plumbing to get responsive-matrix into application proper
- plumbing to watch for simulatin-run-id changes
- small router tweaks

image

## Testing
Navigate to http://localhost:8078/app/#/projects/1/simulation-results/1 

Note there are some minor issues with the sidebar states, you need to click the simulator-run-result icon directly instead of just opening the sidebar-panel, this is a sidebar issue I think and the fix is out of scope for this PR.

## Type of change


packages/client/hmi-client/src/components/Sidebar.vue
```diff
@@ -101,6 +117,10 @@ function openView(view: string, openViewSidePanel: boolean = true): void {
 				<ModelSidebarPanel v-if="selectedView === RouteName.ModelRoute" />
 				<DocumentsSidebarPanel v-else-if="selectedView === RouteName.DocumentRoute" />
 				<ProfileSidebarPanel v-else-if="selectedView === RouteName.ProfileRoute" />
+				<SimulationResultSidebarPanel
+					v-if="selectedView === RouteName.SimulationResultRoute"
```
mj3cheun, line 121: should this `v-if` also be a `v-else-if` like the other panels instead?

packages/client/hmi-client/src/router/index.ts
```diff
@@ -16,8 +17,8 @@ export enum RoutePath {
 	DocView = '/docs/:id?',
 	Project = '/projects/:projectId',
 	ModelView = '/projects/:projectId/model/:modelId',
-	SimulationView = '/projects/:projectId/simulation',
-	Results = '/projects/:projectId/results',
+	SimulationView = '/projects/:projectId/simulations/:simulationId?',
+	SimulationResultView = '/projects/:projectId/simulation-results/:simulationRunId?',
```
YohannParis, line 21: Why the change to _Simulation Result_ instead of _Result_?
Are we planning on showing different kind of results in the future?

packages/client/hmi-client/src/types/Project.ts
```diff
@@ -3,4 +3,11 @@ export type Project = {
 	name: string;
 	description: string;
 	timestamp: string;
+	assets: {
+		publications: number[];
+		intermediates: number[];
+		models: number[];
+		plans: number[];
+		simulation_runs: number[];
+	};
```
YohannParis, line 12: Same comment I gave @amostafa-uncharted: Why not putting those assets on the root? they are as important part of a project than `name` or `timestamp`?
Also, we don't care about respecting the _Data Service_ types on the front-end.

packages/client/hmi-client/src/components/sidebar-panel/simulation-result-sidebar-panel.vue
```diff
@@ -0,0 +1,33 @@
+<script setup lang="ts">
+import { Project } from '@/types/Project';
+import { useRouter } from 'vue-router';
+
+defineProps<{
+	project: Project;
```
YohannParis, line 6: Why not just send `simulation_runs` as prop?

packages/client/hmi-client/src/components/sidebar-panel/simulation-result-sidebar-panel.vue
```diff
@@ -0,0 +1,33 @@
...
+			@click="openSimulationResult(runId)"
+		>
+			{{ runId }}
+		</li>
+	</ul>
+</template>
+
+<style>
+.selectable-item:hover {
+	background-color: #ddd;
```
YohannParis, line 30: We use [Uncharted Design Tokens](https://github.com/unchartedsoftware/design-tokens/blob/main/build/scss/_uncharted-variables.scss#L17)

packages/client/hmi-client/src/views/SimulationResult.vue
```diff
@@ -0,0 +1,69 @@
+<script setup lang="ts">
+import { ref, watch } from 'vue';
+import { useRoute } from 'vue-router';
+import ResponsiveMatrix from '@/components/responsive-matrix/matrix.vue';
+
+const route = useRoute();
+
+watch(
+	() => route.params.simulationRunId,
+	async (id) => {
```
YohannParis, line 10: Shouldn't be `simulationRunId` ?

packages/client/hmi-client/src/views/SimulationResult.vue
```diff
@@ -0,0 +1,69 @@
...
+		{ S: 50, I: 10, R: 40 }
+	],
+	[
+		{ S: 98, I: 2, R: 0 },
+		{ S: 20, I: 60, R: 20 },
+		{ S: 0, I: 30, R: 70 }
+	]
+]);
+
+const fillColorFn = (datum: any, _paramsMin: any, paramsMax: any): String => {
```
YohannParis, line 37: No `any` please.

mj3cheun: if this needs to be typed we can use the following types.

```
import { CellData, ParamMinMax } from '@/types/ResponsiveMatrix';

const fillColorFn = (datum: CellData, _paramsMin: ParamMinMax, paramsMax: ParamMinMax): String => {
```

packages/client/hmi-client/src/components/sidebar-panel/simulation-result-sidebar-panel.vue
mwdchang: It is a balance between least knowledge principle / demeter's law, and to have a unified interface. 

In this case because all the sidebar panels are project-assets:
- it looks more unified to pass in project object to everything rather than their specific asset-classes
- we will probably need to add/remove project-associations, in which case we will want to have a project-id handy

packages/client/hmi-client/src/router/index.ts
mwdchang: Just making it more explicit, because we will have `datasets` that can be simulation-results.

packages/client/hmi-client/src/components/Sidebar.vue
mwdchang: done

packages/client/hmi-client/src/router/index.ts
YohannParis: But if datasets can be simulation results, we can display them on the same view? We will have a control system to switch between tabular and matrix cells views.

packages/client/hmi-client/src/components/sidebar-panel/simulation-result-sidebar-panel.vue
YohannParis: Adding/removing assets within a project should be done via the `services/projects.ts` and not within the component.
I'm just worried of having to much reactivity on massive object when it is not needed for now.
I prefer to keep it simple until we need more info.

packages/client/hmi-client/src/App.vue
```diff
@@ -77,7 +77,12 @@ watch(
 		@show-data-explorer="dataExplorerActivated = true"
 	/>
 	<main>
-		<Sidebar v-if="isSidebarVisible" class="sidebar" data-test-id="sidebar" />
+		<Sidebar
+			v-if="isSidebarVisible && project"
```
YohannParis, line 81: I don't understand the need to add this?
Why not adding it in `isSidebarVisible` test that is built for it.

packages/client/hmi-client/src/components/sidebar-panel/simulation-plan-sidebar-panel.vue
```diff
@@ -0,0 +1,33 @@
...
+			@click="openSimulationPlan(simulationId)"
+		>
+			{{ simulationId }}
+		</li>
+	</ul>
+</template>
+
+<style>
+.selectable-item:hover {
+	background-color: #ddd;
```
YohannParis, line 30: Uncharted Token as well

packages/client/hmi-client/src/components/sidebar-panel/simulation-result-sidebar-panel.vue
mwdchang: Agreed, but I need to do something like
```
projectService.removeSimulation(projectId, simulationPlanId)
```

So the components needs `projectId` and `simulation id array` passed in as properties. So we might as well pass in the project data structure that encapsulates both.

packages/client/hmi-client/src/App.vue
mwdchang: This is a linter error because project can be null, it isn't really the same check semantically as whether the route is valid or not which is what `isSidebarVisible` is checking.

It is more hassle to check at the children components if the project-property is null, and makes a lot more sense for them to be able to make non-null assumptions.

YohannParis (approved):

PR #293, opened 2022-12-02
fix: update the Create a new project form

YohannParis:
# Description

Updated the Create a new project form based on Pascale new design:

Screen Shot 2022-12-02 at 16 51 58
Screen Shot 2022-12-02 at 16 52 08


mwdchang:
Looks good!
Personally I would prefer smaller paddings in the modal and reclaim more available spaces, but can chat about it later.

PR #297, opened 2022-12-05
chore: small props cleanup

mwdchang:
### Description
Small clean up to have a nicer interface where we don't need to constantly check for nulls

### Testing
No functional changes, responsive-matrix should work as before

See http://localhost:8078/app/#/responsive-playground


YohannParis (approved):
LGTM, thanks for cleanin

PR #321, opened 2022-12-07
chore: fix issue preventing adding resources from explorer to project assets

amostafa-uncharted:
# Description

fix issue preventing adding resources from explorer to project assets

Resolves #(issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
+++ b/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
@@ -80,7 +80,7 @@ public class ProjectResource {
 		@PathParam("resource_type") final String type, // ResourceType
 		@PathParam("resource_id") final String resourceId
 	) {
-		return proxy.createAsset(projectId, ResourceType.findByType(type).name(), resourceId);
+		return proxy.createAsset(projectId, type, resourceId); // ResourceType.findByType(type).name()
 	}
 
 	@DELETE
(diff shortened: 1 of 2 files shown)


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
```diff
@@ -80,7 +80,7 @@ public Response createAsset(
 		@PathParam("resource_type") final String type, // ResourceType
 		@PathParam("resource_id") final String resourceId
 	) {
-		return proxy.createAsset(projectId, ResourceType.findByType(type).name(), resourceId);
+		return proxy.createAsset(projectId, type, resourceId); // ResourceType.findByType(type).name()
```
dvince2, line 83: Was `ResourceType.findByType(type).name()` not working as expected here? It was added as a solution to allow for lowercase paths to be used.  Going back to just the String `type` would restrict you to using case-sensitive all-caps paths?

dvince2 (changes requested):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
amostafa-uncharted: ResourceType.findByType(type).name() this was returning for example "DATASETS" while the data service expects "datasets" and this has been breaking the ability to add assets to the project. I do not have time in fix it properly in java, and I am passing the proper string from the client side that should be used as is and it has been fixing the problem!

dvince2: Ok sounds good - I'm going to open an [issue](https://github.com/DARPA-ASKEM/TERArium/issues/322) to revisit this after the demo so that this doesn't block anything.  I had been under the impression that lower case wasn't working and was causing issues but if thats not the case then this may be good as-is :)

dvince2 (approved):

PR #332, opened 2022-12-07
fix: added papers don't show until refresh

bencodrington:
https://user-images.githubusercontent.com/16928557/206299759-11489b51-d41f-440b-a7a2-7513c0e49214.mov

# Description

Just needed to properly propagate the project ID for the document route like the other routes.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/packages/client/hmi-client/src/components/sidebar-panel/documents-sidebar-panel.vue
+++ b/packages/client/hmi-client/src/components/sidebar-panel/documents-sidebar-panel.vue
@@ -19,6 +19,7 @@ import { useRouter } from 'vue-router';
 import { deleteAsset } from '@/services/project';
 import { PUBLICATIONS } from '@/types/Project';
 import { PublicationAsset } from '@/types/XDD';
+import { RouteName } from '@/router/routes';
 import ArtifactList from './artifact-list.vue';
 
 const router = useRouter();
@@ -35,7 +36,10 @@ const documentsAsArtifactList = computed(() =>
 const openDocumentPage = async (xddUri: string) => {
 	// pass this doc id as param
 	documentId.value = xddUri; // track selection
-	router.push({ path: `/docs/${xddUri}` });
+	router.push({
+		name: RouteName.DocumentRoute,
+		params: { projectId: resourcesStore.activeProject?.id, id: xddUri }
+	});
 };
 
 const removeDocument = async (xddUri: string) => {
--- a/packages/client/hmi-client/src/router/index.ts
+++ b/packages/client/hmi-client/src/router/index.ts
@@ -15,7 +15,7 @@ import { RouteName } from './routes';
 
 export enum RoutePath {
 	Home = '/',
-	Document = '/docs/:id?',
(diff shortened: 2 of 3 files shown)


amostafa-uncharted:
A related issue that I would argue it should be fixed here as well:
- opening an existing project that has papers saved with it
- immediately click on the "Documents" button/icon to open the side panel. 
Note that the list is empty because proper (id) is not saved stored properly

packages/client/hmi-client/src/components/sidebar-panel/documents-sidebar-panel.vue
```diff
@@ -35,7 +36,10 @@ const documentsAsArtifactList = computed(() =>
 const openDocumentPage = async (xddUri: string) => {
 	// pass this doc id as param
 	documentId.value = xddUri; // track selection
-	router.push({ path: `/docs/${xddUri}` });
+	router.push({
+		name: RouteName.DocumentRoute,
+		params: { projectId: resourcesStore.activeProject?.id, id: xddUri }
+	});
```
YohannParis, line 42: So clean... so beautiful!

packages/client/hmi-client/src/router/index.ts
```diff
@@ -15,7 +15,7 @@ import { RouteName } from './routes';
 
 export enum RoutePath {
 	Home = '/',
-	Document = '/docs/:id?',
+	Document = '/projects/:projectId/docs/:id?',
```
YohannParis, line 18: We might want to figure something about that. Because we want to be able to share documents without having to open a project, maybe we need a separate page just for that purposes. Anyway good for now.

YohannParis (approved):
Thank you!

* My comment are purely informational, please resolve them.
* I love how your 24s video is 10s of xDD API loading time

YohannParis:
> A related issue that I would argue it should be fixed here as well:
> 
> * opening an existing project that has papers saved with it
> * immediately click on the "Documents" button/icon to open the side panel.
>   Note that the list is empty because proper (id) is not saved stored properly

> Note this only happens the first time (after a page refresh)

This should be done in another PR.

packages/client/hmi-client/src/router/index.ts
bencodrington: Agreed.

bencodrington:
> > > A related issue that I would argue it should be fixed here as well:
> > > 
> > > * opening an existing project that has papers saved with it
> > > * immediately click on the "Documents" button/icon to open the side panel.
> > >   Note that the list is empty because proper (id) is not saved stored properly
> > 
> > 
> > > Note this only happens the first time (after a page refresh)
> > 
> > 
> > This should be done in another PR.
> 
> Noted

@amostafa-uncharted I noticed that as well but I think it occurs when you open the side panel before the assets request gives a response. Could be just a matter of showing a loading spinner and refreshing the pane when the response gets in, rather than a missing ID.

Issue #331, opened 2022-12-07
[BUG]: Fix model page for empty project


Tom-Szendrey:
**Describe the bug**
After creating a new project going to the model page will default to model/1
<img width="1222" alt="image" src="https://user-images.githubusercontent.com/17088680/206296028-95e4e053-f7ba-4ec3-b40e-0a3666ba9184.png">

PR #333, opened 2022-12-07
chore: do not display a model if empty project

Tom-Szendrey:
# Description
Do not display model #1 if no project is provided
image

Resolves #(issue)
#331 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Manually

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/client/hmi-client/src/components/Sidebar.vue
+++ b/packages/client/hmi-client/src/components/Sidebar.vue
@@ -69,7 +69,7 @@ const openView = (view: RouteName) => {
 		}
 
 		if (view === RouteName.ModelRoute) {
-			params.modelId = props?.project?.assets[MODELS]?.[0] ?? 1;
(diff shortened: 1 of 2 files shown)


bencodrington (changes requested):
Very nice! I was just in the process of doing something like this myself.

One nitpick: can we use `undefined` instead of the `''` empty string value?

Tom-Szendrey:
> Very nice! I was just in the process of doing something like this myself.
> 
> One nitpick: can we use `undefined` instead of the `''` empty string value?

When trying to use undefined the if statement did not catch it. `if (props.modelId !== undefined) {`
I am not familiar enough to know why. Will look into this further

bencodrington (approved):
Forget what I said, the prop goes through the URL so I believe it needs to be a string. Thanks, and merge away.

PR #338, opened 2022-12-08
fix: show blank page when no dataset selected

bencodrington:
image

## Description

Just put everything inside a `<template v-if="dataset !== null">`


YohannParis (approved):
Sweet and simple.

PR #340, opened 2022-12-08
ci: ci build optimizations

Nithos:
# Description

This is a double PR that changes the following CI builds:

1. Adding a new group to the bake file for staging. When merging PRs to `main` it will ONLY build the client and regular server. Native server build will only be build when doing a tagged publish.
2. E2E tests will no longer be run on each PR, instead they will be run daily at 5am. Devs should check the action status/slack notification for errors

### DIFF
--- a/.github/workflows/test-client-e2e.yaml
+++ b/.github/workflows/test-client-e2e.yaml
@@ -4,9 +4,10 @@
 # See: https://playwright.dev/
 
 name: Client E2E Tests
-# yamllint disable-line rule:truthy
 on:
-  workflow_call:
+  # schedule tests to run everyday at 5am
+  schedule:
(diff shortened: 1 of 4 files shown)


.github/workflows/test-client-e2e.yaml
```diff
@@ -6,7 +6,9 @@
 name: Client E2E Tests
 # yamllint disable-line rule:truthy
 on:
-  workflow_call:
+  # schedule tests to run everyday at 5am
+  schedule:
+    - cron: '0 5 * * *'
```
YohannParis, line 11: This will make our lives nicer.

YohannParis (approved):
Thank you, I appreciate it.

chris-dickson (approved):

PR #344, opened 2022-12-08
chore: improve doc and example

mwdchang:
### Summary
Better documentation and examples

### DIFF
--- a/packages/client/graph-scaffolder/README.md
+++ b/packages/client/graph-scaffolder/README.md
@@ -12,8 +12,8 @@ The graph data is specified in the format of `{nodes, edges}` and allows for nes
 
 ```
 const graphData = {
-	nodes: [
-		{
+  nodes: [
+    {
       id: 'node1',
       label: 'node1 label'
       x: 100,
@@ -22,7 +22,7 @@ const graphData = {
       height: 100,
       data: { value: 123 },
       nodes: []
-		},
+    },
     {
       id: 'node2',
       label: 'node2 label',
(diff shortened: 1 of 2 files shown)


packages/client/graph-scaffolder/README.md
```diff
@@ -1,28 +1,129 @@
...
+### Usage
+Simple renderer with Typescript
+
+```
+interface NodeData {
+  value: number;
+}
+
+interface EdgeData {
+  foo: string;
```
YohannParis, line 69: You sure this should be `foo`?

packages/client/graph-scaffolder/examples/src/index.ts
```diff
@@ -31,8 +31,18 @@ class SampleRenderer extends BasicRenderer<NodeData, EdgeData> {
 	}
 
 	renderNodes(selection: D3SelectionINode<NodeData>) {
-		const species = selection.filter((d) => d.data.type === 'species');
-		const transitions = selection.filter((d) => d.data.type === 'transition');
+		const species = selection.filter((d) => d.type !== 'custom' && d.data.type === 'species');
+		const transitions = selection.filter(
+			(d) => d.type !== 'custom' && d.data.type === 'transition'
+		);
+		const customs = selection.filter((d) => d.type === 'custom');
```
YohannParis, line 38: I can't wait until we can use [`Array.prototype.group()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/group)
```js
const { species, transitions, customs } = selection.group(({ type }) => type);
```

Until then why not use [`Array.prototype.reduce()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#grouping_objects_by_a_property)
```js
const { species, transitions, customs } = selection.reduce((sorted, d) => {
    const curGroup = sorted[d.type] ?? [];
    return { ...sorted, [d.type]: [...curGroup, d] };
}, {});
```

YohannParis (approved):
LGTM

packages/client/graph-scaffolder/README.md
mwdchang: `NodeData` and `EdgeData` are arbitrary data structures, as in INode<NodeData> and IEdge<EdgeData> respectively, I should add some notes that it supports and curries the typing information.

packages/client/graph-scaffolder/examples/src/index.ts
mwdchang: Because I do not like the reduce -syntax, I find them harder to read, and there is not sufficient data here where a reduce call will have a faster performance than filter 🤷

PR #391, opened 2022-12-14
feat: vuetify integration

YohannParis:
# Description

* Implemented [Vuetify](https://next.vuetifyjs.com/en/) as design system
* Added a property css files to overwrite the Vuetify CSS properties with TERArium ones
* Change the _Data Explorer_ button in the header with Vuetify as first test.


jamiewaese-uncharted:
It's a "component library" not a "design system".... haha

:)

mwdchang:
Looks good code wise, is there an adoption plan? e.g. we do {a, b, c}, then {d, e}, then {f , g, h}?

mwdchang (approved):

PR #404, opened 2022-12-17
chore: refactor explorer optimize performance

### LINKED ISSUE #365: [TASK]: Improve data explorer search performance
**Describe the task**
Currently, any user interaction within the data explorer such as changing search terms and/or applying a facets filter results in a full search across artifacts. In other words, multiple server calls including external calls to xDD are performed with almost every user interaction in the data explorer. This adds significant overhead since much of the data didn't change. For example, if the user has focused on the "Models" view/tab and clicked on one of the facets to apply a filter. Currently, this interaction results in re-fetching all data (xDD publications, models, datasets) and re-computing all facets and aggregations.  Clearly, only the models data should be recomputed. A better/lighter implementation is to not re-execute the query against all types of artifacts especially if the user filter has only been against a specific artifact type. 

Note that we still want to maintain parallel search across artifacts, but we should refactor to avoid performance overhead and unnecessary data fetch. Perhaps only the first ever fetch should be truly done across artifacts, but subsequent search interactions should be more limited in scope to the currently focused tab/view. 

This task is all about better tracking what data is invalidated and only recompute that part of the data. Ideally, caching should also implemented to speed up fetching of data. 

**Task Items**
- [x] implement a way to track which data need invalidation
- [x] fetch only subset of the data and use existing cached data from other artifact types

**Goal**

**Screenshots**

**Additional context**

amostafa-uncharted:
# Description
- Significantly improve the Data Explorer performance ~%50 or more increase
- Avoid timeout issues

No functionality changes, but working with the explorer explorer now should be much faster and smoother. 
A key observation is that the explorer now only fetches data as needed!

Resolves #365

### DIFF
--- a/packages/client/hmi-client/src/services/data.ts
+++ b/packages/client/hmi-client/src/services/data.ts
@@ -98,7 +98,7 @@ const getModels = async (term: string, modelSearchParam?: ModelSearchParams) => 
 	const allModels = modelsList.map((m) => ({ ...m, type: 'model' }));
 
 	// first get un-filtered concept facets
-	let conceptFacets = await getConceptFacets([MODELS]);
+	let conceptFacets = await getConceptFacets([ProjectAssetTypes.MODELS]);
 
 	const modelResults = filterModels(allModels, conceptFacets, term);
 
@@ -270,7 +270,7 @@ const getDatasets = async (term: string, datasetSearchParam?: DatasetSearchParam
 		type: 'dataset'
 	}));
 
-	let conceptFacets = await getConceptFacets([DATASETS]);
+	let conceptFacets = await getConceptFacets([ProjectAssetTypes.DATASETS]);
 
 	const datasetResults = filterDatasets(allDatasets, conceptFacets, term);
 
--- a/packages/client/hmi-client/src/services/data.ts
+++ b/packages/client/hmi-client/src/services/data.ts
@@ -1,10 +1,16 @@
 import { cloneDeep, uniq, uniqBy } from 'lodash';
-import { Facets, ResourceType, SearchParameters, SearchResults } from '@/types/common';
+import {
+	Facets,
+	FullSearchResults,
+	ResourceType,
+	SearchParameters,
+	SearchResults
+} from '@/types/common';
 import API from '@/api/api';
 import { getDatasetFacets, getModelFacets } from '@/utils/facets';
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/services/data.ts
```diff
@@ -562,47 +620,71 @@ const getDocumentById = async (docid: string) => {
...
+	if (resultType === ResourceType.DATASET || resultType === ResourceType.ALL) {
+		const promise3 = new Promise<FullSearchResults>((resolve, reject) => {
+			try {
+				resolve(getDatasets(term, searchParamWithFacetFilters?.dataset));
+			} catch (err: any) {
+				reject(new Error(`Error fetching dataset results: ${err}`));
+			}
+		});
+		promiseList.push(promise3);
+	}
```
shawnyama, line 677: I think it may be worth looping through `ResourceType` instead of having similar if statements.
```suggestion
Object.entries(ResourceType).forEach(([key]) => {
		console.log(ResourceType[key])
		if (resultType === ResourceType[key] || resultType === ResourceType.ALL) {
			// eslint-disable-next-line no-async-promise-executor
			const promise = new Promise<FullSearchResults>(async (resolve, reject) => {
				try {
					switch (ResourceType[key]) {
						case ResourceType.XDD:
							const allData = await searchXDDArticles(term, searchParam?.xdd);
							const allDataFilteredWithFacets = await searchXDDArticles(
								term,
								searchParamWithFacetFilters?.xdd
							);
							resolve({ allData, allDataFilteredWithFacets });
							break;
						case ResourceType.MODEL:
							resolve(getModels(term, searchParamWithFacetFilters?.model));
							break;
						case ResourceType.DATASET:
							resolve(getDatasets(term, searchParamWithFacetFilters?.dataset));
							break;
					}
				} catch (err: any) {
					reject(new Error(`Error fetching ${ResourceType[key]} results: ${err}`));
				}
			});
			promiseList.push(promise);
		}
	});```

shawnyama (changes requested):
Overall, looks good

packages/client/hmi-client/src/services/data.ts
amostafa-uncharted: Agreed, but I would leave this to another time once more focused code refactoring is performed!

YohannParis (approved):

PR #411, opened 2022-12-20
feat: update responsive matrix zoom to toggle between camera and selection modes

echl:
# Description

Alt/Option key toggles between camera and selection modes with cursor changing styles to indicate mode. 

Should there be a label to additionally indicate active cursor mode?

### DIFF
--- a/packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
+++ b/packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
@@ -1,5 +1,5 @@
 <template>
-	<main class="matrix-container" ref="matrixContainer">
+	<main class="matrix-container" ref="matrixContainer" :class="{ cameraCursor: isCameraMode }">
 		<div class="matrix" ref="matrix" :style="matrixStyle">
 			<LabelCols
 				v-if="!disableLabelCol && rendererReady"
@@ -103,6 +103,10 @@ import matrixGridFS from './matrix-grid.fs.glsl';
 
 // EMPTY_POINT used as a fallback value
 const EMPTY_POINT = new Point();
+enum CursorModes {
+	Selection,
+	Camera
+}
 
 export default {
 	// ---------------------------------------------------------------------------- //
@@ -272,7 +276,7 @@ export default {
 			update: 0,
 			move: 0,
 
-			enableDrag: false,
+			cursorMode: CursorModes.Selection,
 			resizeObserver: null as unknown as ResizeObserver
 		};
 	},
@@ -358,6 +362,9 @@ export default {
 			}
 
 			return microColSettings;
+		},
+		isCameraMode(): boolean {
+			return this.cursorMode === CursorModes.Camera;
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -103,6 +103,10 @@ import matrixGridFS from './matrix-grid.fs.glsl';
 
 // EMPTY_POINT used as a fallback value
 const EMPTY_POINT = new Point();
+enum CursorModes {
```
mj3cheun, line 106: we got a type file for the responsive stuff at '@/types/ResponsiveMatrix', this should probably go in there?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<main class="matrix-container" ref="matrixContainer">
+	<main class="matrix-container" ref="matrixContainer" :class="{ cameraCursor: isCameraMode }">
```
mj3cheun, line 2: we should probably have one class property attached to this element rather than multiple. in addition, ive been using kebab case for class properties throughout the responsive matrix stuff, if we want to change that we should probably do it everywhere, but until that happens probably a good idea to keep this consistent and stick with kebab case.

packages/client/hmi-client/src/types/ResponsiveMatrix.ts
```diff
@@ -65,3 +65,8 @@ export type Uniforms = {
 	// cell element color data
 	uColor: Texture;
 };
+
+export enum CursorModes {
+	Select,
```
mj3cheun, line 70: another consistency thing, i have been using ALL_CAPS for enums (which im treating as constants).

mwdchang:
The toggling feels a bit awkward, since there is little status to indicate the modality change. 
What is the original problem we are trying to solve here?

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
echl: What's the rational for only having one class attached to this element?

echl:
> The toggling feels a bit awkward, since there is little status to indicate the modality change. What is the original problem we are trying to solve here?

Original issue was just to use a different modifier key for zooming (and fix the bug where simply pressing shift without scrolling could also zoom). We could definitely have more indication about what the cursor mode is, I'm not sure whether toggle or modifier is more intuitive. Perhaps we could get input from the team.

mwdchang (dismissed):

packages/client/hmi-client/src/components/responsive-matrix/matrix.vue
mj3cheun: on some more looking it appears that having both a `class` and `:class` on the same element is normal, i personally dont really like that but if thats how its supposed to be done its fine. i would prefer if we do something like the following for the class name.

```
<main ref="matrixContainer" class="matrix-container" :class="{ 'camera-cursor': isCameraMode }">

mj3cheun (approved):

PR #413, opened 2022-12-21
refactor: merge getModels and getDatasets function

shawnyama:
# Description

* Purpose: Cleaning up data explorer

* Merged `getModels` and `getDatasets` into `getAssets`. Will probably combine `getXDDAritcles` later to this in some way in another PR.
* Cleaned up `getFacetsDisplayNames` and some smaller miscellaneous clean up.

Resolves #(issue)


packages/client/hmi-client/src/utils/facets.ts
```diff
@@ -116,25 +116,21 @@ function mergeCustomizer(objValue: any, srcValue: any) {
 	// return null;
 }
 
-export const getFacets = (results: SearchResults[], resultType: string) => {
+export const getFacets = (results: SearchResults[], resultType: ResourceType | string) => {
 	let facets = {} as Facets;
 	if (results.length > 0) {
 		results.forEach((resultsObj) => {
+			console.log(resultsObj, resultType);
```
YohannParis, line 123:
```suggestion
```

packages/client/hmi-client/src/services/data.ts
```diff
@@ -693,8 +605,9 @@ export {
 	getXDDDictionaries,
 	getXDDArtifacts,
 	searchXDDArticles,
-	getModels,
-	getDatasets,
+	getAssets,
+	// getModels,
+	// getDatasets,
```
YohannParis, line 610:
```suggestion
```
You are cleaning up, go for it 😄

packages/client/hmi-client/src/services/data.ts
```diff
@@ -633,47 +555,37 @@ const fetchData = async (
...
+						default:
+							break;
+					}
+				} catch (err: any) {
+					reject(new Error(`Error fetching ${ResourceType[key]} results: ${err}`));
+				}
+			});
+			promiseList.push(promise);
+		}
+	});
```
YohannParis, line 588: This will need to be moved to the `hmi-server`.
Once @Tom-Szendrey is done with his first implementation this can be a good candidate for it.

YohannParis (approved):
Thanks for this first pass 🎉

packages/client/hmi-client/src/utils/data-util.ts
```diff
@@ -12,7 +12,39 @@ import { Dataset, FACET_FIELDS as DATASET_FACET_FIELDS } from '@/types/Dataset';
 // source: https://www.crossref.org/blog/dois-and-matching-regular-expressions/
 const DOI_VALIDATION_PATTERN = /^10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i;
 
-// Apply filter to data in place
+// export const applyFacetFilters = (results: Model[] | Dataset[], filters: Filters) => {
```
echl, line 15: Can we just delete this method?

packages/client/hmi-client/src/services/data.ts
```diff
@@ -251,72 +130,98 @@ const filterDatasets = (
...
-	// FIXME: this client-side computation of facets from "dataset" data should be done
+	// FIXME: this client-side computation of facets from "models" data should be done
 	//        at the HMI server
 	//
 	// This is going to calculate facets aggregations from the list of results
-	const datasetFacets = await getDatasetFacets(datasetResults, conceptFacets);
+	let assetResults: Model[] | Dataset[];
+	let assetFacets: Facets;
+
+	switch (resourceType) {
```
echl, line 178: There's a few places where we do a switch and then call a method that is basically identical whether it's for models or datasets. I am tempted to say that where this happens it can be combined into one method (e.g. `getAssetFacets` instead of `getModelFacets` and `getDatasetFacets`, making use of function overloading. 

However, if all this logic is going to be moved to server side then it's probably not worth it? What do you think @YohannParis

echl (dismissed):

packages/client/hmi-client/src/utils/data-util.ts
shawnyama: This is a work in progress

packages/client/hmi-client/src/services/data.ts
shawnyama: Yes there will still be some repetition even after I do more clean up. The `getFacets` functions will be moved to the server side later.

YohannParis (approved):
LGTM

PR #418, opened 2022-12-24
refactor: combine apply facet filter funcs

shawnyama:
# Description
Continuing to refactor Data Explorer related functions. Take a look at places like line `59` in `data.ts` I avoided typecheck issues by using the conditional there but it still results in code that's still a bit repetitive, haven't come across a better way to do it. 

`data-util.ts`:
- Combined `applyFacetFiltersToDatasets` and `applyFacetFiltersToModels` into `applyFacetFilters`

`data.ts`:
- Cleaned up `searchXDDArticles` by using a switch case
- Combined `filterModels` and `filterDatasets` into `filterAssets`


packages/client/hmi-client/src/utils/facets.ts
```diff
@@ -16,7 +16,7 @@ import { groupBy, mergeWith, isArray } from 'lodash';
 
 // FIXME: this client-side computation of facets from "models" data should be done //////////////////no point in editing//////////////////
 //        at the HMI server
-export const getModelFacets = (models: Model[], conceptFacets: ConceptFacets | null) => {
+export const getModelFacets = async (models: Model[], conceptFacets: ConceptFacets | null) => {
```
mwdchang, line 19: I don't understand why
- getDatasetFacets, and
- getModelFacets

Are marked as `async`, is there a particular reason here?

packages/client/hmi-client/src/utils/data-util.ts
```diff
@@ -12,81 +12,53 @@ import { Dataset, FACET_FIELDS as DATASET_FACET_FIELDS } from '@/types/Dataset';
...
+					break;
+				case ResourceType.DATASET:
+					results.splice(
+						0,
+						results.length,
+						...(results as Dataset[]).filter(
+							(asset) =>
+								filterValues.includes(asset[filterField as keyof Dataset].toString()) === isNot
+						)
+					);
```
mwdchang, line 57: This pattern is a bit weird in the sense that a lot of other places we tend to do more of the immutable routes where functional parameters are not modified. So i.e.:

```
result = applyFacetFilters(result, ...)
```

As opposed to

```
applyFacetFilters(result, ...)
```

Here you are removing everything, then injecting items back into the same address-space...

Is there a particular reason why this is done this way? Do we have a preference more one paradigm than the other? 
@YohannParis @amostafa-uncharted

packages/client/hmi-client/src/services/data.ts
```diff
@@ -42,69 +42,39 @@ const getXDDDictionaries = async () => {
...
+				case ResourceType.MODEL:
+					finalAssets = (allAssets as Model[]).filter((d) =>
+						(d[attribute as keyof Model] as string).toLowerCase().includes(term.toLowerCase())
+					);
+					break;
+				case ResourceType.DATASET:
+					finalAssets = (allAssets as Dataset[]).filter((d) =>
+						(d[attribute as keyof Dataset] as string).toLowerCase().includes(term.toLowerCase())
+					);
+					break;
```
mwdchang, line 71: For situations like this, you can consider using generics which can help cut down boiler-plate code. 

In this particularly case there isn't a huge amount of savings because this generally need to be taken into account from the start and not in medias res. Nevertheless something like this is possible:

```
		const filterXYZ = <T>(array: T[], attr: string) =>
			array.filter(d => (d[attr as keyof T] as string).toLowerCase().includes('blah'));
```

Then later on 

```
  finalAssets = filterXYZ(allAssets as Model[], attribute);
```

But yeah, if you really want to reduce replication here the generic pattern needs to be applied across various functions so you don't have the `Model[] | Dataset[]` type declaration that can make wrangling things difficult.

packages/client/hmi-client/src/utils/facets.ts
mwdchang: Generally speaking you'd need async/await pairs if you are dealing with promises, which none were produced here as far as I am able to decipher.

If the idea is that these functions will be replaced with API calls then let's just leave them alone. Otherwise the flow would be a little simpler without the unnecessary async/aait.

packages/client/hmi-client/src/utils/data-util.ts
```diff
@@ -12,81 +12,34 @@ import { Dataset, FACET_FIELDS as DATASET_FACET_FIELDS } from '@/types/Dataset';
...
-			const isNot = !clause.isNot; // is the filter reversed?
-			const filteredDatasets = datasetResults.filter(
-				(dataset) =>
-					// direct query against Dataset fields
-					filterValues.includes(dataset[filterField as keyof Dataset].toString()) === isNot
+			results.splice(
+				0,
+				results.length,
+				...results.filter(
+					(asset) => filterValues.includes(`${asset[filterField as keyof T]}`) === isNot
```
amostafa-uncharted, line 40: Why did you remove the call to "toString()"?

packages/client/hmi-client/src/utils/data-util.ts
amostafa-uncharted: No preference. The code was written without consistency at that time for the sake of quick implementation and perhaps other parts of the code was refactored slightly at other times resulting in such inconsistency with regards to following the im/mutable path. Happy to discuss further what you guys prefer!

packages/client/hmi-client/src/utils/data-util.ts
shawnyama: I ran into this issue so that was my attempt to work around it.
image

Is there a better way?

amostafa-uncharted: I would consider casting to any first

shawnyama: Changed it let me know if it's good

packages/client/hmi-client/src/services/data.ts
```diff
@@ -240,10 +196,10 @@ const getAssets = async (
 
 		switch (resourceType) {
 			case ResourceType.MODEL:
-				applyFacetFiltersToModels(assetResults as Model[], searchParam.filters);
+				applyFacetFilters(assetResults as Model[], searchParam.filters, resourceType);
```
amostafa-uncharted, line 199: I don't think we need the switch statement here

packages/client/hmi-client/src/services/data.ts
```diff
@@ -365,88 +321,60 @@ const getRelatedDocuments = async (docid: string, dataset: string | null) => {
 };
 
 const searchXDDArticles = async (term: string, xddSearchParam?: XDDSearchParams) => {
-	const limitResultsCount = xddSearchParam?.perPage ?? XDD_RESULT_DEFAULT_PAGE_SIZE;
+	let searchParams = `term=${term}`;
+	const url = '/xdd/documents?';
 
-	// NOTE when true it disables ranking of results
-	const enablePagination = xddSearchParam?.fullResults ?? false;
+	if (xddSearchParam) {
```
amostafa-uncharted, line 327: I am a little concerned that the `XDDSearchParams` may contain fields that are not actually valid query parameters. Also, the attempt to abstract how some of the search fields are concerted to proper API query parameter names seems to complicate things.

packages/client/hmi-client/src/utils/facets.ts
mwdchang: I think let's leave this alone for now.

packages/client/hmi-client/src/utils/data-util.ts
```diff
@@ -12,81 +12,35 @@ import { Dataset, FACET_FIELDS as DATASET_FACET_FIELDS } from '@/types/Dataset';
...
-// 			// use splice to filter in place
-// 			results.splice(0, results.length, ...filteredModels);
-// 		}
-// 	});
-// };
-
-// Apply filter to data in place //////////// refactor this
-export const applyFacetFiltersToModels = (modelResults: Model[], filters: Filters) => {
-	if (isEmpty(filters) || isEmpty(modelResults)) {
+export const applyFacetFilters = <T extends Model | Dataset>(
```
mwdchang, line 15: I don't think you need `T extends ...` here, isn't `<T>` by itself sufficient?

packages/client/hmi-client/src/services/data.ts
shawnyama: Do you mean, there are attributes in the `XDDSearchParams` type that shouldn't be there or they aren't handled by the server yet? Also what does the abstraction complicate?

amostafa-uncharted: Some of the XDDSearchParams fields such as "type" is not a valid xDD query parameter and is treated differently to filter xDD extractions. Therefore, I don't see a major value in refactoring the various if statements into a switch statement especially when blindly looping through the XDDSearchParams fields.

In addition, some of the XDDSearchParams fields should be combined with others but this knowledge is kind of hidden away and not well documented. By abstracting or refactoring how all fields are treated via the switch statement fallback mechanism, you risk mistreating some of the fields. 

Overall, this part of the code is actually trivial and should not be a major part of the refactoring nor deserve that many discussions! I am OK with your refactoring but was just pointing out a few potential issues and hence the preference of just leaving this part as is!!

packages/client/hmi-client/src/services/data.ts
```diff
@@ -365,88 +312,60 @@ const getRelatedDocuments = async (docid: string, dataset: string | null) => {
...
 const searchXDDArticles = async (term: string, xddSearchParam?: XDDSearchParams) => {
-	const limitResultsCount = xddSearchParam?.perPage ?? XDD_RESULT_DEFAULT_PAGE_SIZE;
+	let searchParams = `term=${term}`;
+	const url = '/xdd/documents?';
+
+	if (xddSearchParam) {
+		Object.entries(xddSearchParam).forEach(([key]) => {
+			// Loop through searchParam attributes
+			if (xddSearchParam[key]) {
+				const urlKey = key.replace(/[A-Z]/g, (c) => `_${c.toLowerCase()}`); // Later maybe make attributes have underscore format from the get go?
```
amostafa-uncharted, line 322: One reason I am against this change is because the code
`key.replace(/[A-Z]/g, (c) => `_${c.toLowerCase()}`);
` is not too clear on what exactly it is doing as it is not documented and assumes some knowledge about regex (as opposed to the old code which was more explicit).

Also, please note here how the parameters are assumed be passed to the server with underscore in their naming. While this is currently the case, it shouldn't be because the server is supposed to be receiving proper parameter/variable names without the underscore and without forcing the client side to do this naming transformation. The initial implementation, however, was forced to use underscore because the server is not yet able to handle such naming. I would prefer code refactoring that properly fix this issue rather than converting the multiple "if" statements into a switch statement!

amostafa-uncharted (approved):
Other than the comments I mentioned I think this initial refactoring pass is good and can be merged

amostafa-uncharted (approved):

amostafa-uncharted (approved):
Other than the comments I mentioned I think this initial refactoring pass is good and can be merged.

I would highly suggest the next refactoring effort to focus on moving some of the data fetch, filtering, facets-calculation and similar logic to the backend

PR #429, opened 2023-01-03
fix: target identity on scroll click event

### LINKED ISSUE #341: [BUG]: Not all clicks register when scrolling home page cards
**Describe the bug**
Clicking the grey circle behind the arrow scrolls the list, but clicking directly on the arrow itself has no effect.

**To Reproduce**
1. Go to Home page
2. Click on the grey circle behind an arrow, should scroll
3. Click on the arrow, doesn't scroll

**Expected behavior**
Should be able to click anywhere on the button

**Screenshots**

https://user-images.githubusercontent.com/16928557/206480552-bee6a7d0-e201-4af3-8abe-8a22763e6fbd.mov

mwdchang:
### Summary
This address a somewhat rare issue when the click event lands on svg:path instead of svg, which cause the parent element query to misfire

### Test
In the home view, scroll should work regardless of where the click is betting registered on the icon-button.

Close #341

### DIFF
--- a/packages/client/hmi-client/src/views/Home.vue
+++ b/packages/client/hmi-client/src/views/Home.vue
@@ -64,7 +64,11 @@ const close = () => {
 const SCROLL_INCREMENT_IN_REM = 21;
 const scroll = (direction: 'right' | 'left', event: PointerEvent) => {
 	const chevronElement = event.target as HTMLElement;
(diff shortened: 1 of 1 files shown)


echl (approved):
I think we might want to refactor this in the future to do it in a more vue-y way, though probably the whole page will change soon.

PR #456, opened 2023-01-10
fix: bad document id given at load of application

### LINKED ISSUE #455: [BUG]: bad document id given at load of application
**Describe the bug**
If an asset doesn't return correctly, a bad document id is passed into `getRelatedDocuments` rather than the expected emptyh string.  This then echos through the document service and can cause the application to lock when searching documents.

**To Reproduce**
1. Load application URL.
2. Notice in the log for the HMI-Server there is an http-500 error.
3. Search for a term, for example, "covid"
4. Notice how the application hangs with "Loading..."

**Expected behavior**
Results should be returned

**Screenshots**
N/A

**Additional context**
There are a few related issues here like "why can this asset not be found" and "why have we hardcoded to look at the second document in the list and not the first" but those can be addressed later.  For now we're just adding safety code.

dvince2:
# Description

* Adding safety code here to handle `undefined` being passed in as a String value.
* There are a few related issues here like "why can this asset not be found" and "why have we hardcoded to look at the second document in the list and not the first" but those can be addressed later..... this safety code would be needed either way and solves the build issues at hand.

Resolves #455

### DIFF
--- a/packages/client/hmi-client/src/services/project.ts
+++ b/packages/client/hmi-client/src/services/project.ts
@@ -102,12 +102,11 @@ async function deleteAsset(projectId: string, assetsType: string, assetId) {
 async function getRelatedArticles(aProject: Project): Promise<XDDArticle[]> {
 	const resp = await getAssets(aProject.id);
 	try {
+		// Dec 6th 2022:
 		// TODO: Speak with XDD Team about broken doc: 5f6d0e20a58f1dfd52184931
 		// Grab the 2nd of publication for related results because grabbing the first provides a broken doc id: 5f6d0e20a58f1dfd52184931
-		const listOfRelatedArticles = await getRelatedDocuments(
-			String(resp?.[ProjectAssetTypes.PUBLICATIONS][1].xdd_uri),
-			'xdd-covid-19'
-		);
+		const docId = resp ? resp?.[ProjectAssetTypes.PUBLICATIONS][1].xdd_uri : '';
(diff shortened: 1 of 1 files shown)


Tom-Szendrey (dismissed):
LGTM

packages/client/hmi-client/src/services/project.ts
```diff
@@ -102,12 +102,11 @@ async function deleteAsset(projectId: string, assetsType: string, assetId) {
...
 	const resp = await getAssets(aProject.id);
 	try {
+		// Dec 6th 2022:
 		// TODO: Speak with XDD Team about broken doc: 5f6d0e20a58f1dfd52184931
 		// Grab the 2nd of publication for related results because grabbing the first provides a broken doc id: 5f6d0e20a58f1dfd52184931
-		const listOfRelatedArticles = await getRelatedDocuments(
-			String(resp?.[ProjectAssetTypes.PUBLICATIONS][1].xdd_uri),
-			'xdd-covid-19'
-		);
+		const docId = resp ? resp?.[ProjectAssetTypes.PUBLICATIONS][1].xdd_uri : '';
```
echl, line 108: Slightly less verbose way you can do conditional assignment: 
`
const docId = resp?.[ProjectAssetTypes.PUBLICATIONS][1].xdd_uri ?? '';`

Tom-Szendrey (approved):

echl (approved):

PR #461, opened 2023-01-11
feat: basic article result card

### LINKED ISSUE #438: [FEAT]: Restyle/restructure search items in Data Explorer
Put search item in its own component

Search items will have the following features:
- [x] first row: type + icons (lower priority)
- [x] second row: title
- [x] third row: authors, date, journal
- [x] fourth/fifth rows: snippets
- [x] optional last row: already saved in project(s) (lower priority)
- [x] selection toggle (+, check)
- [x] 3 dot menu option when in cart
- [x] asset carousel
- [x] filter asset type in xdd preview
- [x] item states (normal, hover, selected)
- [ ] ~highlighting (xdd to provide class to style the highlight)~ separate issue

Image

shawnyama:
# Description

Putting just the article result in its own card. Starting up the basic layout here. Styling and making other asset types compatible will be other PRs. You can compare the old cards with the new cards, I haven't deleted the old ones since I know some of the functionality is in the process of being moved to other components.

Another small thing I rearranged is where the button groups for viewing different assets is located.

Resolves 438(partially)


packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -0,0 +1,87 @@
...
+import { XDDArticle } from '@/types/XDD';
+import { isXDDArticle } from '@/utils/data-util';
+import IconAdd24 from '@carbon/icons-vue/es/add/24';
+import IconCheckmark24 from '@carbon/icons-vue/es/checkmark/24';
+import { ResultType } from '@/types/common';
+
+const props = defineProps<{
+	d: XDDArticle;
+	previewedArticle: XDDArticle | null;
+	selectedSearchItems: ResultType[];
```
mwdchang, line 11: 1. I would recommend avoid using single letter variables unless you are in a short closure/lambda situation, it makes things difficult to read, and it gets confused with other closures/lambdas where we tend to use this convention.

2. If the only purpose of previewedArticle is to check whether the item is active or not, you may want to consider bubbling it up. so instead of
```
{ XDDArticle, XDDArticle, ResultType }
```
can consider
```
{ XDDArticle, boolean, ResultType }
```
or something

packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -0,0 +1,87 @@
...
+const isSelected = () =>
+	props.selectedSearchItems.find((item) => {
+		if (isXDDArticle(item)) {
+			const itemAsArticle = item as XDDArticle;
+			return itemAsArticle.title === props.d.title;
+		}
+		return false;
+	});
+
+const formatTitle = () => (props.d.title ? props.d.title : props.d.title);
```
YohannParis, line 25: What is this?

packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -0,0 +1,87 @@
...
+const formatArticleAuthors = () => props.d.author.map((a) => a.name).join(', ');
```
YohannParis, line 26: Better do a `formatDetails()` and do a proper job of adding `year` and `journal` with error handling.

packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -0,0 +1,87 @@
...
+			<div class="title">{{ formatTitle() }}</div>
+			<div class="details">{{ formatArticleAuthors() }} ({{ d.year }}) {{ d.journal }}</div>
+			<ul class="snippets" v-if="d.highlight">
+				<li v-for="h in d.highlight" :key="h">...<span v-html="h"></span>...</li>
+			</ul>
+		</div>
+		<div @click.stop="emit('toggle-article-selected')">
+			<IconAdd24 v-show="!isSelected()" />
+			<IconCheckmark24 class="checkmark-color" v-show="isSelected()" />
+		</div>
```
YohannParis, line 42: This should be a button instead of a div.
Even with no design to it (no border, no background colours, etc.)

packages/client/hmi-client/src/components/data-explorer/search-item.vue
shawnyama: I'll get rid of it - blindly copy pasted that from the original component haha.

packages/client/hmi-client/src/components/data-explorer/search-item.vue
shawnyama: What do you mean by error handling in this case?

YohannParis: In case of undefined values, etc. What you did does the trick

YohannParis (approved):

PR #465, opened 2023-01-13
chore: article reno work 2 - tabs, adaptive ui, and other tweaks

mwdchang:
### Summary
This is a 2nd round of refactoring and restyling Document viewer
- Add a tab system similar to that of models, Document is moved to its own component and a "DocumentViewer" view is added.
- Clean up accordion as per design spec
- Make image sizing adaptive depending on available width, for data-explorer images will be smaller, for the full document view images will be larger

I placed the new component within "articles" folder for now,  we will probably want to reorg and have better nomenclatures soon.

In full viewer with tabs:
image

In data explorer:
image

### Testing
- Go into project => papers and select a few documents, should be tabbed and able to switch between tabs
- Go into explorer and search for "covid", select and see documents in the right side pane

### DIFF
--- a/packages/client/hmi-client/src/components/articles/Document.vue
+++ b/packages/client/hmi-client/src/components/articles/Document.vue
@@ -1,5 +1,5 @@
 <template>
-	<section class="doc-view-container">
+	<section class="doc-view-container" ref="sectionElem">
 		<div v-if="doc">
 			<div class="journal">{{ doc.journal }}</div>
 			<div v-if="docLink" class="title">
@@ -32,7 +32,12 @@
 				<AccordionTab header="Figures">
 					<div v-for="ex in figureArtifacts" :key="ex.askemId" class="extracted-item">
 						<div class="img-container">
-							<img id="img" :src="'data:image/jpeg;base64,' + ex.properties.image" :alt="''" />
+							<img
+								id="img"
+								:src="'data:image/jpeg;base64,' + ex.properties.image"
+								:alt="''"
+								:style="{ 'max-width': imageSize }"
+							/>
 							<span>{{
 								ex.properties.caption ? ex.properties.caption : ex.properties.contentText
 							}}</span>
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/components/articles/Document.vue
```diff
@@ -20,31 +20,60 @@
 				<Button label="Open PDF"></Button>
 			</div>
 
-			<Accordion :multiple="true" class="accordian">
+			<Accordion :multiple="true" :active-index="[0, 1, 2, 3, 4, 5, 6, 7]" class="accordian">
```
mwdchang, line 23: Jamie wants all tabs to open by default, I can't find a better way to do this.

packages/client/hmi-client/src/views/DocumentView.vue
```diff
@@ -0,0 +1,121 @@
...
+
+const resourcesStore = useResourcesStore();
+const tabStore = useTabStore();
+
+const newDocumentId = computed(() => props.assetId);
+const openTabs = ref<Tab[]>([]);
+const activeTabIndex = ref(0);
+const documentsInCurrentProject = resourcesStore.activeProjectAssets?.publications; // fixme
+const activeProject = resourcesStore.activeProject;
+const tabContext = `document${activeProject?.id}`;
```
YohannParis, line 28: Should we add a `:` between `document` and the project Id.

YohannParis (approved):

packages/client/hmi-client/src/views/DocumentView.vue
echl: Imo the tab context doesn't really need to be nicely formatted or even human readable.

packages/client/hmi-client/src/views/DocumentView.vue
```diff
@@ -0,0 +1,121 @@
...
+import { useTabStore } from '@/stores/tabs';
+import { isEmpty } from 'lodash';
+import ResourcesList from '@/components/resources/resources-list.vue';
+
+const props = defineProps<{
+	assetId?: string;
+	project: Project;
+}>();
+
+const emit = defineEmits(['show-data-explorer']);
```
echl, line 18: With my recent changes, data explorer is a view, so we can route to it instead of using emits.

YohannParis (approved):

packages/client/hmi-client/src/views/DocumentView.vue
mwdchang: We have a lot of convention-related matters we can talk about, I will make a note, but let's table this for now.

packages/client/hmi-client/src/views/DocumentView.vue
mwdchang: Done

Issue #447, opened 2023-01-09
[TASK]: Fix underscore mapping for Data Service in the Proxy/Ressources


YohannParis:
xDD APIs provide raw data with the names of many fields including underscore and reserved words. For example, fetching xDD documents return a list of documents each with fields such as `_gddid`, `known_terms`, and `abstract`. Note that in the HMI server we ideally want to use mapping (through mp rest client) to re-map such improperly named fields into proper names. However, all the mapping files defined under /document-server/models/xdd are NOT used at all since the defined proxies are configured to return "Response" objects, which means the raw xDD response is returned to the front end as is without actually applying any de/serialization mapping!!
Accordingly, the front end has been applying such field name mappings at the client side after receiving the response from the HMI server as a temporary solution. See for example the following code:

<img width="981" alt="Screen Shot 2023-01-09 at 8 56 52 PM" src="https://user-images.githubusercontent.com/69688226/211444435-8971c014-8942-42be-8005-ff012430d4a4.png">

This task is all about re-visiting all the code at the HMI-server, the document service, and subsequently the client side and fixing all the improperly named fields, which will mostly be fields named after reserved words or those that contain underscore.

No functionality changes is expected and everything should work as is, but the mappings defined at the backend should be in use. 

Note that the aforementioned issue may affect other data services or APIs beyond xDD such as the TERArium data service!

PR #470, opened 2023-01-16
393 home page

Tom-Szendrey:
# Description
Move a lot of the home page's API calls that were nested in projects + related articles ect into the hmi-server

Still todo to remove from draft is just a lot of cleaning up of the code. Removing logs, grabbing correct author + identifier for documents ect 

Resolves #(issue)


packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/models/xdd/Document.java
```diff
@@ -16,13 +16,13 @@
 @Accessors(chain = true)
 public class Document implements Serializable {
 
-	@JsonbProperty("_gddid")
+	@JsonbProperty("gddid")
```
dvince2, line 19: Right now in this PR you won't actually see the results of changing this, as the `Document` object is never used in the serialization (currently this is being addressed in #447 )  However, we can't get rid of `"_gddid" as this is the field name coming to us from xDD, and changing this to not have the underscore results in it not being serialized properly.

dvince2 (changes requested):

packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/models/xdd/Document.java
Tom-Szendrey: Okay I will revert this change. I believe I did something similar for `abstract`/`abstracttext` 
To be clear the remapping of this fieldname should be done on the front end then?

packages/client/hmi-client/src/services/project.ts
```diff
@@ -57,9 +57,15 @@ async function getAll(): Promise<Project[] | null> {
...
  */
-async function getAssets(projectId: string): Promise<ProjectAssets | null> {
+async function getAssets(projectId: string, types?: string[]): Promise<ProjectAssets | null> {
 	try {
-		const response = await API.get(`/projects/${projectId}/assets`);
+		let url = `/projects/${projectId}/assets`;
+		if (types) {
+			types.forEach((type, indx) => {
+				url += `${indx === 0 ? '?' : '&'}types=${type}`;
+			});
```
YohannParis, line 66: It might be more readable with an [`Array.join()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join)
```js
url += `?type=${types.join('&type='}`;
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Project.java
```diff
@@ -25,23 +29,23 @@ public class Project implements Serializable {
...
 	private List<Document> relatedDocuments;
 
-	public Project(String id, String name, String description,String timestamp) {
+	public Project(String id, String name, String description) {
 		this.projectID = id;
 		this.name = name;
 		this.description = description;
 		this.active = true;
-		this.timestamp = Instant.parse(timestamp + "Z");
+		// this.timestamp = Instant.parse(timestamp + "Z");
```
YohannParis, line 48: Remove comments

packages/client/hmi-client/src/components/articles/selected-article-pane.vue
```diff
@@ -66,7 +66,9 @@ const addResourcesToProject = async (projectId: string) => {
 
 // TODO: May need more formatting than just replcing <p></p> in future
 const formatAbstract = (item: XDDArticle) =>
-	item.abstract.replace('<p>', '\n').replace('</p>', '\n') || '[no abstract]';
+	item.abstract !== undefined
+		? item.abstract.replace('<p>', '\n').replace('</p>', '\n')
+		: '[no abstract]';
```
YohannParis, line 71: Wait, why not just render the HTML as is on the HMI?

Tom-Szendrey: I wasnt sure how to do that so I just referenced what I saw other components doing. Let me look into that real quick!

YohannParis: Don't worry about it.

packages/client/hmi-client/src/services/project.ts
Tom-Szendrey: It has to be ?types=A&types=B&types=C unfortunately. I will add comments to this to help its readability

YohannParis:
Can someone approve this? It will be nice to have it merge today.

dvince2 (approved):
There are a few minor clean up tasks here to do just to get rid of some unneeded accessors/setters because we're using Lombok.  I'm going to approve since they're just minor cleanup but if you wouldnt mind swapping that out would be great!

packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/models/xdd/RelatedDocument.java
```diff
@@ -14,4 +14,6 @@ public class RelatedDocument {
 
 	private Number score;
 
+	public Document getDocument(){ return this.bibjson; }
```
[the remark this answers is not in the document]
Tom-Szendrey: I will be leaving this for readability. I dont know how to customize the getter's name via annotations

PR #473, opened 2023-01-16
feat: task fix underscore mapping for data service in the proxyressources

### LINKED ISSUE #447: [TASK]: Fix underscore mapping for Data Service in the Proxy/Ressources

dvince2:
# Cleanup of document service and related API

* Removal of unused document models from hmi-server
* Document-service now correctly uses the model and response objects rather than just `Response` type
* Corrected a serialization bug, allowing for document-service to deserialize xDD json properties as one name and reserialize as a different name to allow our hmi-server and client the ability to refer to objects without underscores.
* Added ASCII art to the document service startup console printout
* Updated hmi-client to use the new properties on the front end.

Resolves #447


Tom-Szendrey (commented):
Thank you for this!!

YohannParis:
Assign someone to this PR, and then we can bug them to approve it.

Tom-Szendrey (approved):

PR #480, opened 2023-01-17
feat: filters in doc card

shawnyama:
# Description

You can interact with the filter icons now. Also fixed click bubbling on card buttons.

image

Resolves 438 partially


mwdchang:
@shawnyama why are there so many commits in your PRs, only the last 4-5 are relevant but it looks like you are carrying your own version of commit history since last October. 

Would like to ensure that
1. you are not somehow working off an abandoned/corrupted base, and
2. this listing does not get longer and longer and create issues with Github UI/interactions

packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -17,18 +17,24 @@ const props = defineProps<{
...
+		? props.asset.relatedExtractions?.filter((ex) => {
+				if (chosenExtractionFilter.value === null) {
+					return (
+						ex.askemClass === XDDExtractionType.Figure ||
+						ex.askemClass === XDDExtractionType.Table ||
+						ex.askemClass === XDDExtractionType.Equation ||
+						ex.askemClass === XDDExtractionType.Section || // remove this later just for testing pagination
+						ex.askemClass === XDDExtractionType.Document || // remove this later just for testing pagination
+						ex.askemClass === undefined // remove this later just for testing pagination
+					);
```
mwdchang, line 34: Why is there an extra `undefined` clause now? and the associated comments don't seem to make any sense??

Anyway, if we can exclude needing `undefined` you can probably do something like this

```
return ex.askemClass in XDDExtractionType
```

packages/client/hmi-client/src/components/data-explorer/search-item.vue
```diff
@@ -120,17 +142,20 @@ const formatFeatures = () => {
...
 				<div class="asset-nav-arrows">
-					<i class="pi pi-arrow-left" @click="paginationMovement(-1)"></i>
-					Asset {{ relatedAssetPage + 1 }} of {{ extractionsWithImages.length }}
-					<i class="pi pi-arrow-right" @click="paginationMovement(1)"></i>
+					<i class="pi pi-arrow-left" @click.stop="paginationMovement(-1)"></i>
+					<template v-if="extractionsWithImages.length > 0">
+						Asset {{ relatedAssetPage + 1 }} of {{ extractionsWithImages.length }}
+					</template>
+					<template v-else> No {{ chosenExtractionFilter }}s </template>
+					<i class="pi pi-arrow-right" @click.stop="paginationMovement(1)"></i>
```
mwdchang, line 158: Nitpick: I would recommend we don't use the pagination nomenclature here, maybe `preview` is a better term?

Tabular data will typically not show up well in small sizes, and to some extent equations because of with/height ratios ... I wonder if we need to go back to Jamie for this for more thoughts.

YohannParis (approved):

packages/client/hmi-client/src/components/data-explorer/search-item.vue
shawnyama: xdd has been returning undefined classes as of late. It's just there so I can test the pagination. Once Derek fixes the backend stuff I'll only be filtering through table, figure, url, equation

packages/client/hmi-client/src/components/data-explorer/search-item.vue
shawnyama: Similar situation as above, they are undefined anyways so I'll apply proper filters once we have chosen which ones/once the backend works

YohannParis (dismissed):

YohannParis (approved):

PR #482, opened 2023-01-18
feat: add endpoints for persisting and getting user events

chris-dickson:
This PR adds Hibernate connection properties for persisting values to a database.  We add a simple `EventResource` endpoint that allows for creating of events and listing/search for previous events.  

The `EventType` can be specified as `persistent` meaning that records are stored in the database for the given user.  This is useful for recording things like searches, logins, logouts, etc.  Non-persistent events are logged for metrics and/or dashboarding after the fact (eg/ facet filter, add X to project, etc)

To create an event:
```bash
curl --location --request POST 'http://localhost:3000/api/events' \
--header 'Authorization: Bearer ...
--data-raw '{
    "type": "SEARCH",
    "projectId": 1,
    "value": "covid-19"
}'
```
Results in the following event persisted to the database:
```json
{
    "id": "11d773fb-62a8-44d3-98c9-64efcbd3767f",
    "timestampMillis": 1674058852832,
    "projectId": 1,
    "username": "adam",
    "type": "SEARCH",
    "value": "covid-19"
}
```
A unique id is generated, along with the timestamp of event creation, the projectId, username, type, and value of the event are persisted.

We additionally provide and endpoint for fetching events:
```bash
curl --location --request GET 'http://localhost:3000/api/events?projectId=1&type=SEARCH' \
--header 'Authorization: Bearer ... '
```
Results in:
```json
[
    {
        "id": "a604313e-8a97-4607-987a-aa53c3db8368",
        "timestampMillis": 1674059193970,
        "projectId": 1,
        "username": "adam",
        "type": "SEARCH",
        "value": "covid-19"
    }
]
```

The `projectId` and `type` are required parameters. We inherently filter to only events created by the requesting user. We can optionally provide a `search` term that will filter the results based on the search string being contained in the `value` field of the event.  

eg:
```
```bash
curl --location --request GET 'http://localhost:3000/api/events?projectId=1&type=SEARCH&value=cov' \
--header 'Authorization: Bearer ... '
```

### DIFF
--- a/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
+++ b/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
@@ -1,0 +1,63 @@
+package software.uncharted.terarium.hmiserver.entities;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import io.quarkus.hibernate.orm.panache.PanacheQuery;
+import io.quarkus.panache.common.Sort;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.experimental.Accessors;
+import software.uncharted.terarium.hmiserver.models.EventType;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+
+@Entity
+@Data
+@Accessors(chain = true)
+@NoArgsConstructor
+public class Event extends PanacheEntityBase implements Serializable {
+	@Id
+	private String id = UUID.randomUUID().toString();
+
+	@Column(nullable = false)
+	private Long timestampMillis = Instant.now().toEpochMilli();
+
+	@Column(nullable = false)
+	private Long projectId;
+
+	@Column(nullable = false)
+	private String username;
+
+	@Column(nullable = false)
+	private EventType type;
+
+	@Column(length = 2047)
+	private String value;
+
+	/**
+	 * Gets events by type and an option search string for values
+	 * @param type				the {@link EventType}
+	 * @param projectId		the project id
+	 * @param username		the username of the current user
+	 * @param like				optional search string for values of matching events
+	 * @param limit				the maximum number of events to return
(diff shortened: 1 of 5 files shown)


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+
+@Entity
+@Data
+@Accessors(chain = true)
+@NoArgsConstructor
+public class Event extends PanacheEntityBase implements Serializable {
```
mwdchang, line 23: Do we need indices?

chris-dickson: Added

dvince2 (approved):

dvince2 (dismissed):
LGTM

tangopapatime (dismissed):

YohannParis (commented):
Could you add information to the README.md and the [Technical Specifications](https://docs.google.com/document/d/1h_CGQhB0aouIfsOVz79pfuqrgqm7EuBUzaYcekU74aU/edit?usp=sharing)

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/EventResource.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	/**
+	 * Create an event
+	 * @param event	the {@link Event} instance
+	 * @return			the persisted event instance
+	 */
+	@POST
+	@Transactional
+	public Response putEvent(final Event event) {
```
dvince2, line 57: Should this be `postEvent` with that annotation?

dvince2 (commented):
Unsure if the method name here is intentional or not

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/EventResource.java
chris-dickson: Fixed

dvince2 (approved):

PR #484, opened 2023-01-18
chore: extend document service implementation to give more control when fetching extractions

### LINKED ISSUE #468: [TASK]: Extend document service implementation to give more control to users when fetching extractions
**Describe the task**
Currently, fetching paper extractions is done via the function `getXDDArtifacts` at the client side which calls the xdd proxy path `/xdd/extractions`. The current implementation however does not expose two parameters to either fetch ALL extractions associated with a given paper (via `doc_doi`) or to search for extractions as part of keyword search (via the `term` parameter).

IT looks like that xDD has extended the API that work with extractions on their side and now other undocumented flags can be used such as "`page`" and "`ASKEM_CLASS`"
For example, the following will returns only figure extractions that match a given search term
https://xdddev.chtc.io/askem/object?query_all=COVID-19&ASKEM_CLASS=Figure

We need to update the extraction end point that is part of the document service to allow the user to specify the type of extractions to be returned as an optional parameter. The client side code should also be updated to leverage this flag to only return perhaps Figures, Tables, and Equations.
Note that if so much data is returned, the "page" flag can be used as a query parameter to retrieve paginated result and should also be integrated in the HMI server/client.

**Task Items**
- [ ] list of items for completion
- [ ] another item

**Goal**

**Screenshots**

**Additional context**

dvince2:
# Description

Updating xdd extraction backend to pass through new options for user when fetching extractions

Resolves #468


amostafa-uncharted (approved):
LGTM

PR #486, opened 2023-01-18
feat: temporary context menu

mwdchang:
### Summary 
Temporary context menu for doing related-xyz actions. This will be replaced at some point by drag-and-drop interactions


mwdchang:
Subsumed by another task.

PR #488, opened 2023-01-18
feat: search by example

### LINKED ISSUE #446: [TASK] Data explorer - APIs for document similarity
Given an document artifact, we need the logic and plumbings to
- forward citation search
- backward citation search
- content similarity
- models and dataset ??

NOTE 1: that we know how to find similar document given its DOI.
NOTE: 2: we can utilize the provenance graph to find models and assets related to a given publication

Image

amostafa-uncharted:
# Description
Implementation of the search by example feature allowing users to find similar papers and related TERArium artifacts.

Adding a paper to a cart
Screen Shot 2023-01-18 at 5 53 16 PM

Issuing a search by example to find related content
Screen Shot 2023-01-18 at 5 53 39 PM

Similar papers for a given paper utilizing xDD
Screen Shot 2023-01-18 at 5 54 08 PM

Resolves #446


packages/client/hmi-client/src/types/Provenance.ts
```diff
@@ -0,0 +1,37 @@
+export enum ProvenanceType {
+	// eslint-disable-next-line @typescript-eslint/no-shadow
+	Dataset = 'Dataset',
+	// eslint-disable-next-line @typescript-eslint/no-shadow
+	Model = 'Model',
```
YohannParis, line 5: Why do you need those `eslint-disable-...` if `Dataset` and `Model` aren't defined outside the export scope?

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -17,6 +17,7 @@
 			></i>
 		</span>
 		<i class="pi pi-history" />
+		<i class="pi pi-image" title="Search by Example" @click="toggleSearchByExample" />
```
YohannParis, line 20: This should be a `<Button />`

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,203 @@
...
+			</div>
+			<div @click.stop="searchOptions.relatedContent = !searchOptions.relatedContent">
+				Related Content
+				<span v-show="searchOptions.relatedContent">
+					<IconCheckboxChecked20 />
+				</span>
+				<span v-show="!searchOptions.relatedContent">
+					<IconCheckbox20 />
+				</span>
+			</div>
```
YohannParis, line 104: Those needs to be [multiple select button](https://www.primefaces.org/primevue/selectbutton) or at least [checkboxes](https://www.primefaces.org/primevue/checkbox).
@jamiewaese-uncharted could you weigh in?

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,203 @@
...
+			<Button :class="{ disabled: isSearchDisabled }" :disabled="isSearchDisabled" @click="search"
+				>Search</Button
+			>
+		</footer>
+	</div>
+</template>
+
+<style scoped>
+.search-by-example-card {
+	border: 1px solid var(--un-color-body-stroke);
```
YohannParis, line 114: Please replace `--un-*` with the newer CSS properties:
https://github.com/DARPA-ASKEM/TERArium/blob/main/packages/client/hmi-client/src/assets/css/theme/_variables.scss#L834

YohannParis:
@derekryansnider and @Tom-Szendrey could you please review this PR?

dvince2:
@derekryansnider I believe thats me :)

packages/client/hmi-client/src/types/Provenance.ts
amostafa-uncharted: that was needed earlier but not anymore so I have removed it

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
amostafa-uncharted: this is a temp placeholder so I would no worry about it since it is actually not hooked to anything!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ProvenanceQueryParameters.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.experimental.Accessors;
+
+import javax.json.bind.annotation.JsonbProperty;
+import java.io.Serializable;
+
+@Data
+@Accessors(chain = true)
+public class ProvenanceQueryParameters implements Serializable {
+
+	@JsonbProperty("root_id")
```
dvince2, line 13: In its current form, these annotations aren't doing anything as the java objects are also named with underscores.  These objects should be renamed to camel case (for example`rootId`).

If you want to access these properties by camel case in the front end instead of the underscored version, you would do something like...

```
private Number rootId;
.
.
.
@JsonbProperty(root_id)
public void setRootId(Number rootId){
     this.rootId = rootId;
}
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/ProvenanceProxy.java
```diff
@@ -20,11 +21,12 @@ Response createProvenance(
...
 	);
 
-	@GET
-	@Path("/derived_from")
-	Response searchProvenance(
-		@QueryParam("artifact_id") String artifactId,
-		@QueryParam("artifact_type") String artifactType
+	@POST
+	@Path("/search")
+	@Consumes(MediaType.APPLICATION_JSON)
```
dvince2, line 26: I don't believe this Consumes is needed - that should be default type

dvince2 (changes requested):
I've done a look at the back end and made a few suggestions - I'll let @Tom-Szendrey and @YohannParis give a 👍 for the front end work though!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ProvenanceQueryParameters.java
amostafa-uncharted: Good point as I noticed some other places in the code such as xdd_uri was not missed and still uses the underscore. I will fix both. Thanks

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/ProvenanceProxy.java
amostafa-uncharted: It doesn't hurt being a bit more specific but I will remove to simplify things

amostafa-uncharted: done

amostafa-uncharted:
> I am getting error when trying to use this on: POST http://localhost:8078/api/provenance/connected_nodes Is this using data-service's main?

This is a new API added in TDS and in this PR I have added the backend for accessing it. An example of using it from the frontend is in the provenance.ts file

Tom-Szendrey (approved):

Tom-Szendrey:
> I am getting error when trying to use this on: POST http://localhost:8078/api/provenance/connected_nodes Is this using data-service's main?

FYI this is because you need to use dev branch of data-service not main branch

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -71,6 +72,9 @@ function addSearchTerm(term) {
 	// @ts-ignore
 	inputElement.value?.$el.focus();
 }
+const toggleSearchByExample = () => {
+	emit('toggle-search-by-example');
```
echl, line 76: If this is the only thing this method does, you can just emit directly from `@click` instead of having a function that just calls another function

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -122,6 +126,16 @@ watch(suggestedItems, (newItems) => {
 	color: var(--text-color-secondary);
 }
 
+.pi-image {
```
echl, line 129: Can you use a custom class name for this behaviour, since `pi-image` should just be used to define what icon is displayed

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,170 @@
+<script setup lang="ts">
+import IconClose16 from '@carbon/icons-vue/es/close/16';
+import { ref, PropType, computed } from 'vue';
+import { ResultType, SearchByExampleOptions } from '@/types/common';
+import { Model } from '@/types/Model';
+import { XDDArticle } from '@/types/XDD';
+import Checkbox from 'primevue/checkbox';
+import Button from '@/components/Button.vue';
```
echl, line 8: Are we able to use primevue button here instead?

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,170 @@
+<script setup lang="ts">
+import IconClose16 from '@carbon/icons-vue/es/close/16';
```
echl, line 2: Can we use prime icon here instead?

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,170 @@
...
+const emit = defineEmits(['hide', 'search']);
+
+const searchOptions = ref<SearchByExampleOptions>({
+	similarContent: true,
+	forwardCitation: false,
+	bakcwardCitation: false,
+	relatedContent: false
+});
+
+const onClose = () => {
```
echl, line 26: You can just do `emit` wherever `onClose` is being called.

packages/client/hmi-client/src/components/data-explorer/search-by-example.vue
```diff
@@ -0,0 +1,170 @@
...
+	emit('hide');
+};
+
+function search() {
+	emit('search', searchOptions.value);
+	onClose();
+}
+
+const isSearchDisabled = computed(() => {
+	if (!props.item) return true;
```
echl, line 36:
```suggestion
	return (!props.item) ||
	(!searchOptions.value.similarContent &&
		!searchOptions.value.forwardCitation &&
		!searchOptions.value.bakcwardCitation &&
		!searchOptions.value.relatedContent)
```

packages/client/hmi-client/src/components/drilldown-panel/selected-resources-options-pane.vue
```diff
@@ -107,6 +120,15 @@ const getType = (item: ResultType) => {
 	return ResourceType.ALL;
 };
 
+const findRelatedContent = (item: ResultType) => {
```
echl, line 123: Same as above re: don't need to wrap emit in a function.

packages/client/hmi-client/src/components/drilldown-panel/selected-resources-options-pane.vue
```diff
@@ -107,6 +120,15 @@ const getType = (item: ResultType) => {
...
+	emit('find-related-content', item);
+};
+
+// only available for publications
+const findSimilarContent = (item: ResultType) => {
```
echl, line 128: Same as above re: don't need to wrap emit in a function.

packages/client/hmi-client/src/services/data.ts
```diff
@@ -25,6 +27,8 @@ import {
 import { getFacets as getConceptFacets } from './concept';
 import * as DatasetService from './dataset';
 import { getAllModelDescriptions } from './model';
+// eslint-disable-next-line import/no-cycle
```
echl, line 30: Should not ignore cyclical imports in my opinion. If it's unavoidable then maybe they should be in the same file.

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -19,6 +19,7 @@
 				></i>
 			</span>
 			<i class="pi pi-history" />
+			<i class="pi pi-image" title="Search by Example" @click="toggleSearchByExample" />
```
YohannParis, line 22: This should be a `<Button>`

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
amostafa-uncharted: sure but this is temporary placeholder code that is not being used at this time and will likely be refactored later!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Publication.java
```diff
@@ -12,8 +12,12 @@ public class Publication implements Serializable {
 
 	private String id;
 
-	@JsonbProperty("xdd_uri")
 	private String xddUri;
+	@JsonbProperty("xdd_uri")
+	public void setXddUri(String xddUri){
+			this.xddUri = xddUri;
+	}
+	public String getXddUri(){ return this.xddUri; }
```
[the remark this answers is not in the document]
amostafa-uncharted: Noted. Thanks

YohannParis (approved):
Let's do it!

Issue #462, opened 2023-01-12
[FEAT]: add client log shipping from hmi-client to hmi-server


chris-dickson:
The hmi-client generates a lot of important logs (both info and errors). Because they're running within the clients browser, we have no way to track errors that occur or anything other information that's crucial to understanding the research activities performed, specifically for the hackathon.  

We want to be able to ship client logs to the server so they can be echoed and shipped to our centralized logging solution. Currently, the `hmi-client` module is using `console.log` everywhere.  Once this framework is in place, a follow on task will be to replace the console.log statements with the logger and add a lint rule that forbids using the console logger.

We will need to choose a logging framework that integrates well with vue3. Please do some research on what the vue3 community is using.  It doesn't seem like vue3 has logging as a "built-in".

Some candidates:
- https://www.npmjs.com/package/vue-logger-plugin
- https://www.npmjs.com/package/vuejs3-logger
- https://github.com/unjs/consola
- Write a very simple utility ourself that wraps `console.log`

Requirements:
- Should be easy for developers to use/inject. Ideally, it should not be more complicated than `console.log`
- Should incorporate the concept of "hooks" as we'll need to buffer client logs before shipping them to the server

Client Side implementation:
- The post-log hook should buffer each log level (info, debug, error, warn) and periodically (every 5 seconds?) make a post back to the server.

Server side:
- Create a simple `/log` controller that accepts a `POST` which is simply an array of the following payload:
```json
[
  {
    "level" : "info",
    "message": "This is a log from the hmi-client"
  },
  ...
]  
```
The `/log` controller then simples echos the logs on the server so they may be collected by filebeats/prometheus/etc.

PR #489, opened 2023-01-19
feat: add client log shipping from hmi client to hmi server

tangopapatime:
# Description

First part of the setting up a logging framework to capture user logs in a structured manner.  This will allow to measure user interactions and issues that occurred during usage.

Currently uses `vue-logger-plugin` and it's `afterHook` to buffer log messages and then post the messages periodically every 5 seconds.

Resolves #(issue)
Resolves the first part of setting up the instrumentation to

ship client calls to the server
the server saves structured logs

https://github.com/DARPA-ASKEM/TERArium/issues/462#event-8213619647

### DIFF
--- a/packages/client/hmi-client/src/main.ts
+++ b/packages/client/hmi-client/src/main.ts
@@ -1,7 +1,7 @@
 import { createApp } from 'vue';
 import { createPinia } from 'pinia';
 import PrimeVue from 'primevue/config';
-import { createLogger } from 'vue-logger-plugin';
+import { logger } from '@/utils/logger';
 import useAuthStore from './stores/auth';
 import router from './router';
 import App from './App.vue';
@@ -11,7 +11,7 @@ const app = createApp(App);
 app.use(createPinia());
 app.use(router);
 app.use(PrimeVue, { ripple: true });
-app.use(createLogger({}));
+app.use(logger);
 
 const auth = useAuthStore();
 await auth.fetchSSO();
--- a/packages/client/hmi-client/src/main.ts
+++ b/packages/client/hmi-client/src/main.ts
@@ -1,6 +1,7 @@
 import { createApp } from 'vue';
 import { createPinia } from 'pinia';
 import PrimeVue from 'primevue/config';
+import { createLogger } from 'vue-logger-plugin';
 import useAuthStore from './stores/auth';
 import router from './router';
 import App from './App.vue';
@@ -10,6 +11,7 @@ const app = createApp(App);
 app.use(createPinia());
 app.use(router);
 app.use(PrimeVue, { ripple: true });
+app.use(createLogger({}));
 
 const auth = useAuthStore();
 await auth.fetchSSO();
--- a/packages/client/hmi-client/src/utils/logger.ts
+++ b/packages/client/hmi-client/src/utils/logger.ts
@@ -1,0 +1,75 @@
+// HMI-Client logging service
+import API from '@/api/api';
+import { createLogger, LogEvent, LoggerHook, LogLevel } from 'vue-logger-plugin';
+// TODO: add logic for different modes
+// const env = process.env.NODE_ENV;
+
+// LogDetails Interface
+interface LogDetails {
+	level: LogLevel;
+	message: string;
+}
+
+// Usage:
+// import logger from '@/utils/logger';
+class LogBuffer {
+	logs: LogDetails[];
+
+	constructor() {
+		this.logs = [];
+	}
+
+	clearLogs = () => {
+		this.logs = [];
+	};
+
+	add = (log: LogDetails) => {
+		this.logs.push(log);
+	};
+
+	isEmpty = (): boolean => {
+		if (this.logs.length > 0) {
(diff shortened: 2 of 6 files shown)


yarn.lock
```diff
@@ -11377,6 +11378,13 @@ send@latest:
   languageName: node
   linkType: hard
 
+"vue-logger-plugin@npm:2.2.3":
```
tangopapatime, line 11381: Definitely don't really know how/what exactly needs to be checked in to get a new package working.

YohannParis: What do you mean?

tangopapatime: Just meant that I'm not sure I added the package in all the places it needs to be.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
```diff
@@ -0,0 +1,36 @@
...
+	@Inject
+	SecurityIdentity securityIdentity;
+
+	@Inject
+	LoggingService loggingService;
+
+	@POST
+	@Path("/logs")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Consumes(MediaType.APPLICATION_JSON)
```
dvince2, line 31: The default produce/consume type is APPLICATION_JSON, so you won't need these here (especially with the Produce/Consume at the top of the class too)

yarn.lock
YohannParis: The right way to add packages is:
- go within `TERArium/packages/client/hmi-client`
- install package with `yarn add --exact [package_name]`
- make sure `yarn.lock` is part of the commit

Then you are all good, and it looks like this is what you did.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
```diff
@@ -0,0 +1,36 @@
...
+	public Response letsDoThis(LoggingService logData) {
```
dvince2, line 32: Also, may need a more description function name ;)

dvince2 (changes requested):

yarn.lock
tangopapatime: gotcha.  Thanks @YohannParis .

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
tangopapatime: Definitely didn't know this as Quarkus virgin :P

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
tangopapatime: lol, oops, yeah forgot.  glad I didn't use my usual temp naming of using curse words.

dvince2 (dismissed):

packages/client/hmi-client/src/main.ts
```diff
@@ -10,6 +11,7 @@ const app = createApp(App);
 app.use(createPinia());
 app.use(router);
 app.use(PrimeVue, { ripple: true });
+app.use(createLogger({}));
```
mwdchang, line 14: This looks a little strange, is the usage correct? I would have expected that createLogger to be called only once but we are calling it twice?

We have
```
import { createLogger } from 'vue-logger-plugin';
app.use(createLogger({}));
```

shouldn't this be 

```
import { logger } from '@/utils/logger-something-something';
app.use(logger);
```

where the logger instance is created in the util?

packages/client/hmi-client/src/utils/logger.ts
```diff
@@ -0,0 +1,75 @@
...
+		if (!this.isEmpty()) {
+			await API.post(`/logs/`, {
+				logs: this.getLogBuffer()
+			})
+				.then(() => {
+					this.clearLogs();
+				})
+				.catch((error) => {
+					console.error(error);
+				});
```
mwdchang, line 55: Would prefer if we can stick with either async/await-style, or promise-style.  But not both at the same time as it makes it a bit harder to understand.

I think the code here, as-is,  it technically works but the await is redundant.

packages/client/hmi-client/src/main.ts
tangopapatime: no you're right.  found it strange that this worked in the first place.

dvince2 (approved):
lgtm

packages/client/hmi-client/src/utils/logger.ts
tangopapatime: fixed.

PR #501, opened 2023-01-20
ci: update target extraction from bakefile

Nithos:
# Description

CI was failing to build and publish docker images. Yesterday there was a GH outage which caused some issues which also seems to have been a time where they updated their docker/buidx version. As a result the output of the bake file we were using changed and our script to extract the appropriate targets was incorrect. This update should resolve the issue to match the new output.

NOTE: not sure how we would be notified of buildx output format changes but this is a possible issue for future as well if it changes again.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -46,12 +46,13 @@ jobs:
       - name: Create matrix
         id: targets
         run: |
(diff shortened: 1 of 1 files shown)


YohannParis (approved):
Thank you 🙏 
If it happens again without warning, it might be time to move to Circle CI.

PR #507, opened 2023-01-20
feat: use asset-card component in cart and preview

mj3cheun:
# Description

* extract asset card component from search-item and use it for preview and cart
* add context menu to cart asset card allowing for deletion


YohannParis (approved):

dgauldie (approved):

YohannParis (approved):
Thanks @shawnyama

PR #510, opened 2023-01-20
feat(hmi-server): rowlimit added for csv preview

### LINKED ISSUE #464: [TASK]: add support for limiting the download of datasets for preview to a reasonable number of records
**Describe the task**
Fetching datasets from TDS returns the full dataset which can be large and cause app freeze. The current implementation is done via the function `downloadRawFile` in dataset.ts.
TDS has recently added a query parameter to control the max number of records to return, via a parameter `row_limit` which should be added to the API and used with some reasonable default by the HMI client. 
Remember the main use of downloading datasets at the moment is for preview purposes so downloading the full dataset is not required!

**Task Items**
- [ ] update the API for downloading datasets with a specified limit
- [ ] update the client side to use the updated API with some reasonable limit value

**Goal**

**Screenshots**

**Additional context**

dvince2:
Adding the `row_limit` to the csv download api to limit the number of rows shown in the front end to not lock up the application.

Resolves #464


amostafa-uncharted (approved):
LGTM

PR #514, opened 2023-01-23
feat: add sticky footer to bottom of all pages

echl:
image


packages/client/hmi-client/src/App.vue
```diff
@@ -118,4 +121,17 @@ main {
 .data-explorer {
 	z-index: 3;
 }
+
+footer {
+	width: 100%;
+	height: 48px;
+	background-color: #344054;
```
YohannParis, line 128: Could you use a CSS property for it? One of the `--grey-X0`

YohannParis (approved):
LGTM

packages/client/hmi-client/src/App.vue
echl: we don't have an existing property but I will create one

echl: or do you just want to use an existing colour? I grabbed the colour directly from figma

YohannParis: `--gray-800` should be the closest one.

YohannParis (approved):

Issue #528, opened 2023-01-24
[BUG]: Unable to add a paper to a project


hvasque14:
**Describe the bug**
I tried adding a paper from my cart to my project. When I select the paper and add to "my current project" or one from the list, when I go to the project section, the paper is not in the project. 

**To Reproduce**
Add a paper to the cart
Add the paper in the cart to the project
Go to the project
Paper is not in the list

**Expected behavior**
When I add a paper from my cart to my project, I should be able to access/see the paper in my project. 

**Screenshots**

**Additional context**

PR #532, opened 2023-01-24
chore: correct naming for expected output

Tom-Szendrey:
# Description
Changing publication's naming to allow us to hit the /external/publications POST with the expected JSON formatting

https://user-images.githubusercontent.com/17088680/214401897-a077e3dc-f92d-4fd8-86fb-ebb35f79991d.mov

Resolves #(issue)
#528


dvince2 (approved):
lgtm

When the dust settles here we'll have to see how to have a nicer look for the front end but still retain the correct field info for the services we use.

PR #534, opened 2023-01-25
feat: add request logging for document-service

chris-dickson:
This adds the standard request/response logging to the document-service so we can see the proxied requests being sent to it


dgauldie (approved):
LGTM

PR #558, opened 2023-01-26
style: adjust nav dropdown/suggested terms and add cart number

shawnyama:
# Description

- Amount of items in cart is shown

- nav dropdown is adjusted

- Suggested terms list works and is restyled. 
 
There is a more backend related issue that should be solved in a future PR. When switching to a model or dataset you won't see suggested terms (as you should) but when you switch back to papers you won't see them anymore because the suggested terms are called along with the papers. Since the papers are already here the suggested terms are assumed to be here too but they are not. New suggested terms will only appear if something is freshly searched.

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -37,7 +37,7 @@ function updateSearchBar(newQuery) {
 }
 
 function updateRelatedSearchTerms(newTerms) {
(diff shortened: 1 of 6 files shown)


YohannParis (approved):
Looks great 🎉

YohannParis (dismissed):

mwdchang (approved):

Issue #523, opened 2023-01-24
[BUG]: Facet result numbers are not accurate


hvasque14:
**Describe the bug**
Facet result numbers

**To Reproduce**
Search for a term (e.g., COVID hospitalizations)
Review the side panel (e.g., publications for 2020 are 0 of 64697).

**Expected behavior**
Results should add up. 0/64697 should just be 0.

**Screenshots**
<img width="238" alt="image" src="https://user-images.githubusercontent.com/116816354/214345557-9334ca54-aebc-4829-a1ff-a5156dbdd989.png">

**Additional context**

Issue #521, opened 2023-01-24
[BUG]: Filtering on Facets not working


hvasque14:
**Describe the bug**
Results do not filter when facet is applied.

**To Reproduce**
I've searched for covid hospitalizations. When I select papers from 2021, I still see results from 2020. 

**Expected behavior**
When a facet is selected this should filter the results.

**Screenshots**
<img width="1436" alt="image" src="https://user-images.githubusercontent.com/116816354/214342359-f852cfa4-6b08-4fed-9c4b-bc4d6ec2c333.png">

**Additional context**

Issue #497, opened 2023-01-19
[TASK]: Update Facets for xdd articles on explorer page.


Tom-Szendrey:
Facets should be created and displayed for papers the same way they are currently for models and datasets
Only show facets relevant to the results we are currently looking at. we can make these on our side (should be HMI, will be frnt end for now). Clicking on a facet should not call xdd again but just filter our current results we are looking at

<img width="708" alt="image" src="https://user-images.githubusercontent.com/17088680/213559558-eebcb5d5-6977-4779-b5cb-c8282352aba8.png">

<img width="630" alt="image" src="https://user-images.githubusercontent.com/17088680/213559597-1dc33237-d587-4894-9172-c91914377fad.png">

PR #559, opened 2023-01-26
feat: 497 task update facets for xdd articles on explorer page

Tom-Szendrey:
# Description
Correcting colour on facets
Correct Model + Dataset usage of facets when clicked
Update Facets for Papers tab to display similar to other two tabs

Resolves #(issue)
#523 
#521
#497


packages/client/hmi-client/src/components/facets/categorical-facet.vue
```diff
@@ -232,7 +232,9 @@ export default defineComponent({
 });
 </script>
 
-<style scoped>
+<style lang="scss" scoped>
+@import '@/styles/variables.scss';
```
echl, line 236: You can probably remove this now since you don't seem to be using this file? Or if this is needed for `facets-overrides.scss` you can import that file instead.

packages/client/hmi-client/src/components/facets/numerical-facet.vue
```diff
@@ -157,7 +157,9 @@ export default defineComponent({
 });
 </script>
 
-<style scoped>
+<style lang="scss" scoped>
+@import '@/styles/variables.scss';
```
echl, line 161: You can probably remove this now since you don't seem to be using this file? Or if this is needed for `facets-overrides.scss` you can import that file instead.

packages/client/hmi-client/src/views/DataExplorer.vue
```diff
@@ -345,12 +345,19 @@ const executeSearch = async () => {
...
 		}
 	});
-	const modelSearchParams = searchParamsWithFacetFilters?.[ResourceType.MODEL] || {
-		filters: clientFilters.value
-	};
-	const datasetSearchParams = searchParamsWithFacetFilters?.[ResourceType.DATASET] || {
-		filters: clientFilters.value
-	};
+
+	let modelSearchParams;
```
echl, line 349: Can we use the previous terser syntax here?

packages/client/hmi-client/src/components/facets/categorical-facet.vue
Tom-Szendrey: This file is required at the moment for variables ect inside the override. This is based off of the WM design, so not sure if in the future we will be overriding completely differently

packages/client/hmi-client/src/components/facets/numerical-facet.vue
Tom-Szendrey: Same as above

packages/client/hmi-client/src/views/DataExplorer.vue
Tom-Szendrey: The previous syntax is what was breaking the use of facets previously.
It looks cleaner but is not correct

What is says is: if searchParamsWithFacetFilters?.[ResourceType.DATASET]
What we want is: if searchParamsWithFacetFilters?.[ResourceType.DATASET]?.filters

If the if/else statements are too ugly I am happy to try a cleaner approach. Let me know if you have suggestions!

echl: Does 
```
const modelSearchParams = searchParamsWithFacetFilters?.[ResourceType.DATASET]?.filters || {
		filters: clientFilters.value
	};
``` work?

Tom-Szendrey: No, then we are assigning the type Filters to the type SearchParameters.

echl: Ah okay, so the two different possible values have different types?

echl (approved):

PR #560, opened 2023-01-26
feat(hmi-server): exception handling

dvince2:
There is a new class here, `HmiResponseExceptionMapper` which is hooked up to `ProjectProxy`.  It right now is more or less a placeholder class which will throw a java exception if the response is an http error 400 or higher.  In that class I have seperate paths for 400s and 500s as an example to show that we can do specific things for specific errors (for example, different exception types could be thrown for a 404 error vs a 500).

The `HomeResource` endpoint was updated to take advantage of these new exceptions being thrown...

1. The initial call to `ProjectProxy#getProjects` now catches an exception and returns a `Response.Status.NO_CONTENT` to the front end instead of bubbling up the initial 400+ error. There is also a record of the failure made to hmi-server's log.

1. Later on, inside of a for loop we ask for all of the publication assets of each project.  Previously a failure would have broken this loop. Now when the assets for a given project can't be given that specific project is just skipped over and we log out to hmi-server's log.

There is still a discussion to be had about how the front end should let users know about some errors, etc, but what I'm looking for here is if this the approach we want to take on the back end and any feedback on taht

Previously, the 400/500s errors would bubble straight up to the front end or lock up our back end as we weren't handling anything but success. By translating these HTTP errors to java exceptions we can do better error handling on the server and make better decisions about what to pass to the front end in these cases.

For use, developers would need to include `@RegisterProvider(HmiResponseExceptionMapper.class)` on the proxy classes to convert http errors to exceptions.  The `HmiResponseExceptionMapper` class may need future expansion or we may need different exception mappers based on our goals.
The resource classes would then have to handle these exceptions and return what is needed to the front end.

Resolves #no_issue


Tom-Szendrey (approved):
Thank you for all of the cleanup along with the logging!

chris-dickson (changes requested):
On hold until a filled in description for the PR is provided.

tangopapatime (approved):
Looks good.  As you mentioned, an additional discussion or two about error handling and codes is needed which should also be properly documented :P

chris-dickson (approved):

Issue #569, opened 2023-01-27
[FEAT]: apply facets as filters to what is being currently searched


shawnyama:
**The problem.**
Right now when a facet is chosen a new search query is being called. Instead a facet should just be filtering the search results we already have. PrimeVue chips similar to the ones in Suggested Terms should be added in the header (beside Showing (x) items) to see the chosen facets and remove them.

PR #599, opened 2023-01-30
feat: 569 apply facets as filters to what is being currently searched

Tom-Szendrey:
# Description
Upon clicking facets for papers do not call XDD to requery, Instead just filter what results are already grabbed and update new facets on left page

https://user-images.githubusercontent.com/17088680/215792231-42630d48-4b2a-4c0a-95f3-ea666405d5f7.mov

Resolves #(issue)
#569


YohannParis:
Looks great, but why do you have a loading spinner if the objects are filtered locally?

packages/client/hmi-client/src/services/data.ts
```diff
@@ -158,96 +192,137 @@ const getAssets = async (
...
-							(conceptResult) =>
-								conceptResult.name === conceptNameOrCurie ||
-								conceptResult.curie === conceptNameOrCurie
-						);
-						// update the clause value by mapping concept/curie to model id
-						clauseValues.push(...matching.map((m) => m.id));
-						curies.push(...matching.map((m) => m.curie));
+	if (searchParam) {
+		// Filtering for model/dataset data
+		if (searchParam && searchParam.filters && searchParam.filters.clauses.length > 0) {
```
YohannParis, line 197: You can use lodash [`isEmpty()`](https://lodash.com/docs/4.17.15#isEmpty)
```js
...
if (!isEmtpy(searchParam?.filters?.clauses) {
    ...
```

Tom-Szendrey: Awesome! Ill fix this
Thanks for the tip I will have to keep this library in mind

Tom-Szendrey: Unfortunately I still need the first two arguments for the code inside to not freak out thinking the object could be empty.

Tom-Szendrey:
> Looks great, but why do you have a loading spinner if the objects are filtered locally?

I had not touched the loader spinner. It exists for the models and dataset filtering as well. The other two are smaller objects and smaller sets. They may also be filtered more efficiently im not certain
Is this something you want me to remove?

packages/client/hmi-client/src/services/data.ts
```diff
@@ -158,96 +192,137 @@ const getAssets = async (
...
-			const finalAssetIDs = assetResults.map((m) => m.id);
-			conceptFacets.results.forEach((conceptFacetResult) => {
-				if (finalAssetIDs.includes(conceptFacetResult.id)) {
-					curies.push(conceptFacetResult.curie);
-				}
-			});
-		}
+			// re-create the concept facets if the user has applyied any concept filters
+			const uniqueCuries = uniq(curies);
+			if (uniqueCuries.length > 0) {
```
shawnyama, line 261: I guess you could use the `isEmpty` in these other checks too

shawnyama (commented):
Lookin good

shawnyama (approved):
I'll see about dealing with the spinner on my end

PR #606, opened 2023-01-31
fix: navbar tweak

### LINKED ISSUE #589: [BUG]: Top nav suggested terms line should not get clipped
The suggested terms line in the top nav should never get clipped, even if the badges are long.

It looks to me like the suggested terms are in the same div as the search field. They should be in a separate row so they can be as wide as they need to be.

CURRENT:
suggested terms line should not get clipped

DESIGN:
suggested terms line should not get clipped - DESIGN

YohannParis:
# Description

* Menu dropdown padding
* Change hover state on list of assets
* Dataset Preview mistake solve no 😞 
* Add a grid layout for the App.vue, tired of playing with flex
* Fixed some CSS issues with the Suggested terms
* Simplify the Suggested terms API call
* Move the Suggested terms to the navbar and out of the search bar
* Fix single space highlight on search 
* Make the footer white
* Make the side-panels narrowers
* Make the header-height fixed and remove unused css property

fix #589


packages/client/hmi-client/src/components/Navbar.vue
```diff
@@ -96,14 +99,18 @@ function updateProjectNavItem(id, name) {
...
+			icon: 'pi pi-clone',
 			routeName: RouteName.ProjectRoute
 		}
 	};
 	navItems.value = { ...initialNavItems, ...projectNavItem };
 	selectedPage.value = navItems.value[currentRoute.value.path];
 }
 
+function closeLogoutDialog() {
+	isLogoutDialog.value = false;
```
echl, line 111: In my opinion we should do this assignment on the cancel button's `onClick` since it is only one line and only used once.

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -57,17 +53,15 @@ const execSearch = () => {
 // 	emit('toggle-search-by-example');
 // };
 
+function addToQuery(term: string) {
+	query.value = query.value ? query.value.concat(' ').concat(term).trim() : term;
+}
+defineExpose({ addToQuery });
```
YohannParis, line 59: Mind exploded 🤯 
Never seen this, will read about it, genius,

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -16,97 +15,79 @@
...
+
+const emit = defineEmits(['query-changed', 'toggle-search-by-example']);
+
+const route = useRoute();
+const resources = useResourcesStore();
 
 const props = defineProps<{
-	text?: string;
-	suggestedTerms?: string[];
+	text: string;
```
shawnyama, line 46: `text` prop isn't used anywhere

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -16,97 +15,79 @@
...
 	resultType: string;
```
shawnyama, line 47: I want all the `resultType`s to be named `resourceType` eventually. Also you can make it of type `ResourceType`

YohannParis: @edwin what is the `resultType` doing?

packages/client/hmi-client/src/components/Navbar.vue
```diff
@@ -238,7 +258,46 @@ i {
...
+	white-space: nowrap;
+}
+
+.suggested-terms,
+.p-chip {
+	font-size: small;
+	font-weight: bold;
+	color: var(--text-color-subdued);
+}
+
```
shawnyama, line 279: Not sure how `p-chip` styles are back in here. Remove them and for the above style move those into just `.suggested terms`

shawnyama:
There is still this issue if the chips are maxed out.
image

According to Jamie's design the terms should be in their own div: https://github.com/DARPA-ASKEM/TERArium/issues/589

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
echl: `resulttype` is passed from `dataexplorer`, it's whether or not the user is on papers, models, or datasets; for the latter two, there is no call to the suggestions api since it requires a dataset parameter, which is not present for the latter two

shawnyama (commented):
Randomly while typing in the searchbar I can't input anymore text, I have to click the bar again to get it back in focus. Can't find why yet

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
```diff
@@ -1,88 +1,67 @@
...
-	searchText.value = searchText.value ? searchText.value.concat(' ').concat(term) : term;
-	// @ts-ignore
-	inputElement.value?.$el.focus();
+// const toggleSearchByExample = () => {
+// 	emit('toggle-search-by-example');
+// };
+
+function addToQuery(term: string) {
+	query.value = query.value ? query.value.concat(' ').concat(term).trim() : term;
+	execSearch();
```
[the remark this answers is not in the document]
YohannParis: This is how I sees it, otherwise what is the point, and it doesn't seems intuitive to click enter again.

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
YohannParis: ... this will need to be reviewed with @pascaleproulx and @hvasque14, because that's an UX issue.

YohannParis: @shawnyama I will not change this, because resources and results are still.

shawnyama:
> > There is still this issue if the chips are maxed out. image
> > According to Jamie's design the terms should be in their own div: #589
> 
> This is a really tricky one to solve... @jamiewaese-uncharted wants the suggested-terms to be centered with the off-centered search bar, but want it to expand to the whole page width.

Oh the search bar is supposed to be off-center now. I attempted this in my PR you closed. The issue with my attempt was that when you remove a chip everything moves to the center which isn't good either. I did get it anchored to the left if it overflows though.

YohannParis:
> > > There is still this issue if the chips are maxed out. image
> > > According to Jamie's design the terms should be in their own div: #589
> > 
> > 
> > This is a really tricky one to solve... @jamiewaese-uncharted wants the suggested-terms to be centered with the off-centered search bar, but want it to expand to the whole page width.
> 
> Oh the search bar is supposed to be off-center now. I attempted this in my PR you closed. The issue with my attempt was that when you remove a chip everything moves to the center which isn't good either. I did get it anchored to the left if it overflows though.

Yep, I need to talk to @jamiewaese-uncharted because this design doesn't work on so many level.

packages/client/hmi-client/src/App.vue
```diff
@@ -4,7 +4,7 @@
 		:active="!isErrorState"
 		:project="project"
 		:query="searchBarText"
-		:resultType="resultType"
+		:resourcesType="resourcesType"
 	/>
```
shawnyama, line 8: `resourceType` not `resourcesType` 🥲

packages/client/hmi-client/src/components/data-explorer/search-bar.vue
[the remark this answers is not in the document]
echl: If you want to add multiple search terms before running the search would be the case I can think of. @YohannParis

packages/client/hmi-client/src/components/Navbar.vue
```diff
@@ -116,120 +176,81 @@ watch(activeProjectName, (newProjectName) => {
...
-	</header>
-</template>
+/*
+ * Search
+ */
+const terms = ref<string[]>([]);
+
+async function queryChanged(q: string | null) {
+	// Empty the related terms when the query is over
+	if (!q) terms.value = [];
```
echl, line 187: In this pr: https://github.com/DARPA-ASKEM/TERArium/pull/615/files I made `getRelatedWords` just return an empty array if the term is blank, so we won't need to do these conditional checks.

packages/client/hmi-client/src/components/Navbar.vue
```diff
@@ -238,7 +259,27 @@ i {
...
+.suggested-terms {
+	align-items: center;
+	color: var(--text-subdued);
+	display: flex;
+	column-gap: 0.5rem;
+	font-size: var(--font-caption);
+	grid-area: suggested-terms;
+	justify-content: center;
+	margin-top: 0.5rem;
+	overflow: hidden;
```
echl, line 272: Would `overflow: visible` fix the clipping issue?

packages/client/hmi-client/src/components/Navbar.vue
YohannParis: I agree, but I know from that point on that if the query is empty I can return an empty array.
So better do it here that trusting a function that might change later.

mwdchang (approved):

PR #610, opened 2023-01-31
feat: suggested terms and preview are back and facet chips for easy removal

shawnyama:
# Description

Edited `data.ts` to return suggested terms and asset previews once again. 
Facet chips for searching is now available.


YohannParis (approved):
Well done 🎉

PR #613, opened 2023-01-31
feat: reformat suggested terms

### LINKED ISSUE #589: [BUG]: Top nav suggested terms line should not get clipped

shawnyama:
# Description

Suggested terms are reformatted to fill up to width of the screen

Resolves #589


YohannParis:
Sorry I'm working on this one already with @jamiewaese-uncharted... my bad.

PR #615, opened 2023-01-31
fix: links in reference should be clickable; fix searchbar losing focus when autocomplete occurs

### LINKED ISSUE #557: [BUG]: Links in reference should be clickable
**Describe the bug**
When reviewing references in the preview panel, URLs should be clickable so that the user does not need to copy and paste. 

**To Reproduce**
Open the preview panel for a paper, open the references, references with URLs are not clickable

**Expected behavior**
References with URLs should be clickable

**Screenshots**
image

**Additional context**

### LINKED ISSUE #629: [BUG]: Searchbar loses focus after autocomplete.
Searchbar loses focus after autocomplete.

echl:
https://user-images.githubusercontent.com/16358728/215898232-ddcc3bce-9ae1-43d4-9db8-03f76ec53ef5.mov

Resolves #557 
Resolves #629


packages/client/hmi-client/src/components/articles/Document.vue
```diff
@@ -253,8 +253,13 @@ const formatCitation = (obj: { [key: string]: string }) => {
 	let citation: string;
 	if (Object.keys(obj).length <= 1) {
 		citation = obj.unstructured_citation ?? '';
+		citation = citation.replace(
+			/\bhttps?:\/\/\S+/,
+			`<a style='color: var(--text-color-subdued);' href=$&>$&</a>`
```
YohannParis, line 258: Really neat, I like when RegEx are nicely implemented.

It will be nicer to not add styling within the JS, and just add a class, or target the link within the AccordionTab directly.
```suggestion
			`<a href="$&">$&</a>`
```

YohannParis (approved):

packages/client/hmi-client/src/components/articles/Document.vue
echl: I agree about styling, but that styling is already present in the css for this component and did not seem to get applied here (maybe because the html is dynamically generated?), so I had to add it inline.

YohannParis: You might have to add a `>>> a`  selector somewhere.

YohannParis (approved):
Great job!

PR #621, opened 2023-02-01
feat: clickable chips

shawnyama:
# Description

Chips are clickable, I like this approach since it keeps editing the query string more in the search-bar component


YohannParis (approved):
Amazing

PR #628, opened 2023-02-02
feat: accordion headers show how many assets are under them

### LINKED ISSUE #592: [BUG]: Accordion label should have badge to indicate number of items inside
Accordion labels should have a badge to indicate the number of items they contain.

The two exceptions to this are the Abstract section and the Provenance sections, which will only have one block of contents.

Also, when an accordion with zero contents is opened, the container should have some helper text to indicate that it is empty so the user doesn't wonder if something is broken.

CURRENT:
add helper text to empty accordions

TO BE:
add helper text to empty accordions

shawnyama:
# Description

Can't seem to make the number have `--text-color-subdued` because the header string is put into a header parameter within the AccordianTab PrimeVue component. Instead of having the No References found message I decided to not show the tab if there are none. This keeps things consistent since other tabs aren't shown if they have no artifacts.

You'll see Yohann's commits because I originally branched this off of `navbar-tweak` (in that branch I could see the preview properly). I merged it back with main though

image

Resolves #592


shawnyama:
> 
> If you want real controls of the header, try [pseudo-elements `::after`](https://developer.mozilla.org/en-US/docs/Web/CSS/::after)
> 

See my attempt to do this, I have trouble accessing the proper child (`.p-accordion-header-text`). 

image

The parameter is inserted in the parent div (`p-accordion-tab`). When I do it this happens:

image

Maybe if we use one of the AccordionTab properties here https://primevue.org/accordion it wouldn't give us this warning message. Though I am not sure exactly what they accept syntax wise for an object, HTMLElement etc.

YohannParis (approved):

YohannParis (approved):
LGTM

shawnyama:
> LGTM

It doesn't work yet...

YohannParis (approved):

PR #634, opened 2023-02-02
refactor: solidify searchbar component

### LINKED ISSUE #633: Refactor API calls to be handled by `services/data.ts` directly within the search bar component
make sure that the services/data.ts API call used by the SearchBar.vue are called by this component. As opposed of the calls made another component (i.e. DataExplorer.vue) and their results sent via props.

shawnyama:
# Description

Moved a few functions from `data.ts` to `search-bar.vue`. Also did some cleanup

Resolves #633


YohannParis:
Sorry, what I meant is to make sure that the `services/data.ts` API call used by the `SearchBar.vue` are called by this component. As opposed of the calls made another component (i.e. `DataExplorer.vue`) and their results sent via props.

PR #677, opened 2023-02-08
feat(document-service): adding chaching to xDD sets and dictionaries

### LINKED ISSUE #374: [TASK]: Refactor static xDD calls to improve the Explorer performance
**Describe the task**
The data explorer performs various xDD related calls to fetch mostly static data such as the list of dictionary names and datasets used as indexing collections. 
This task is all about moving the fetch of such calls to the backend and cache them, which will improve the overall performance. see getXDDDictionaries  and getXDDSets at data.ts

**Task Items**
- [ ] move common and mostly static xDD calls to HMI server
- [ ] Let the project or the data explorer fetch those lists once and re-use them by avoiding repeated fetch

dvince2:
Adding caching to xDD sets and dictionaries.  Adding some super basic exception/error handling to invalidate the cache (as that doesn't happen automatically).  It will be part of the exceptions task later to fill these out more.  Cash will invalidate once a day.

# Description

Resolves #374

### DIFF
--- a/packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/DictionariesResource.java
+++ b/packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/DictionariesResource.java
@@ -5,6 +5,7 @@ import io.quarkus.cache.CacheResult;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
+import software.uncharted.terarium.documentserver.caching.CacheClearService;
 import software.uncharted.terarium.documentserver.responses.xdd.XDDDictionariesResponseOK;
 import software.uncharted.terarium.documentserver.responses.xdd.XDDResponse;
 import software.uncharted.terarium.documentserver.proxies.xdd.DocumentProxy;
@@ -28,18 +29,10 @@ public class DictionariesResource {
 	@Produces(MediaType.APPLICATION_JSON)
 	@Tag(name = "Get available XDD dictionaries")
 	@Path("/dictionaries")
-	@CacheResult(cacheName = CACHE_NAME)
+	@CacheResult(cacheName = CacheClearService.CacheName.Constants.XDD_DICTIONARIES_NAME)
 	public XDDResponse<XDDDictionariesResponseOK> getAvailableDictionaries(@QueryParam("all") @DefaultValue("") String all) {
-		try {
-			XDDResponse<XDDDictionariesResponseOK> response = proxy.getAvailableDictionaries(all);
-		} catch (RuntimeException e) {
-			log.error("Unable to get dictionaries. Invalidating cache.", e);
-			invalidateXDDCache(all);
-		}
-		return null;
-	}
+		return proxy.getAvailableDictionaries(all);
 
-	@CacheInvalidate(cacheName = CACHE_NAME)
-	public void invalidateXDDCache(@QueryParam("all") @DefaultValue("") String all) {
 	}
+	
 }
--- a/packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/DictionariesResource.java
+++ b/packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/DictionariesResource.java
@@ -1,5 +1,8 @@
 package software.uncharted.terarium.documentserver.resources.xdd;
 
+import io.quarkus.cache.CacheInvalidate;
+import io.quarkus.cache.CacheResult;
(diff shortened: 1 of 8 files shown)


packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/DictionariesResource.java
```diff
@@ -22,7 +28,18 @@ public class DictionariesResource {
 	@Produces(MediaType.APPLICATION_JSON)
 	@Tag(name = "Get available XDD dictionaries")
 	@Path("/dictionaries")
+	@CacheResult(cacheName = CACHE_NAME)
 	public XDDResponse<XDDDictionariesResponseOK> getAvailableDictionaries(@QueryParam("all") @DefaultValue("") String all) {
-		return proxy.getAvailableDictionaries(all);
+		try {
+			XDDResponse<XDDDictionariesResponseOK> response = proxy.getAvailableDictionaries(all);
+		} catch (RuntimeException e) {
+			log.error("Unable to get dictionaries. Invalidating cache.", e);
```
chris-dickson, line 36: I don't think that's what you want here. I'm guessing if this behaves like other caches that it will not cache the result if the wrapped method throws an exception.

packages/services/document-service/src/main/resources/application.properties
```diff
@@ -18,3 +18,8 @@ quarkus.native.builder-image=quay.io/quarkus/ubi-quarkus-native-image:22.3.0-jav
 ########################################################################################################################
 xdd-document-service/mp-rest/url=https://xdd.wisc.edu
 xdd-extraction-service/mp-rest/url=https://xdddev.chtc.io
+########################################################################################################################
+# Caching (these names correspond to names defined in their java classes - don't change)
+########################################################################################################################
+quarkus.cache.caffeine."xDD-Dictionaries".expire-after-write=86400
```
chris-dickson, line 24: See my comment above on how to manage this.  This will likely become fragile over time.  Additionally, we'll likely want a fixed interval to clear all data caches as having each result expire after a given amount of time could lead to incorrect results.  
eg/ sets is expired and returns updated values and a subsequent call to documents returns a cached value.

packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/SetResource.java
```diff
@@ -1,29 +1,44 @@
...
 import javax.ws.rs.core.MediaType;
 
 @Path("/sets")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
 @Tag(name = "XDD Sets REST Endpoint")
+@Slf4j
 public class SetResource {
 
+	private static final String CACHE_NAME = "xDD-Sets";
```
chris-dickson, line 21: I would move these to an enum so you can do grouped cached clearing. Eg:
```java
public enum CacheCollection {
	XDD
}

public enum CacheName {
	XDD_SETS("xdd-sets", XDD),
	XDD_DICTIONARIES("xdd-dictionaries", XDD);


	@Getter
	private String name;
	@Getter
	private CacheCollection collection;

	CacheName(String name, CacheCollection collection) {
		this.name = name;
		this.collection = collection;
	}
	
	public static List<CacheName> findAllByCollection(CacheCollection collection) {
		return Stream.of(values())
			.filter(cacheName -> cacheName.collection.equals(collection))
			.collect(Collectors.toList());
	}
}
```

Then a simple scheduled method:
```java
@ApplicationScoped
public class CacheClearService {
  @Inject
  CacheManager cacheManager;

  @Scheduled(cron="0 * * * *") // every hour on the hour
  void clearXDDCaches(ScheduledExecution execution) {
        CacheNames.findAllByCollection(XDD).forEach(cacheName -> {
           cacheManager.clear(cacheName.getName());
        }
    }
```

chris-dickson (changes requested):

packages/services/document-service/src/main/resources/application.properties
YohannParis: We should set caching to only an hour for now until we think it's a viable solution.
Or maybe this could be an ENV variable so it's different between `local`, `staging`, and `production`.

packages/services/document-service/src/main/java/software/uncharted/terarium/documentserver/resources/xdd/SetResource.java
dvince2: Updated.  There are some rules about using enums in annoation names that required the Strings to be stored separately, however, it is still way less fragile.

Tom-Szendrey:
Is this still required with the docu service being taken out back?
https://github.com/DARPA-ASKEM/TERArium/pull/677/files

Tom-Szendrey (approved):

Issue #679, opened 2023-02-08
[BUG]: Filtering by facets removes the options


YohannParis:
**Describe the bug**
As of know if you filter by facets, the unselected options disapears.
Now that we have keep track of the facets results, let's keep them displayed.

**Screenshot**
![Screen Shot 2023-02-08 at 11 32 13](https://user-images.githubusercontent.com/636801/217592617-fe2d6c9e-727a-412b-9dd7-10beed4c4bdf.png)

PR #682, opened 2023-02-08
fix: correcting return data for document facet

Tom-Szendrey:
# Description

image

    Changed how we are returning data after filtering by a document facet.
     Previously the facets bar would mistakenly remove facets that had no relevance to selected facets

Resolves #(issue)
#679


YohannParis:
Can you fill the PR with your changes, even one liner is better than the default text.

YohannParis (approved):
LGTM

PR #686, opened 2023-02-09
feat: converted console.log to use the new logging service

tangopapatime:
# Description

Converted almost* all console logs so that they're sent to the backend and not displayed in prod/staging.  In development, these logs should also remain in the console.

Resolves #(issue)

https://github.com/DARPA-ASKEM/TERArium/issues/624
https://github.com/orgs/DARPA-ASKEM/projects/3?pane=issue&itemId=17790439


YohannParis (commented):
I wonder if this will be better achieved with [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties)?

Like that there is no need to import `loggers` everywhere and we can just use `$logger()` when needed.

packages/client/hmi-client/src/stores/auth.ts
```diff
@@ -92,7 +92,7 @@ const useAuthStore = defineStore('auth', {
 			window.location.assign('/logout');
 		},
 		autoRenew() {
-			console.log('RENEW SSO');
+			console.info('RENEW SSO');
```
dvince2, line 95: Is this meant to stay `console` or should it be logger?

tangopapatime:
> I wonder if this will be better achieved with [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties)?
> 
> Like that there is no need to import `loggers` everywhere and we can just use `$logger()` when needed.

It certainly can.  Will do that.

packages/client/hmi-client/src/services/event.ts
```diff
@@ -21,7 +22,7 @@ async function get(
 		const response = await API.get(`/events?${toQueryString(params)}`);
 		return response?.data ?? [];
 	} catch (e) {
-		console.log('Error fetching events: ', e);
+		logger.error(`Error fetching event: ${e}`);
```
dvince2, line 25: Is there a significance in using ` here as a quotation mark?  I see a mix of that and ' in this PR, and in some cases like this one, turning one into the other.

tangopapatime: This is just es7 string literals?

packages/client/hmi-client/src/stores/auth.ts
YohannParis: This is only for the front-end to now that the SSO has been renewed.
I don't think we need to call it anymore. @bigglesandginger what do you think?

packages/client/hmi-client/src/services/event.ts
YohannParis: It's because we need them for [Template Literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

packages/client/hmi-client/src/components/projects/NewProjectCard.vue
```diff
@@ -19,7 +18,7 @@ async function createNewProject() {
 	const author = auth.name ?? '';
 	const project = await ProjectService.create(name.value, description.value, author);
 	if (project) {
-		router.push(`/projects/${project.id}`);
+		logger.info(`Project ${project.title} created`);
```
mwdchang, line 21: You removed `router.push` logic, is this intentional?

mwdchang:
Sorry I don't quite understand the description. `logger.[info | warn | error\` does not send anything to the backend API right?  As most of these console.logs are not events we want to capture.

tangopapatime:
> Sorry I don't quite understand the description. `logger.[info | warn | error\` does not send anything to the backend API right? As most of these console.logs are not events we want to capture.



> I wonder if this will be better achieved with [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties)?
> 
> Like that there is no need to import `loggers` everywhere and we can just use `$logger()` when needed.

I can't seem to get this to work as expected when trying to add to the global properties.  Anyone know what might gives?

`app.config.globalProperties.$logger = ...` but I can't seem to access $logger in any capacity.

tangopapatime:
> Sorry I don't quite understand the description. `logger.[info | warn | error\` does not send anything to the backend API right? As most of these console.logs are not events we want to capture.

Right now it captures all these logs into a buffer to be sent back to the server.  I agree that a lot of these are likely not desirable.

packages/client/hmi-client/src/components/projects/NewProjectCard.vue
tangopapatime: no not all.  didn't mean to stage this change.  thx.

tangopapatime:
> > I wonder if this will be better achieved with [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties)?
> > Like that there is no need to import `loggers` everywhere and we can just use `$logger()` when needed.
> 
> It certainly can. Will do that.

Definitely spoke too soon.  As I mentioned yesterday, using global properties for this purposes seems to be frowned upon in VueJS3.  You can do it by loading the app context and pulling on the variable itself but it's the last thing from elegant lol.

tangopapatime:
> Okay, I agree with your initial judgement. Nuke the idea and revert to your original PR. Thank you for doing the research on the matter. It was a gret learning experience.

Don't need to revert.  The original PR was importing it incorrectly.  The right way to use the logger service is to do it like this.

```
import { useLogger } from 'vue-logger-plugin';

const logger = useLogger();```

This is the way :P  We can/should do a proper audit of messages we want the backend to receive.

YohannParis (approved):

PR #693, opened 2023-02-09
chore: correcting property name

Tom-Szendrey:
# Description
Correcting jsonbproperty name for expected output for frontend to work with
Allows home page to correctly pull relatedDocuments for example
image


dvince2 (changes requested):
Hmm this may actually be a problem.  I believe that `data-service` is giving this to us as `relatedArticles`.  The reason why its working right now is because `ProjectProxy.java` is returning `Response` instead of `Project`.  When that is fixed i don't believe the serialization will work any more.
This may have to use `@JsonAlias("relatedDocuments") to give us access to both?

dvince2 (approved):
looks good!

PR #698, opened 2023-02-09
feat: don't show empty detail panel

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/data-explorer/search-item.vue
+++ b/packages/client/hmi-client/src/components/data-explorer/search-item.vue
@@ -58,8 +58,8 @@ button {
 }
 
 button i {
-	padding: 0.2rem;
-	border-radius: 3px;
+	padding: 0.5rem;
+	border-radius: 2rem;
 	font-size: 14px;
 }
 


packages/client/hmi-client/src/components/widgets/Slider.vue
```diff
@@ -1,9 +1,9 @@
 <template>
 	<aside
 		:class="`slider ${isOpen ? 'open' : 'closed'} ${direction}`"
-		:style="{ width: isOpen ? contentWidth : tabWidth }"
+		:style="`width: ${isOpen ? contentWidth : tabWidth}`"
```
YohannParis, line 4: What's the difference here?

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/Slider.vue
YohannParis: No worries, it doesn't matter at the end. I was curious if it make things faster/better.

YohannParis (approved):

PR #705, opened 2023-02-10
feat: initial setup to allow for running/development in docker compose

mattprintz:
This appears to be fully functional but may have missed certain services.
This stack does not use Keycloak and just uses anonymous/dev users throughout.

# Description

Resolves #(issue)


docker-compose.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - QUARKUS_OIDC_ENABLED=false
+      - QUARKUS_KEYCLOAK_POLICY_ENFORCER_ENABLE=false
+      - QUARKUS_SECURITY_AUTH_ENABLED_IN_DEV_MODE=false
+      - QUARKUS_HTTP_HOST=0.0.0.0
+      - terarium-db-host=terarium-db
+      - terarium-db-port=5432
+      - data-service-host=data-service
+      - data-service-port=8000
+  terarium-db:
+    image: postgres:14.4-alpine
```
YohannParis, line 32: What is `terarium-db` is it what we call `hmi-postgres`?
Also we try to use `ghcr.io/darpa-askem/postgres:15.1` to avoid issue in the long run,

YohannParis (dismissed):
Works for me 🎉

mwdchang (dismissed):

docker-compose.yaml
mattprintz: Pretty sure I called it that because that's what it's called in the config file:

https://github.com/DARPA-ASKEM/Terarium/blob/main/packages/services/hmi-server/src/main/resources/application.properties#L19

YohannParis: I see, then, yes, they are the same thing!

YohannParis (approved):

PR #711, opened 2023-02-13
feat: model view resembles design and generalized styling between asset components

shawnyama:
# Description

image

- `Model.vue` resembles design now. Asset components now share the same stylesheet: `_asset.scss`.
- Show more/less styling is adjusted
- Adjusted some HTML to be more semantic
- Added a rule to `DEVELOPMENT.md` about enabling editing on asset components


packages/client/hmi-client/src/assets/css/theme/designer/components/data/_datatable.scss
```diff
@@ -25,20 +25,23 @@
...
         padding: $tableHeaderCellPadding;
         border: $tableHeaderCellBorder;
         border-width: $tableHeaderCellBorderWidth;
         font-weight: $tableHeaderCellFontWeight;
+        font-size: $fontCaption;
         color: $tableHeaderCellTextColor;
         background: $tableHeaderCellBg;
         transition: $listItemTransition;
+        text-transform: uppercase;
     }
```
mwdchang, line 42: What is the impact of this `uppercase` change? Do we have a screen shot? 
Asking because when you have math, equations and stuff when you make text uppercase it can mean a different thing than the lowercase counterpart.

shawnyama: The uppercase just makes sure that the headers on the table are uppercase.

image

packages/client/hmi-client/src/components/widgets/Slider.vue
```diff
@@ -4,7 +4,10 @@
 		:style="{ width: isOpen ? contentWidth : tabWidth }"
 	>
 		<div class="slider-content-container" :style="`width: ${contentWidth}`">
-			<section class="slider-content" :style="sidePanelContentStyle">
+			<section
+				class="slider-content"
+				:style="`height: ${hasFooter ? 'calc(100% - 5rem)' : '100%'}; ${sidePanelContentStyle}`"
```
mwdchang, line 9: Nitpick: Can the `100% - 5rem` thingy be incorporated into sidePanelContentStyle itself? It makes things a little easier if the logic is not scattered.

packages/client/hmi-client/DEVELOPMENT.md
```diff
@@ -67,3 +67,13 @@ Basic rules to write organised code.
...
     </style>
     ```
+
+* Some components (e.g. `Model.vue`) will showcase an asset's attributes and allows users to edit them. However there are cases where we just want to show a more compact read-only version of these components. In these cases assign the flag `:is-editable="false"` for these components:
+    ```html
+    <model 
+        :asset-id="previewItemId"
+		:project="resources.activeProject" 
+        :highlight="searchTerm"
+        :is-editable="false"
```
YohannParis, line 77: I don't think this prop needs to be reactive, and as a boolean it can be just passed as is or omitted.
https://vuejs.org/guide/components/props.html#prop-passing-details

packages/client/hmi-client/src/assets/css/theme/_asset.scss
```diff
@@ -0,0 +1,95 @@
+.asset {
```
YohannParis, line 1: Could you add a header comment explaining the purpose of this stylesheet.

Also, it shouldn't be in the `/theme` folder, that's for _PrimeVue_, you can place this file  next `css/styles.scss` and import it (like `reset.css`).

packages/client/hmi-client/src/assets/css/theme/_asset.scss
```diff
@@ -0,0 +1,95 @@
...
+                margin-bottom: 0.5rem;
+            }
+        }
+
+        p {
+            margin: 0.5rem 0;
+        }
+
+        strong {
+            font-weight: $fontWeightSemiBold;
```
YohannParis, line 59: I know you use SCSS to leverage nesting, and that's fine here, but we should really use only CSS custom properties.
That will make our life easier if in the future we can get rid of SCSS completely.

packages/client/hmi-client/src/components/documents/Document.vue
```diff
@@ -1,40 +1,36 @@
 <template>
-	<section class="doc-view-container" ref="sectionElem">
-		<div v-if="doc">
+	<section class="asset" ref="sectionElem">
+		<template v-if="doc">
```
YohannParis, line 3: This should be done at the parent level, not here.

packages/client/hmi-client/src/components/widgets/Slider.vue
YohannParis: Nope, the footer should be part of the slider, we need to add a footer slot and keep the design in here.

packages/client/hmi-client/src/assets/css/theme/designer/components/data/_datatable.scss
mwdchang: got it, thanks 👍

packages/client/hmi-client/DEVELOPMENT.md
shawnyama: Apparently not 
image

YohannParis (approved):

YohannParis (approved):

PR #712, opened 2023-02-13
feat: update homepage design

### LINKED ISSUE #645: [FEAT]: Move add new project button to the top right
Remove current `Add new project` button off the project list as a button to the top right button.

[Screen Shot 2023-02-06 at 14 07 56](https://www.figma.com/file/lOoGAsDwLfMmP3u5qzKHRh/TERArium-HMI-Designs---PrimeVue-version?node-id=8591%3A68849&t=kJHciFYHVvDyAkRQ-4)

### LINKED ISSUE #646: [FEAT]: Change design of project and papers section
Change the homepage design to match the new two section layout (white/projects, grey/papers)

[
Screen Shot 2023-02-06 at 14 09 08
](https://www.figma.com/file/lOoGAsDwLfMmP3u5qzKHRh/TERArium-HMI-Designs---PrimeVue-version?node-id=8591%3A68849&t=kJHciFYHVvDyAkRQ-4)

echl:
https://user-images.githubusercontent.com/16358728/219461710-9812f6c8-bced-4dca-bf95-5795b4e67d8a.mov

Resolves #645 
Resolves #646


YohannParis:
Did you updated from `main` first? It looks like the card are backs in your commits.

echl:
> Did you updated from `main` first? It looks like the card are backs in your commits.

Merged main!

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -28,6 +34,12 @@ const selectedDocument = ref<DocumentType>();
 
 const resourcesStore = useResourcesStore();
 const queryStore = useQueryStore();
+const router = useRouter();
+const auth = useAuthStore();
+
+const isNewProjectModelVisible = ref(false);
```
mwdchang, line 40: nitpick: Did you mean?
- newProjectModal, or
- newProjectMode

Not sure if there is a typo here or not.

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -102,81 +127,139 @@ const scroll = (direction: 'right' | 'left', event: PointerEvent) => {
...
-		</div>
-		<!-- Show related documents for the top 5 projects -->
-		<div v-for="(project, index) in projectsToDisplay" :key="index" class="carousel">
+			<TabView>
+				<TabPanel header="Recents">
+					<div class="carousel">
+						<i class="pi pi-chevron-left" @click="scroll('left', $event)" />
+						<i class="pi pi-chevron-right" @click="scroll('right', $event)" />
+						<ul>
+							<li v-for="(project, index) in projects.slice().reverse()" class="card" :key="index">
```
mwdchang, line 171: I realize this is copy-and-paste, but since we are doing reno in the area:
-This is not reliable, if you need a specific ordering it needs to be explicit. Neither the HMI-server nor TDS is guaranteeing ordering by created time. So we need to sort it by hand before variable assignment to "project", or the ordering needs to be requested as such.
- I suspect copying a reactive in this manner (via slice) may have quirky things down the road if we alter project properties/fields.

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -72,14 +84,27 @@ const scroll = (direction: 'right' | 'left', event: PointerEvent) => {
...
-	cardListElement.style.marginLeft = `${newMarginLeft > 0 ? 0 : newMarginLeft}rem`;
+	cardListElement.style.marginLeft = `${newMarginLeft > 0 ? 0.5 : newMarginLeft}rem`;
 };
+
+async function createNewProject() {
+	const author = auth.name ?? '';
+	const project = await ProjectService.create(
+		newProjectName.value,
+		newProjectDescription.value,
+		author
```
mwdchang, line 101: Notes for @YohannParis: we should decide if user attributes should be passed as part of the API input, or to be determined via the auth-token within the server itself.

packages/client/hmi-client/src/page/Home.vue
YohannParis: I agree, this should be a computed, I don't know if we want to order them by timestamp yet. Or check with @fivegrant if the timestamp for last-updated works.

packages/client/hmi-client/src/page/Home.vue
YohannParis: Chatted with @chris-dickson: the `hmi-server` should be doing it.

packages/client/hmi-client/src/page/Home.vue
echl: Good catch, should be `newProjectModalVisible`

packages/client/hmi-client/src/page/Home.vue
mwdchang: Yeah I thought as much. Let's keep a mental note of that.

packages/client/hmi-client/src/page/Home.vue
YohannParis: Good enough for now.
Ping @fivegrant about adding a new last_updated field, etc.

Also, now that the projects are available on App.vue, I don't think we should fetch them again on the home page.

mwdchang: TDS is ordering by id currently - typically you wouldn't want to rely on it or logical orderings.
https://github.com/DARPA-ASKEM/data-service/blob/main/tds/db/helpers.py#L34-L45


The easiest thing to do for now I think, if we want to sort it, is to hook on a sort at the end of the request

```
  projects.value = ((await API.get('/home')).data as Project[])
    .sort( ...sort logic ...);
```

packages/client/hmi-client/src/page/Home.vue
YohannParis: Yes, and we should add it to `DEVELOPMENT.md`.
I will create a new Issue to fix this problem in a new PR and move on with this one.

packages/client/hmi-client/src/page/Home.vue
echl: @YohannParis Since you don't know if we want to order by timestamp yet (and timestamp isn't working) should we shelve this discussion?

echl: Or I can add updating timestamp in the backend as part of this PR

echl: Actually that's in data-service which is the domain of jataware right?

packages/client/hmi-client/src/page/Home.vue
echl: `hmi-server` just proxies to data service so should this be done there?

packages/client/hmi-client/src/page/Home.vue
YohannParis: Let's create a new issue about this later.
But cleaning, ordering and those things should be done on a computed value, not on the <ul> itself.
Also the App.vue can provide the Projects now.

echl: @YohannParis Currently the `/home` endpoint also grabs and populates the projects' related documents. So I will have to move that functionality into `ProjectResource`

echl: Unless we have a separate endpoint that gets related documents only on home page... thoughts?

echl: As discussed, let's resolve the above in a separate task.

YohannParis (approved):

PR #716, opened 2023-02-14
chore: 707 feat improve error handling on back end of hmi

### LINKED ISSUE #707: [FEAT]: Improve error handling on back end of HMI
Currently there is no exception or error handling on the back or front end of the HMI. This means that when our proxied servers return a problem (for example a 404) this is bubble directly up to the front end as-is often resulting in a hanging UI

- [x] All of the _Resource_ type objects on the back end will need to standardize errors to a handful of responses for the front end to handle:

-  404 representing an issue with a proxied resource that stops us from giving performing an action
-  500 for an error in our code, for example, an exception thrown by our java back end
-  204 Nothing is wrong but there is no content to return (for example a search that yields no results)
-  200 Standard response
- [x] API end points should be annotated with `@APIResponses` outlining what each endpoint could return for our swagger docs.
- [ ] Its beyond the scope of this issue to design and implement actual error messaging systems on the front end, but, consider adding some safety in the front end to fail gracefully when a 404/500 error is returned
- [x] Update README with instructions on how to access swagger docs.

dvince2:
# Description

* Adding exception and error handling to the back end.
* HTTP errors in the 400s and 500s from proxy services are funnelled into error 404s by hmi-server for our front end to handle gracefully.
* In some cases, we will also return 204s (no content) or 400 (bad request, when incorrect parameters are used) to the front end in addition to the standard 200 response.
* Swagger docs have been updated with explanations of why these errors are had.
* Filling out the autocomplete back end object. @echl this may be of interest to you, as I've found the annotation that allows us to ignore all of the other json properties and just deserialize what we care about.

Right now this error handling is just on the back end not the front (@tangopapatime is working on this separately) and it is focused primarily on the old document service and home resource.  This was done to keep the PR a bit smaller and make sure everyone is on the same page before i bring it to all of the other endpoints.

Resolves #707


mwdchang:
I am not entirely sure if we want to interpret a RuntimeException as 404s, there is something a bit odd about treating external errors as not being able to find resources. What are people's thoughts here?

As an aside, what exactly is the logging mechanism we are using in the HMI-server? I see `Slf4j` getting injected via Lombok, but Quarkus also has its own logging eccentricities no? Do we prefer one over another?

dvince2:
We had a lot of back and forth about the error to send up to the client, and landed on 404.

In this case here, the exception mapper is taking 400s and 500s given to us by the external services and throwing runtime exceptions in our back end to catch and send how we want to the front end.

YohannParis:
@mwdchang the errors is from the hmi-client perspective.
A 400/500 from xDD should return a 404, because we should get something back, but didn't for X cause.
from the hmi-client point of view, the hmi-server has not crashed, etc.

mwdchang:
@YohannParis @dvince2 You are effectively treating 404s like 500s as even **PUT**, **POST**, and **PATCh** API-routes returns 404s on errors (eg. Adding a new dictionary => Not_found). 

Semantically it is a bit confusing for me personally, and I wonder if we are better served (and more aligned with stanard) to use the 500 generic server error instead.

YohannParis:
I see your point, those should be [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503).
Anyway, this is a first pass to start the process of sending proper errors to the client, not just passing what the services returns.

mwdchang:
My general rule of thumb is: 

If server error or unexpected error
- of some indeterminable origin, send 500 with message
- otherwise send the appropriate 5xx code with message, 

If there is something wrong wrong with the input:
- then 4xx series applies as they apply to clients.

You'd boil down the 4xx/5xx series down to a few generic ones and re-throw them to avoid the catch-them-all Pokemon kind of thing. You can always tailor a specific error message to an exception to differentiate within a single type.


What I see us doing here is obfuscating the true causes, while people intentionally do this some times (e.g. Github obfuscate 401s/403s with 404s for I guess privacy reasons) I just don't see a strong incentive to do this here.

Go give a concrete example: I would rather see

```
500:  external service XYZ failed
```

then

dvince2:
I'm personally in favour of the 500's for most of these errors as well for those reasons - if others agree I can switch most of them over to that? @chris-dickson  @YohannParis @dgauldie

tangopapatime:
I think @mwdchang concrete example says it all.

`500:  external service XYZ failed` is the way to go.  From the client's point of view, the message is what matters.

echl (dismissed):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/documentservice/DictionariesResource.java
```diff
@@ -22,7 +27,31 @@ public class DictionariesResource {
...
-		return proxy.getAvailableDictionaries(all);
+	@APIResponses({
+		@APIResponse(responseCode = "404", description = "An error getting available dictionaries"),
+		@APIResponse(responseCode = "204", description = "Request received successfully, but there are no dictionaries")})
+	public Response getAvailableDictionaries(@QueryParam("all") @DefaultValue("") String all) {
+		try {
+			XDDResponse<XDDDictionariesResponseOK> response = proxy.getAvailableDictionaries(all);
+
+			if (response.getErrorMessage() != null) {
+				Response.status(Response.Status.INTERNAL_SERVER_ERROR)
```
YohannParis, line 38: I'd prefer [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) than 500, in that case.
What do you think?

dvince2: From our discussion, we're opting to settle on 500 for now

YohannParis (approved):
Thank you for doing this boring but oh so necessary task.

PR #726, opened 2023-02-15
fix: undo temporary hackathon changes, fix grid duplicates

mwdchang:
### Summary
Small changes and fixes based on chatter with Jamie
- Remove a temporary LR (left-right) vs TD (top-down) layout from hackathon, by default the layout will do LR.
- Fix a grid element duplication issue if useGrid option is enabled

image

### Test
Go to any model view, the graph structure should be flowing left-ight instead of top-down


mj3cheun (approved):

YohannParis (approved):
LR is so much better to me.

PR #736, opened 2023-02-16
feat(hmi-client): fixed logging and added toast service

tangopapatime:
# Description

The vue logger didn't really work very well in our use case so I created our own.  I added a little toast system that can be activated globally when creating a new Logger:

```
export const logger = new Logger({
	consoleEnabled: !isProduction,
	callerInfoEnabled: true,
	**showToast: false**
});
```

By default, all logs will be buffered and sent to hmi-server but you can pass in an additional message option:

```
interface ILoggerMessageOptions {
	showToast?: boolean; // will force a show or surpression of toast
	toastTitle?: string; // change default title of toast
	silent?: boolean; // do not transmit to backend
}
```

logger.error(error, { showToast: true, toastTitle: `Authentication Error` });

The toast service can also be used independent of the logger:

```
import Toast from 'primevue/toast';
import { ToastSummaries, ToastSeverity, useToastService } from '@/services/toast';

const toast = useToastService();

toast.showToast(
			ToastSeverity.error,
			`${ToastSummaries.NETWORK_ERROR} (${status})`,
			'Unauthorized',
			5000
		);
```

Resolves #(issue)


tangopapatime:
error

packages/client/hmi-client/src/api/api.ts
```diff
@@ -20,7 +22,29 @@ API.interceptors.request.use(
 		return config;
 	},
 	(error) => {
-		console.error(error);
+		logger.error(error, { showToast: true, toastTitle: `Authentication Error` });
```
YohannParis, line 25: I don't think we want to do this here, we should do like on the `App.vue` and redirect the user to the login page.

Also can we merge those two interceptors (from `App.vue` and here)? I don't understand why we should have two.

packages/client/hmi-client/src/services/toast.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export const useToastService = () => {
+	const showToast = (
+		severity: ToastSeverity,
+		summary: string,
+		detail: string,
+		life: number = 4000
+	) => {
+		ToastEventBus.emit('add', {
+			severity: severity,
```
YohannParis, line 31: In JS now you don't need to explicitely write the value if it's the same name as the property.
i.e. `{ severety, summary, detail, life }` will do it.

packages/client/hmi-client/src/utils/log-buffer.ts
```diff
@@ -0,0 +1,83 @@
...
+			config.headers.Authorization = `Bearer ${auth.token}`;
+		} else {
+			config.headers = { Authorization: `Bearer ${auth.token}` };
+		}
+		return config;
+	},
+	(error) => {
+		console.warn(error);
+	}
+);
```
YohannParis, line 27: We already have an Axios call API for that same purpose, and the same kind of intereceptors on the App.vue.
I think all of those needs to be refactored.

packages/client/hmi-client/src/utils/log-buffer.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const toast = useToastService();
+
+// LogDetails Interface
+interface LogDetails {
```
YohannParis, line 32: Now we end types with the word `Type`, [see documentation](https://github.com/DARPA-ASKEM/TERArium/blob/main/packages/client/hmi-client/DEVELOPMENT.md#types)

packages/client/hmi-client/src/utils/log-buffer.ts
```diff
@@ -0,0 +1,83 @@
...
+	add = (log: LogDetails) => {
+		this.logs.push(log);
+	};
+
+	isEmpty = (): boolean => {
+		if (this.logs.length > 0) {
+			return false;
+		}
+		return true;
+	};
```
YohannParis, line 57: We are using lodash in the app, so you can use their `isEmtpy()` instead

packages/client/hmi-client/src/utils/log-buffer.ts
```diff
@@ -0,0 +1,83 @@
...
+			try {
+				const resp = await LOGS.post(`/logs/`, {
+					logs: this.getLogBuffer()
+				});
+				const { status } = resp;
+				if (status === 200) {
+					this.clearLogs();
+				}
+			} catch (error: any) {
+				console.log(error);
```
YohannParis, line 78: This should be a `console.error()`?

packages/client/hmi-client/src/utils/logger.ts
```diff
@@ -1,86 +1,210 @@
...
-					console.warn('POST to /logs did not return a 200');
-				} else {
-					this.clearLogs();
-				}
-			} catch (error) {
-				console.error(error);
-			}
+
+		if (this.options.consoleEnabled) {
+			console.log(`[${level}] ${message} ${this.callerInfo}`, ...optionalParams);
```
YohannParis, line 111: Why not using the level for the console itself?  
```js
console.[level](`${message} ${this.callerInfo}`, ...optionalParams)
```

packages/client/hmi-client/src/utils/logger.ts
```diff
@@ -1,86 +1,210 @@
...
+	INFO = 'info',
+	ERROR = 'error',
+	WARN = 'warn',
+	DEBUG = 'debug'
+}
+
+/**
+ * @interface ILoggerMessageOptions
+ */
+interface ILoggerMessageOptions {
```
YohannParis, line 23: Same comment as above about naming types [per documentation](https://github.com/DARPA-ASKEM/TERArium/blob/main/packages/client/hmi-client/DEVELOPMENT.md#types)

YohannParis (changes requested):
Thanks, could you also update the documentation on how to use the Toast/Log system on the DEVELOPMENT.md.

tangopapatime:
Thanks for the feedback @YohannParis .  Will look into addressing them!

packages/client/hmi-client/src/utils/log-buffer.ts
tangopapatime: Yeah.  I wasn't exactly sure of what the best approach was.  The one on `app.vue` re routes you to the login page.  The `api.vue` might not have access to this vue router yet when api is being used.  I'm all ears if there's a particular pattern that solves this.

packages/client/hmi-client/src/utils/log-buffer.ts
tangopapatime: Yeah I remember this but maybe my comment about thinking it's weird to force `Interface` usage and `Type` nomenclature lol.  Will change it to `Type` cause I don't have big feelings over it :)

packages/services/hmi-server/docker/jvm/Dockerfile.jvm
```diff
@@ -1,7 +1,7 @@
 ####
 # This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
 ###
-FROM registry.access.redhat.com/ubi8/openjdk-17:1.14-10
+FROM registry.access.redhat.com/ubi8/openjdk-17:1.14-10.1675788279
```
YohannParis, line 4: Are we sure we want to do that.
I wonder if your branch is out of date with `main`?

packages/client/hmi-client/src/utils/logger.ts
```diff
@@ -126,30 +126,30 @@ class Logger {
 	/**
 	 *
 	 * @param {string} level
-	 * @param {*} msg
+	 * @param {*} message
```
YohannParis, line 129: This is amazing!
I love long variables name

YohannParis (approved):

YohannParis (approved):

tangopapatime:
I apologize for the extremely linty PR.

YohannParis (approved):

PR #740, opened 2023-02-17
chore: start reorganizing models/petrinets

mwdchang:
### Summary
There are a couple of things in this PR, the main gist is to consolidate scattered and (arguably) misplaced code fragment under the same roof - we have been sort of intermingling "petrinet" and "model" both conceptually and within code logic. As a precursor to more advanced model-editor work, this PR addresses these inconsistencies:

1. Extract out the skeletons of a de-facto PetriNet editor that we will reuse across the application. Current each page has it's own renderer to do the job

2. Consolidate types/services. We have a blurry line of what a Model is and what a PetriNet is, likewise between types/utilities/services. In this PR we try to sort this out:
   - Create a src/petrinet folder to hold all things specifically related to petrinet
   - Created a "ConcreateModel<type>" to accommodate various model-frameworks. The current "Model" is aliased as "ConcreateType<any>".  This makes it a little easier to _not_ mishmash petrinet and model together...though subject to change pending exactly how we'd adopt new frameworks down the road.

### Testing
There should be no functional changes. App works as before:
- Check model page
- Check the experimental model-editor page


mj3cheun (approved):

mwdchang:
Maybe `TypedModel` instead of `ConcreteModel` makes more sense ... any thought?

shawnyama:
> Maybe `TypedModel` instead of `ConcreteModel` makes more sense ... any thought?

To keep things consistent with `DEVELOPMENT.md` you call it `ModelType` and make it an interface

mwdchang:
> > Maybe `TypedModel` instead of `ConcreteModel` makes more sense ... any thought?
> 
> To keep things consistent with `DEVELOPMENT.md` you can call it `ModelType` and make it an interface

The thing is Model is already an interface, I don't want to go down the Java rabbit hole of calling something ModelTypeType 😬

packages/client/hmi-client/src/page/ModelEditor.vue
```diff
@@ -60,13 +60,12 @@
 </template>
 
 <script setup lang="ts">
-import * as d3 from 'd3';
 import _ from 'lodash';
-import graphScaffolder, { IGraph } from '@graph-scaffolder/index';
-import { runDagreLayout, D3SelectionINode, D3SelectionIEdge } from '@/services/graph';
+import { IGraph } from '@graph-scaffolder/index';
```
YohannParis, line 64: Can we do `{ IGraph as GraphType }` to be consistent with our coding style?

packages/client/hmi-client/src/types/Model.ts
```diff
@@ -1,12 +1,7 @@
-import { PetriNet } from '@/utils/petri-net-validator';
 import { CONCEPT_FACETS_DISPLAY_NAME, CONCEPT_FACETS_FIELD } from './Concept';
 import { Filters } from './Filter';
 
-// FIXME: other model content types will be supported depending on
-//	Model.framework
-export type ModelContent = PetriNet;
-
-export type Model = {
+export type ConcreteModel<T> = {
```
YohannParis, line 4: Should we name it `GenericModel` instead? What does _concrete_ means in this context?

packages/client/hmi-client/src/page/ModelEditor.vue
mwdchang: I would avoid this kind of in-situ renaming of things of the sake of consistency: graph-scaffolder is it's own stand-alone library and should be treated as such in my opinion, like, we would not do `type HTMLElementType = HTMLElement`.

If you think that graph-scaffolder itself should conform to terarium naming standard, that will need to be a different PR.

packages/client/hmi-client/src/types/Model.ts
mwdchang: @YohannParis @shawnyama 

Sorry about the confusion, It is actually the reverse of what you are inferring, this is a specific model-type, as it takes in a typed-generic.

I do agree that concrete sounds weird. I offered from above that we should renamed ConcreteModel
- TypeModel<T> to indicate a model of specific type. e.g. TypedModel<PetriNet>, TypedModel<ABM> ..., and
- Model, as a "generic" model reference which is an alias for TypedModel<any>....to be used when you don't care about the underlying framework semantics, for example in facets.

shawnyama (approved):

YohannParis (approved):

PR #749, opened 2023-02-21
chore: clean up add to cart

Tom-Szendrey:
# Description

https://user-images.githubusercontent.com/17088680/220458481-e122624c-7485-4f5f-92bf-acafcf0b5b80.mov

  Separated the header and data from the explorer's pane. added a bunch of styling

   Clean up for issue 687

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/slider-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/slider-panel.vue
@@ -12,10 +12,10 @@
 					@click="emit('update:isOpen', false)"
 				/>
 				<slot name="header"></slot>
-				<div class="column">
+				<section>
 					<h4 class="slider-header-item">{{ header }}</h4>
 					<slot name="subHeader"></slot>
-				</div>
+				</section>
 			</header>
 			<slot name="content"></slot>
 		</template>
@@ -91,7 +91,7 @@ i {
 	align-items: start;
 }
 
-.column {
+section {
 	display: flex;
 	flex-direction: column;
 }
--- a/packages/client/hmi-client/src/components/widgets/slider-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/slider-panel.vue
@@ -11,7 +11,11 @@
 					:class="`slider-header-item pi ${directionMap[direction].iconOpen}`"
 					@click="emit('update:isOpen', false)"
 				/>
-				<h4 class="slider-header-item">{{ header }}</h4>
+				<slot name="header"></slot>
+				<div class="column">
+					<h4 class="slider-header-item">{{ header }}</h4>
+					<slot name="subHeader"></slot>
+				</div>
 			</header>
 			<slot name="content"></slot>
 		</template>
@@ -84,7 +88,12 @@ i {
 
 .slider-header {
 	display: flex;
-	align-items: center;
+	align-items: start;
+}
+
+.column {
+	display: flex;
+	flex-direction: column;
 }
 
 .slider-header.content {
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/widgets/slider-panel.vue
```diff
@@ -11,7 +11,11 @@
 					:class="`slider-header-item pi ${directionMap[direction].iconOpen}`"
 					@click="emit('update:isOpen', false)"
 				/>
-				<h4 class="slider-header-item">{{ header }}</h4>
+				<slot name="header"></slot>
+				<div class="column">
```
Tom-Szendrey, line 15: Please feel free to give me a better class name than "column" if you have any

packages/client/hmi-client/src/components/widgets/slider-panel.vue
```diff
@@ -11,7 +11,11 @@
 					:class="`slider-header-item pi ${directionMap[direction].iconOpen}`"
 					@click="emit('update:isOpen', false)"
 				/>
-				<h4 class="slider-header-item">{{ header }}</h4>
+				<slot name="header"></slot>
```
YohannParis, line 14: Why do you need this here?

packages/client/hmi-client/src/page/data-explorer/components/selected-resources-header-pane.vue
```diff
@@ -0,0 +1,180 @@
...
+
+onMounted(async () => {
+	const all = await ProjectService.getAll();
+	if (all !== null) {
+		projectsList.value = all;
+	}
+});
+</script>
+
+<style lang="scss" scoped>
```
YohannParis, line 121: We do not use `lang="scss"` on component.

packages/client/hmi-client/src/page/data-explorer/components/selected-resources-header-pane.vue
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	addResourcesToProject(projectId);
+
+	emit('close');
+	router.push(`/projects/${projectId}`);
+};
+
+onMounted(async () => {
+	const all = await ProjectService.getAll();
```
YohannParis, line 114: This is better to be called `projects` instead of `all`

YohannParis (dismissed):

packages/client/hmi-client/src/components/widgets/slider-panel.vue
Tom-Szendrey: This is the dropdown button that is provided from the ..header.. file
This allows us to have the `CART` the `dropdown` and the `>>` on the same line.

Previously the header div containing what is now `CART` and `>>` wouldnt contain the drop down button so I could not make them all level with each other

Tom-Szendrey: I can use v-slot instead to make it more consistent with what we have

packages/client/hmi-client/src/page/data-explorer/components/selected-resources-header-pane.vue
Tom-Szendrey: makes sense! I didnt even catch this had just moved the code I will fix that

packages/client/hmi-client/src/components/widgets/slider-panel.vue
YohannParis: I mean you can just make it a `<section>` and in then 
```html
<style>
   ...
   section { ... } /* instead of .column */
   ...
</style>
```

YohannParis (approved):

YohannParis (approved):

PR #754, opened 2023-02-22
chore: fix failing e2e test

chris-dickson:
The name change caused a (rather silly) test to fail


mwdchang (approved):

mwdchang:
Yohann fixed this in https://github.com/DARPA-ASKEM/Terarium/pull/747

Issue #755, opened 2023-02-22
[BUG]: Home Page Related Break


Tom-Szendrey:
**Describe the bug**
The backend `HomeResource.java` is breaking line 81 and 82.
This is for 2 reasons 
1) `Model.java` and `Dataset.java` have `Instant` for their timestamps but are receiving format: 2023-01-31**T**16:26:48.451859
    We need to tell java how to parse with json 
    Solution: Change type to `LocalDateTime` or add `@JsonbDateFormat(ISO_LOCAL_DATE_TIME)`
2) `Model.java -> framework` to be type String

<img width="1393" alt="image" src="https://user-images.githubusercontent.com/17088680/220704127-896936a6-c68f-4541-951f-12302b8266fa.png">

Im not sure if we should change our `Model.java -> framework` or `ModelFramework.java` or if we need to bug data service to change their API

PR #760, opened 2023-02-23
fix: deserialization breaks 

Tom-Szendrey:
# Description
Could not parse models or datasets correctly.
One clear place to see this way home page's related to section was not being populated because of this
image

Resolves #(issue)
#755


packages/services/hmi-server/src/main/resources/application.properties
```diff
@@ -20,6 +20,7 @@ quarkus.datasource.db-kind=postgresql
 %dev.quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5021/terarium
 # add new columns but do not drop
 quarkus.hibernate-orm.database.generation=update
+quarkus.hibernate-orm.jdbc.timezone=UTC
```
Tom-Szendrey, line 23: Pretty sure this is defaulted but figured having it explicit couldnt hurt.

PR #784, opened 2023-03-01
feat: add structured logging for hmi-server

chris-dickson:
This PR adds the ability to enable structured logging to hmi-server. We need to be able to parse logs in Loki, to do this we can enable json logging in staging and production.  With the logs provided I was able to create the following dashboard:

Screenshot 2023-03-01 at 4 38 28 PM

We can provide a breakdown of
  - Request duration by user (top graph)
  - Request duration by endpoint (bottom graph)

This allows us to see who's using (and potentially taxing) the system, and what endpoints are slower than others

Screenshot 2023-03-01 at 4 38 46 PM

We also add some tables that allows filtering/sorting by request as well as a simple table of exceptions that occur in the backend.


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/configuration/RequestLoggingResponseFilter.java
```diff
@@ -16,6 +18,12 @@ public class RequestLoggingResponseFilter implements ContainerResponseFilter {
 	public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
 		final String user = requestContext.getSecurityContext().getUserPrincipal() != null ? requestContext.getSecurityContext().getUserPrincipal().getName() : "Anonymous";
 		final long durationMs = Instant.now().toEpochMilli() - (long)requestContext.getProperty(RequestLoggingRequestFilter.REQUEST_START_TIMESTAMP_MS);
-		log.info("REQUEST COMPLETED | " + requestContext.getRequest().getMethod() + " | " + requestContext.getUriInfo().getRequestUri() + " | " + user + " | " + durationMs + "ms");
+		log.info(JsonString.write(Map.of(
+			"log_message", "REQUEST COMPLETED",
```
dvince2, line 22: These same fields show up in a few places.  To avoid mistakes that could affect detecting these logs (`log_message` vs `logMessage`) should we through these into an enum or just static Strings somewhere to access?

dvince2 (dismissed):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/configuration/RequestLoggingResponseFilter.java
chris-dickson: I moved this into a `StructuredLog` application scoped service. This will cover most cases but there may be instances where we need structured logs and this is not `@Inject`able in which case we'll have to rely on convention and uses the bare `log.info` function.

tangopapatime:
Sorry for jumping in :P.  Wanted to make a comment on whether an update to `LoggingService.java` is also needed.

chris-dickson:
> Sorry for jumping in :P. Wanted to make a comment on whether an update to `LoggingService.java` is also needed.

As we discussed offline, this isn't needed and is `just handled` already.

YohannParis (approved):
LGTM

PR #792, opened 2023-03-03
feat: fix/simplify tab functionality and take care of router-view warnings

shawnyama:
# Description

- fix/simplify tab functionality (less code, fixed tab storage, project context persists)
- give container to components rendered in `router-view` so that class and props can be passed properly (just wrapped them with a `main` tag)
- resourceType is passed from DataExplorer to Navbar using `pageRef` (`App.vue`) - ideally the `assetType` should be in the data explorer route/query instead, this is just temporary clean up and clarifies what `resourceType` is being used for in `search-bar.vue`

Grab child variables/functions from parent with `router-view`: https://stackoverflow.com/questions/67053956/vue-3-get-access-to-router-view-instance-in-order-to-call-child-methods

Resolves #(issue)


packages/client/hmi-client/src/App.vue
```diff
@@ -9,10 +9,12 @@
 		:active="!isErrorState"
 		:current-project-id="project?.id ?? null"
 		:projects="projects"
-		:resourceType="resourceType"
+		:resourceType="pageRef?.resourceType ?? ResourceType.XDD"
```
YohannParis, line 12: I'll prefer that you make this a computed, and that you explain where `resourceType` is coming from with the `defineExpose` function in `DataExplorer.vue`.
I'm not even sure I like this patern; but for now we need to be explicit as to not search for it.

shawnyama: I decided to take this a bit further by passing `showSuggestions` which is the only thing the children need from `resourceType`

YohannParis (approved):

PR #806, opened 2023-03-08
Drag exp

mj3cheun:
Model editor rework

https://user-images.githubusercontent.com/14062458/223769956-9a6b8398-5361-4968-a8a5-8b2bcc6cd33d.mov

### DIFF
--- a/packages/client/graph-scaffolder/src/core/renderer.ts
+++ b/packages/client/graph-scaffolder/src/core/renderer.ts
@@ -27,6 +27,9 @@ export interface Options {
 
 	// Whether to show grid
 	useGrid?: boolean;
+
+	// The css class for custom dragging event
+	dragSelector?: string;
 }
 
 export const pathFn = d3
@@ -67,6 +70,8 @@ export abstract class Renderer<V, E> extends EventEmitter {
 	zoom: d3.ZoomBehavior<Element, unknown> | null = null;
 
 	zoomTransformObject: d3.ZoomTransform | null = null;
+
+	customDrag = false;
 
 	constructor(options: Options) {
 		super(); // Event emitter
@@ -410,6 +415,8 @@ export abstract class Renderer<V, E> extends EventEmitter {
 
 		let sufficientlyMoved = false;
 
+		const dragSelector = this.options.dragSelector || null;
+
 		function collisionFn(p: IPoint) {
 			const buffer = 10;
 			for (let i = 0; i < nodes.length; i++) {
@@ -429,13 +436,16 @@ export abstract class Renderer<V, E> extends EventEmitter {
 			return false;
 		}
 
-		function nodeDragStart(evt: any): void {
+		function nodeDragStart(evt: d3.D3DragEvent<any, any, any>): void {
 			evt.sourceEvent.stopPropagation();
 
 			// @ts-ignore: D3 "this"
 			node = d3.select(this) as D3SelectionINode<V>;
 
-			if (evt.sourceEvent.shiftKey) {
+			renderer.customDrag = (dragSelector &&
+				d3.select(evt.sourceEvent.target).classed(dragSelector)) as boolean;
+
+			if (renderer.customDrag) {
 				emitWrapper('node-drag-start', evt, node, renderer);
 				return;
 			}
@@ -450,7 +460,8 @@ export abstract class Renderer<V, E> extends EventEmitter {
 			const dy = evt.dy;
 
 			if (!node) return;
-			if (evt.sourceEvent.shiftKey) {
+
+			if (renderer.customDrag) {
 				emitWrapper('node-drag-move', evt, node, renderer);
 				return;
 			}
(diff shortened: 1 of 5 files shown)


packages/client/graph-scaffolder/src/core/renderer.ts
```diff
@@ -429,13 +436,15 @@ export abstract class Renderer<V, E> extends EventEmitter {
...
 
-		function nodeDragStart(evt: any): void {
+		function nodeDragStart(evt: d3.D3DragEvent<any, any, any>): void {
 			evt.sourceEvent.stopPropagation();
 
 			// @ts-ignore: D3 "this"
 			node = d3.select(this) as D3SelectionINode<V>;
 
-			if (evt.sourceEvent.shiftKey) {
+			renderer.isDragEnabled = d3.select(evt.sourceEvent.target).classed(dragSelector ?? 'no-drag');
```
mwdchang, line 445: I don't understand why `no-drag` is introduced here - what are you trying to do?

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -44,25 +46,107 @@ export class PetrinetRenderer extends graphScaffolder.BasicRenderer<NodeData, Ed
...
+
+		transitions
+			.append('circle')
+			.classed('shape', true)
+			.classed('no-drag', true)
+			.attr('cx', (d) => -d.width * 1.2)
+			.attr('r', (d) => d.width * 0.2)
+			.attr('fill', '#FFF')
+			.attr('stroke', '#AAA')
+			.attr('stroke-width', 2)
```
mwdchang, line 92: There are quite a bit of repetitions here, I think you can cut them down by something like:
```
transitions.append('circle').classed('blahblah', true).attr('cx', ...).attr('cy', ...)
transitions.append('circle').classed('blahblah', true).attr('cx', ...).attr('cy', ...)
transitions.append('circle').classed('blahblah', true).attr('cx', ...).attr('cy', ...)
...

transition.selectAll('.blahblah').attr('fill', ...). attr('stroke', ...) ....
```

Also how come there are only three handles instead of 4?

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -12,6 +12,8 @@ const pathFn = d3
 	.y((d) => d.y);
 
 export class PetrinetRenderer extends graphScaffolder.BasicRenderer<NodeData, EdgeData> {
+	displayedHandles: D3SelectionHandles<NodeData> | null = null;
```
mwdchang, line 15: It might be cleaner to cache the node instead of the node's handles, just so there aren't too many points of references that point to the same DOM/data-binds.

```
nodeSelection: D3SelectionINode<NodeData> | null = null
...
on('node-click',  (_eventName, _event, selection: D3SelectionINode<NodeData>)  => {
  if (nodeSelection) nodeSelection.selectAll('.blahblah').transition().duration(200) ....  
  nodeSelection = selection;
  ndoeSelection.selectAll('.blahblah').transition().duration(200) ...
})

File: packages/client/graph-scaffolder/src/core/renderer.ts (hunk 1 above)
Comment (mj3cheun) [reply packages/client/graph-scaffolder/src/core/renderer.ts:445]: so what im trying to do here is use 'no-drag' as a default selector for disabling drag

File: packages/client/hmi-client/src/petrinet/petrinet-renderer.ts (hunk 2 above)
Comment (mj3cheun) [reply packages/client/hmi-client/src/petrinet/petrinet-renderer.ts:92]: ok so you would prefer that the handles were all in one line? there are only 3 handles because if there was a handle up top it might overlap with the node name

File: packages/client/hmi-client/src/petrinet/petrinet-renderer.ts (hunk 3 above)
Comment (mj3cheun) [reply packages/client/hmi-client/src/petrinet/petrinet-renderer.ts:15]: ok yes thats a good idea

File: packages/client/hmi-client/src/petrinet/petrinet-renderer.ts (hunk 2 above)
Comment (mwdchang) [reply packages/client/hmi-client/src/petrinet/petrinet-renderer.ts:92]: @mj3cheun I meant you create and set the positions of X-number of handles across X-number of lines, but since they all have the same style you can do it in one shot. Maybe a better example:

```
// Create circle and their position
svg.append('circle').attr('cx', 10).attr('cy', 10).attr('r', 10);
svg.append('circle').attr('cx', 50).attr('cy', 50).attr('r', 10);
svg.append('circle').attr('cx', 100).attr('cy', 100).attr('r', 10);

// Set their style
svg.selectAll('circle').attr('fill', '#f80').attr('stroke', '#888').style('cursor', 'pointer');
```

File: packages/client/graph-scaffolder/src/core/renderer.ts (hunk 1 above)
Comment (mwdchang) [reply packages/client/graph-scaffolder/src/core/renderer.ts:445]: I think if the idea is to opt-in, then there shouldn't be any defaults - you should explicitly acknowledge that you want the feature.

File: packages/client/hmi-client/src/petrinet/petrinet-service.ts [hunk 4]
```diff
@@ -120,15 +120,16 @@ export const parsePetriNet2IGraph = (model: PetriNet) => {
 	// Reset current Graph.
 	result.nodes = [];
 	result.edges = [];
-	const nodeHeight = 20;
-	const nodeWidth = 20;
+	const nodeHeight = 40;
+	const nodeWidth = 40;
+	const nodeLayoutSpacing = 140;
```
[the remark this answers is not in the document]
mwdchang: Hrmm., these are my temporary hacks that should really be fixed instead of continuing to be built upon .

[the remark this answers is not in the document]
mwdchang: The idea here was to center the graph on the canvas but I never got around the to write that function. I will take a swing at this now.

mwdchang (approved):

PR #830, opened 2023-03-09
feat: add an axios cache for API calls

YohannParis:
# Description

* Add `axios-cache-interceptor` https://axios-cache-interceptor.js.org/
* Set it up only for `/api` calls
* Setup to use LocalStorage, with a default of 5 minutes, or less based on TTL header


chris-dickson:
Why is this necessary and how is it different from the browser cache?

YohannParis:
> Why is this necessary and how is it different from the browser cache?

It is not necessary at all, just trying to see if we can speed up some heavy part of the /api.
It was an old branch I wanted to see working and up to discussion.
Mostly some /api calls can be really heavy, like the homepage.

mwdchang:
I think before we go ahead and merge this we should have a better understanding of what are the endpoints that are slow and if we can do anything about them, it is better to see if we can cure the root-causes rather than just treating the symptoms. 

There are already several levels of caching here and dealing with coherency can be a head scratching exercise....even when we agree that some of the items are not critical it is n't clear to everyone which oddities can be ignored and which oddities need to be chased down. Just to take an inventory, I think we have
- An application cache on some specific routes
- Whatever magic cache-headers quarkus adds
- This axios cache
Are we missing anything else?

YohannParis:
This has been deemed useless.

PR #832, opened 2023-03-09
feat: generate shared models automatically

chris-dickson:
# Motivation
This is an experiment where we use a gradle plugin to automatically generate typescript types that need to be shared between the client and server.  It's common that we need to defined "mirrored" types in the client and server.  These can go out of date and cause bugs. This PR allows us to specify such types *only* on the server to ensure they're stay in sync.

# Example
Given the following server type definition
```java
@ClientModel
public class DataRequest {
  private String id;
  private String targetId;
  @TypescriptOptional
  private String description;
  private DataRequestType type;
}
```
We would generate the following entries in `client/hmi-client/src/types/Types.ts`:
```typescript
export interface DataRequest {
    id: string;
    targetId: string;
    description?: string;
    type: DataRequestType;
}

export enum DataRequestType {
    Signed = "SIGNED",
    Sealed = "SEALED",
    Delivered = "DELIVERED",
}
```

Things to note:
- Typescript generation is *opt-in* via the `@ClientModel` annotation to avoid cruft
- Types that are imported and used in any `@ClientModel` annotated class will be automatically included in TS generation
- Fields are by default required unless the `@TypescriptOptional` annotation is added

# Usage
This automatically runs when running `yarn dev`.  Anytime a model changes on the backend it will trigger a regeneration which then automatically kicks off a rebuild in vite. Ideally, this should require no though.

The exception (for now) is when you're only running the backend.  Developers should run the `yarn generateTypes` command if this is necessary (although I suspect it won't be)


YohannParis (commented):
Could you add your nice PR description within thé `docs` folder?

dvince2:
Is the hope then to bring this to more of the types in the future (With event just for now?

chris-dickson:
> Is the hope then to bring this to more of the types in the future (With event just for now?

Yes I didn't want to make a massive PR. We can pull them over incrementally to make sure we're not breaking anything.

dvince2 (approved):
Awesome - this is going to be great.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
```diff
@@ -30,34 +32,37 @@
 	@Index(columnList = "type"),
 	@Index(columnList = "value")
 })
-@EqualsAndHashCode(callSuper = true)
 public class Event extends PanacheEntityBase implements Serializable {
```
YohannParis, line 35: Is `implements Serializable` needed? I remember you mentioning that it was not necessary

YohannParis (approved):
Why make them `TSIgnore` instead of `TypeScriptIgnore`?
It would be better to be verbose and avoid confusion about why this annotation is here on the first place.

PR #834, opened 2023-03-10
feat: model edit api integration

mwdchang:
### Summary
Add and amend API-endpoints for model creation, update, and copy operation.

I added the "temporary" ModelStub class to model create and update operations, because the two of them take in a `content` as string, but the way we are modelling content is of the type ModelContent object.

Using object-mapper to serialize content-object into a json-string has other caveats that are probably not worthwhile to over come, given that to support a new modelling framework (e.g. PDEs) we will likely need quite a bit of shake-up across the board. So to keep things simple and to convey a transient nature I introduced the "ModelStub".

### Testing
API Testing through cURL.
```
# Create a model, you should get a model id
curl -v -H "content-type: application/json" -XPOST "http://localhost:3000/api/models" -H "${AUTHORIZATION_HEADER}" -d'
{
	"name": "dc-test",
	"framework": "Petri Net",
	"description": "dc-test",
	"content": "{ \"S\": [], \"T\": [], \"I\": [], \"O\": [] }"
}
'

# Now we can copy it, replace <id> from above, we should get a new model id
curl -v -H "content-type: application/json" -XPOST "http://localhost:3000/api/models/<id>/operation-copy" -H "${AUTHORIZATION_HEADER}" -d'
	{
		"left": <id>,
		"name": "New model",
		"description": "just a test",
		"content": "[\"abc delicious\"]",
		"parameters": [],
		"framework": "Petri Net"
	}
'

# Now we can update the new model, replace <newid>
curl -v -H "content-type: application/json" -XPOST "http://localhost:3000/api/models/<id>" -H "${AUTHORIZATION_HEADER}" -d'
{
	"name": "dc-test",
	"framework": "Petri Net",
	"description": "dc-test",
	"content": "{ \"S\": [ \"test S\" ], \"T\": [], \"I\": [], \"O\": [\"test O\"] }"
}
'

# Now we can verify the content has been updated from before via:
http://localhost:3020/models/<id>
http://localhost:3020/models/<newid>
```


dvince2 (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
```diff
@@ -0,0 +1,16 @@
...
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+import java.io.Serializable;
+
+@Data
+@Accessors(chain = true)
+public class ModelOperationCopy implements Serializable {
+	private Long left;
```
YohannParis, line 11: Why the `left`?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
```diff
@@ -0,0 +1,16 @@
...
+	private String name;
+	private String description;
+	private String framework;
```
YohannParis, line 14: Shouldn't the `framework` be the one from the copied model?
If we want to change the framework from PetriNet to ODE for example, we should call another endpoint and not mix them.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelStub.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.experimental.Accessors;
+
+import java.io.Serializable;
+
+/**
+ * This is used to mirror model-creation and model-update
+**/
+@Data
+@Accessors(chain = true)
+public class ModelStub implements Serializable {
```
YohannParis, line 13: From what I understand you do not need `Serializable`

mwdchang: Can you point to a reference of not using serializable? 
Is this some Lombok magic because Java beans had always had to implement Serializable interface.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
mwdchang: TDS uses a somewhat RDF-like schema for tracking, that being left/right/operation. RDF is subject/object/relation.

```
askem=# \d+ provenance
                                              
    Column     |            Type             | 
---------------+-----------------------------+-
 id            | integer                     | 
 timestamp     | timestamp without time zone | 
 relation_type | relationtype                | 
 left          | integer                     | 
 left_type     | provenancetype              | 
 right         | integer                     | 
 right_type    | provenancetype              | 
 user_id       | integer                     | 
 concept       | character varying           | 

```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
mwdchang: This is part of the minimal copy-operation payload as specified by TDS. Given we had almost no write-operations until now there are bound to be some quirky things we have missed or have become irrelevant. 

CC @mattprintz @fivegrant if there are any comments.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
YohannParis: I agree, but why does the `hmi-client` needs to provide this when `id` is part of the URI already?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelStub.java
YohannParis: From what I understand `Serializable` is necessary if we need for example to cache those object in Redis, etc.
But we do not need it yet.

chris-dickson: Correct, it's not strictly necessary at this point.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelOperationCopy.java
mwdchang: This is a proxy object, the `/id` path param doesn't exist on TDS.

The API point we are hitting is
```
/models/opts/copy
```

dvince2 (approved):
LGTM

PR #837, opened 2023-03-13
feat: project card placeholder

YohannParis:
# Description

* New automatically generated image placeholder for the project-card
* Based on _p5_ library, and @jamiewaese-uncharted code
* Moved to TypeScript and implement into our codebase

# Sceenshot
Screenshot 2023-03-13 at 14 56 34


packages/client/hmi-client/src/utils/project-card.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Define variables
+let placeholderCanvas: p5 | null = null;
+const imageHeight = 133;
+const imageWidth = 208;
+let contributors: number;
+let models: number;
+let datasets: number;
+let papers: number;
```
mwdchang, line 31: These (contributors, models, datasets ...etc) would be better served if they are not at the global scape
- It would be easier to follow the logic, and
- it reduces the odds of creating dangling or orphaned DOM trees which might lead to memory leaks

At a glance, it seems like we just need to change a couple of function signatures to slot them in as parameters

packages/client/hmi-client/src/utils/project-card.ts
```diff
@@ -0,0 +1,218 @@
...
+	datasets = Math.min(args.datasets, 25);
+	papers = Math.min(args.papers, 25);
+
+	// Create the p5 sketch
+	if (!placeholderCanvas) {
+		placeholderCanvas = new p5(sketch); /* eslint-disable-line new-cap */
+	}
+
+	// Add all of this within the canvas and export the base64 image
+	return placeholderCanvas.draw() as unknown as string;
```
mwdchang, line 217: What does it mean to say `foobar as unknown as string` ?

mwdchang:
What is the memory footprint of including `p5.js`? 
If we only use it to do procedurally-generated textures and it is a beast, we should have some plans to replace the logic with native-canvas equivalent.

packages/client/hmi-client/src/utils/project-card.ts
```diff
@@ -0,0 +1,218 @@
...
+
+			spacerY *= 1.1;
+			spacerX *= 1.1;
+		}
+	}
+}
+
+// Squiggly lines across the screen for each dataset
+function drawDatasets(p: p5) {
+	function squigglyLine(y, number, lineNumber) {
```
mwdchang, line 69: Nitpick: might be cheaper (computationally) to use Perlin-noises or some kind of discrete-fourier-transform to generate squigglylines.

packages/client/hmi-client/src/utils/project-card.ts
YohannParis: You can't transform a type to another directly, even if the kind of match, so you have to change it via unknown

packages/client/hmi-client/src/utils/project-card.ts
YohannParis: Sure we can do that.

YohannParis:
> What is the memory footprint of including `p5.js`? If we only use it to do procedurally-generated textures and it is a beast, we should have some plans to replace the logic with native-canvas equivalent.

Same answer as above, feel free to do so.

mwdchang (approved):

PR #838, opened 2023-03-14
feat: prototype opening code from github

shawnyama:
# Description

Prototype for importing github code.

https://user-images.githubusercontent.com/28237258/225028932-5d243af5-adca-4afe-a870-085f8d0ded66.mov

Choose a document in the project view and under URLs you can choose a repo you want to import from. Then choose the file you want to open.

I kept things such as the conversion from the rawgit url to the jsdelivr url brittle since it's a prototype but let me know if it's worth improving. Followed this conversion:  https://lukasgamper.medium.com/how-to-import-files-directly-from-github-1a41c72a3ad3


packages/client/hmi-client/src/components/documents/Document.vue
```diff
@@ -67,15 +67,28 @@
...
+							<a :href="url.html_url" target="_blank" rel="noreferrer noopener">{{
+								url.html_url
+							}}</a>
+						</li>
+					</template>
 					<li v-for="ex in urlArtifacts" :key="ex.url">
 						<b>{{ ex.resourceTitle }}</b>
 						<div>
-							<a :href="ex.url" rel="noreferrer noopener">{{ ex.url }}</a>
+							<a :href="ex.url" target="_blank" rel="noreferrer noopener">{{ ex.url }}</a>
```
YohannParis, line 91: Remove all `target="_blank"`.  This is not a normal behaviour, links should open right there or people should right-click, etc.

packages/client/hmi-client/src/components/documents/Document.vue
```diff
@@ -67,15 +67,28 @@
...
 				</template>
 				<ul>
+					<template v-if="isEditable">
+						<li class="github-link" v-for="(url, index) in githubUrls" :key="index">
+							<Button
+								label="Import"
+								class="p-button-sm p-button-outlined"
+								icon="pi pi-cloud-download"
+								@click="openCodeBrowser(url)"
+							/>
```
YohannParis, line 82: I wonder if you should create a component just for this, that will include the modal and all the GitHub integration you did, just to make things tidy and in one place.

packages/client/hmi-client/src/components/documents/Document.vue
```diff
@@ -161,6 +199,31 @@ function highlightSearchTerms(text: string | undefined): string {
...
+	if (url.download_url === null) {
+		chosenRepositoryName.value = `${chosenRepositoryName.value}/${url.name}`;
+		filesToSelect.value = await getGithubRepositoryContent(url.url);
+		return;
+	}
+
+	emit(
+		'open-asset',
+		{ assetName: 'New file', assetType: ProjectAssetTypes.CODE },
+		// { assetName: url.name, assetId: url.name, assetType: ProjectAssetTypes.CODE },
```
YohannParis, line 222: Why the comment line?

packages/client/hmi-client/src/page/project/components/code-editor.vue
```diff
@@ -30,16 +30,27 @@ import FileUpload from 'primevue/fileupload';
...
+import { isEmpty } from 'lodash';
 
 const DEFAULT_TEXT = '# Paste some python code here or import from the controls above';
-const content = ref(DEFAULT_TEXT);
+const content = ref<string>(DEFAULT_TEXT);
 const editor = ref<VAceEditorInstance['_editor'] | null>(null);
 const selectedText = ref('');
 
+const props = defineProps<{
+	importedGithubCode?: string;
```
YohannParis, line 45: You can just call it `code` we might use it for other purpose in the future

packages/client/hmi-client/src/utils/github-import.ts
```diff
@@ -0,0 +1,44 @@
+const header = {
+	Authorization: `Bearer ${import.meta.env.VITE_CR_PAT}`
```
YohannParis, line 2: I don't see where you set this up

packages/client/hmi-client/src/utils/github-import.ts
YohannParis, line 1: * This file should be in `services` not `utils`
* Add some `try { ... } catch (error) { ... }` in there.

packages/client/hmi-client/src/components/documents/Document.vue
shawnyama: That wouldn't open a tab properly since it's not registered as an asset in the resourceStore/resource sidebar yet. So for now the New File resource is opened and the code is pasted into it. I thought adding a new project asset within the project view like this (such as creating a code file) should be in a separate PR since it comes with its own challenges.

packages/client/hmi-client/src/utils/github-import.ts
shawnyama: I just have my own `.env` file in my `/hmi-client`. I don't think we will need one unless there are over 100 requests within an hour or something. I set it up for myself in case

YohannParis: Okay, in that case we should remove it, or comment it for now.

packages/client/hmi-client/src/page/project/components/code-editor.vue
```diff
@@ -30,16 +30,27 @@ import FileUpload from 'primevue/fileupload';
...
 
 const DEFAULT_TEXT = '# Paste some python code here or import from the controls above';
-const content = ref(DEFAULT_TEXT);
+const content = ref<string>(DEFAULT_TEXT);
 const editor = ref<VAceEditorInstance['_editor'] | null>(null);
 const selectedText = ref('');
 
+const props = defineProps<{
+	code?: string;
+}>();
```
mwdchang, line 46: Personally I find optional properties annoying to deal with, you end up sometimes doing extra checks that are seeming unnecessary just to satisfy the linter.

How do we feel about making `code` required and default to empty string instead?

packages/client/hmi-client/src/services/github-import.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function getGithubUrls(query: string = 'SIR model') {
+	// Search for potential repositories
+	try {
+		const response = await fetch(
+			`https://api.github.com/search/repositories?q=${query}`, // {&page,per_page,sort,order}
+			requestOptions
+		);
+		const result = await response.json();
+		if (!response.ok) return null;
+		return result.items.slice(0, 3); // Just show 3 for the prototype
```
mwdchang, line 21: Why not just state the item limitation in the request via `per_page=X`?
```
https://api.github.com/search/repositories?q=astar&per_page=3
```

packages/client/hmi-client/src/page/project/components/code-editor.vue
YohannParis: I disagree.
A simple test if better, they can be null and that's it.

YohannParis (dismissed):

YohannParis (approved):

PR #844, opened 2023-03-14
fix: error determining provenance

### LINKED ISSUE #843: [BUG]: Exception thrown when trying to get provenance info for docs
**Describe the bug**
An error toast appears with 404 error when clicking on a document

**To Reproduce**
1. Do a search for documents and extractions, say for the term `Covid`
2. Click on a document
Notice the toast error saying error 404

**Expected behavior**
No toast is shown.  If there is provenance information it should be shown

**Screenshots**

**Additional context**
The problem here is that we are directly sending a gddid to tds.  Instead we should be sending the correct ID, obtained by a get publication call to tds first.

dvince2:
# Description

• Updating back end to use same terminology as front end.
• Using new typescript generation to keep this file in sync with front end.
• Now when retrieving publications if a 404 is hit we change this to a 204 for our front end.
• Updating exception being used by back end to pass on response code.
• Front end now asks TDS for correct document asset before proceeding.
• Using generated DocumentAsset

Resolves #843

### DIFF
--- a/packages/client/hmi-client/src/services/provenance.ts
+++ b/packages/client/hmi-client/src/services/provenance.ts
@@ -5,11 +5,12 @@
 import API from '@/api/api';
 import { logger } from '@/utils/logger';
 import { ResultType } from '@/types/common';
+import { DocumentAsset } from '@/types/Types';
 import { ProvenanceResult, ProvenanceQueryParam, ProvenanceType } from '@/types/Provenance';
 // eslint-disable-next-line import/no-cycle
 import { getBulkDocuments } from './data';
 import { getBulkDatasets } from './dataset';
-import { getBulkDocumentAssets } from './external';
+import { getBulkDocumentAssets, getDocument } from './external';
 import { getBulkModels } from './model';
 
 //
@@ -28,16 +29,19 @@ const MAX_RELATED_ARTIFACT_COUNT = 5;
 /**
  For a document, find similar content
 	Find related documents (xDD)
-*/
+ */
 
 // API helper function for fetching provenance data
 async function getConnectedNodes(
-	id: string | number,
+	id: string,
 	rootType: ProvenanceType
 ): Promise<ProvenanceResult | null> {
+	const publication: DocumentAsset | null = await getDocument(id);
+	if (!publication) return null;
+
 	// FIXME: all underscore naming should be fixed
 	const body: ProvenanceQueryParam = {
-		root_id: Number(id),
+		root_id: publication.id,
 		root_type: rootType
 	};
 	const connectedNodesRaw = await API.post('/provenance/connected_nodes', body).catch((error) =>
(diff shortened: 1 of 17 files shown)


dvince2:
@chris-dickson You don't need to PR this, but, Typescript plugin in action

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
```diff
@@ -57,12 +59,13 @@ public class Event extends PanacheEntityBase implements Serializable {
...
-	 * @param like				optional search string for values of matching events
-	 * @param limit				the maximum number of events to return
-	 * @return						a list of {@link Event} matching the input
+	 *
+	 * @param type      the {@link EventType}
+	 * @param projectId the optional project id
+	 * @param username  the username of the current user
+	 * @param like      optional search string for values of matching events
+	 * @param limit     the maximum number of events to return
+	 * @return a list of {@link Event} matching the input
```
YohannParis, line 68: @dvince2 and @chris-dickson you need to decide how your Java Doc is setup.
Every other PR there is a change just for that.

packages/client/hmi-client/src/services/provenance.ts
```diff
@@ -28,16 +29,19 @@ const MAX_RELATED_ARTIFACT_COUNT = 5;
...
 // API helper function for fetching provenance data
 async function getConnectedNodes(
-	id: string | number,
+	id: string,
 	rootType: ProvenanceType
 ): Promise<ProvenanceResult | null> {
+	const publication: DocumentAsset | null = await getDocument(id);
+	if (!publication) return null;
+
 	// FIXME: all underscore naming should be fixed
```
YohannParis, line 42: Does this FIXME still necessary, has it been fixed now?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/DocumentAsset.java
```diff
@@ -3,13 +3,17 @@
...
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import software.uncharted.terarium.hmiserver.annotations.TSOptional;
 
 import java.io.Serializable;
 
 @Data
 @Accessors(chain = true)
-public class Publication implements Serializable {
+@TSModel
+public class DocumentAsset implements Serializable {
```
YohannParis, line 14: `implements Serializable` we don't need them.

YohannParis (approved):

packages/client/hmi-client/src/services/provenance.ts
dvince2: This hasn't been fixed yet:
```
const body: ProvenanceQueryParam = {
		root_id: Number(id),
		root_id: publication.id,
		root_type: rootType
		root_type: rootType
	};
	};
```
The corresponding java object is also the same.
I've opened a new issue to clean up the underscores and move this to the TS generator: https://github.com/DARPA-ASKEM/Terarium/issues/849

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Event.java
dvince2: @chris-dickson I'm using the default settings from IntelliJ

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/DocumentAsset.java
dvince2: Will do this as seperate issue

PR #858, opened 2023-03-16
feat: model view params

shawnyama:
# Description

Resolves #(issue)


packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -150,6 +141,13 @@ const fetchRelatedTerariumArtifacts = async () => {
 	}
 };
 
+function updateParamaterRow(attribute: string, newParameterRow) {
+	if (model.value && model.value[attribute]) {
```
YohannParis, line 145: If you want to avoid the double test you can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax) on array too.

```suggestion
	if (model?.value?.[attribute]) {
```

YohannParis (approved):
Beautiful, I couldn't do better!

PR #867, opened 2023-03-18
feat: resolve ontology curies from models

YohannParis:
# Description

* Add an ontology resolver to the `ModelConcept`
* Edit `api/models/{id}` endpoint to resolve those concept before sending back the model
* Call MIRA APIs _(in the future we should host these services)_
* Create a couple model classes for Petri Net models

# Screenshot

Screenshot 2023-03-22 at 14 19 00

# Resolves

Screenshot 2023-03-17 at 22 30 17

https://docs.google.com/presentation/d/18macftUe871RshB87QhEgNXY_14YrW0jVlaW-NRY1lQ/edit#slide=id.g21da5d9f705_0_67


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ModelContent.java
YohannParis, line 1: @chris-dickson this is the part of Java where I'm block, if you could help me with that one. Maybe some pair-programming.

I try to transform that string `('disease_severity', 'ncit:C25269')` into a proper object 
```java
{
     name: 'disease_severity',
     curie: 'ncit:C25269',
     title: 'Symptomatic',
     description: 'Exhibiting the symptoms of a particular disease.',
     url: 'https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI%20Thesaurus&code=C25269'
}
```

chris-dickson: Options:

1. Regex:
```java
final String input = "('disease_severity', 'ncit:C25269')";
final Matcher matcher =  Pattern.compile("\\(\\'(.+?)\\', \\'(.+?)\\'\\)").matcher(string);

final String name = matcher.group(1);
final String curie = matcher.group(2);
```

2. String splitting
```java
final String nameCurie = "('disease_severity', 'ncit:C25269')".split(", ");
final String name = nameCurie[0].substring(2, nameCurie[0].length - 1);
final String curie = nameCurie[1].substring(1, nameCurie[1].length - 2);
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/petrinet/Ontology.java
```diff
@@ -0,0 +1,54 @@
...
+			this.name = matcher.group(1);
+			this.curie = matcher.group(2);
+		} else {
+			this.name = this.curie = null;
+		}
+
+		if (this.curie == null) return;
+
+		try {
+			final Response entity = proxy.getEntity(this.curie);
```
YohannParis, line 41: @dvince2 or @dgauldie, I'm unable to call the _MiraProxy_ from within the _Ontology_ model. 
Any clue? I can't find any documentation about proxies on _Quarkus_ website.

dvince2: You may be missing the `@Inject` annotation on that

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/petrinet/Ontology.java
```diff
@@ -0,0 +1,30 @@
...
+@Data
+public class Ontology implements Serializable {
+	private final String name;
+	private final String curie;
+	private String title;
+	private String description;
+	private String link;
+
+	public Ontology (String input) {
+		final Matcher matcher = Pattern.compile("\\(\\'(.+?)\\', \\'(.+?)\\'\\)").matcher(input);
```
dvince2, line 20: This may be worth a comment on what you're looking to do - if we hadn't spoke yesterday this would leave me scratching my head :D

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/petrinet/SpeciesJsonDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+	 */
+
+	@Override
+	public Species deserialize(final JsonParser p, final DeserializationContext ctxt) throws IOException, JacksonException {
+		JsonNode node = p.getCodec().readTree(p);
+
+		final Species species = new Species();
+		species.setName(node.get("sname").asText());
+		species.setMiraIds(new ArrayList<>());
+		species.setMiraContext(new ArrayList<>());
```
dvince2, line 28: This is totally correct and doesn't have to change, but as someone who has been recently introduced to lombok, if you add the annotation `@Accessors(chain = true)` to your `Species` class you would be able to chain these together like...
`final Species species = new Species().setName(foo).setMiraIds(bar).setMiraContext(fooBar);`

Again... totally not needed but kind of cool!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/DKGProxy.java
```diff
@@ -0,0 +1,25 @@
...
+
+@RegisterRestClient(configKey = "mira-dkg")
+@Produces(MediaType.APPLICATION_JSON)
+public interface DKGProxy {
+	@GET
+	@Path("/api/entity/{curie}")
+	DKG getEntity(
+		@PathParam("curie") final String curie
+	);
+
```
dvince2, line 19: Also completely optional, but used elsewhere in our application.

If you annotation the interface with the common portions of the API paths (`@Path("/api")`) you can then annotate these methods with just the unique portion (for example, `@Path("/entities/{curies}")`).  Optional, but if your class had a lot of endpoints it makes it a bit easier to read

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/DKGProxy.java
```diff
@@ -0,0 +1,25 @@
...
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+import software.uncharted.terarium.hmiserver.models.mira.DKG;
+
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@RegisterRestClient(configKey = "mira-dkg")
+@Produces(MediaType.APPLICATION_JSON)
```
dvince2, line 12: For the exception handling in the resource classes to work you will need to annotate this class with `@RegisterProvider(HmiResponseExceptionMapper.class)`. That will pass the responses through the HmiResponseExceptionMapper class, which converts 400 and 500 series errors to `ResponseRuntimeException`

dvince2 (changes requested):
Looks great Yohann!
Most of the comments added here are just FYI - I think the only thing that needs to be looked at is just the exception mapper annotation (And anyone elses PR feedback that I maybe missed)

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -75,15 +75,35 @@
 				</AccordionTab>
 				<AccordionTab header="State variables">
 					<DataTable
-						:value="model?.content.S"
+						:value="model?.content?.species"
```
dvince2, line 78: I think that this is coming from `petrinet-service.ts` and now needs to be changed there too.  May be worth using the typescript generation tool on this?

dvince2 (dismissed):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/petrinet/SpeciesJsonDeserializer.java
mwdchang: @dvince2 There is also `@builder` ... if we want to fully lombok there is also `val` keyword 🤔 ... any thoughts on where we draw the line?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/petrinet/Species.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.Serializable;
+import java.util.List;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(using = SpeciesJsonDeserializer.class)
+public class Species implements Serializable {
+
+	@JsonAlias("sname")
+	private String name;
```
mwdchang, line 17: Per discussion, it is hard to reason with petrinet acsets, and it would be good if we keep the naming as close as possible to the transport format. That is calling it sname instead of name.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MetaRegistryProxy.java
```diff
@@ -0,0 +1,12 @@
+package software.uncharted.terarium.hmiserver.proxies.mira;
+
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+
+@RegisterRestClient(configKey = "mira-metaregistry")
+@Produces(MediaType.APPLICATION_JSON)
+public interface MetaRegistryProxy {
```
dvince2, line 10: What is this empty interface for?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ModelResource.java
```diff
@@ -114,12 +126,97 @@ public Response updateParameters(
...
+			try {
+				entities = dkgProxy.getEntities(curies);
+			} catch (RuntimeException e) {
+				log.error("Unable to get the ontology entity for curies: " + curies, e);
+			}
+
+			if (!entities.isEmpty()) {
+				// FIXME - For now we need to resolve ourselves this link, HMS will fix this
+				// I don't know how to read application.properties to get that URL directly
+				final String metaRegistryURL = "http://34.230.33.149:8772/";
```
dvince2, line 183: You should be able to set this as a property then use
```
@ConfigProperty(name = "mira-metaregistry/mp-rest/url") 
String metaRegistryURL;
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MetaRegistryProxy.java
YohannParis: It exist, I wanted to use it, but that was simple to send the URL to the front end instead of resolving in the back-end.

dvince2 (approved):
Ship it!

Issue #868, opened 2023-03-19
[TASK]: Blind Stratification fix ID future issue 


Tom-Szendrey:
Blind Stratification fix ID future issue 
Currently resulting graph just has the Id has A.id , B.id which can cause duplicate problems

PR #869, opened 2023-03-19
feat: blind stratification in a playground

Tom-Szendrey:
# Description
Go to /stratifyplayground and play around with blind stratification 

https://user-images.githubusercontent.com/17088680/226202937-c8252b1d-c1c0-4b5c-af6b-120da235216f.mov

https://github.com/DARPA-ASKEM/Terarium/issues/868 

Also has the potential sample typing solution in ta-2-playground. I figure this is just a playground no harm in adding this as well but if you prefer just the blind strat let me know

https://user-images.githubusercontent.com/17088680/226203581-32b2fbba-6ce2-457a-8e19-00693cb49382.mov


packages/client/hmi-client/src/temp/StratifyPlayground.vue
```diff
@@ -0,0 +1,568 @@
...
+								width: graphOne.nodes[j].width,
+								x: graphOne.nodes[j].x,
+								y: graphOne.nodes[j].y,
+								nodes: []
+							});
+						}
+						//Create this transition's edge(s)
+						let newSource = graphOne.nodes[j].id + ',' + graphTwo.edges[i].source;
+						let newTarget = graphOne.nodes[j].id + ',' + graphTwo.edges[i].target;
+						resultGraph.edges.push({ source: newSource, target: newTarget });
```
mwdchang, line 435: This part looks a bit suspicious, that it being outside of the nested ifs.
Can you provide a small textual description for clarify?

Tom-Szendrey: rereading this there is something weird going on with looking at edges vs nodes with the same iterator 
Give me a few seconds to see what I was thinking and comment it better or change it a tad

PR #883, opened 2023-03-21
feat: project page tweaks from slides

shawnyama:
# Description

- Empty page behaviour when there are no tabs + remove empty tab case
- Remove buttons that aren't used (just keep delete button)
- Add "Are you sure" popup for removing an asset from project
- Tabs are identified more uniquely, some used to be identified as the same (eg. in the case where they have the same name)

https://user-images.githubusercontent.com/28237258/226956405-f6808f87-9a16-4c32-9fd6-9199d05672fb.mov

### DIFF
--- a/packages/client/hmi-client/src/page/project/tera-project.vue
+++ b/packages/client/hmi-client/src/page/project/tera-project.vue
@@ -8,8 +8,9 @@
 		>
 			<template v-slot:content>
 				<tera-resource-sidebar
-					:tabs="tabs"
 					:project="project"
+					:tabs="tabs"
+					:opened-asset-route="openedAssetRoute"
 					@open-asset="openAsset"
 					@close-tab="removeClosedTab"
 				/>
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -104,7 +114,7 @@ const props = defineProps<{
 	project: IProject;
 	assetName?: string;
 	assetId?: string;
-	assetType?: ProjectAssetTypes;
+	assetType?: ProjectAssetTypes | 'overview' | '';
```
YohannParis, line 117: Maybe create an `IAssetType` in the `/types` folder

YohannParis (approved):

packages/client/hmi-client/src/page/project/tera-project.vue
shawnyama: I think this would be worth dealing with while we merge `ResourceType` and `AssetType`.

PR #884, opened 2023-03-21
Model editor tweak2

mwdchang:
### Summary
Added a "cancel" button to cancel existing save buffer - currently this only applies to the model's structural changes

Better? default zoom and alignment

### Testing
1. In model view, go into edit-mode, remove a few nodes then click cancel. The graph should restore to the default loaded state.

2. Browser several models and observing the initial zoom state, this is rather subjective but they shouldn't look too weird.

### DIFF
--- a/packages/client/graph-scaffolder/src/core/renderer.ts
+++ b/packages/client/graph-scaffolder/src/core/renderer.ts
@@ -346,6 +346,7 @@ export abstract class Renderer<V, E> extends EventEmitter {
 		// Zoom control
 		// FIXME: evt type
 		const zoomed = (evt: any) => {
+			console.log('zoomed................');
 			if (this.options.useZoom === false) return;
 			if (chart) chart.attr('transform', evt.transform);
 
@@ -373,7 +374,6 @@ export abstract class Renderer<V, E> extends EventEmitter {
 
 		const minZoom = 0.05;
 		const maxZoom = Math.max(2, Math.floor((this.graph.width as number) / this.chartSize.width));
-		let zoomLevel = Math.min(1, 1 / ((this.graph.height as number) / this.chartSize.height));
 
 		if (this.options.zoomRange) {
 			this.zoom = d3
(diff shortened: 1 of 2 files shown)


packages/client/graph-scaffolder/src/core/renderer.ts
```diff
@@ -386,17 +385,28 @@ export abstract class Renderer<V, E> extends EventEmitter {
...
-		let zoomX =
-			(-((this.graph.width as number) * zoomLevel * 0.5) + 0.5 * this.chartSize.width) / zoomLevel;
-		let zoomY =
-			(-((this.graph.height as number) * zoomLevel * 0.5) + 0.5 * this.chartSize.height) /
-			zoomLevel;
+		// let zoomX =
+		// 	(-((this.graph.width as number) * zoomLevel * 0.5) + 0.5 * this.chartSize.width) / zoomLevel;
+		// let zoomY =
+		// 	(-((this.graph.height as number) * zoomLevel * 0.5) + 0.5 * this.chartSize.height) /
+		// 	zoomLevel;
```
YohannParis, line 392: Remove comments

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -3,8 +3,14 @@
 		<header>
 			<div class="framework">{{ model?.framework }}</div>
 			<div class="header-and-buttons">
-				<h4 v-html="title" />
+				<h4>{{ title }}</h4>
```
YohannParis, line 6: This needs to be a v-html because of the `highlightSearchTerms()` used on the computed `title`.
`highlightSearchTerms()` returns HTML. This was the only quick way I was able to add highlighting of results during the hackthon.
[Here a good example of what I've done in the Vue.js doc](https://vuejs.org/guide/essentials/template-syntax.html#raw-html)

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -3,8 +3,14 @@
...
 				<span v-if="isEditable">
+					<Button
+						v-if="isEditing"
+						@click="cancelEdit"
+						label="Cancel"
+						class="p-button-sm p-button-outlined"
+					/>
 					<Button
 						@click="toggleEditMode"
 						:label="isEditing ? 'Save model' : 'Edit model'"
```
YohannParis, line 16: I wonder if we should reverse the buttons here, so that the toggle edit should be `Edit model` or `Cancel edit model` and the first conditional button is `Save edited model`.

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -377,6 +383,26 @@ const toggleEditMode = () => {
...
 };
 
+// Cancel existing edits, currently this will:
+// - Resets changs to the model structure
+const cancelEdit = async () => {
+	isEditing.value = false;
+	if (!model.value) return;
+
+	// Convert petri net into a graph with raw input data
+	const g: IGraph<NodeData, EdgeData> = parsePetriNet2IGraph(model.value.content, {
```
YohannParis, line 393: `graph` instead of `g` will be better.
Those mathematicians can't spell out variables, but we can 😄

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/models/Model.vue
mwdchang: Yeah I suspected there may be some fishy injections going here.  I was mucking around a couple of things out of place thinking they may have upset the linter ... will put it back.

packages/client/hmi-client/src/components/models/Model.vue
mwdchang: done

PR #885, opened 2023-03-22
feat: blind stratification in ta2-playground

Tom-Szendrey:
# Description
Provided 2 petrinets (not typed) and no ontology, stratify them together by multiplying everything together.  

https://user-images.githubusercontent.com/17088680/226913176-ca9b0b29-f64c-4a8a-a3c3-cd8c0cadbdf2.mov

### DIFF
--- a/packages/client/hmi-client/src/petrinet/petrinet-service.ts
+++ b/packages/client/hmi-client/src/petrinet/petrinet-service.ts
@@ -323,7 +323,6 @@ function cloneFirstGraph(
 	graphOne: IGraph<NodeData, EdgeData>,
 	graphTwo: IGraph<NodeData, EdgeData>
 ) {
-	console.log('Start clone first graph');
 	const resultGraph: IGraph<NodeData, EdgeData> = { nodes: [], edges: [] };
 	for (let i = 0; i < graphTwo.nodes.length; i++) {
 		if (graphTwo.nodes[i].data.type === NodeType.State) {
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/petrinet/petrinet-service.ts
```diff
@@ -311,3 +312,77 @@ export const petriToLatex = async (petri: PetriNet) => {
 	const response = await API.post('/transforms/acset-to-latex', petri);
 	return response.data;
 };
+
+/*
+Given two IGraphs, clone the first X times where X is the length of nodes of the 2nd one.
+	Example, SIR + QNQ leads to two distinct SIR graphs. One with Q and one with NQ
+return a graph
+TODO: https://github.com/DARPA-ASKEM/Terarium/issues/868
+*/
```
YohannParis, line 321: Let's use [jsdoc](https://jsdoc.app/about-getting-started.html) formatting of function comment to standardize them. We don't need to add @params and @return, etc. but more like
```js
/**
 * Given two IGraphs, clone the first X times where X is the length of nodes of the 2nd one.
 * Example, SIR + QNQ leads to two distinct SIR graphs. One with Q and one with NQ
 * @return graph
 * @todo https://github.com/DARPA-ASKEM/Terarium/issues/868
 */
```

packages/client/hmi-client/src/temp/TA2Playground.vue
```diff
@@ -766,6 +777,28 @@ export default defineComponent({
...
+					<option :value="2">QNotQModel</option>
+					<option :value="3">typeModel</option>
+					<option :value="4">SIRD</option>
+				</select>
+				<select v-model="stratifyModelB" type="text">
+					<option :value="0">generic</option>
+					<option :value="1">generic2</option>
+					<option :value="2">QNotQModel</option>
+					<option :value="3">typeModel</option>
+					<option :value="4">SIRD</option>
```
YohannParis, line 794: You don't need to bind the value as they are not reactive.
```vue
<option :value="0">generic</option>

<option value="0">generic</option>
```

YohannParis (approved):

PR #888, opened 2023-03-22
feat: drag asset directly into searchbar

### LINKED ISSUE #786: [FEAT]: Drag & Drop Assets in the search by example
Same process as for within the Data Explorer, drag an Asset, from anywhere, into the search bar to start a _Search By Example_ query.

echl:
# Description

Dragging and dropping an asset into search bar opens search by example modal. 

https://user-images.githubusercontent.com/16358728/226952019-95a16513-868b-4b70-a4b9-351a0310eee3.mov

Resolves #786

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/components/search-bar.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/components/search-bar.vue
@@ -16,6 +16,14 @@
 					@complete="fillAutocomplete"
 					@keyup.enter="initiateSearch"
 					@item-select="initiateSearch"
+					@dragover.prevent
+					@dragenter.prevent
+					@drop="
+						{
+							isSearchByExampleVisible = true;
+							onDrop();
+						}
+					"
 				>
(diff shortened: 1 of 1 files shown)


mwdchang:
It isn't obvious where you would drop the search-item, or if the action is even possible. Do we think it is worthwhile to make the affordances better @YohannParis ?

YohannParis:
> It isn't obvious where you would drop the search-item, or if the action is even possible. Do we think it is worthwhile to make the affordances better @YohannParis ?

No, it's not necessary for now. People can access this feature in three ways:
- click on the file icon in the search bar
- drag & drop (this PR)
- or click on the three vertical dot on the top right of an asset

YohannParis (approved):

echl:
> It isn't obvious where you would drop the search-item, or if the action is even possible. Do we think it is worthwhile to make the affordances better @YohannParis ?

I agree, but in my opinion it's a "bonus" interaction for now

PR #892, opened 2023-03-22
feat: tooltip for model extractions

shawnyama:
# Description

Tooltip that contains model extractions appear when hovering over extractions chip in model view

### DIFF
--- a/packages/client/hmi-client/src/components/models/Model.vue
+++ b/packages/client/hmi-client/src/components/models/Model.vue
@@ -262,6 +262,7 @@ watch(
 		if (props.assetId !== '') {
 			const result = await getModel(props.assetId);
 			model.value = result;
+			console.log(result);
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/models/Model.vue
YohannParis, line 1: You can remove those console log test

packages/client/hmi-client/src/components/models/model-parameter-list.vue
YohannParis, line 1: What this is for?

shawnyama: See where I use exampleIndex in the child. I use it to go through every unique model extraction in the example data

YohannParis (approved):

Issue #889, opened 2023-03-22
[BUG]: Search by related/similar is not working in data-explorer


Tom-Szendrey:
**Describe the bug**
When clicking on search by related and search by similar content the explorer's results are not getting updated

**Expected behavior**
A new result set based off of the results from this related/similar search

PR #896, opened 2023-03-22
fix: search by related similar is not working in data explorer

Tom-Szendrey:
# Description
Fixing search by similar and search by related in data-explorer.

In terms of search by related it sends back a regular 100 with facets ect. I think these may be the same results but it is hitting the correct call

In terms of search by similar it is getting 10 back (I am assuming this is expected) and no facets (again assume expected) so I have just closed the facet bar when this is clicked as it wont be necessary 

https://user-images.githubusercontent.com/17088680/227028617-c2ac92e2-185b-4595-bb16-39a4d6317b02.mov

https://user-images.githubusercontent.com/17088680/227028672-bed0c316-061e-456b-ab30-44ad65b53a28.mov

Resolves #(issue)
#889


YohannParis (approved):
Nice!
It will be great if we can add a pill next to the results numbers saying we search from that `similar content XXXX`. Like when you click on 2022 in the facet a `Year 2022` pill shows up.

PR #899, opened 2023-03-22
feat(hmi): delete a project

YohannParis:
# Description

* Update `hmi-server` to not return _soft-deleted_ projects for the `/home` and `/projects` endpoints.
* Added a _Project Menu_ with a `Remove` option.
* Added a Confirmation dialog to prompt the user to remove the project.
* Show a toast in case of success or error of removal.
* Remove the project from the homepage by listening to the `removed` event emitted by the project card.

# Video

https://user-images.githubusercontent.com/636801/227049317-df358ac2-c097-4e8a-9314-3839d3196767.mov

### DIFF
--- a/packages/client/hmi-client/src/components/projects/ProjectCard.vue
+++ b/packages/client/hmi-client/src/components/projects/ProjectCard.vue
@@ -1,25 +1,3 @@
-<script setup lang="ts">
-import { IProject, ProjectAssetTypes } from '@/types/Project';
-import Button from 'primevue/button';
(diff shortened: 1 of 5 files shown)


mwdchang (approved):

mwdchang:
Seems fine, I have a slight pet peeve about the ProjectCard getting doubly encoded as ProjectCard and "SkeletonProjectCard" which I think makes subsequent logic harder to read because we constantly have to check for nulls/undefined but we can take that up a different day.

PR #910, opened 2023-03-24
fix: parsing mira_ids and mira_context are optional

mwdchang:
### Summary
A somewhat temporary fix so we don't assume mira_ids and mira_contexts are always there. 

The caveat is we have not settled upon a metadata representation so this is likely going to change, but the PR here will prevent unexpected errors in the meantime.

### Testing
Should be able to fetch all models

```
curl -v -H "content-type: application/json" -XGET -H "${AUTHORIZATION_HEADER}" http://localhost:3000/api/models/{id}
```


YohannParis (approved):
Perfect!

PR #927, opened 2023-03-28
chore: polish mit api integration

echl:
Polish previous pr for mit api integration. 

https://user-images.githubusercontent.com/16358728/228358012-8e4859ea-2061-4c60-bf76-8304597058fb.mov

https://user-images.githubusercontent.com/16358728/228358160-e3681f87-86c2-49bb-9896-a709c465ed8a.mov

https://user-images.githubusercontent.com/16358728/228358270-7e675057-d51d-4204-854a-b5f3656b0871.mov


mwdchang:
I am getting empty response with no errors when I tried to submit this fragment of python code

```
        def SIR(t, y):
            S = y[0]
            I = y[1]
            R = y[2]
            y0=-beta*S*I
            y1=beta*S*I-gamma*I
            y2=max(0,1.-(y0+y1))
            return [y0,y1,y2]
```

packages/client/hmi-client/src/page/project/components/code-editor.vue
```diff
@@ -476,22 +167,28 @@ async function initialize(editorInstance) {
...
+				name: newModelName,
+				framework: 'Petri Net',
+				content: JSON.stringify({ ...acset.value, ...response.data })
+			};
+			const model = await createModel(newModel);
+			if (model) {
+				emit('on-model-created', model.id, newModelName);
+			} else {
+				logger.error(`Something went wrong.`);
+			}
```
mwdchang, line 188: This part doesn't look right, we created a model but no project information had been submitted that I can see. Were do we associate the model with the project we are on?

mwdchang: oh I see the project-asset stuff is done outside .... kind of weird, would prefer to have the transactional actions in one place, but okay for now.

packages/client/hmi-client/src/page/project/components/code-editor.vue
```diff
@@ -431,33 +141,14 @@ async function onFileOpen(event) {
...
+async function onExtractModel() {
 	extractPetrinetLoading.value = true;
-	// const response = await API.post(`code/to_acset?code=${selectedText.value}`);
+	const response = await API.post(`code/to_acset?code=${selectedText.value}`);
 	extractPetrinetLoading.value = false;
-	// acset.value = response.data;
-	acset.value = mockAcset;
+	acset.value = response.data;
 	codeExtractionDialogVisible.value = true;
 }
```
YohannParis, line 150: This should be in a service file

packages/client/hmi-client/src/page/project/components/code-editor.vue
```diff
@@ -171,14 +171,27 @@ async function createModelFromCode() {
...
+				.setPollAction(async () => {
+					const response = await API.get(`/code/response?id=${textVars.data}`);
+					if (response) {
+						return {
+							data: response.data.data
+						};
+					}
+					return { error: logger.info(`Linking metadata to model...`) };
+				});
+			const metadata = await poller.start();
```
YohannParis: Any functions that calls an API should be in a service.

packages/client/hmi-client/src/services/mit-askem.ts
```diff
@@ -0,0 +1,30 @@
...
+	const id = await API.post(`/code/annotation/find_text_vars`, text);
+	const poller = new Poller<object>()
+		.setInterval(2000)
+		.setThreshold(90)
+		.setPollAction(async () => {
+			const response = await API.get(`/code/response?id=${id.data}`);
+			if (response) {
+				return {
+					data: response.data.data
+				};
```
mwdchang, line 19: Since response is not null regardless of success or error, it is probably prudent to be more pedantic here, both to guard against errors and to convey intention.

```
const { status, data } = response;
if (status >= 400) return { data: null, progress: null, error: "failed" };
return { data: data.data, progress: null, error: null };
```

packages/client/hmi-client/src/services/mit-askem.ts
```diff
@@ -0,0 +1,30 @@
...
+			}
+			return { error: logger.info(`Linking metadata to model...`) };
```
mwdchang, line 21: Not sure what you are trying to do here, are you trying to pass the logger into the poller?

echl: Just trying to display a status message to indicate that the action is still happening.

echl: Should change the message though.

packages/client/hmi-client/src/services/mit-askem.ts
echl: The response is null actually, otherwise the poller wouldn't work at all!

mwdchang: Look like the hmi-server resource is not catching the errors. Will patch this in a different PR.

YohannParis (approved):

PR #928, opened 2023-03-28
feat: add simulation-service test runs

mj3cheun:
# Description

* Adds simulation service


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/simulationservice/SimulationServiceProxy.java
```diff
@@ -0,0 +1,34 @@
...
+	Response getRunStatus(
+		@PathParam("runId") String runId
+	);
+
+	@GET
+	@Path("/runs/{runId}/result")
+	@Consumes(MediaType.APPLICATION_JSON)
+	Response getRunResult(
+		@PathParam("runId") String runId
+	);
```
YohannParis, line 33: `/status` doesn't return the data once it's ready? We have to call `/result` for it?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationResource.java
```diff
@@ -0,0 +1,56 @@
...
+	@POST
+	@Path("/forecast")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Consumes(MediaType.APPLICATION_JSON)
+	@Tag(name = "Make new forecast simulation")
+	public Response makeForecastRun(
+		final SimulationParams simulationParams
+	) {
+		System.out.println(simulationParams);
+		System.out.println(proxy);
```
YohannParis, line 33: I think we should remove logs, unless, if you want to see this information it will be better to log them properly to see them in Grafana.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationResource.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@GET
+	@Path("/{runId}/result")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Tag(name = "Get simulation result associated with run ID")
+	public Response getRunResult(
+		@PathParam("runId") final String runId
+	) {
+		return proxy.getRunResult(runId);
+	}
```
YohannParis, line 55: Why not make it one endpoint, that return the results once the status is `done`?
We have a chance to simplify the calls here, instead of doing it in the front-end.

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -0,0 +1,42 @@
+import { logger } from '@/utils/logger';
+import API from '@/api/api';
+
+interface ForecastParametersType {
+	petri: string;
+	initials: { [key: string]: number };
+	tspan: [number, number];
+	params: { [key: string]: number };
+}
```
YohannParis, line 9: You should use the `@TSModel` annotations in your `/models/simulationservice/SimultionParams.java` to export those automatically within `/types/Types.ts`.

packages/client/hmi-client/src/components/models/forecat-launcher.vue
YohannParis, line 1: The filename has a type `forecat` should be `forecast`

packages/client/hmi-client/src/components/models/forecat-launcher.vue
```diff
@@ -0,0 +1,117 @@
...
+const emit = defineEmits(['close', 'launch-forecast']);
+
+const launch = async () => {
+	// FIXME: current need to strip out metadata
+	const cleanedModel: PetriNet = {
+		S: props.model.content.S.map((s) => ({ sname: s.sname })),
+		T: props.model.content.T.map((t) => ({ tname: t.tname })),
+		I: props.model.content.I,
+		O: props.model.content.O
+	};
```
YohannParis, line 65: This should be done on the back-end when doing the call.

packages/client/hmi-client/src/components/models/forecat-launcher.vue
```diff
@@ -0,0 +1,117 @@
...
+	const payload = {
+		petri: JSON.stringify(cleanedModel),
+		initials: initialValues.value,
+		params: parameterValues.value,
+		tspan: [0, 50]
+	};
+
+	const run = await makeForecast(payload);
+
+	// FIXME: Cache into sessionStorage
```
YohannParis, line 76: You are right... we need to have a user management/database to handle this properly now.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationResource.java
mj3cheun: whoops forgot to remove these, was debugging

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/simulationservice/SimulationServiceProxy.java
mwdchang: I suspect separate endpoints should be fine.

Imagine that we have a listing of runs that we want to check (e.g. a summary table) and we just want to know if they are done or not, having a dedicated status end point accomplishes that without incurring the cost of sending the data.

packages/client/hmi-client/src/components/models/forecat-launcher.vue
mwdchang: noted, added a fixme, will circle back when data structure is more mature.

mwdchang (approved):

packages/client/hmi-client/src/services/models/simulation-service.ts
mwdchang: will do this in a separate clean up PR

PR #930, opened 2023-03-28
feat: update parameter/state table design

shawnyama:
# Description

https://user-images.githubusercontent.com/28237258/228615452-49c17d06-17fd-4a87-87b2-1c7d59a7bf44.mov


packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -291,14 +284,29 @@ const mathmode = ref('mathLIVE');
 // 	\beta IS - \gamma I \frac{d\color{red}{R}}{dt} = \gamma I`
 // };
 
+const betterStates = computed(() => {
```
YohannParis, line 287: We have to agree between `State` and `Species`. Which one do you prefer @mwdchang ?

YohannParis (approved):

packages/client/hmi-client/src/components/models/Model.vue
mwdchang: I prefer `states`.

PR #933, opened 2023-03-29
fix: handle null values in model description property

YohannParis:
# Description

* Add a Jackson setter to ignore non-existing and null values for the `description` property.
* Did a bit of clean-up here and there
* Added `yarn install` in the `.husky/pre-commit` file, as sometimes, `lint-staged` is not installed. @Nithos is this okay?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn install && yarn lint-staged
+yarn lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn lint-staged
+yarn install && yarn lint-staged
--- a/packages/client/hmi-client/src/services/data.ts
+++ b/packages/client/hmi-client/src/services/data.ts
@@ -31,6 +31,30 @@ import { getAllModelDescriptions } from './model';
 // eslint-disable-next-line import/no-cycle
 import { getRelatedArtifacts } from './provenance';
 
+/**
+ * fetch list of extractions data from the HMI server
+ */
+const getXDDArtifacts = async (term: string, extractionTypes?: XDDExtractionType[]) => {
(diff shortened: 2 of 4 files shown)


packages/client/hmi-client/src/services/data.ts
```diff
@@ -49,6 +73,143 @@ const getXDDDictionaries = async () => {
...
+
+	// "full_results": "Optional. When this parameter is included (no value required),
+	//  an overview of total number of matching documents is returned,
+	//  with a scan-and-scroll cursor that allows client to step through all results page-by-page.
+	//  NOTE: the "max" parameter will be ignored
+	//  NOTE: results may not be ranked in this mode
+	let searchParams = `term=${term}`;
+	const url = '/documents?';
+
+	if (xddSearchParam?.docid) {
```
chris-dickson, line 90: You should be able to use the `query-string` utility for this.
```typescript
import { toQueryString } from '@/utils/query-string';

const url = `/documents?${toQueryString(xddSearchParam)}`
```

chris-dickson (changes requested):

packages/client/hmi-client/src/services/data.ts
YohannParis: Yes, this part of the code needs to be corrected. I just moved functions around to remove `@ts-ignore` and `eslint-disable-line` that people used instead of organizing their code properly.

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn lint-staged
+yarn install && yarn lint-staged
```
mwdchang, line 4: I don't think yarn install is desirable here - it can potentially create unexpected changed and affect yarn.lock, which would necessitate another commit cycle.

The lint-staged came from renovate-bot, perhaps we can ignore it and update this dep manually?

YohannParis: Nope, `lint-staged` is used by us.
We can simply to that with a normal script. let me remove this and make another PR about this.

mwdchang (approved):

.husky/pre-commit
YohannParis: Yes, just need to run `yarn install` after every merge from main

Issue #925, opened 2023-03-28
[FEAT]: new search bar for model page


Tom-Szendrey:
Add a search bar for the tables in the model page. This is just a front-end solution for now as we currently have data from 3 places that we will need to ground in the future

PR #934, opened 2023-03-29
feat: adding table filters to model page

Tom-Szendrey:
# Description

https://user-images.githubusercontent.com/17088680/228641457-2d17f331-5a0a-40ae-846a-4f90a86b8616.mov

Resolves #(issue)
#925


YohannParis:
I think it will be better if the search was for all the state/parameters instead of one per accordion.

Tom-Szendrey:
> I think it will be better if the search was for all the state/parameters instead of one per accordion.

I would need to move the bar to the outside, but youre saying something like this right?:
image

Are these not in their own distinct tables/accordions because theyre all different things? Im not familiar enough with using this workflow to argue youre lucky!

YohannParis:
> > I think it will be better if the search was for all the state/parameters instead of one per accordion.
> 
> I would need to move the bar to the outside, but youre saying something like this right?: image
> 
> Are these not in their own distinct tables/accordions because theyre all different things? Im not familiar enough with using this workflow to argue youre lucky!

Yes, maybe this is worth asking @jamiewaese-uncharted in #askem-hmi

Issue #902, opened 2023-03-23
[TASK]: make the math-editor great again


tangopapatime:
**Describe the task**
Update the math-editor
Change the model split panel layout based on the size of the viewing area.
Add edit/save buttons.

PR #937, opened 2023-03-30
fix: make the math editor great again

tangopapatime:
# Description

I made several changes under this task including a bunch of re-factoring including renaming the component to `tera-math-editor`
* The main split panel changes orientation based on its % of the screen real estate
* Moved configuration options into the component
* Added and moved the edit buttons for the petrinet editor and math equation editor
* Added some utility functions to help strip out the mathml that is being produced by mathlive
* Added some documentation :P

Resolves #(issue)
[TASK]: make the math-editor great again #902


tangopapatime:
This is mostly done.  I haven't set it to `save` an edited equation yet but some error handling is going to be needed.

packages/client/hmi-client/src/components/mathml/tera-math-editor.vue
```diff
@@ -0,0 +1,262 @@
...
+	if (props.mathMode) {
+		mathMode.value = props.mathMode;
+	}
+});
+
+onUpdated(() => {
+	if (mathMode.value === MathEditorModes.LIVE) {
+		mathLiveField.value?.setValue(latexEquation.value);
+	}
+});
```
mwdchang, line 133: Is this life-hook cycle necessary? Seems like you are doing the equivalent checks in the watchers already?

tangopapatime: It actually is.  I spent way too time trying to get this to work without doing this 😭

PR #938, opened 2023-03-31
chore: icons instead of text pill

YohannParis:
# Description

This started with the fact that we should use icons instead of text to represent a project assets in tabs and sidebar.
But it quickly brought me into a rabbit hole of cleaning and tidying a messy front-end with tabs, etc.

# Tasks
* Use Icons instead of text, icons names are defined in the types.
* Use Accordion to display assets instead of the tree component.
* Refactor the sidebar to use props instead of store.
* Rename slider-panel to tera-slider-panel.
* Clean up the tab group, and use CSS properties and CSS grid to organize them.
* Update the project services to add a boolean to the get project to fetch assets information at the same time.
* Remove intermediates from the hmi-server, as we do not use it.
* Handle remove asset back into tera-project.

# Video

https://user-images.githubusercontent.com/636801/229008395-b5c6fbdd-97c5-41be-b2ae-5f9a625c9333.mov

_Still TODO in another PR_
* Use props for tabs and not the store.
* handle all tabs and asset within the tera-project.
* provide projet types from the back-end.


packages/client/hmi-client/src/services/project.ts
```diff
@@ -156,4 +176,34 @@ async function home(): Promise<IProject[] | null> {
...
+	[ProjectAssetTypes.MODELS, 'share-alt'],
+	[ProjectAssetTypes.DATASETS, 'sliders-v'],
+	[ProjectAssetTypes.SIMULATIONS, 'cog'],
+	[ProjectAssetTypes.SIMULATION_RUNS, 'eye'],
+	[ProjectAssetTypes.CODE, 'code'],
+	['overview', 'inbox']
+]);
+function iconClassname(type: ProjectAssetTypes | string | null): string {
+	if (type && icons.has(type)) {
+		return `pi pi-${icons.get(type)}`;
```
mwdchang, line 193: Nitpick: How come this is a two step process? As in why not just store `pi pi-foobar` in the map?

mwdchang:
didn't perform a deep test, but looks fine!

mwdchang (approved):

packages/client/hmi-client/src/services/project.ts
YohannParis: I could make it in the enum, but it's just repetitive of `pi pi-` so I put it in one place.
Also, we test, because sometimes we can request icons of none exisiting types, so that's why we have a default.

PR #941, opened 2023-03-31
feat: add a pdf download button

chris-dickson:
This grabs the pdf if it's available using a public service to find it.  Given a DOI, we can recursively follow URL redirects until we either:
1. Get the PDF itself
2. Get to a webpage that links to the PDF

If 2. happens, we attempt to find a link to the PDF and save that instead.

We provide a simple button in the preview panel to initiate this. 
Screenshot 2023-03-31 at 10 42 31 AM

Screenshot 2023-03-31 at 10 42 52 AM


packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -454,6 +458,23 @@ const toggleDataItemSelected = (dataItem: { item: ResultType; type?: string }) =
 	}
 };
 
+const downloadPDF = async (dataItem: { item: ResultType; type?: string }) => {
+	const doi = dataItem.item?.identifier[0]?.id;
```
chris-dickson, line 462: @YohannParis this is throwing a lint error.  Am I doing this wrong?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/DownloadResource.java
```diff
@@ -0,0 +1,108 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+
+@Path("/api/download")
+@Authenticated
+@Tag(name = "User Annotation REST Endpoints")
```
dvince2, line 29: name looks wrong

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
YohannParis: If this is only for Documents and not Model or Dataset. You can do:

```typescript
import { DocumentType } from '@/types/Document';
...
const downloadPDF = async (dataItem: { item: DocumentType; type?: string }) => {
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/DownloadResource.java
```diff
@@ -0,0 +1,108 @@
...
+			}
+		} else {
+			// We actually have a document, if it's an HTML page with the content, look for a link to the pdf itself and follow
+			// it
+			final String contentType = response.getEntity().getContentType().getValue();
+			if (contentType.contains("html")) {
+				final String html = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
+				final Document document = Jsoup.parse(html);
+				final Elements links = document.select("a");
+				final String pdfUrl = links.stream()
```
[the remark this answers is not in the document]
chris-dickson: Yep.  This isn't an exact science but it's "good enough" for now.

dvince2 (approved):

YohannParis (approved):

PR #942, opened 2023-03-31
chore: clean up sidebar

YohannParis:
# Description

* Add numbers of assets in accordion
* Fix issue of opening two _overview_ tabs for no reasons
* Remove unused components
* Added assets count to overview tab
* Fix issue where tabs where using `xdd_uri` instead of `assetId`.
* Rename props of `Document.vue` to match content.

# Details

https://user-images.githubusercontent.com/636801/229195586-0c49c170-6637-4a2a-8504-9e3e029621f6.mov


packages/client/hmi-client/src/services/project.ts
```diff
@@ -195,6 +196,17 @@ function iconClassname(type: ProjectAssetTypes | string | null): string {
 	return 'pi pi-circle';
 }
 
+/**
+ * Get the xdd_uri of a Project Document
+ */
+function getDocumentAssetXddUri(project: IProject, assetId: Tab['assetId']): string | null {
```
mwdchang, line 202: Where would we expect assetId to be null?

mwdchang:
Code looks fine, didn't test - I assume you resolved the TDS integer vs uuid issue?

mwdchang (approved):

packages/client/hmi-client/src/services/project.ts
YohannParis: For new assets, when creating them on the front end.
Or just simply when the document.vue first mount and the props aren't filled yet.

PR #946, opened 2023-04-03
fix: use config property for ontology resolver

YohannParis:
# Description

* Thanks @dvince2 for showing me how to use config property
* Now use appilcation.properties for the mira ontology metaregistry resolver.


dvince2 (approved):
Nice!

PR #949, opened 2023-04-03
fix: drag- start anchor at the handle instead of of center

mwdchang:
### Summary
Start drag at node's drag anchor instead of node centre

https://user-images.githubusercontent.com/1220927/229611166-2f7047df-1280-4d52-86e8-ae9c99a24ebc.mov

### Testing
Go into edit mode, select a node and try to create an edge

### DIFF
--- a/packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
+++ b/packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
@@ -275,11 +275,15 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 		this.removeAllEvents('node-drag-end');
 
 		// (Re)create dragging listeners
-		this.on('node-drag-start', (_eventName, _event, selection: D3SelectionINode<NodeData>) => {
+		this.on('node-drag-start', (_eventName, event, selection: D3SelectionINode<NodeData>) => {
 			if (!this.isDragEnabled) return;
 			sourceData = selection.datum();
 			start.x = sourceData.x;
 			start.y = sourceData.y;
+
+			const target = event.sourceEvent.target;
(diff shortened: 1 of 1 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -275,11 +275,15 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
...
-		this.on('node-drag-start', (_eventName, _event, selection: D3SelectionINode<NodeData>) => {
+		this.on('node-drag-start', (_eventName, event, selection: D3SelectionINode<NodeData>) => {
 			if (!this.isDragEnabled) return;
 			sourceData = selection.datum();
 			start.x = sourceData.x;
 			start.y = sourceData.y;
+
+			const target = event.sourceEvent.target;
+			start.x += +d3.select(target).attr('cx');
+			start.y += +d3.select(target).attr('cy');
```
YohannParis, line 286: Why not? Just out of curiosity?
```suggestion
			const target = d3.select(event.sourceEvent.target);
			start.x += +target.attr('cx');
			start.y += +target.attr('cy');
```

YohannParis (approved):

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
mwdchang: No particular reason other than I was trying to use target.getBBox() in an earlier attempt.

mwdchang: changed

PR #950, opened 2023-04-03
feat: open pdf button uses generated link as fallback

shawnyama:
# Description

Open PDF button uses the generated link as a fallback

https://user-images.githubusercontent.com/28237258/229615788-c001f4ad-c8f4-4166-9fe4-10e6cad5219c.mov


YohannParis (approved):
Smart!

PR #951, opened 2023-04-03
chore: buttons to footer section

jamiewaese-uncharted:
- I set up an account at Canny to manage the `Report an issue` and `Request a feature` forums.
- Added a modal with information about the version number and copyright notice. Got the language from Will Brooke. Holly will be providing new words soon.

You might want to convert the version number in the About dialog into a variable that can be modified separately. That's a bit beyond my skillset at the moment but the UI is blocked in.


packages/client/hmi-client/src/App.vue
```diff
@@ -22,14 +22,59 @@
...
+					visualization and analysis software.
+				</p>
+				<br />
+				<p>&copy; Copyright Uncharted Software {{ new Date().getFullYear() }}</p>
+			</div>
+		</div>
+		<template #footer>
+			<Button label="Close" @click="visible = false" />
+		</template>
+	</Dialog>
```
YohannParis, line 70: We have a `<Modal>` component `/src/components/widgets/Modal.vue`

jamiewaese-uncharted: Switched to the Modal component. Thanks.

YohannParis (approved):

PR #960, opened 2023-04-04
chore: 944 renderer updates 3 5 6

Tom-Szendrey:
# Description
Some of the styling for mode editor:
-Click to focus
-Click away to unfocus
-Green border on dragging node
-Enter to exit typing

https://user-images.githubusercontent.com/17088680/229910760-4e0822bb-e603-43ad-8a6e-c2a89197d181.mov

Resolves #(issue)


packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -480,6 +480,20 @@ const editorKeyHandler = (event: KeyboardEvent) => {
...
+				.selectAll('.no-drag')
+				.style('opacity', 0)
+				.style('visibility', 'hidden');
+			renderer.nodeSelection = null;
+		}
+		if (renderer?.edgeSelection) {
+			renderer.deselectEdge(renderer.edgeSelection);
+			renderer.edgeSelection = null;
+		}
+	}
```
mwdchang, line 496: If you are already checking `if (renderer && ..)` you shouldn't need to use the option-chaining thingy. I think you can just do directly
```
if (renderer.nodeSelection) { ... }
```

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -329,6 +335,11 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 		});
 
 		this.on('node-click', (_eventName, _event, selection: D3SelectionINode<NodeData>) => {
+			// Set focus on node:
+			this?.chart?.selectAll('.node-ui').style('opacity', '0.3');
+			this?.chart?.selectAll('.edge').style('opacity', '0.3');
+			selection.style('opacity', '1');
```
mwdchang, line 341: Easier to read if you pass in numbers `0.3`, `1.0` or whatnot.

The optional-chaining is weird .... why do you have `this?`?

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -369,6 +380,10 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 		});
 
 		this.on('background-click', () => {
+			// Reset opacity from focus:
+			this?.chart?.selectAll('.node-ui').style('opacity', '1');
+			this?.chart?.selectAll('.edge').style('opacity', '1');
+
```
mwdchang, line 386: Easier to read if you pass in numbers `0.3`, `1.0` or whatnot.

The optional-chaining is weird .... why do you have `this?`?

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
```diff
@@ -276,6 +277,8 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 
 		// (Re)create dragging listeners
 		this.on('node-drag-start', (_eventName, event, selection: D3SelectionINode<NodeData>) => {
+			selection.select('circle').attr('stroke', highlightedStrokeColour);
+			selection.select('rect').attr('stroke', highlightedStrokeColour);
```
mwdchang, line 281: This almost works but not quite right, you cannot arbitrarily select circle/rect as I never give any guarantee which circle/rect you are going to get. e.g. you are changing the style of things you don't want. Note below the right anchor is larger.

image


So if we take a transition-node for example, it is made up of
- 1 rect
- 4 circles (the handles)
- text

Your selection logic would change the rect and the circle, in this case the first circle which happens to be the east-facing one.

So you probably want to add a specific class selector such that the shape can be uniquely identified.

packages/client/hmi-client/src/components/models/Model.vue
Tom-Szendrey: Totally correct thanks!

packages/client/hmi-client/src/petrinet/petrinet-renderer.ts
Tom-Szendrey: Good catch, thanks!

mwdchang (approved):

PR #966, opened 2023-04-05
feat: adobe pdf viewer integration

shawnyama:
# Description

You can view the PDF in Document.vue

https://user-images.githubusercontent.com/28237258/230186514-8de1e405-c005-45bb-9dc0-d84ac749fa05.mov

Followed this and related articles: https://www.raymondcamden.com/2022/09/30/using-the-adobe-pdf-embed-api-with-vue-3


YohannParis (commented):
Looks good, but can you add info in the `DEVELOPMENT.md` file about adding the API key in the `.env` file.

YohannParis (commented):
this is almost good to go, I'm just thinking because this will be in the front-end, we don't care about encoding the API key.
So we will get one and just add it to this PR.

packages/client/hmi-client/src/components/widgets/pdf-embed.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+	<div id="adobe-dc-view"></div>
```
mwdchang, line 2: Does having an `id=xyz` hardwired like this cause a problem if we have multiple pdf tabs open at the same time? or the case where they are opened across multiple panels/windows?

YohannParis: @shawnyama can you look into it please.

shawnyama: I think it's fine. One document component is shown at a time so there aren't going to be duplicate ids

YohannParis (approved):
Cheers !

PR #969, opened 2023-04-05
feat: multiple design tweaks to the notes side panel

### LINKED ISSUE #807: [TASK]: Multiple design tweaks to the notes side panel
### Code

Clean up the code so that it:
* uses services instead of calling it from within `<tera-project>`.
* Saves notes per project/assets in `data-service`
* Add proper error handlings on all API calls, etc.

### Design

Nice work on the notes side panel. Here are some design tweaks to make it pretty:
- The width of the side panel should be 240 px
- There should be 16 px of padding on the left and right
- There should be 16 px of padding between different notes
- The font size should be caption (12 px) for everything
- There should be a borderless dropdown selector at the top of each note for users to select which section of their notes document they want to assign each note to. The options should be: Unassigned, Abstract, Intro, Methods, Results, Discussion, References, and "+ Add section" where you can add another section and name it whatever you want.
- The author and creation date should be on one line
- The "Add note" button should just be a text button (see design)
- The empty text input field should not be shown by default. It should only be revealed when the user clicks "Add note". (See design).
- The text input field border should be the same grey as all other input fields, not black. 
- There should be an 3-dot menu button in the corner. Clicking it should open a menu with two options: Edit and Delete. If you click delete, another option appears with "Are you sure?". (See design)

CURRENT
image

DESIGN
Overall
image

Default note
image

Options menu
image

Delete confirmation
image

Edit mode
image

echl:
https://user-images.githubusercontent.com/16358728/230941991-6c7d7488-2cbc-4151-aae7-24ee439ccf4c.mov

- No add section yet; suggest using the vue dropdown menu's existing option to allow the user to enter new sections into the dropdown menu directly i.e. it functions as a dropdown and input field
- Input text field loses text if it regains focus from somewhere else

Resolves #807


YohannParis:
Please add information and screenshot to the description

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/entities/Annotation.java
```diff
@@ -55,6 +55,9 @@ public class Annotation extends PanacheEntityBase implements Serializable {
 	@JsonProperty("artifact_type")
 	private String artifactType;
 
+	@Column(nullable = true)
+	private String section;
```
mwdchang, line 59: What is this field meant to convey? Where the annotation points to? 

I wonder if we want to have a a more generic data representation now to avoid changing this a few more times in the future? For example one might imagine you may have a starting-characer-number, ending-character-number kind of thing. 

Any thoughts?

echl: They are predefined sections. See https://github.com/DARPA-ASKEM/Terarium/issues/807

YohannParis (approved):

PR #971, opened 2023-04-10
chore: add vue feather icons and some custom icons

shawnyama:
# Description

Added proper icons

Two resources for icons have been added:
- [feather icons](https://github.com/fengyuanchen/vue-feather)
- our own custom icons which are SVGs imported as Vue components (these components are automatically generated by the newly added [vite-svg-loader](https://github.com/jpkleemans/vite-svg-loader) package)


packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
```diff
@@ -92,8 +114,10 @@ header {
 	display: none;
 	visibility: hidden;
 }
+
 .p-button {
-	flex-shrink: 0; /* Avoid the highlight to be squished */
+	flex-shrink: 0;
+	/* Avoid the highlight to be squished */
```
YohannParis, line 120: If it can't be in one line, then the comment should be above

```suggestion
	/* Avoid the highlight to be squished */
	flex-shrink: 0;
```

packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
```diff
@@ -2,7 +2,25 @@
...
+						stroke="rgb(16, 24, 40)"
+					/>
+				</template>
+				<component
+					v-else
+					:is="getAssetIcon(tab.assetType ?? null)"
+					class="p-button-icon-left"
+					fill="rgb(16, 24, 40)"
+					style="overflow: visible"
+				/>
```
YohannParis, line 23: Okay this is too much, it would be simpler to use only _Feather Icons_ for all assets, and stop using _Prime Icons_ `pi pi-*` to avoid this kafkaesque rendering.
I know you spend a good amount of time on this solution, and it's nice, but this is too complex just for some icons.

packages/client/hmi-client/vite.config.ts
```diff
@@ -50,7 +51,8 @@ export default defineConfig({
 					isCustomElement: (tag) => tag.startsWith('facet-') || tag === 'math-field'
 				}
 			}
-		})
+		}),
+		svgLoader({ defaultImport: 'url' })
```
YohannParis, line 55: Could you add a comment about why we are using it.

packages/client/hmi-client/src/services/project.ts
```diff
@@ -181,17 +184,17 @@ async function home(): Promise<IProject[] | null> {
...
-	[ProjectAssetTypes.SIMULATION_RUNS, 'eye'],
-	[ProjectAssetTypes.CODE, 'code'],
-	['overview', 'inbox']
+	[ProjectAssetTypes.DOCUMENTS, 'pi pi-file'],
+	[ProjectAssetTypes.MODELS, 'pi pi-share-alt'],
+	[ProjectAssetTypes.DATASETS, DatasetIcon],
+	[ProjectAssetTypes.SIMULATIONS, 'pi pi-cog'],
+	[ProjectAssetTypes.SIMULATION_RUNS, ResultsIcon],
+	[ProjectAssetTypes.CODE, 'pi pi-code'],
+	['overview', 'layout']
```
YohannParis, line 193: What I mentioned above, let's drop the _Prime Icons_, just find the ones that looks close enough for now.

packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
```diff
@@ -61,22 +59,23 @@
...
+						<vue-feather
+							class="p-button-icon-left"
+							type="git-merge"
+							size="1.25rem"
+							stroke="rgb(16, 24, 40)"
+						/>
+						<span class="p-button-label">New workflow</span>
+					</Button>
+					<Button size="large" class="p-button p-button-secondary quick-link-button">
+						<compare-models-icon fill="rgb(16, 24, 40)" style="scale: 1.25" />
```
YohannParis, line 72: Because this is an SVG, it might be better to add a class, and style with CSS for both `fill` and `style` attributes.

Issue #944, opened 2023-03-31
[TASK]: Misc UX & design tweaks for the Petri net renderer


jamiewaese-uncharted:
- [x] 1) Move the "Edit" button to the top right corner of its container. 
![image](https://user-images.githubusercontent.com/120480244/229229907-676b8796-87d8-4871-9ec8-968bddb592d3.png)

- [x] 2) Add a toolbar in edit mode. The right click menu isn't easily discoverable. 
** note the "Reset zoom" button on the left. This should restore the zoom and layout to the "fit all" default.
![image](https://user-images.githubusercontent.com/120480244/229232178-dfe33533-2eb3-4e5e-adfa-dc90b76ca53f.png)

- [x] 3) When editing a label, if the user presses “return” the label editing mode should close.
![image](https://user-images.githubusercontent.com/120480244/229230700-d1674376-ee53-4e67-bd7b-67b8cc8881fb.png)

- [x] 4) Add “click anywhere” to close / cancel the popup menu if you change your mind. 
![image](https://user-images.githubusercontent.com/120480244/229230955-63b823c6-75e2-479b-a718-25d0e916e82b.png)

- [x] 5) Change border color while object is being dragged (onmousedown?)
![image](https://user-images.githubusercontent.com/120480244/229231128-a9a2cd53-ef95-4b3f-80f7-72bfa1c71990.png)

- [x] 6) Click on an object to focus on it. This should change the border color and reduce the opacity of everything else to 30%. Click anywhere to unselect
![image](https://user-images.githubusercontent.com/120480244/229231451-27dc4aa6-7734-4810-8299-e21f51812879.png)

- [x] 7) While drawing links, the line should appear to come from the handle you clicked on
![image](https://user-images.githubusercontent.com/120480244/229231623-002b06b2-191d-41ec-b867-79d2ef579fc4.png)

- [ ] 8) Line should be curvy, not straight
![image](https://user-images.githubusercontent.com/120480244/229231789-0ccf046b-7eaf-46bb-9b31-782eb39497c6.png)

- [ ] 9) Adjust line endings so they stop before arrowhead
![image](https://user-images.githubusercontent.com/120480244/229231985-6d629e95-4d9a-49a9-91b6-f8f46cd0ecd8.png)

PR #973, opened 2023-04-10
feat: 944 renderer updates 2

Tom-Szendrey:
# Description
-Adding model task bar to add states and transitions + reset zoom
Setting graph's width + height attributes post layout
Changing zoom to be based off of width instead of height

Resolves #(issue)
#944

### DIFF
--- a/packages/client/hmi-client/src/services/graph.ts
+++ b/packages/client/hmi-client/src/services/graph.ts
@@ -124,5 +124,19 @@ export const runDagreLayout = <V, E>(graphData: IGraph<V, E>, lr: boolean = true
 		dupe.add(hash);
 	}
 
+	// Find new width and height
+	let minX = Number.MAX_VALUE;
+	let maxX = Number.MIN_VALUE;
+	let minY = Number.MAX_VALUE;
+	let maxY = Number.MIN_VALUE;
+	for (let i = 0; i < graphData.nodes.length; i++) {
+		if (graphData.nodes[i].x < minX) minX = graphData.nodes[i].x;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/services/graph.ts
```diff
@@ -124,5 +124,19 @@ export const runDagreLayout = <V, E>(graphData: IGraph<V, E>, lr: boolean = true
...
+	let minX = Number.MAX_VALUE;
+	let maxX = Number.MIN_VALUE;
+	let minY = Number.MAX_VALUE;
+	let maxY = Number.MIN_VALUE;
+	for (let i = 0; i < graphData.nodes.length; i++) {
+		if (graphData.nodes[i].x < minX) minX = graphData.nodes[i].x;
+		if (graphData.nodes[i].x > maxX) maxX = graphData.nodes[i].x;
+		if (graphData.nodes[i].y < minY) minY = graphData.nodes[i].y;
+		if (graphData.nodes[i].y > maxY) maxY = graphData.nodes[i].y;
+	}
```
YohannParis, line 143: Just to be simpler to read
```suggestion
	graphData.nodes.forEach({x, y} => {
		if (x < minX) minX = x;
		if (x > maxX) maxX = x;
		if (y < minY) minY = y;
		if (y > maxY) maxY = y;
	}
```

YohannParis (approved):

packages/client/hmi-client/src/services/graph.ts
Tom-Szendrey: `{x,y}` did not work unfortunately. But I rewrote to use `foreach(node) `to help the readability!

PR #982, opened 2023-04-11
fix: model renderer small tweaks

Tom-Szendrey:
# Description
When a graph has no nodes do not try to recalculate the sizing

### DIFF
--- a/packages/client/hmi-client/src/services/graph.ts
+++ b/packages/client/hmi-client/src/services/graph.ts
@@ -131,18 +131,21 @@ export const runDagreLayout = <V, E>(graphData: IGraph<V, E>, lr: boolean = true
 	}
 
 	// Find new width and height
-	let minX = Number.MAX_VALUE;
-	let maxX = Number.MIN_VALUE;
-	let minY = Number.MAX_VALUE;
-	let maxY = Number.MIN_VALUE;
-	graphData.nodes.forEach((node) => {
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/graph.ts
```diff
@@ -131,18 +131,21 @@ export const runDagreLayout = <V, E>(graphData: IGraph<V, E>, lr: boolean = true
...
+		let maxY = Number.MIN_VALUE;
+		graphData.nodes.forEach((node) => {
+			if (node.x < minX) minX = node.x;
+			if (node.x > maxX) maxX = node.x;
+			if (node.y < minY) minY = node.y;
+			if (node.y > maxY) maxY = node.y;
+		});
+
+		graphData.width = Math.abs(maxX - minX) + 2 * nodeWidth;
+		graphData.height = Math.abs(maxY - minY) + 2 * nodeHeight;
```
YohannParis, line 147: Those properties `width, height` can be non-existant and it works fine?

YohannParis (approved):

packages/client/hmi-client/src/services/graph.ts
Tom-Szendrey: They have a default set in the constructor. If there are no nodes we will just leave them as these default values rather than recalculating when we call this runLayout function

PR #983, opened 2023-04-11
feat: update simulation scenario persistence implementation

echl:
# Description

* Implement hmi-server endpoints for saving simulations ahead of changes to data-service
* Temporarily save new simulation object to deprecated `SimulationRun.parameters`
* Retain current saving of simulation data to session storage due data-service not having an endpoint to get all `SimulationRuns`, therefore not being able to load all simulations for a given modelId without saving a list; neither is there an object I can save AND update that info into


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/SimulationProxy.java
```diff
@@ -15,6 +16,28 @@
...
+
+	@POST
+	Response createSimulation(
+		Simulation simulation
+	);
+
+	@PATCH
+	Response updateSimulation(
+		Simulation simulation
+	);
```
YohannParis, line 33: Missing a `@Path("/{id})`

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -32,11 +26,59 @@ export async function getRunStatus(runId: number) {
...
+}
+
+export async function createSimulation(name, description, simulationParams, result) {
+	try {
+		const response = await API.post('/simulation', {
+			name,
+			description,
+			simulationParams,
+			result
+		});
```
YohannParis, line 55: Would be better to pass a `simulation Simulation` as argument.
```suggestion
export async function createSimulation(simulation: Simulation) {
	try {
		const response = await API.post('/simulation', simulation);
```

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -32,11 +26,59 @@ export async function getRunStatus(runId: number) {
...
+		const resp = await API.get(`simulation/${runId}/result`);
 		const output = resp.data;
 		return output;
 	} catch (err) {
 		logger.error(err);
 		return null;
 	}
 }
+
+export async function getSimulation(id) {
```
YohannParis, line 38: Let's leverage the Types as much as possible

```suggestion
export async function getSimulation(id: Simulation["id"]): Simulation {
```

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -32,11 +26,59 @@ export async function getRunStatus(runId: number) {
...
+			result
+		});
+		return response.data;
+	} catch (error) {
+		logger.error(error);
+		return null;
+	}
+}
+
+export async function updateSimulation(id, name, description) {
```
YohannParis, line 63: Here too you can pass a simulation object, and check if it has an id before calling the endpoint

YohannParis (commented):
Great first pass!

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -47,7 +47,17 @@ export async function getSimulation(id: Simulation['id']): Promise<Simulation |
 
 export async function createSimulation(simulation: Simulation): Promise<Simulation | null> {
 	try {
-		const response = await API.post('/simulation', simulation);
+		// const response = await API.post('/simulation', simulation);
+		const request = {
+			simulator_id: 1,
```
YohannParis, line 52: Why do you need to send a `simulator_id`?

echl: It's simply a required field by data-service

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/SimulationProxy.java
```diff
@@ -15,6 +16,30 @@
 @Produces(MediaType.APPLICATION_JSON)
 public interface SimulationProxy {
 
+	@GET
+	@Path("/{id}")
+	Response getSimulation(
```
dvince2, line 21: Rather than `Response`, should these return a `Simulation`? Looking at the front end it looks like that is what is being expected.  For just generic responses that we don't really care about we can probably avoid using POJOs but when we have them made may as well use them :)

dvince2: (This goes for any of the other endpoints here - I'm not sure what is being returned from these end points or what java objects we have but if we have them lets use them)

dvince2 (changes requested):
Looks good! just one change i think to use java objects where possible

PR #989, opened 2023-04-13
feat: added ability to create a new model

tangopapatime:
# Description
This will allow a user to create a new model within the project.
Fields that the user can edit are as follows:

1) Title of model
2) Description
3) Formula

Resolves #(issue)
988-feat-add-ability-to-create-new-model

[newmodel.webm](https://user-images.githubusercontent.com/98172806/231870979-3c50ceaa-8456-4430-b85d-2e560bbab3ee.webm)

### DIFF
--- a/packages/client/hmi-client/src/components/models/Model.vue
+++ b/packages/client/hmi-client/src/components/models/Model.vue
@@ -315,12 +315,10 @@ const extractions = ref([]);
 const props = defineProps({
 	assetId: {
 		type: String,
-		default: '',
 		required: true
 	},
 	isEditable: {
 		type: Boolean,
-		default: false,
 		required: true
 	},
 	highlight: {
--- a/packages/client/hmi-client/src/components/models/Model.vue
+++ b/packages/client/hmi-client/src/components/models/Model.vue
@@ -3,8 +3,17 @@
 		<header>
 			<div class="framework">{{ model?.framework }}</div>
 			<div class="header-and-buttons">
-				<h4 v-html="title" />
-				<span v-if="isEditable">
+				<InputText
+					v-if="assetId === ''"
+					v-model="newModelName"
+					class="model-title-text-area"
+					placeholder="Title of New Model"
+				/>
+				<h4 v-else v-html="title" />
+				<span v-if="assetId === ''">
+					<Button @click="createNewModel" label="Create New Model" class="p-button-sm" />
+				</span>
+				<span v-else-if="isEditable">
 					<Button
 						@click="launchForecast"
 						label="Open simulation space"
@@ -18,7 +27,16 @@
 		</header>
 		<Accordion :multiple="true" :active-index="[0, 1, 2, 3, 4]">
 			<AccordionTab header="Description">
-				<p v-html="description" class="constrain-width" />
+				<p v-if="assetId !== ''" v-html="description" class="constrain-width" />
+				<section v-else>
+					<label for="placeholder"></label
+					><Textarea
+						v-model="newDescription"
+						class="model-description-text-area"
+						rows="5"
+						placeholder="Description of New Model"
+					/>
+				</section>
 			</AccordionTab>
 			<AccordionTab header="Model diagram">
 				<section class="model_diagram">
@@ -169,7 +187,11 @@
 					/>
 				</template>
 			</AccordionTab>
-			<AccordionTab :header="`Extractions ${extractions?.length}`">
+			<AccordionTab
+				:header="`Extractions ${
+					extractions?.length ? extractions?.length : ': No Extractions Found'
+				}`"
+			>
 				<DataTable :value="extractions">
 					<Column field="name" header="Name" />
 					<Column field="id" header="ID" />
@@ -244,7 +266,7 @@
 <script setup lang="ts">
 import { remove, isEmpty, pickBy, isArray } from 'lodash';
 import { IGraph } from '@graph-scaffolder/index';
-import { watch, ref, computed, onMounted, onUnmounted } from 'vue';
+import { watch, ref, computed, onMounted, onUnmounted, defineEmits } from 'vue';
 import { runDagreLayout } from '@/services/graph';
 import { PetrinetRenderer } from '@/petrinet/petrinet-renderer';
 import {
@@ -257,6 +279,8 @@ import {
 	petriToLatex,
 	NodeType
 } from '@/petrinet/petrinet-service';
+import Textarea from 'primevue/textarea';
+import InputText from 'primevue/inputtext';
 import { separateEquations, MathEditorModes } from '@/utils/math';
 import { getModel, updateModel } from '@/services/model';
 import { getRelatedArtifacts } from '@/services/provenance';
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/page/Home.vue
```diff
@@ -140,7 +140,7 @@
 				</template>
 				<template #default>
 					<form>
-						<label for="new-project-name">Name</label>
+						<label for="new-project-name">Name</label>F
```
YohannParis, line 143: That must be a typo

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -834,4 +902,23 @@ section math-editor {
...
+}
+
+.model-title-text-area {
+	border: 1px solid var(--surface-border-light);
+	border-radius: 4px;
+	padding: 5px;
+	resize: none;
+	overflow-y: hidden;
+	width: 100%;
+	font-size: Bold 1.25rem;
```
YohannParis, line 922: Nope, we don't do this, there are multiple font-size you can use.

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -834,4 +902,23 @@ section math-editor {
 .constrain-width {
 	max-width: 60rem;
 }
+
+.model-description-text-area {
+	border: 1px solid var(--surface-border-light);
+	border-radius: 4px;
```
YohannParis, line 908: We have border-radius values you can choose from _variables.scss

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -646,24 +701,37 @@ const hasNoEmptyKeys = (obj: Record<string, unknown>): boolean => {
 	return Object.keys(nonEmptyKeysObj).length === Object.keys(obj).length;
 };
 
+const createNewModel = async () => {
+	const newModel = {
+		name: newModelName.value,
+		framework: 'Petri Net',
+		description: newDescription.value,
+		content: JSON.stringify(newPetri.value ?? { S: [], T: [], I: [], O: [] })
```
YohannParis, line 709: This is going to bite us in the *** @mwdchang I know it.

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -359,16 +371,42 @@ function openAsset(asset: Tab = tabs.value[activeTabIndex.value], newCode?: stri
...
+		const model = await getModel(newAssetId.value);
+
+		if (model) {
+			resources.activeProjectAssets?.[ProjectAssetTypes.MODELS].push(model);
+		} else {
+			logger.warn('Could not add new model to project.');
+		}
+		isNewModel.value = false;
+	}
+};
```
YohannParis, line 408: This should be in `services/model.ts`

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -388,6 +426,12 @@ async function openNewModelFromCode(modelId, modelName) {
 	});
 }
 
+// create the new sset
```
YohannParis, line 429: Typo

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -359,16 +371,42 @@ function openAsset(asset: Tab = tabs.value[activeTabIndex.value], newCode?: stri
...
+// Create the new model
+const createNewModel = async (newModel) => {
+	const newModelResp = await createModel(newModel);
+	if (newModelResp) {
+		newAssetId.value = newModelResp.id.toString();
+	}
+
+	if (newAssetId.value) {
+		// add to sidebar
+		await ProjectService.addAsset(props.project.id, ProjectAssetTypes.MODELS, newAssetId.value);
```
YohannParis, line 397: I wonder if this endpoint return the asset already, I should write an issue about it.

packages/client/hmi-client/src/components/models/Model.vue
tangopapatime: Likely won't hurt in the meantime in the case of creating a new empty model.  We should however make sure we can handle the case when it is empty.

mwdchang: @YohannParis can you please elaborate? 

Is this because you want to hide the S/D/I/O data structures? Or is it something else?

I don't think it will be the case that the UI will not need to known about STIO stuff - especially with the way the properties and observation functions are going with ensembles that forces you to acknowledge where things are. I think for now it is fine ... probably in the longer term we want to have a `model-interface` to encapsulate some of these yuckiness. Something like

```
content: Petrinet.emptyModel()...
content: ReactionNet.emptyMode()
```

Kind of thing .... and `gene reaction network` is rumoured to be the next modelling framework we will take a crack at.

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -283,24 +307,41 @@ import Splitter from 'primevue/splitter';
...
+		type: Boolean,
+		default: false,
+		required: true
+	},
+	highlight: {
+		type: Boolean,
+		default: false,
+		required: false
+	}
+});
```
mwdchang, line 331: I think generally if you have required=true, you don't want to specify defaults.

packages/client/hmi-client/src/components/models/Model.vue
YohannParis: Yes, but I wonder if this should be handle in the back-end, the model object does't have to know in the front-end what is the content of a graph, only the graph-editor. I just think less places where we need to know the format the better.

packages/client/hmi-client/src/services/model.ts
```diff
@@ -52,3 +71,19 @@ export async function updateModel(model: Model) {
 	});
 	return response?.data ?? null;
 }
+
+export async function addModelToProject(
+	projectId: string,
+	assetId: string,
+	resources: ResourceType
+) {
+	await ProjectService.addAsset(projectId, ProjectAssetTypes.MODELS, assetId);
```
YohannParis, line 80: We should handle the response from this call

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -266,7 +267,7 @@ const tabStore = useTabStore();
 const router = useRouter();
 const resources = useResourcesStore();
 
-const newAssetId = ref<string>('');
+const newModelId = ref<string>('');
```
YohannParis: This is used by all the other assets type, we cannot just make it for model

packages/client/hmi-client/src/page/project/tera-project.vue
tangopapatime: It doesn't right now.

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -3,8 +3,17 @@
 		<header>
 			<div class="framework">{{ model?.framework }}</div>
 			<div class="header-and-buttons">
-				<h4 v-html="title" />
-				<span v-if="isEditable">
+				<InputText
+					v-if="assetId === ''"
```
echl, line 7: You could simplify to `v-if="assetId"`

packages/client/hmi-client/src/components/models/Model.vue
```diff
@@ -18,7 +27,16 @@
...
 			<AccordionTab header="Description">
-				<p v-html="description" class="constrain-width" />
+				<p v-if="assetId !== ''" v-html="description" class="constrain-width" />
+				<section v-else>
+					<label for="placeholder"></label
+					><Textarea
+						v-model="newDescription"
+						class="model-description-text-area"
+						rows="5"
+						placeholder="Description of New Model"
```
echl, line 37: Same thing about sentence case

packages/client/hmi-client/src/components/models/Model.vue
tangopapatime: yeah, I was toying around with another falsy state so I kept this explicit.  Didn't remove it.

packages/client/hmi-client/src/components/models/Model.vue
mwdchang: I suspect that will not be case because of the discussions around metadata ... e.g. the metadata are embedded in S and T directly so you have to acknowledge that.

I think we can minimize UI's manipulation of the inner workings ... kind of like a read-only mode unless in the model-editor.

tangopapatime: I'll let you guys resolve this amongst yourselves.

YohannParis (approved):

PR #990, opened 2023-04-14
feat: search by example pops up while asset is being dragged

shawnyama:
# Description

The search by example popup is on if 
- an asset is being dragged
- an asset has just been successfully dropped in
- if the popup toggle is on.

https://user-images.githubusercontent.com/28237258/232125300-af2b9671-607e-42d5-a7f3-99604d4cd35b.mov


YohannParis:
Based on the video, you should empty the input when opening it.

YohannParis (approved):

PR #1002, opened 2023-04-18
draft: misc tab bug fixes

echl:
# Description

* Make tera-tab-group and tera-project more "in sync" such that the tab header and page content switch at the same time on tab change (same behaviour as browsers) plus a small animation for more responsive feel
* Remember opened tabs when navigating away from project 

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/documents/Document.vue
+++ b/packages/client/hmi-client/src/components/documents/Document.vue
@@ -301,7 +301,7 @@ const doc = ref<DocumentType | null>(null);
 const pdfLink = ref<string | null>(null);
 const documentView = ref(DocumentView.EXRACTIONS);
 
-const emit = defineEmits(['open-asset']);
+const emit = defineEmits(['open-asset', 'asset-loaded']);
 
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/documents/Document.vue
YohannParis, line 1: Why do we have a `Document.vue` and a `tera-document.vue`?

packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
```diff
@@ -113,6 +141,34 @@ i {
...
+	}
+
+	50% {
+		width: calc(100% - 2px);
+	}
+
+	100% {
+		/* opacity: 0; */
+	}
+}
```
YohannParis, line 165: You can remove the comments and the 100%.

packages/client/hmi-client/src/components/documents/Document.vue
echl: whoops, this happened when merging main. will fix

YohannParis (approved):

PR #1004, opened 2023-04-19
feat: 967 task dataset viewer add column summary histograms just above the datatable

Tom-Szendrey:
# Description
Change `hmi-server`'s `getCsv` endpoint to now return an object with some more information in it that just a raw file.
Add some histograms to our datatables in our `datatable` tab
Theres still some styling to do to make this a bit prettier. That will come in next PR
image

Resolves #(issue)


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvAsset.java
```diff
@@ -0,0 +1,33 @@
...
+@Accessors(chain = true)
+@Slf4j
+@TSModel
+public class CsvAsset implements Serializable {
+	List<List<String>> csv;
+	@TSOptional
+	List<CsvColumnStats> stats;
+	List<String> headers;
+
+	public CsvAsset(List<List<String>> csv, List<CsvColumnStats> stats, List<String> headers){
```
dvince2, line 22: Having explicit constructors is fine, but with using Lombok we can technically avoid these.  If you add the annotation `@AllArgsConstructor` Lombok will auto create a constructor for all the fields on this class (`csv`, `stats`, `headers`)

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvAsset.java
```diff
@@ -0,0 +1,33 @@
...
+		this.csv = csv;
+		this.stats = stats;
+		this.headers = headers;
+	}
+
+	public CsvAsset(List<List<String>> csv, List<String> headers){
```
dvince2, line 28: Similarly here...
If `csv` and `headers` are always required to be present on this object, you can annotate those specific fields with `@NonNull`, and then add the annotation `@RequiredArgsConstructor` to the class itself. This will....

1) Add null checking to the all-args constructor
2) Create a second constructor for just `csv` and `headers` 

Again, both of these suggestions are optional - what you have here works fine - but Lombok can save you some typing :)

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvColumnStats.java
```diff
@@ -0,0 +1,31 @@
...
+@TSModel
+public class CsvColumnStats implements Serializable {
+	List<Integer> bins; 
+	double minValue; 
+	double maxValue;
+	double mean;
+	double median;
+	double sd;
+
+	public CsvColumnStats(List<Integer> bins, double minValue, double maxValue, double mean, double median, double sd){
```
dvince2, line 23: Simlarly here with `@AllArgsConstructor`

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -168,9 +178,30 @@ public Response getCsv(
 		@PathParam("id") final String id,
 		@DefaultValue("true") @QueryParam("wide_format") final Boolean wideFormat,
 		@DefaultValue("false") @QueryParam("data_annotation_flag") final Boolean dataAnnotationFlag,
-		@DefaultValue("50") @QueryParam("row_limit") final Integer rowLimit
+		@DefaultValue("50") @QueryParam("row_limit") final Integer rowLimit,
+		@DefaultValue("0") @QueryParam("binCount") final Integer binCount
 	) {
-		return proxy.getCsv(id, wideFormat, dataAnnotationFlag, rowLimit);
+		
+		log.info("Getting csv content");
```
dvince2, line 185: debug statement

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -168,9 +178,30 @@ public Response getCsv(
...
+		Response returnResponse = proxy.getCsv(id, wideFormat, dataAnnotationFlag, rowLimit);
+		String rawCsvString = proxy.getCsv(id, wideFormat, dataAnnotationFlag, rowLimit).readEntity(String.class);
```
dvince2, line 187: You may want to have error handling here if the proxied service is down/returns a bad value or empty string, etc.  As an example, see `HomeResource.java` starting around line 84 and 105

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -183,4 +214,87 @@ public Response uploadFile(
...
+			// Fill bins:
+			for (int i = 0; i < numberList.size(); i++){
+				Integer index = (int)Math.abs(Math.floor((numberList.get(i) - numberList.get(0)) / stepSize));
+				Integer value = bins.get(index);
+				bins.set(index,value + 1);
+			}
+			
+			return new CsvColumnStats(bins,minValue,maxValue,meanValue,medianValue,sdValue);
+		
+		}catch(RuntimeException e){
```
dvince2, line 270: The `RuntimeException` is mostly being thrown by our proxy handler error catching, which I don't see a call to proxy here. You'll likely want a different exception type, or just use `Exception` to catch anything

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -183,4 +214,87 @@ public Response uploadFile(
...
+			//Cannot convert column to double, just return empty list.
+			return new CsvColumnStats(bins,0,0,0,0,0);
+		}	
+	}
+
+	//Assume list is already sorted because i know when its being called its already sorted..
+	private double getMedianValue(List<Double> aList){
```
dvince2, line 277: I wonder if for this and standard deviation we could use Google Guava - then we don't need to worry about sorted or not

Tom-Szendrey: I wasnt sure about adding any libraries in general so i figured Id just write the helper functions myself. I know I am calling this with a sorted list so I dont need to worry about resorting it, but if we want to use Guava let me know

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvAsset.java
Tom-Szendrey: Annotation magic scares cavemen like me

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvAsset.java
Tom-Szendrey: Annotation magic scares cavemen like me

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/CsvColumnStats.java
Tom-Szendrey: Annotation magic scares cavemen like me

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -168,9 +178,48 @@ public Response getCsv(
...
-		return proxy.getCsv(id, wideFormat, dataAnnotationFlag, rowLimit);
+		
+		log.debug("Getting csv content");
+		Response returnResponse;
+		String rawCsvString; 
+		try {
+			returnResponse = proxy.getCsv(id, wideFormat, dataAnnotationFlag, rowLimit);
+			rawCsvString = returnResponse.readEntity(String.class);
+			if (rawCsvString.length() == 0){
+				log.debug("No csv assosiated with this ID");
```
YohannParis, line 192:
```suggestion
				log.debug("No CSV associated with this ID");
```

packages/client/hmi-client/src/components/dataset/Dataset.vue
```diff
@@ -108,39 +108,56 @@
...
+								<Chart
+									type="bar"
+									:height="800"
+									:data="chartData?.at(index)"
+									:options="chartOptions"
+								/>
+								<div class="histogram-label-max">Max: {{ csvMaxsToDisplay?.at(index) }}</div>
+								<div class="histogram-label-other">Mean: {{ csvMeansToDisplay?.at(index) }}</div>
+								<div class="histogram-label-other">Median: {{ csvMedianToDisplay?.at(index) }}</div>
+								<div class="histogram-label-other">SD: {{ csvSdToDisplay?.at(index) }}</div>
```
YohannParis, line 140: All of this and it's attached JS and CSS should be on its own component.
Even if its only used by this dataset, it will make Dataset more compact.

Tom-Szendrey: Good point

YohannParis (commented):
Don't forget to rename the component by prefixing with `tera-` in the filename and within the HTML
- tera-dataset.vue
- tera-dataset-datatable.vue

Tom-Szendrey:
> 

Good catch, I totally missed that.

YohannParis (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
```diff
@@ -183,4 +232,87 @@ public Response uploadFile(
...
+			Collections.sort(numberList); 
+			double minValue = numberList.get(0);
+			double maxValue = numberList.get(numberList.size() - 1);
+			double meanValue = numberList.stream()
+				.mapToDouble(d -> d)
+				.average()
+				.orElse(0.0);
+
+			double medianValue = getMedianValue(numberList);
+			double sdValue = getSdValue(numberList, meanValue);
```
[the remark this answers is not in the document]
dvince2: I put a comment down below there - there are a few but I think that one of the more accepted ones is Guava by google.  That would give us median (and could be for a sorted or unsorted list) and standard deviation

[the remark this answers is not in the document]
Tom-Szendrey: Using Guava, thanks Derek!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java
Tom-Szendrey: Using Guava thanks Derek!

dvince2 (approved):

Issue #1019, opened 2023-04-24
[TASK]: Add Calibration Endpoint to HMI-Server


Tom-Szendrey:
add simulation service's /calls/calibrate to hmi-server
https://simulation-service.staging.terarium.ai/docs

PR #1022, opened 2023-04-24
chore: adding calibration end point to hmi-server

Tom-Szendrey:
# Description
Adding calibration endpoint to hmi-server.
Added a test button on ta2-playground just to run their test.

Resolves #(issue)
#1019


dvince2 (approved):
lgtm

PR #1025, opened 2023-04-24
feat: first pass at server side events

### LINKED ISSUE #435: [TASK] Investigate leveraging server-send-events
Investigate leveraging server-send-events and/or websocket technologies. The idea here is to be able to coordinate a user session across multiple windows/dialogs. This is relevant to our ideas of using "popups" ( #432 )

Potential references
- https://quarkus.io/guides/amqp

YohannParis:
# Description

* Connection with local rabbitMQ
* Create a UserEvent listener for SSE endpoint
* Added a message emitter on model-service proxy (because it will not brake staging)
* Added a test page on hmi-client http://localhost:8078/app/#/sse

Resolves #435


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/user/UserEvent.java
```diff
@@ -0,0 +1,22 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import lombok.Data;
+import software.uncharted.terarium.hmiserver.models.EventType;
+
+import java.util.UUID;
+
+@RegisterForReflection
+@Data
+public class UserEvent<S> {
```
mwdchang, line 12: What is `S` meant to do? It doesn't seem to be used or registered anywhere.

YohannParis: @chris-dickson why is this here?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/modelservice/ModelServiceProxy.java
```diff
@@ -16,6 +16,7 @@
 @Produces(MediaType.APPLICATION_JSON)
 public interface ModelServiceProxy {
 	@PUT
+	@Path("/api/models")
```
dvince2, line 19: whoops! good catch

dvince2 (approved):
This looks good, but should follow up with Daniels comment on the generic <S> for `UserEvent`

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/user/UserEvent.java
chris-dickson: Curious why you made a new class for this.  Why not use `Event` that is nearly a clone to this.

PR #1031, opened 2023-04-26
feat: canvas props config

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/temp/tera-infinite-canvas.vue
+++ b/packages/client/hmi-client/src/temp/tera-infinite-canvas.vue
@@ -6,21 +6,35 @@
 		<div class="data-layer" ref="dataLayerRef">
 			<slot name="data" />
 		</div>
-		<svg ref="backgroundLayerRef" :width="width" :height="height">
+		<svg class="background-layer" ref="backgroundLayerRef" :width="width" :height="height">
 			<slot name="background" />
 		</svg>
 	</main>
 </template>
 
 <script setup lang="ts">
-import { ref, onMounted } from 'vue';
+import { ref, onMounted, onUnmounted } from 'vue';
 import * as d3 from 'd3';
+
+const props = withDefaults(
+	defineProps<{
+		debugMode?: boolean;
+		scaleExtent?: [number, number];
+		lastTransform?: d3.ZoomTransform;
+	}>(),
+	{
+		debugMode: false,
+		scaleExtent: () => [0.1, 10]
+	}
+);
+
+const emit = defineEmits(['save-transform']);
 
 let x: d3.ScaleLinear<number, number, never>, y: d3.ScaleLinear<number, number, never>;
 let xAxis: d3.Axis<d3.NumberValue>, yAxis: d3.Axis<d3.NumberValue>;
 let gX: d3.Selection<SVGGElement, any, null, any>;
 let gY: d3.Selection<SVGGElement, any, null, any>;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -0,0 +1,162 @@
...
+
+const props = withDefaults(
+	defineProps<{
+		debugMode?: boolean;
+		scaleExtent?: [number, number];
+		lastTransform?: { k: number; x: number; y: number };
+	}>(),
+	{
+		debugMode: false,
+		scaleExtent: () => [0.1, 10],
```
mwdchang, line 27: We can just return `[0.1, 10]` directly not not wrap it in a function?

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -0,0 +1,162 @@
...
+	} else {
+		// Default position - triggers handleZoom which in turn sets currentTransform
+		svg.transition().call(zoom.transform as any, d3.zoomIdentity);
+	}
+
+	container.append('circle').attr('cx', 400).attr('cy', 300).attr('r', 20).attr('fill', 'red');
+});
+
+onUnmounted(() => {
+	emit('save-transform', currentTransform);
```
mwdchang, line 135: I am not sure what the intention is here - if you mean send update when zoom changes you should consider hooking it into the `end` event of d3.zoom.

For example in graph-scaffolder we have this code: 
```
zoom = d3
				.zoom()
				.scaleExtent(this.options.zoomRange)
				.on('zoom', zoomed)
				.on('end', zoomEnd);
```
and the zoomEnd function looks like
```
		const zoomEnd = () => {
			if (!this.graph || !chart) return;
			this.zoomTransformObject = d3.zoomTransform(chart.node() as Element);
			emit('zoom-end', {});
		};
```

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -0,0 +1,162 @@
...
+
+<style scoped>
+main {
+	width: 100%;
+	height: 100%;
+}
+
+.data-layer {
+	position: absolute;
+}
```
mwdchang, line 147: We need to fix a overlay problem here, we have 3 layers and only one of them is absolute, it means the other layers can shift around as we append items into their slotted content.

For example if I were to inject things into the foreground layer, everything will shift

image

In order for the layering to work, at least 2 of the layers need to be absolutely positioned .... or you can just do all 3 and call it a day.

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
shawnyama: I did that because I get this type issue if I don't. 
image

In the docs if it's an array they make it a function: https://vuejs.org/guide/typescript/composition-api.html#typing-component-props

There may be a way of setting up the props in the runtime way which is the very first example on that page. I just had trouble declaring proper typing like `[Number, Number]`

mwdchang: okay, let's not worry about this now.

mwdchang (approved):

PR #1032, opened 2023-04-26
docs: create a change log

YohannParis:
# Description

* First pass at creating a change log document to log our technologic choices.
* Mostly for Pantera, or for a good _"told you so!"_ when needed.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,0 +1,24 @@
+# Change log
+
+A log of _Terarium_ infrastructure to records all the changes made to the technology stack. 
+The changes may include updates or upgrades to programming languages, tools, libraries, or frameworks. 
+
+## Goal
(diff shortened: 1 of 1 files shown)


mwdchang:
Should we consider just making releases and generate release notes instead? 

Or rather, it feels like if we want to track the technological journey and discourse throughout ASKEM the thoughtss should be somewhere else ... like an askem-blog or something and not a apart of the source code.

YohannParis:
> Should we consider just making releases and generate release notes instead?
> 
> Or rather, it feels like if we want to track the technological journey and discourse throughout ASKEM the thoughtss should be somewhere else ... like an askem-blog or something and not a apart of the source code.

That's a fair point! 

@dgauldie what do you think about making this a `HISTORY.md` file in the root of the project, and then note the version numbers and why they are numbered that way.

PR #1035, opened 2023-04-27
feat: moving quarkus to reactive

YohannParis:
# Description

* Update `hmi-server` to use reactive development
* Tested, and the app works fine as is.


mwdchang:
Is it as simple as that?

It seems like aside from dependencies we also need to swap the annotations?
https://quarkus.io/guides/resteasy-reactive-migration

YohannParis:
> Is it as simple as that?
> 
> 
> 
> It seems like aside from dependencies we also need to swap the annotations?
> 
> https://quarkus.io/guides/resteasy-reactive-migration

Yes, there is more work to do. The plan is that this branch will be were all reactive development PR will merge into.

Once we are good we will be able to merge it back to main.

packages/services/hmi-server/build.gradle
```diff
@@ -68,7 +67,7 @@ dependencies {
 }
 
 group "software.uncharted"
-version "1.0.0-SNAPSHOT"
+version "2.0.0"
```
dvince2, line 70: Should this stay -SNAPSHOT since this is the development branch?

YohannParis: I don't really now, `main` branch is the actual version we are deploying.

dvince2 (approved):

YohannParis:
@dvince2 I was able to move to Quarkus 3.2.2, but it can't find `import javax.inject.Inject;` so I'm not able to build the project.

YohannParis:
This is now irrelevant as we move to Spring

PR #1036, opened 2023-04-27
feat: add workflow objects definitions, add sample unit workflow unit…

mwdchang:
### Summary
Add definitions for workflow and related objects.

Keeping things to types/interface for now as it makes it a little easier to experiment with. We can consider moving to structured classes after a few iterations of resolving quirky kinks.

### Testing
Run `yarn workspace hmi-client run test`


packages/client/hmi-client/src/workflow/workflow.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface WorkflowEdge {
+	id: string;
+	workflowId: string;
+	points: { x: number; y: number }[];
+
+	source: string;
+	sourcePort: number;
+
+	target: string;
+	targetPort: number;
```
YohannParis, line 56:
```suggestion
	source: WorkflowNode["id"];
	sourcePort: number; // Are we really using integer like Petri Net?

	target: WorkflowNode["id"];
	targetPort: number;
```

packages/client/hmi-client/src/workflow/workflow.ts
```diff
@@ -0,0 +1,72 @@
...
+	inputs: OperationData[];
+	outputs: OperationData[];
+}
+
+// Defines the data-exchange between WorkflowNode
+// In most cases the value here will be an assetId
+export interface WorkflowPort {
+	type: string;
+	value?: any;
+}
```
YohannParis, line 23: an `id` would be useful, just to name them.
```js
inputs: [{ id: 'model-a', type: 'model' }, { id:'model-b', type: 'model' }],
```

echl (approved):

packages/client/hmi-client/src/workflow/workflow.ts
```diff
@@ -0,0 +1,72 @@
...
+	y: number;
+	width: number;
+	height: number;
+
+	inputs: WorkflowPort[];
+	outputs: WorkflowPort[];
+
+	// FIXME: The section below is slated to be further spec'ed out later.
+	// State and progress, tracking of intermediate results
+	statusCode?: string; // not-run, completed, in-progress, error, invalid
```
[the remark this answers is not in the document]
echl: I hear daniel hates enums

packages/client/hmi-client/src/workflow/workflow.ts
```diff
@@ -0,0 +1,72 @@
...
+	type: string;
+	value?: any;
+}
+
+// Node definition in the workflow
+// This is the graphical operation of the operation defined in operationType
+export interface WorkflowNode {
+	id: string;
+	workflowId: string;
+	operationType: string;
```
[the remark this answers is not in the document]
mwdchang: I don't think an enum works - enums are good for a fixed set but annoying to extend.

if you want it to be an extendable system an Enum does not work well. Imagine a 3rd party wants to write a model-optimization operation. We won't be able to accommodate that easily unless we already have the token as part of the enumeration already.

packages/client/hmi-client/src/workflow/workflow.ts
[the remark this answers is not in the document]
mwdchang: premature engineering/optimization makes one lose focus on what is important? 😄

packages/client/hmi-client/src/workflow/workflow.ts
mwdchang: I will add it as an optional field now now, I thought about this before, but backed off ... I didn't think an id-scheme would be manageable given that the number of ports are dynamic at runtime, but will have to rethink it.

packages/client/hmi-client/src/workflow/workflow.ts
mwdchang: Nope. This is just direct numbering. Petrinet ascet's number system is indirect. e.g. to get value of index-2 in ascet is like `X[Y[2]]`

YohannParis:
You could at least explain why you do not agree with the comments!

packages/client/hmi-client/src/workflow/workflow.ts
YohannParis: This is going to be so annoying

PR #1039, opened 2023-04-28
chore: update workflow related definitions per latest dicussion points

mwdchang:
### Summary
Update definitions after discussion on generic vs specialized run cases. Mostly it introduces an `isRunnable` flag to allow flagging operations for custom executions. #1041 

Not sure if adding ids to ports made things better, the constructions are a bit more complicated so we will see.

### Testing
`yarn workspace hmi-client run test`

### DIFF
--- a/packages/client/hmi-client/src/types/workflow.ts
+++ b/packages/client/hmi-client/src/types/workflow.ts
@@ -1,3 +1,11 @@
+export enum WorkflowStatus {
+	INVALID = 'invalid',
+	FAILED = 'failed',
+	COMPLETED = 'completed',
+	IN_PROGRESS = 'in progress',
+	ERROR = 'error'
+}
+
 // Defines the type of data an operation can consume and output
 export interface OperationData {
 	type: string;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/types/workflow.ts
```diff
@@ -51,10 +63,10 @@ export interface WorkflowEdge {
 	points: { x: number; y: number }[];
 
 	source: WorkflowNode['id'];
-	sourcePort: number;
+	sourcePortId: string;
```
Tom-Szendrey, line 66: Ports will now be alias'd by a string so i can more easily find my inputs
am i understanding this correctly?

Tom-Szendrey: I had to reread a bit my bad, ignore above

packages/client/hmi-client/src/types/workflow.ts
```diff
@@ -1,3 +1,11 @@
+export enum WorkflowStatus {
```
Tom-Szendrey, line 1: i think we need to add a status for a node that has not been triggered to run?

mwdchang: I was thinking along the same lines, but I don't want to have a plethora of different status code where some of them are kind of similar.....I think for now a node that has not been run can use `invalid` and we can revisit adding more enum-tokens as needed.

PR #1041, opened 2023-04-28
fix: fixed logging resource

tangopapatime:
# Description

Updated the Logging endpoint in Quarkus

Resolves #(issue)

### DIFF
--- a/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
+++ b/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
@@ -16,7 +16,7 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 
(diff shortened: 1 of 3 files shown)


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
```diff
@@ -14,22 +17,24 @@
 import javax.ws.rs.Produces;
 
 @Path("/api/")
```
dvince2, line 19: I imagine that this is testing and works, but I noticed that the other classes omit the last / and just have `@Path("/api")`

dvince2 (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/loggingservice/LoggingResource.java
tangopapatime: yeah definitely don't need that :P

PR #1043, opened 2023-04-28
feat: draw nodes in workflow canvas

echl:
Right click to place a generic test node on the canvas. 

image


packages/client/hmi-client/src/temp/tera-simulation-workflow.vue
```diff
@@ -1,22 +1,86 @@
...
+		:last-transform="canvasTransform"
+		@save-transform="saveTransform"
+	>
 		<template #foreground></template>
 		<template #data>
-			<div style="font-size: 24px; padding: 10px; background: #9ef">This is a DataLayer DIV</div>
+			<ContextMenu ref="contextMenu" :model="contextMenuItems" />
+
+			<ul v-for="node in nodes">
+				<tera-workflow-node :node="node" :initialTransform="canvasTransform"></tera-workflow-node>
```
mwdchang, line 13: Is this related to what we were discussing yesterday about needing transforms x/y/k? 
I don't see initialTransform getting used or declared within tera-workflow-node.

packages/client/hmi-client/src/temp/tera-simulation-workflow.vue
```diff
@@ -1,22 +1,86 @@
 <template>
-	<infinite-canvas debug-mode :last-transform="lastTransform" @save-transform="saveTransform">
+	<infinite-canvas
+		debug-mode
+		@contextmenu="toggleContextMenu"
+		:last-transform="canvasTransform"
+		@save-transform="saveTransform"
```
mwdchang, line 6: There is a bit of a self-referential logic here that we should try to untangle.
- canvasTransform is sent as a prop into the component
- saveTransform modifies canvasTransform 
- canvasTransform is sent as a prop into the component ... etc etc

I think we should just take out `:last-transform` for now to avoid some inadvertent watcher going to infinite loops

packages/client/hmi-client/src/temp/tera-simulation-workflow.vue
```diff
@@ -1,22 +1,86 @@
...
+import { ref } from 'vue';
 import InfiniteCanvas from '@/components/widgets/tera-infinite-canvas.vue';
+import { Operation, WorkflowNode, WorkflowStatus } from '@/types/workflow';
+import TeraWorkflowNode from './tera-workflow-node.vue';
+import ContextMenu from 'primevue/contextmenu';
+
+const nodes = ref<WorkflowNode[]>([]);
+const contextMenu = ref();
+const newNodePosition = ref<{ x: number; y: number }>({ x: 0, y: 0 });
+const canvasTransform = ref<d3.ZoomTransform>();
```
shawnyama, line 29: Make this of type `{ k: number; x: number; y: number }` (same as `lastTransform` prop in the infinite canvas). Also in `saveTransform()` see about extracting the x, y, z from the `d3.ZoomTransform` type and assigning it to the `canvasTransform`.  Or better yet in `handleZoomEnd` pass `{ e.transform.x, e.transform.y, e.transform.k }` to `save-transform`. That way we just have to deal with d3 types in the infinite canvas.

mwdchang, line 29: Is wrapping `d3.ZoomTransform` inside a ref necessary here? 

We have had a lot of problems in making 3rd party data structures reactive in the past, some times they become massive memory-hogs, and sometimes they cause call-stack explosions. It seems like we are just treating it as a normal variable so I think we can take out the `ref` part.

shawnyama (commented):
I think it would make sense to move these out of the temp folder and place them in `components/workflow/`. @mwdchang ?

packages/client/hmi-client/src/temp/tera-simulation-workflow.vue
echl: Good point!

mwdchang:
Looks great!, thanks!

mwdchang (approved):

PR #1045, opened 2023-04-28
feat: add button to project

Tom-Szendrey:
# Description
Adding 1 canvas page from temp to source so we can work on the same wavelength a bit easier
Adding 1 button to go here very easily

https://user-images.githubusercontent.com/17088680/235216604-8522b957-20de-45fc-a0e3-ed238afb8e99.mov

Resolves #(issue)


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -241,6 +246,7 @@ import CodeEditor from '@/page/project/components/code-editor.vue';
 import SimulationPlan from '@/page/project/components/Simulation.vue';
 import TeraResourceSidebar from '@/page/project/components/tera-resource-sidebar.vue';
 import TeraProjectOverview from '@/page/project/components/tera-project-overview.vue';
+import teraSimulationWorkflow from '@/temp/tera-simulation-workflow.vue';
```
shawnyama, line 249: import the one from the workflow folder

Tom-Szendrey: thanks

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -267,7 +273,7 @@ const props = defineProps<{
 	project: IProject;
 	assetName?: string;
 	assetId?: string;
-	assetType?: ProjectAssetTypes | 'overview' | '';
+	assetType?: ProjectAssetTypes | 'overview' | 'workflow' | '';
```
shawnyama, line 276: to fix the typecheck issue add 'workflow' to the `Tab` type in 'common.ts`

Tom-Szendrey: yepp just saw that now, why is this defined in individually in the project page as well as the common.ts?

shawnyama: Yeah it's kinda weird. Ideally it should just be `ProjectAssetTypes` so its just one type that can be easily shared but then more tab types came around that weren't necessarily assets. Perhaps a TabType could be made to hold ProjectAssetTypes + other tabs like overview or ''. Doubt it's important to worry about now though

shawnyama (approved):

PR #1049, opened 2023-05-01
feat: v0 model operation node

shawnyama:
# Description

Model selection + config node

https://user-images.githubusercontent.com/28237258/235509225-adb762df-4d72-4dc2-8204-539d0fb96424.mov


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -646,7 +647,7 @@ section {
 
 :deep(span.p-dropdown-trigger-icon.pi.pi-chevron-down) {
 	color: var(--text-color-subdued);
-}
+} */
```
mwdchang: What and why are we commenting out a section here? This change seems unrelated?

It would be good to keep PRs as single purpose as possible - while it is convenient to fix things as you come across it does make tracking things extra hard, especially since we are squashing commits it makes reverting commits and cherry-picking commits very difficult.

shawnyama: I deleted those styles later, it was making the dropdown in the model-node look bad

mwdchang (approved):

PR #1055, opened 2023-05-02
feat:  data layer and background layer for interactive simulation workflow: nodes, links, interactions

### LINKED ISSUE #1054: Data layer and background layer for interactive simulation workflow: nodes, links, interactions

echl:
Resolves #1054


mwdchang:
We should make plans to remove the notion of graphs/edges from `tera-infinite-canvas`, as they are two separate ideas. You can have a workflow-graph in an infinite-canvas, but having an infinite-canvas does not necessitate having a graph.

We had some other discussion threads about using an infinite-canavas like environment for other things like document-organization and knowledge synthesis. So let's decouple these two ideas while it is still early .... maybe for the next PR.

packages/client/hmi-client/src/types/workflow.ts
```diff
@@ -60,13 +60,13 @@ export interface WorkflowNode {
 export interface WorkflowEdge {
 	id: string;
 	workflowId: string;
-	points: { x: number; y: number }[];
+	points: Position[];
```
shawnyama, line 63: Maybe to constrain it to two points you can make it `[Position, Position]`

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -142,6 +157,26 @@ onMounted(() => {
...
+		const sourcePoint = edge.points[0];
+		const targetPoint = edge.points[1];
+		const path = d3.path();
+		path.moveTo(sourcePoint.x, sourcePoint.y);
+		path.lineTo(targetPoint.x, targetPoint.y);
+		path.closePath();
+		return path.toString();
+	}
+	return 'M0,0';
+}
```
mwdchang, line 179: There are better ways to do this since you are already hooked into D3's system. D3 has generator functions.

```
// The generator function, leave out the curve part if you don't want splines
const pathFn = d3
	.line<{ x: number; y: number }>()
	.x((d) => d.x)
	.y((d) => d.y)
	.curve(d3.curveBasis);
```

then

```
:d="pathFn(edge.points)"
```

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -45,18 +61,23 @@ const nodes = ref<WorkflowNode[]>([]);
 const contextMenu = ref();
 const newNodePosition = ref<{ x: number; y: number }>({ x: 0, y: 0 });
 let canvasTransform = { x: 0, y: 0, k: 1 };
+const mousePosition = ref<Position>({ x: 0, y: 0 });
+const isCreatingNewEdge = ref<boolean>(false);
```
mwdchang, line 65: It doesn't look like mousePosition nor isCreatingNewEdge need to be `ref`.

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -111,8 +132,78 @@ function saveTransform(newTransform: { k: number; x: number; y: number }) {
...
+function onPortMouseover(position: Position) {
+	currentPortPosition = position;
+}
+
+function mouseUpdate(event) {
+	if (newEdge.value && newEdge.value.points && newEdge.value.points.length === 2) {
+		mousePosition.value = {
+			x: (event.pageX - canvasTransform.x) / canvasTransform.k,
+			y: (event.pageY - 57 - canvasTransform.y) / canvasTransform.k
+		};
```
mwdchang, line 179: If we need to use a magic number `57` here ... we are probably dong something weird.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -76,6 +89,21 @@ onMounted(() => {
 	workflowNode.value.addEventListener('mouseup', stopDrag);
 });
 
+function selectPort(port: WorkflowPort) {
```
shawnyama, line 92: I can connect multiple outputs with the same input - perhaps you can check if the input is available?

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -4,31 +4,44 @@
...
 		<slot name="body" />
 		<section class="outputs">
-			<li v-for="(output, index) in node.outputs" :key="index">
+			<li v-for="(output, index) in node.outputs" :key="index" ref="outputs">
 				{{ output.label }}
-				<div class="port"></div>
+				<div
+					class="port"
+					@click.stop="selectPort(output)"
+					@mouseover="mouseoverPort(index, false)"
```
mwdchang, line 24: This feels a bit weird, you wouldn't care about the position of the port until you click on it right? 

So why bother with the mouseover-event? Isn't it just firing a lot of emits we don't necessarily need? 
Moreover, even if we need it, why wouldn't we just do `mouseoverPort(output)` instead of passing in the index and true|false?

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -111,8 +132,78 @@ function saveTransform(newTransform: { k: number; x: number; y: number }) {
...
+		} else {
+			newEdge.value.points[1] = mousePosition.value;
+		}
+	}
+}
+
+onMounted(() => window.addEventListener('mousemove', mouseUpdate));
+onUnmounted(() => window.removeEventListener('mousemove', mouseUpdate));
+
+function updateEdgePositions(node: WorkflowNode, { x, y }) {
```
shawnyama, line 191: May be cleaner to combine `updateEdgePositions` and `updatePosition` to something like `updateObjectPositions`

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -111,8 +132,78 @@ function saveTransform(newTransform: { k: number; x: number; y: number }) {
...
+	edges.value.forEach((edge) => {
+		if (edge.source === node.id) {
+			edge.points[0].x += x / canvasTransform.k;
+			edge.points[0].y += y / canvasTransform.k;
+		}
+		if (edge.target === node.id) {
+			edge.points[1].x += x / canvasTransform.k;
+			edge.points[1].y += y / canvasTransform.k;
+		}
+	});
```
mwdchang, line 201: Nitpick: You might want to change the first and last elements in `points`, just in case we have curved edges with more than 2-elements and someone doesn't get totally surprised.

Can probably move `x / canvasTransform.k;` and `y / canvasTransform.k` outside of the loop to be a tad more efficient.

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -106,7 +119,7 @@ const resizeObserver = new ResizeObserver(() => updateDimensions());
 
 onMounted(() => {
 	const svg = d3.select(backgroundLayerRef.value as SVGGElement); // Parent SVG
-	const container = svg.append('g'); // Pan/zoom area
+	const edges = d3.select(edgesRef.value as SVGGElement); // Pan/zoom area
```
mwdchang, line 122: There are a few things here not quite right - namely the slot system is being bypassed in a strange manner. Let's set up a bit of time to clear up our understandings, maybe we can setup a pair-programming session kind of thing.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
echl: Going to leave it for v1, yeah.

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -111,8 +132,78 @@ function saveTransform(newTransform: { k: number; x: number; y: number }) {
...
+	if (newEdge.value && newEdge.value.points && newEdge.value.points.length === 2) {
+		mousePosition.value = {
+			x: (event.pageX - canvasTransform.x) / canvasTransform.k,
+			y: (event.pageY - 57 - canvasTransform.y) / canvasTransform.k
+		};
+		if (event.target.className === 'port') {
+			newEdge.value.points[1] = currentPortPosition;
+		} else {
+			newEdge.value.points[1] = mousePosition.value;
+		}
```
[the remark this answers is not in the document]
echl: How can I replicate this situation?

mwdchang (approved):

Issue #369, opened 2022-12-13
[TASK]: Fix count of xDD articles displayed in Explorer and refine pagination


amostafa-uncharted:
**Describe the task**
xDD supports some form of Pagination for fetching large amounts of results. Early data explorer implementation provided a toggle that controls whether xDD pagination should be enabled or not through the query param "full_results". However, it was noted that when xDD pagination is enabled the search results are not sorted by relevancy and thus the feature is currently disabled in the data explorer. Ideally, we should reach out to the xDD team to clarify the pagination feature while still fetching ranked results. Nonetheless, with the pagination disabled only a max number of articles is set (currently to 100) and no pagination is provided. This can be confusing since the user can see that a maximum 100 article exist but the facets aggregations reflect much larger size of results. The fix should refactor the label that displays the results count to say something like: "showing 100 articles out of total x results" wherein x is the number of hits provided by xDD. This refactoring work may result in completely removing the unused simple-pagination component.

**Task Items**
- [x] refactor the displayed count of search results
- [ ] communicate with xDD team re proper pagination or to retrieve the true number of hits even if the results only returns a fixed number of matching papers
- [ ] remove the unused simple-pagination component

**Goal**
A clear and concise description of what the end result must have for it to be approved as complete.

**Screenshots**
If applicable, add screenshots or designs

**Additional context**
Add any other context about the task here

dvince2:
From Ian: I also added a `hits` count on the articles route

dvince2:
For now, we've added the "X of Y documents" option. xDD has been made aware of the sorting issue and are working to resolve it.

PR #1057, opened 2023-05-02
feat: align front/back end and fix facets

### LINKED ISSUE #373: [TASK]: Refactor and Move Facets Implementation to HMI server
**Describe the task**
Facets for models and datasets are currently constructed on the client side. Also, applying facet filters is done at the client side. This is very inefficient because ideally applying a facet filter should result in an API call that recalculates both the data and the facets aggregation in a way similar to what is currently done with XDD articles. 

For models and datasets, concept facets are served through a separate endpoint that makes the integration at the client side harder because fetching models, for example, is not aware of any concept filters applied. The currently implemented workaround utilizes concepts-facet filtering by filtering the list of models, identifying related curies and pass them as query parameter filters to the concepts facet API, which is less optimal. This is due to the fact that the endpoint for fetching concept facets does NOT receive filter parameters. So it is not possible to tell the concepts facets API to only provide concepts for models matching a given search term or criteria.

On a similar note, the client side filtering of models/datasets against search terms (by manaully searching the name and description fields) as well as against concept names. The endpoints for fetching models and datasets need to take parameters such as search-terms and concepts in order to facilitate such filtering and all such logic should be moved into a dedicated service module within the HMI server that hides all the search details and provide a thin interface for the client consumption

NOTE this require changes to the data service and alignment on how concept facets shall be provided.

**Task Items**
- [ ] Move the computation of facets data/aggregation for models and datasets to the HMI server (see for example the comment that says "simulate applying filters to the model query" at data.ts)
- [ ] Move logic that filter models and datasets via search criteria to the HMI server, e..g, applyFacetFiltersToModels
- [ ] Consolidate concept facets as part of the models/datasets search results response similar to how xDD provides it

dvince2:
Currently, the front end for hmi-server is creating its own facets for document searches.  This was not overly performant, but more importantly, meant that any changes we wanted to make to facets to use elastic search was going to be an absolute nightmare to accomplish as we'd have to first move everything to the back end, then work on search. Part of this too is that the front end and back end had slightly different representations for this data.
xDD is providing us facet data which we were ignoring in favour of our own facet creation.  This was being done because xdd facets are over all data sets yet our facets were only on the 100 returned that we cared about. Facets are now again over all the data and users are shown how many they're seeing on theirs screen at that time as a temporary fix until prioritization is working correctly by xDD.

# Description

- Front and back end now use common objects to represent documents and related responses (Extractions, facets etc)
- Removing front end implementation of Document, etc.
- Updating xDD api to be noted as version 1, as they're updating their endpoint/api currently for version 2.
- Adding dummy endpoints for ES searching which is not being used by the front end at all
- Adding secret for ES search api key
- Fixing various reference/property bugs in front end

# Screen shots
In this example here, there are our capped 100 results plus 40 extractions.  Those 40 extractions currently mess with the numbers a bit, but, that is going to be addressed soon and we can show just the proper 100 here.
image

The facet max number now matches with the "of" value, which makes a lot of sense when you look at it - that first hit for example, 1769 of the 12489 come from the Journal "cold spring harbour laboratory press"
[13:51](https://unchartedsoftware.slack.com/archives/D04E7HJQKKK/p1683049875774889)
clicking on that facet, the "of" value changes to reflect the smaller number of hits since its only documents from this journal
image
image

In addition, all of the other facets now appear filtered as well and show this 1769 number
image

If you click on nextstrain/ncov now, you'll end up seeing just the results you'd expect
image

(again, its showing 46 here not 6 because of the 40 extractions that are just being tacked on to the top)
The 6 hits that come from Cold Spring Harbour Lab and have a github link for nextstrain/ncov
image

Resolves #373 (partial), #369


dvince2:
Currently an issue with Datasets facets (fixing documents broke that...). Looking to resolve that now before flipping this to a non-draft.

Tom-Szendrey:
Im not understanding your comments above about the "40 tacked on top"
Is this expected?
1 facet match but 48 results:

image

Tom-Szendrey:
I know this is just the first step but I wanted to flag that the facets for models and datasets are weird. 
All of them for both are showing 0/214
I just wanted to make sure you were aware of this

image

mwdchang:
I scanned the Java parts, looks fine there 👍

dvince2:
@Tom-Szendrey for your first point - Presently when an xdd search is being done we do both an extractions search and then a documents search and combine the two in the results panel.  What we _should_ be doing is just doing a single document search then using the extractions metadata on the documents to enrich our results with extractions... however xDD isn't _quite_ there yet.  As a result we'll always have some number of extractions tacked on to our document search... for a popular/basic search like "covid" this is usually about 30-40 results.

On the first unfaceted search you'll see something like "132 of 123456 results" - that will mean that you'll always see n+32 returned... so in the case highlighted here there are 8 documents returned (correct) PLUS the 40 extractions we just always tack on the top.

Tom-Szendrey (approved):
Ill bet $5 on 0 attribute and type issues are found after we all pull to main

dvince2:
After every merge of main into this branch I've squashed a few... I'm confident that right after this merge we'll be ok but after the next pr goes in....

PR #1058, opened 2023-05-02
feat: dragging cavas objects

mwdchang:
### Summary
Add drag-related event listeners for the workflow nodes, you will be able to drag the nodes around the canvas.

Note this is sort of a stand-in replacement for using the vue3-draggable-resizable library, as it doesn't account for the transforms on the parent-container (dragging with zoom <> 1.0 is weird). We may switch out back to the lib later on, or to a different lib.

### Testing
Create nodes on the workflow, you can drag nodes around.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
@@ -21,7 +21,7 @@
 
 <script setup lang="ts">
 import { WorkflowNode } from '@/types/workflow';
-import { ref, onMounted, onBeforeUnmount } from 'vue';
+import { ref, computed, onMounted, onBeforeUnmount } from 'vue';
 
 const emit = defineEmits(['dragging']);
 
@@ -29,12 +29,12 @@ const props = defineProps<{
 	node: WorkflowNode;
 }>();
 
-const nodeStyle = ref({
+const nodeStyle = computed(() => ({
 	minWidth: `${props.node.width}px`,
 	minHeight: `${props.node.height}px`,
 	top: `${props.node.y}px`,
 	left: `${props.node.x}px`
-});
+}));
 
 const workflowNode = ref<HTMLElement>();
 
@@ -55,7 +55,6 @@ const drag = (evt: MouseEvent) => {
 	const dx = evt.x - tempX;
 	const dy = evt.y - tempY;
 
-	console.log('move', dx, dy);
 	emit('dragging', { x: dx, y: dy });
 
 	tempX = evt.x;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -21,20 +21,68 @@
...
+	tempX = 0;
+	tempY = 0;
+	dragStart = false;
+};
+
+onMounted(() => {
+	if (!workflowNode.value) return;
+
+	workflowNode.value.addEventListener('mousedown', startDrag);
+	document.addEventListener('mousemove', drag);
```
mwdchang, line 75: Not too happy about hooking into document listener for every workflow-node, but it should work and we are not expected to have hundreds of nodes anyway.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -21,20 +21,68 @@
...
+}));
+
+const workflowNode = ref<HTMLElement>();
+
+let tempX = 0;
+let tempY = 0;
+let dragStart = false;
+
+const startDrag = (evt: MouseEvent) => {
+	console.log('start', evt.x, evt.y);
```
echl, line 46: Can we remove these logs

echl (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
shawnyama: Hmm I am fiddling with `{defineExpose}` if you want to go that route.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
mwdchang: I would like to keep the console.log around for a bit longer for debugging, will clean up with we have a full version-0 or version-1.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
mwdchang: @shawnyama I am not sure what you are trying to say - can you elaborate or have a small code-snippet that we can use to base the conversation?

mwdchang: Hmm I see, I have not played with defineExpose before so I am not sure how to work with it or the caveats. Might also able to do this as a currying function going parent=>child instead of child=>parent

Anyway I am going to merge to PR and we can have this as a side discussion.

PR #1059, opened 2023-05-02
feat: update dataset layout

shawnyama:
# Description

Added table of contents and data view to `tera-dataset`. In the data/pdf view I made a css rule that stretches the content to fill space that would've been taken up by the table contents (matches the design).

https://user-images.githubusercontent.com/28237258/235774949-4aabb7d0-7f82-4e35-9b6d-787c6368fb70.mov


packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -165,6 +220,7 @@ watch(
 					}
 				});
 				dataset.value = datasetTemp;
+				console.log(datasetTemp);
```
mwdchang, line 223: This is the entire dataset? Is it really necessary to print it out?

mwdchang:
Looks good, I didn't perform a super thorough check as some things might change in light of trying to adopt the notion of data-sheets and model-cards.

mwdchang (approved):

PR #1062, opened 2023-05-03
feat: create new output port when model node is run

shawnyama:
# Description

Creates a new output port when model node is run. The node itself and its model operation is updated to support a new output.

https://user-images.githubusercontent.com/28237258/235957704-c7dd1bc5-157b-4467-aed4-104f37ee34d4.mov


packages/client/hmi-client/src/components/workflow/tera-model-node.vue
```diff
@@ -48,15 +53,18 @@ const model = ref<Model | null>();
...
+	if (thisModelOperation.action) {
 		console.log(
-			ModelOperation.action({
+			thisModelOperation.action({
 				model: model.value,
 				initialValues: initialValues.value,
 				parameterValues: parameterValues.value
 			})
 		);
+		emit('append-output-port', props.node.id, thisModelOperation.outputs[0].type);
```
echl, line 67: I think we should refactor our node components, I don't think `tera-model-node` should have to emit an event to add a port, nor should the node object be passed as a prop into *both* `tera-workflow-node` and `tera-model-node`

packages/client/hmi-client/src/components/workflow/model-operation.ts
```diff
@@ -15,5 +15,8 @@ export const ModelOperation: Operation = {
 		model: Model;
 		intialValues: StringValueMap;
 		parameterValues: StringValueMap;
-	}) => [{ type: 'modelConfig', value: modelConfig }]
+	}) => {
+		modelOperation.outputs.push({ type: 'modelConfig' });
+		return [{ type: 'modelConfig', value: modelConfig }];
+	}
```
mwdchang, line 21: There is a misunderstanding here. The ModelOperations' output does not change, so the cloneDeep you have elsewhere is not necessary.

It is kind of like saying: Tim Hortons outputs coffee, but how many pick up windows of coffee is a property of each location, and not that of the definition of the franchise.

packages/client/hmi-client/src/components/workflow/tera-model-node.vue
mwdchang: Two separate observations.

@echl: 
Because we generally treat props as a one-way data-bind, we don't typically want to directly change `props.workflowNode.xyz`. So what we have here follows the data-down-action-up paradigm ... which is fine in my opinion. In leu of the emit, we can consider doing a v-model paradigm that is closer to a two-way-data-bind. Do you have examples you want to share in terms of how you want to do things or are we okay with this for now?


Let's talk about the tera-model-node  vs era-workflow-node issue offline, it is largely not relevant to this PR.

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -88,6 +90,13 @@ const testOperation: Operation = {
 	isRunnable: true
 };
 
+function appendOutputPort(index: number, type: string) {
+	nodes.value[index].outputs.push({
+		id: nodes.value[index].outputs.length.toString(),
+		type
+	});
+}
+
```
mwdchang, line 99: Still not quite right. 

1. "node.id" is not the same as integer-index, node.id is actually defined as a string
2. Here we would actually want the value produced (model-config) in the value field 

So in short, your emit should pass back and store at least the model-config output you get from "run".

```
export interface WorkflowPort {
	id: string;
	type: string;
	label?: string;
	value?: any;
}```

**Review by mwdchang — CHANGES_REQUESTED** [review]

File: packages/client/hmi-client/src/components/workflow/tera-model-node.vue (hunk 1 above)
Comment (echl) [reply packages/client/hmi-client/src/components/workflow/tera-model-node.vue]: I'm okay with it for now. Something about it seems unintuitive to me but I'll explore it further.

File: packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue [hunk 4]
```diff
@@ -88,6 +90,18 @@ const testOperation: Operation = {
...
+	// Find node and assign outport data to its output port
+	const node = nodes.value[nodes.value.findIndex(({ id }) => id === nodeId)];
+	node.outputs[node.outputs.length - 1] = outputPortData;
+
+	// Create new output port
+	node.outputs.push({
+		id: node.outputs.length.toString(),
+		type: outputPortData.type
+	});
+}
```
mwdchang, line 103: I am not sure if this is the pattern we want - that there is always an "empty" output-port that cannot be used.

I think it is good for now but we should have a bit of discussion around it. 
CC @echl @mwdchang @Tom-Szendrey

mwdchang (approved):

PR #1076, opened 2023-05-05
chore: a round of design-tweaks, add edge arrow marker

mwdchang:
### Summary 
A around of tweaks:
- Adding edge arrow marker, make edges more distinct
- Make tera-workflow-node a bit closer to the design spec

image

### Testing
No functional changes, workflow canvas should work as before.

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
@@ -6,6 +6,9 @@
 			:width="width"
 			:height="height"
 		>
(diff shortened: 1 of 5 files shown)


echl (approved):
I think arrows might be a bit redundant since we should know the direction of the data flow implicitly since they always go from input to output?

mwdchang:
> I think arrows might be a bit redundant since we should know the direction of the data flow implicitly since they always go from input to output?

The arrows/edges go from output-to-input 😄 

Arrows help in situation like this where your node are not necessarily ordered left-to-right
image

PR #1077, opened 2023-05-08
feat: split view for workflow

shawnyama:
# Description

What I added:
- A panel that will be used to view an asset/process from a workflow node - this has been added into `tera-project.vue`
- I have also made a component called `tera-project-page.vue` to house the possible asset/process components that are viewed in a panel in the project view

How it works/how you can add to this:
- `tera-workflow-node.vue` is where I pass the node to be saved in pinia, I use`openedWorkflowNodeStore` (if you have a better name let me know)
- In the store I save the node as well as its `assetId` and `pageType` for easy access
- I get that node in `tera-project.vue` and pass its `assetId` attribute into `tera-project-page.vue` to view a model asset
- So far, whatever the last output is of your model node will be displayed in the drilldown panel
- If you need to add more attributes from the node or even the node itself feel free to pass it into `tera-project-page.vue` as new prop(s) according to your needs
- Also add your components for new processes such as calibrate into `tera-project-page.vue`

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/4e5b033e-bcb1-47b5-9619-ba4b618a82d8


packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
shawnyama, line 1: Why I got rid of these: 

This 100%s were shifting the whole project component up when opening the new panel. They are redundant anyways since the width and height of the svg is assigned in the HTML

packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -164,8 +164,6 @@ main > * {
 
 .background-layer {
 	cursor: grab;
-	width: 100%;
-	height: 100%;
```
shawnyama, line 168: Why I got rid of these:

The 100%s were shifting the whole project component up when opening the new panel. They are redundant anyways since the width and height of the svg is assigned in the HTML

YohannParis:
You need to add tabs in that panel, exactly like in tera-project, kind of a tera-project within a tera-project (inception much).
Tell me if that makes sense?

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -276,21 +229,29 @@ const props = defineProps<{
...
 
 const tabStore = useTabStore();
-const router = useRouter();
-const resources = useResourcesStore();
 
-const newModelId = ref<string>('');
-const isNewModel = ref<boolean>(true);
+// Makes asset from workflow node appear
+const openedWorkflowNodeStore = useOpenedWorkflowNodeStore();
+const workflowNodeAssetId = computed(() => {
```
echl, line 235: Can you add some comments to explain what is happening here? I find it a bit hard to follow.

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -276,21 +229,29 @@ const props = defineProps<{
...
+	let index: number = 0;
+
+	if (openedWorkflowNodeStore?.workflowNode?.outputs.length) {
+		index = openedWorkflowNodeStore.workflowNode.outputs.length - 2;
```
echl, line 239: Why `-2`? If length is 1, won't index be -1?

packages/client/hmi-client/src/page/project/components/tera-tab-content.vue
```diff
@@ -0,0 +1,167 @@
...
+			:project="project"
+			@asset-loaded="emit('asset-loaded')"
+		/>
+	</template>
+	<code-editor
+		v-else-if="assetType === ProjectAssetTypes.CODE"
+		:initial-code="code"
+		@on-model-created="openNewModelFromCode"
+	/>
+	<tera-model
```
echl, line 45: Why are there two `tera-models`? Is one for an existing model and one for a new one?

shawnyama: Yeah the one on the outside is for when you're creating one.

packages/client/hmi-client/src/page/project/tera-project.vue
shawnyama: Once the model is ran a new output port is added. That new port is empty though so I am just grabbing the one before that. I'll write this better though as I haven't accounted for one empty output port.

packages/client/hmi-client/src/page/project/components/tera-tab-content.vue
echl: Just wondering if there is any value to consolidating them into the same component. Intuitively I would think `tera-model` would display either a new model or saved model instead needing two instances of `tera-model`.

echl:
Is there a way currently to close the split view?

shawnyama:
> Is there a way currently to close the split view?

No, I thought once tabs are implemented you'd close the last tab and that would close the view. At the moment you can just drag it out of sight

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -276,21 +229,31 @@ const props = defineProps<{
...
+
+// Grabs the model id from the last model config output from the node
+const workflowNodeAssetId = computed(() => {
+	let index: number = 0;
+
+	if (openedWorkflowNodeStore?.workflowNode?.outputs.length) {
+		index = openedWorkflowNodeStore.workflowNode.outputs.length - 1; // Grab index of last output port
+	}
+	return (
+		openedWorkflowNodeStore?.workflowNode?.outputs[index].value.model.id.toString() ?? undefined
```
echl, line 244: Thinking about how to make this more generic to support the other drilldowns... maybe `openedWorkflowNodeStore` should just store a `assetType` and `assetId`?

packages/client/hmi-client/src/page/project/components/tera-tab-content.vue
```diff
@@ -0,0 +1,161 @@
...
+import CodeEditor from '@/page/project/components/code-editor.vue';
+import SimulationPlan from '@/page/project/components/Simulation.vue';
+import SimulationRun from '@/temp/SimulationResult3.vue';
+import TeraProjectOverview from '@/page/project/components/tera-project-overview.vue';
+import TeraSimulationWorkflow from '@/components/workflow/tera-simulation-workflow.vue';
+
+const props = defineProps<{
+	project: IProject;
+	assetId?: string;
+	assetType?: ProjectAssetTypes | 'overview' | 'workflow' | '';
```
echl, line 85: We might want to just add `overview` and `workflow`  and `` to `ProjectAssetTypes`

shawnyama: Workflow makes sense. But overview isn't an asset and if we were to add a process like calibrate it wouldn't make sense either. What do people think of having an `AssetType`, a `ProcessType` and a `ProjectPageType` that would hold the `AssetType` + `ProcessType` and any other project page like 'overview'.

echl (approved):

PR #1078, opened 2023-05-09
chore: workflow v1 updates

mwdchang:
### Summary
- Adding and start to leverage port-status code to indicate connected/not-connected information
- Start to leverage node status code (invalid, running, in-progress....) 
- Start with 0-output port instead of 1 invalid output port


Tom-Szendrey (approved):
nice ports getting their own status should add some good flexibility

packages/client/hmi-client/src/types/workflow.ts
```diff
@@ -6,6 +6,11 @@ export enum WorkflowStatus {
 	ERROR = 'error'
 }
 
+export enum WorkflowPortStatus {
+	CONNECTED = 'connected',
+	NOT_CONNECTED = 'not connected'
+}
```
YohannParis, line 12: 🎉

PR #1087, opened 2023-05-09
fix(hmi-server): document service resources are unauthenticated

### LINKED ISSUE #1086: [BUG]: fix authentication on document resources
**Describe the bug**
The document service resources are not protected by Authentication - likely a copy/paste side effect of decommissioning the old document service

dvince2:
Adding the @Authenticated annotation to our document resources 

Resolves #1086


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ConceptResource.java
```diff
@@ -19,7 +19,7 @@
 import javax.ws.rs.core.Response;
 
 @Path("/api/concepts")
-@Authenticated
+
```
YohannParis, line 22: Are you sure you want a white space here?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/documentservice/DictionariesResource.java
```diff
@@ -1,6 +1,7 @@
 package software.uncharted.terarium.hmiserver.resources.documentservice;
 
 
+import io.quarkus.security.Authenticated;
```
YohannParis, line 4: Where this package being used?

YohannParis (commented):
I don't really get what's happening here, you remove the annotation but you added the packages without the annotation.

YohannParis (approved):

dvince2:
> 

Originally we had @Authenticated on like 80% of our resources.  This PR was initially to add it to the remaining 20%... however after the PR was issued I realized that by default quarkus has authentication enabled on everything and this annotation isn't needed at all. Rather than leave it to cause confusion, I removed it all together... but forgot to remove the import statements everywhere that it was :)

PR #1088, opened 2023-05-09
fix: misc workflow fixes and polish

echl:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/54dfb166-29f1-4925-aabb-b6703b88d1b2

* Fix workfow nodes overlapping ui elements
* Fix overview and workflow tabs not being in a "loaded" state (i.e. tab header shows they are not selected)
* Fix `@click.stop` on canvas component preventing you from exiting node creation context menu 
* Stop node drag when cursor leaves canvas to avoid some weird behaviours (e.g. if you close/open a side panel while dragging, cursor may be outside of the node but you are still dragging and unable to do a 'mouseup' event to end the drag action.


packages/client/hmi-client/src/components/widgets/tera-slider.vue
```diff
@@ -64,6 +64,10 @@ const sidePanelTabStyle = computed(
 </script>
 
 <style scoped>
+.slider {
+	z-index: 1;
+}
```
YohannParis, line 69: Please handle z-index within the parent component that's using `tera-slider.vue`.
And when doing so add `isolation: isolate;` to make sure that those z-indexes are not conflicted with other components.
Take a look at `App.vue` for how to implement it.

packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
```diff
@@ -74,7 +74,9 @@ nav {
 	display: grid;
 	grid-auto-columns: min(calc(100% / var(--nb-tabs, 1)), 20%);
 	grid-auto-flow: column;
-	margin: 0.2rem 0.2rem 0 0.2rem;
+	padding: 0.2rem 0.2rem 0 0.2rem;
+	background-color: var(--surface-ground);
+	z-index: 1;
```
YohannParis, line 79: Same comment as above.

YohannParis (approved):

YohannParis (approved):
Thank you

PR #1092, opened 2023-05-10
refactor: move emit functions where they belong and pass appropriate props (merges into split view)

shawnyama:
I have transferred some functions that were emitted from `tera-tab-content.vue` into an asset component into the actual component (`tera-model` and `code-editor`). Instead of emitting the functions the `project` prop had to be used in these asset components. Also some of the asset components were passed props that aren't even used so I got rid of those.

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/components/tera-preview-panel.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/components/tera-preview-panel.vue
@@ -6,15 +6,27 @@
 		:is-open="Boolean(previewItem)"
 	>
 		<template v-slot:content>
-			<component
-				v-if="assetComponent"
-				:is="assetComponent"
-				:is-editable="false"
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/page/data-explorer/components/tera-preview-panel.vue
YohannParis, line 1: I'm not sure I'm seeing the value in here. Each component will have their own set of props and events and trying to standardize it will not bring any benefits.

shawnyama: Yeah I see your point. I thought it would be easy to standardize but when I tried doing so it didn't get much better. Preview panel is clean but not as much with tera-tab-content. I will keep the emit transfers I did though.

YohannParis (approved):

PR #1095, opened 2023-05-11
chore: 1066 task calibration updates

Tom-Szendrey:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/c957e2d7-2f47-420b-8230-59410e233846

Letting the user set up their own calibration.
Note not all calibrations will converge. I also need to understand the implicit rules such as do initial params always need to sum to 1? This just means at the moment the user can easily make a calibration that returns null values

To Test:

I am running simulation service locally (It needs the following flag set to false to not send IDs) `ENABLE_TDS=false`
Then create a calibration node and connect it to a model config and a dataset
align the timestamp column with a drop down as well as pair up the model nodes with their column names
click run
click get status (like already finished)
click get result


packages/services/hmi-server/src/main/resources/application.properties
```diff
@@ -52,6 +52,7 @@ jsdelivr/mp-rest/url=https://cdn.jsdelivr.net
 mira-dkg/mp-rest/url=http://34.230.33.149:8771
 pdf/mp-rest/url=http://localhost:5000
 mira-metaregistry/mp-rest/url=http://34.230.33.149:8772
+# simulation-service/mp-rest/url=http://localhost:4321
```
YohannParis, line 55: I think you can delete this line.

Tom-Szendrey: Yepp will do

YohannParis (approved):

packages/client/hmi-client/src/types/ModelConfig.ts
```diff
@@ -0,0 +1,9 @@
+import { Model } from '@/types/Model';
+
+// This is temp as it will be replaced by types/Types when this is more finalized and moved to hmi-server
+export interface ModelConfig {
```
dvince2, line 4: If this is a new file just for this PR, is it worth it to just do the back end file and `@TSModel` it?

mwdchang: Maybe for next week? This is in flux because
- TDS has no model-config yet, and
- Waiting for everyone's api/interface (TDS, sim-service, hmi-server) to catch up with the new model-representation

Tom-Szendrey: I will gladly add this file and @TSModel it when TDS can create this

PR #1100, opened 2023-05-12
feat: model space config first pass

shawnyama:
# Description

Edit a model configuration and add new ones within the model view. Following [this design](https://www.figma.com/proto/5uzTEU3jtIfY62fp6I7SYo/Terarium---Workflow-%26-etc?page-id=1%3A1563&type=design&node-id=55-112262&viewport=898%2C-2371%2C0.14&scaling=min-zoom&starting-point-node-id=55%3A109576).

When a model node is opened in the workflow you can edit the values from there as well. You can switch the output that you want to edit by clicking on the corresponding output label. The value changes you make on the drilldown will be reflected on the node (vice-versa). **Note that when you change something in the node you have to press Enter or click away from the input box for the variable to update in the drilldown.** Not sure why it's like that. However adding a new model config in the drilldown doesn't add a new output in the node yet. I wasn't sure if we would want to have them all in the node and if I am even going about using the global storage elegantly. 

Things I am leaving for next iterations include:
- implement more detailed styling (eg. table borders in Jamie's design)
- adding config in drilldown adds to a node output?
- proper output names
- "Show in workflow" toggles in drilldown table

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/9522f4fa-3008-41cf-8438-be51dc4182e0


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -487,25 +492,27 @@ function formatAuthorTimestamp(username, timestamp) {
...
 	flex: 1;
-	overflow-x: auto;
-	overflow-y: hidden;
+	background: none;
+	border: none;
 }
 
-.p-splitter {
+section,
+.p-splitter:deep(.project-page) {
```
echl, line 510: Why do we need the `project-page` class if all Splitters have it?

packages/client/hmi-client/src/stores/opened-workflow-node.ts
```diff
@@ -1,16 +1,28 @@
 import { defineStore } from 'pinia';
 import { ProjectAssetTypes } from '@/types/Project';
 
+interface StringValueMap {
```
echl, line 4: Any reason to not just use a `Map<string, number>` here?

packages/client/hmi-client/src/stores/opened-workflow-node.ts
```diff
@@ -1,16 +1,28 @@
...
+
 // Probably will be an array later
 export const useOpenedWorkflowNodeStore = defineStore('opened-workflow-node', {
 	state: () => ({
 		assetId: null as string | null,
-		pageType: null as ProjectAssetTypes | null
+		pageType: null as ProjectAssetTypes | null,
+		openedOutputIndex: 0 as number,
+		// model node
+		initialValues: null as StringValueMap[] | null,
```
echl, line 15: Nitpick; maybe we name these `modelConfigInitialValues` and `modelConfigParameterValues` so that we know it's explicitly related to model node functionality? Or maybe wrap the two fields in a `modelConfig` object?

packages/client/hmi-client/src/stores/opened-workflow-node.ts
```diff
@@ -1,16 +1,28 @@
...
+		parameterValues: null as StringValueMap[] | null
 	}),
 	actions: {
 		set(assetId: string | null, pageType: ProjectAssetTypes | null) {
```
echl, line 19: Nitpick; since there's more than one setter now, maybe we can rename this one to be more clear about what it is setting.

echl (approved):

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -386,6 +380,124 @@ let renderer: PetrinetRenderer | null = null;
...
+function generateModelConfigValues() {
+	// Sync with workflow
+	if (
+		model.value &&
+		openedWorkflowNodeStore.assetId === model.value.id.toString() &&
+		openedWorkflowNodeStore.initialValues !== null &&
+		openedWorkflowNodeStore.parameterValues !== null
+	) {
+		// Shallow copy
+		initialValues.value = openedWorkflowNodeStore.initialValues as any; // Not sure why the typing doesn't match
```
echl, line 460: For typing, try defining `const initialValues = ref<StringValueMap[] | null>([{}]);`

packages/client/hmi-client/src/page/project/tera-project.vue
shawnyama: Because :deep effects the children components regardless and `tera-model.vue` is one of them and `tera-model.vue` has a splitter :(

packages/client/hmi-client/src/components/models/tera-model.vue
shawnyama: That makes it want to check if `initialValues` is null everywhere lol. I'll see if changing things to just a map can yield anything

packages/client/hmi-client/src/stores/opened-workflow-node.ts
shawnyama: The problem that I run into is that if I want to edit a value in an input field `v-model` doesn't play nice when it comes to setting a Map value. I use this StringValueMap interface in various places so a lot of syntax will change when I start using .set and .get - I'll investigate this in a second pass.

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -138,7 +145,7 @@ function showNodeDrilldown() {
 		if (pageType && assetId) {
 			openedWorkflowNodeStore.set(assetId, pageType);
 		}
-	} else alert('Node needs a valid output');
+	} else logger.error('Node needs a valid output');
```
tangopapatime, line 148: I'd probably make this silent `logger.error('Node needs a valid output', { silent: true }`

tangopapatime (approved):
LGTM

PR #1108, opened 2023-05-16
chore: clean up unused files

mwdchang:
### Summary
Cleans up unused files and routes to make things easier to navigate.

- SimulationPlan: The notion of a wiring simulation-plan was axed in the February hackathon


dvince2 (approved):
Nice! Always nice to have cleaner code

PR #1109, opened 2023-05-16
fix: 1097 task create hmi extraction service as a microservice in askem

tangopapatime:
# Description

Made several updates to the pdf extraction to use the new `hmi-extraction-service` properly including fixing that annoying multipart issue that @dvince2 managed to figure out.  Also updated the hmi logging to work a little better*.


packages/services/hmi-server/build.gradle
```diff
@@ -29,13 +29,19 @@ dependencies {
...
+	implementation 'org.glassfish.jersey.core:jersey-server:3.1.1'
 
 	// Persistence
 	implementation "io.quarkus:quarkus-hibernate-orm"
 	implementation "io.quarkus:quarkus-jdbc-postgresql"
 	implementation "io.quarkus:quarkus-hibernate-orm-panache"
 	annotationProcessor "io.quarkus:quarkus-panache-common"
-
+	// Multipart file upload/download
+	implementation 'org.glassfish.jersey.media:jersey-media-multipart:3.1.1'
```
dvince2, line 42: phew we're going to have a massive amount of conflicts....

Unrelated to above, why is this imported twice?

dvince2 (approved):

packages/services/hmi-server/src/main/resources/application.properties
```diff
@@ -50,7 +50,8 @@ mit/mp-rest/url=https://mit-proxy.staging.terarium.ai
 github/mp-rest/url=https://api.github.com
 jsdelivr/mp-rest/url=https://cdn.jsdelivr.net
 mira-dkg/mp-rest/url=http://34.230.33.149:8771
-pdf/mp-rest/url=http://localhost:5000
+%dev.pdf/mp-rest/url=http://localhost:5000
+%prod.pdf/mp-rest/url=https://hmi-extraction.staging.terarium.ai:5000/
```
YohannParis, line 54: I wonder if we should do like simulation-service and only use staging, unless needed to run locally.

tangopapatime: Yeah I was wondering the same thing.  I'm good either/or.  lmk.

tangopapatime: can I make this switch later?  caught in this chicken/egg paradox right now since we don't have anything on staging yet.

PR #1121, opened 2023-05-19
feat: misc workflow polish

echl:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/5aac16ed-99bc-4758-861e-d76eea609cf1

* Update edge and port styling
* The entire port label is hover-able for port interactions
* Edges snap to ports
* You can drag an edge from output to input


YohannParis (approved):
I would prefer the styling to be in CSS mostly and using classes, instead of all the convoluted JS inline `style:`

echl:
> I would prefer the styling to be in CSS mostly and using classes, instead of all the convoluted JS inline `style:`

I tend to agree though some of the styling needs to be dynamic

YohannParis:
> > I would prefer the styling to be in CSS mostly and using classes, instead of all the convoluted JS inline `style:`
> 
> I tend to agree though some of the styling needs to be dynamic

True, but the dynamic styling can select classes instead of writing new CSS.

echl:
> > > I would prefer the styling to be in CSS mostly and using classes, instead of all the convoluted JS inline `style:`
> > 
> > 
> > I tend to agree though some of the styling needs to be dynamic
> 
> True, but the dynamic styling can select classes instead of writing new CSS.

I'll try it!

PR #1128, opened 2023-05-24
chore: 1098 calibrate detail pane

Tom-Szendrey:
# Description
-Create a placeholder file that will contain the drill down components for the calibration
-update the drill down to allow us to send more than just asset information
-move model diagram to its own component so it can be reused by calibration drill down and others
-Adding ENUM: WorkflowOperationTypes

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/tera-calibration-side-panel.vue
Tom-Szendrey, line 1: This is mostly just a placeholder file to be updated a lot in the future

Tom-Szendrey, line 1: This is mostly just a placeholder file to be updated a lot in the future

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -135,9 +135,7 @@ function showNodeDrilldown() {
 			default:
 				break;
 		}
-		if (pageType && assetId) {
-			openedWorkflowNodeStore.set(assetId, pageType);
-		}
+		openedWorkflowNodeStore.set(assetId, pageType, props.node);
```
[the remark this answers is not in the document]
shawnyama: you should use `setDrilldown`, `set` doesn't work

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -44,13 +44,19 @@
 			<SplitterPanel
 				class="project-page"
 				v-if="
-					openedWorkflowNodeStore.assetId &&
-					openedWorkflowNodeStore.pageType &&
-					pageType === ProjectAssetTypes.SIMULATION_WORKFLOW
+					(openedWorkflowNodeStore.assetId &&
```
shawnyama, line 47: Change it to this so it appears only when the workflow is open

`v-if="pageType === ProjectAssetTypes.SIMULATION_WORKFLOW &&
				((openedWorkflowNodeStore.assetId && openedWorkflowNodeStore.pageType) || openedWorkflowNodeStore.node)"`

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -0,0 +1,624 @@
...
+		relatedTerariumArtifacts.value = [];
+	}
+};
+// Whenever selectedModelId changes, fetch model with that ID
+watch(
+	() => [props.assetId],
+	async () => {
+		updateLatexFormula('');
+		if (props.assetId !== '') {
+			const result = await getModel(props.assetId);
```
shawnyama, line 253: Okay so it looks like the model should be passed into this instead of the id. The model is needed in the parent because we need to display the model name, description etc which is missing

shawnyama (approved):

shawnyama (approved):

PR #1154, opened 2023-05-30
fix: model search when a model has null for description/name was breaking

Tom-Szendrey:
# Description
Thanks Shawn for the help on your local!!!

Resolves #(issue)


shawnyama (approved):
Perfection

PR #1171, opened 2023-06-02
fix: removed navigation panel from Dataset, Model and Document pages

jamiewaese-uncharted:
I removed the table of contents with links to scroll to different sections of the page. It occupied too much space.

fyi @shawn I discussed your idea of showing/hiding that feature depending on window width with @pascaleproulx and we decided to remove it (for now) because we felt the content switcher should be on the right side of the title. We'll revisit the scroll-to-section feature down the road.

MODEL PAGE
image
image

DOCUMENT PAGE
image

DATASET PAGE
image

WHEN IN A WORKFLOW
image


dvince2 (commented):
lgtm

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -638,9 +615,12 @@ onUpdated(() => {
 	}
 });
 
+/*
+// apparently this is never used?
 const launchForecast = () => {
 	showForecastLauncher.value = true;
 };
+*/
```
mwdchang, line 623: You removed the invocation above in the template section, so it is definitely no longer being used. 

Let's just remove it altogether instead of having the comment/code laying around.

PR #1191, opened 2023-06-06
feat: add workflow entity as part of projet resource

mwdchang:
### Summary
Putting together basic CRUD APIs for saving and loading workflows


Tom-Szendrey (approved):
Just a heads up if you leave the tab before it saves it will not save the contents of the tab.

PR #1192, opened 2023-06-07
feat: add LLM agent integration

mattprintz:
# Description

* Includes a fully working, but limited, hmi client integration to Terarium.
* Provide some new widgets and controllers for communicating with the new jupyter service that also must be running.
* Due to security around websockets, we need to proxy the websocket connection to the jupyter service using the same url/port as the hmi-service is using.


tangopapatime:
As previously mentioned, this PR should be good to go for a first viewing for those reviewing.  Please let me know if you see any high priority items that need changing.  @YohannParis

mattprintz:
Taken care of via PR #1373 .

Issue #1168, opened 2023-06-02
[FEAT]: Show Provenance in github file importer


dvince2:
**Is your feature request related to a problem? Please describe.**
See comments in issue #1142 for more information and screengrabs.

Show the provenance for files imported from github below the preview panel

Issue #1164, opened 2023-06-02
[FEAT]: Allow users to impor unknown file type as a known type


dvince2:
**Is your feature request related to a problem? Please describe.**
Spin-off task from https://github.com/DARPA-ASKEM/Terarium/issues/1142

**Describe the solution you'd like**
When using the github file browser, if there is a file of unknown type users should be able to tell the hmi what to treat the file as (for example, csv, txt, pdf, etc) and then import it as that

**Additional context**
See screenshots on issue 1142 for more information

Issue #1144, opened 2023-05-29
[FEAT]: Import files from the overview page


jamiewaese-uncharted:
This task uses the same component as we're building for https://github.com/DARPA-ASKEM/Terarium/issues/1142

When users import files from the overview page, in most cases we can assume the resource type (document / code / dataset) from the file extension. But not all files will have recognized extensions. So we need to provide a mechanism for the user to assign a file type. Here are some designs for how this might work:

![Uploading files from the overview page 1](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/cbf1d2a1-741e-45d6-b146-0ae38538d5b2)

![Uploading files from the overview page 2](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/5e3929de-84bb-43ef-ad79-e2d1263f5b8f)

![Uploading files from the overview page 3](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/350f5d99-09a5-413f-a5f0-48f395938d66)

![Uploading files from the overview page 4](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/e75400e6-7d9e-479a-ad3f-a252f7eb5e8b)

PR #1194, opened 2023-06-08
feat: made several visual updates to the import github modal and populated the resource datatable

tangopapatime:
# Description
Update the github import modal with the suggest visual enhancements
Populated resource manager datatable

Resolves #(issue)

https://github.com/DARPA-ASKEM/Terarium/issues/1164
https://github.com/DARPA-ASKEM/Terarium/issues/1168

which are related to https://github.com/DARPA-ASKEM/Terarium/issues/1144

Sorry merged main and get things back to create the video :P


mwdchang:
Food for thoughts: 

I am a bit skeptical about all the z-index changes, because the numbers are relative to the context where they are set and they are not compared against each other in a absolute sense. 

It may be better if we have a layout schematic and we can use that as a reference rather than tweaking z-indices all over the place. @jamiewaese-uncharted

tangopapatime:
> Food for thoughts:
> 
> I am a bit skeptical about all the z-index changes, because the numbers are relative to the context where they are set and they are not compared against each other in a absolute sense.
> 
> It may be better if we have a layout schematic and we can use that as a reference rather than tweaking z-indices all over the place. @jamiewaese-uncharted

absolutely.  I knew out of all changes there might a comment about this.  So, maybe I did it wrong, but it was the only way I was able to get the `DropDown` list to work within the modal.  I just scaled the indexes so the order should be in theory preserved, at least for the elements who's z-index was predefined.

packages/client/hmi-client/src/components/widgets/tera-import-github-file.vue
```diff
@@ -99,13 +101,13 @@
 									:key="index"
 									@click="previewTextFile(content)"
 								>
-									<input
-										type="checkbox"
-										:id="content.name"
-										:value="content"
+									<Checkbox
```
dvince2, line 104: Out of curiosity, what is the difference here between a `Checkbox` and a `input type="checkbox`?

dvince2 (approved):

packages/client/hmi-client/src/components/widgets/tera-import-github-file.vue
tangopapatime: one is a `primevue` component and the latter is the native version.

tangopapatime:
[updates.webm](https://github.com/DARPA-ASKEM/Terarium/assets/98172806/364679e9-b2cf-41aa-aa7c-935c468fc265)

PR #1201, opened 2023-06-12
fix(hmi): unable to add datasets

### LINKED ISSUE #1200: [BUG]: Unable to create new datasets
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

dvince2:
This fix correctly sends up the expected snake_case to TDS instead of camelCase. A few other odds and ends fixes here as well.

NOTE that this still requires a pt 2 fix in orchestration for setting up buckets correctly for local dev.

Resolves #1200

### DIFF
--- a/packages/client/hmi-client/src/services/project.ts
+++ b/packages/client/hmi-client/src/services/project.ts
@@ -80,6 +80,8 @@ async function getAll(): Promise<IProject[] | null> {
 
 /**
  * Get project assets for a given project per id
+ * @param projectId projet id to get assets for
+ * @param types optional list of types. If none are given we assume you want it all
  * @return ProjectAssets|null - the appropriate project, or null if none returned by API
  */
 async function getAssets(projectId: string, types?: string[]): Promise<ProjectAssets | null> {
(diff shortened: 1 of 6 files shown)


tangopapatime (approved):

packages/client/hmi-client/src/services/project.ts
```diff
@@ -90,6 +92,9 @@ async function getAssets(projectId: string, types?: string[]): Promise<ProjectAs
 				// add URL with format: ...?types=A&types=B&types=C
 				url += `${indx === 0 ? '?' : '&'}types=${type}`;
 			});
+		} else {
+			url +=
+				'?types=datasets&types=model_configurations&types=models&types=publications&types=simulations&types=workflows';
```
mwdchang, line 97: Can we to opt-out semantics? If none is specified assume we want to retrieve everything? This makes it easier to add additional asset types.

shawnyama (approved):

packages/client/hmi-client/src/services/project.ts
dvince2: I believe that TDS's old API took no types as "return everything"... where as now it returns an error and requires an explicit type(s).  There is absolutely better ways to do this that I want to circle back to - right now i just need to get this working :)

PR #1210, opened 2023-06-13
feat: smarter workflow save behaviour

mwdchang:
### Summary
- Buffer save state in a workflowDirty flag to avoid unnecessary saves
- Save check when component in unmounted

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
@@ -169,6 +169,7 @@ let canvasTransform = { x: 0, y: 0, k: 1 };
 let currentPortPosition: Position = { x: 0, y: 0 };
 let isMouseOverPort: boolean = false;
 let saveTimer: any = null;
+let workflowDirty: boolean = false;
 
 const newEdge = ref<WorkflowEdge | undefined>();
 const newAssetId = ref<string | null>(null);
@@ -201,6 +202,7 @@ function appendOutputPort(node: WorkflowNode, port: { type: string; label?: stri
 		value: [port.value],
 		status: WorkflowPortStatus.NOT_CONNECTED
 	});
+	workflowDirty = true;
 }
 
 // Run testOperation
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -418,11 +427,16 @@ watch(
...
-		workflowService.updateWorkflow(wf.value);
+		if (workflowDirty) {
+			workflowService.updateWorkflow(wf.value);
+			workflowDirty = false;
+		}
 	}, 8000);
 });
 onUnmounted(() => {
+	if (workflowDirty) {
+		workflowService.updateWorkflow(wf.value);
```
echl, line 438: IIRC this won't trigger if the user switches between two different workflow tabs, or closes a workflow tab if there is still an existing workflow tab.

mwdchang: Hmmm I don't see it (I put a console.log in umount and it seems to trigger). 

@echl can you outline the steps to reproduce? 

Also if this is the case we are going to have pretty severe memory issues with each tab holding onto a chunk of the memory and the DOM-tree. So hopefully this isn't the case.

echl: https://github.com/DARPA-ASKEM/Terarium/assets/16358728/422420d5-d78d-484a-8be8-d673aa2a115c

Steps to reproduce:
1. Open 2 workflows in different tabs
2. Switch between tabs (`console.log` in `onUnmounted` does not execute)
3. Close one of the workflow tabs (`console.log` in `onUnmounted` does not execute)
4. Close the other workflow tab (`console.log` in `onUnmounted` *does* execute

echl: Checking the value of `workflowDirty` in `onBeforeUpdate` should work for these cases.

mwdchang: `onBeforeUpdate` is an overkill because it is hooked into DOM events, it is kind of like the sledge-hammer solution 😄 

The point I was making is that we don't actually unmount the memory pressure is monotonically increasing, and that is irrelevant of how many instances of `tera-simulation-workflow` we have instantiated.

Anyway I patched up the workflow-2-workflow edge case.

echl (approved):

PR #1219, opened 2023-06-13
feat: adding button to route to resource

Tom-Szendrey:
# Description
Asset on overview should be able to route to its asset page

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/17b368d0-cdc5-4047-8326-1782f06532e8


packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
```diff
@@ -523,4 +539,27 @@ ul {
 .modal:deep(main) {
 	width: 50rem;
 }
+::v-deep(.asset-button.p-button) {
```
shawnyama, line 542: I think you can just do `:deep` instead of `::v-deep`. I think `::v-deep` was deprecated

Tom-Szendrey: Thanks will change

PR #1228, opened 2023-06-14
Simulate api update

mwdchang:
### Summary
Update to reflect simulation-service API changes. There are now two parts
- Creation of the simulate request, which returns a jobId
- Creation of the simulation object referring to the jobId


mwdchang:
Going to do this in Tom's branch https://github.com/DARPA-ASKEM/Terarium/pull/1229

PR #1238, opened 2023-06-15
feat: model config/node update 

mwdchang:
# Description

Model configuration editing/saving as well as model node design update

### Shawn's issue:
I don't think I am going about updating the side panel properly. I think it makes sense to update it once the change has been made to the backend but as you can see in the video the table rows reload slowly when I do that. In the video I change the value in the config - this doesn't reload funny because I am using a frontend ref to keep track of it. I am wondering what would be the best way to go about this - should I also use a frontend ref for when a model config is added? On the other hand grabbing stuff just from the backend makes the source of truth easier to handle.

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/63241866-6515-41d0-89fe-4f824fac7e73


packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -122,56 +148,140 @@ import TabView from 'primevue/tabview';
...
 
 const selectedModelConfig = ref();
 const fakeExtractions = ref(['Resource 1', 'Resource 2', 'Resource 3']);
 
-const initialValues = ref<NumericValueMap[]>([{}]);
-const parameterValues = ref<NumericValueMap[]>([{}]);
 const openValueConfig = ref(false);
 const cellValueToEdit = ref({ data: {}, field: '', index: 0 });
 
+// const
```
dvince2, line 176: this can be removed

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -122,56 +148,140 @@ import TabView from 'primevue/tabview';
...
-const openedWorkflowNodeStore = useOpenedWorkflowNodeStore();
+// const variableTypes = computed(() => props.model.semantics ? Object.keys(props.model.semantics.ode) : []);
 
-const modelConfiguration = computed(() =>
-	modelConfigNames.value.map((name, i) => ({
-		name,
-		...initialValues.value[i],
-		...parameterValues.value[i]
-	}))
+const configurations = computed(
```
dvince2, line 184: What is the expected type on this on?

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -122,56 +148,140 @@ import TabView from 'primevue/tabview';
...
+	() => editableModelConfigs.value?.map((m) => m.configuration.semantics.ode) ?? null
 );
 
+// TODO: Clean this up and use appropriate loops
+const modelConfigurationTable = computed(() => {
+	if (editableModelConfigs.value && configurations.value) {
+		console.log('Configuration', configurations.value);
```
dvince2, line 191: There are a bunch of log statements inside of this loop - Unless this is a super transient change we should remove those to reduce logspam

dvince2: Similarly there is a lot of `eslint-disable-next-line` in there too - what warnings are those suppressing?

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -122,56 +148,140 @@ import TabView from 'primevue/tabview';
...
+		for (let i = 0; i < odes.length; i++) {
+			variables[i] = {};
+			// eslint-disable-next-line
+			for (const [key, values] of Object.entries(odes[i])) {
+				const flattenedObj = {};
+
+				if (isArray(values)) {
+					// @ts-ignore
+					// eslint-disable-next-line
+					for (let j = 0; j < values.length; j++) {
```
dvince2, line 220: This looks like a O(n^3) loop which gives me pause :D

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
shawnyama: Thanks, I'll cast that to an `object[]` since the names of the keys could vary in the future

packages/client/hmi-client/src/components/workflow/tera-calibration-node.vue
```diff
@@ -56,6 +55,7 @@ watch(
 	() => datasetId.value, // When dataset ID changes, update datasetColumnNames
 	async () => {
 		if (datasetId.value) {
+			// console.log(modelConfig);
```
dvince2, line 58: temp code

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ModelConfigurationResource.java
```diff
@@ -0,0 +1,63 @@
...
+	ModelConfigurationProxy proxy;
+
+	@GET
+	public Response getModelConfigurations() {
+		return proxy.getModelConfigurations();
+	}
+
+	@POST
+	@Consumes(MediaType.APPLICATION_JSON)
+	public Response createModelConfiguration(Object config) {
```
dvince2, line 33: Do we have a `config` POJO we can use?

dvince2: Actually I see below `ModelConfiguration` if thats the same can we use that here and anywhere else there is an `Object` reference?

dvince2 (changes requested):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ModelConfigurationResource.java
shawnyama: When I create - the config has to lack some attributes of the ModelConfig object such as id etc. So I just made it an object.

dvince2: I think you can use `@TSOptional` on those fields to have it not complain about them being issing

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
mwdchang: ODE is actually an object with a "fixed-number" of keys so it is more like `O(N^2)`.  But the parsing could be clearer and better, perhaps this is something we can flag with a FIXME and we can work offline @shawnyama to improve - the modelConfiguration change is holding back other PRs so I think we don't want to dwell too much time unless it is a critical error.


See example here: https://github.com/DARPA-ASKEM/Model-Representations/blob/main/petrinet/examples/sir.json

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ModelConfigurationResource.java
shawnyama: True but what I am passing is more like a payload to create a model config, everywhere else those attributes shouldn't be optional. Do you think I should rename just the `config` variable to `createConfigPayload` or should I create another POJO along with that?

mwdchang: Note ModelConfiguration actually doesn't quite work because down-stream systems cannot consume AMR format just yet. So we need to do a bit of hacky things and transforms  on the fly .... so in light of that not too horribly worried about Object-typings and related matters, but we should revisit next week.

PR #1240, opened 2023-06-15
fix: update dataset columns to reflect change in data service

mattprintz:
I don't think this is being used anywhere yet in Terarium, so hopefully is safe to change.
Should match changes in
https://github.com/DARPA-ASKEM/data-service/pull/259.

Let me know if anything else needs to be fixed.


dvince2 (approved):
Thanks!

PR #1259, opened 2023-06-21
feat: stratify drilldown step 1

### LINKED ISSUE #1258: Drilldown step 1 as per designs
- [x] https://github.com/DARPA-ASKEM/Terarium/issues/1159
- [x] Display connected model and form to select strata type, list of state labels
- [x] Render strata model (may need new renderer)

echl:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/6843678b-6613-4f23-9dba-d6898189c40f

Resolves #1258


packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -25,7 +25,7 @@
 			>
 				<div
 					class="input-port-container"
-					@mouseover="(event) => mouseoverPort(event)"
+					@mouseenter="(event) => mouseoverPort(event)"
```
echl, line 28: This is to fix an error that was occuring on mouseover

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
```diff
@@ -56,7 +56,7 @@
 			>
 				<div
 					class="output-port-container"
-					@mouseover="(event) => mouseoverPort(event)"
+					@mouseenter="(event) => mouseoverPort(event)"
```
echl, line 59: This is to fix an error that was occuring on mouseover

mwdchang:
`src/petrinet/petrinet-service` is slated to be decommissioned - you should move what you need to `src/model-representations/petri-net/petrinet-service`.

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -128,6 +147,7 @@ const props = defineProps<{
 	model: Model | null;
 	isEditable: boolean;
 	nodePreview?: boolean;
+	strataView?: boolean;
```
mwdchang, line 150: My feel is that you are overloading too much functionalities into tera-model-diagram. 

It becomes difficult to understand, and to use because the number of combinations you need to be aware of grows geometrically. For example you don't need a math-editor for strata, but it is included anyway .... etc etc

We had something similar before (albeit with API) that takes in like 5-6 different options, some mutually exclusive and some complementary,  and it is very tedious to get things right. So I think an independent strata-component would be better suited.

echl: Good point

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -325,7 +325,7 @@ const contextMenuItems = ref([
 // Render graph whenever a new model is fetched or whenever the HTML element
 //	that we render the graph to changes.
 watch(
-	[props.model, graphElement],
+	[() => props.model, graphElement],
```
shawnyama, line 328: Are you sure it isn't this instead: `() => [props.model, graphElement]`

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -200,24 +207,33 @@ export const parseAMRPetriNet2IGraph = (
...
+			},
 			nodes: []
 		});
 	}
 
 	// Add each node found in T
 	for (let i = 0; i < petrinet.transitions.length; i++) {
 		const aTransition = petrinet.transitions[i];
+		const typeMap = typing?.type_map.find((map) => map.length === 2 && aTransition.id === map[0]);
+		const strataType = typeMap?.[1] ?? '';
```
mwdchang, line 236: Same as above, hard to to understand 
```
map.length === 2 && aNode.id === map[0]
```

Maybe consider a helper function and give it a meaningful name.

mwdchang:
Unfortunately there seemed to be some copy and paste snafus int he petri => AMR transition work. There are function that does similar things and we ended up working against a version of the function that should have been commented out or removed.... 🤦 

`parseAMRPetriNet2IGraph` should not be used because while it transforms the AMR into an IGraph, it is lossy and we cannot transform it back. 

The one that should be used I think, is `convertToGraph`

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -200,24 +207,33 @@ export const parseAMRPetriNet2IGraph = (
 		edges: []
 	};
 	const petrinet = amr.model as PetriNetModel;
+	const typing: TypingSemantics | undefined = amr.semantics?.typing;
 	// add each nodes in S
 	for (let i = 0; i < petrinet.states.length; i++) {
 		const aNode = petrinet.states[i];
+		const typeMap = typing?.type_map.find((map) => map.length === 2 && aNode.id === map[0]);
+		const strataType = typeMap?.[1] ?? '';
```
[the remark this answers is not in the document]
echl: It finds the first value of `type_map` that satisfies the condition `map.length === 2 && aNode.id === map[0]`. Seems self explanatory to me!

The structure of `type_map` is an array of arrays, where each inner array has 2 elements, the first element is a state or transition id, the second element is the type id. So we're grabbing that inner array and assigning it to `typeMap`, then we're getting the second element of `typeMap` to get the corresponding type id.

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
echl: I think a helper function for a one liner would be too verbose but I can definitely add an explanatory comment.

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
echl: Yeah, Vue will throw a warning for this if props.model is null, if you don't use the getter function syntax.

packages/client/hmi-client/src/utils/color-schemes.ts
```diff
@@ -0,0 +1,16 @@
...
+	'#25ac82',
+	'#1f998a',
+	'#24878e',
+	'#2b748e',
+	'#34618d',
+	'#3d4d8a',
+	'#453581',
+	'#481c6e',
+	'#440154'
+];
```
mwdchang, line 16: Did you get the palette some where? if so can you leave a link? 
If you create this yourself can you give a brief description/comment of the characteristics of this palette?

packages/client/hmi-client/src/services/models/stratification-service.ts
```diff
@@ -39,28 +43,78 @@ export function generateAgeStrataModel(stateNames: string[]) {
...
+	};
 	return {
+		id: '',
+		framework: '',
 		name: 'Age-contact strata model',
 		description: 'Age-contact strata model',
+		schema:
+			'https://raw.githubusercontent.com/DARPA-ASKEM/Model-Representations/petrinet_v0.1/petrinet/petrinet_schema.json',
+		schema_name: 'petrinet',
+		model_version: '0.1',
```
mwdchang, line 91: This amr creation is getting repeated over and over again, I would maybe suggest that we have a function in petrinet-service that builds a basic model skeleton, something like 

```
const amr = petrinetService.newModel(id, name, description ....)
```

packages/client/hmi-client/src/components/models/tera-strata-model-diagram.vue
```diff
@@ -0,0 +1,384 @@
...
+	isEditable: boolean;
+	nodePreview?: boolean;
+}>();
+
+const menu = ref();
+
+const newModelName = ref('New Model');
+
+const equationLatex = ref<string>('');
+const equationLatexOriginal = ref<string>('');
```
mwdchang, line 91: Do we still care about showing LaTex and math equation in this context? Can we just remove them and simplify?

echl: I have already removed them for the next pr, if you don't mind waiting until then.

packages/client/hmi-client/src/utils/color-schemes.ts
echl: This is just temporary, I used the same ones for the simulation graph. I'll add a comment explaining as such.

packages/client/hmi-client/src/services/models/stratification-service.ts
echl: I agree. If you don't mind, I can address that in the next pr.

packages/client/hmi-client/src/utils/color-schemes.ts
mwdchang: Okay, If you mean you used VIRDIS colour palette that is probably not the right solution. VIRDIS is meant for a continuous range of values. Here, with the graph colouring, what we would rather have is a categorical colouring palette. E.g. Susceptible is twice as high as Recovered makes no sense.

Colorbrewer here https://colorbrewer2.org/#type=sequential&scheme=BuGn&n=3a provide some nice categorical (qualitative) palette you can select.

mwdchang (approved):

PR #1270, opened 2023-06-23
feat: add observables in model config

shawnyama:
# Description

Note: Not sure why `observableMenu` ref doesn't work for the Menu - it should be recognized as an array.


packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -178,10 +213,54 @@ const cellValueToEdit = ref({ data: {}, field: '', index: 0 });
...
+
+function addObservable() {
+	for (let i = 0; i < editableModelConfigs.value.length; i++) {
+		if (!editableModelConfigs.value[i].configuration.semantics.ode.observables) {
+			editableModelConfigs.value[i].configuration.semantics.ode.observables = [];
+		}
+
+		editableModelConfigs.value[i].configuration.semantics.ode.observables.push({
+			id: `Obs ${editableModelConfigs.value[0].configuration.semantics.ode.observables.length + 1}`,
+			expression: 'expr'
```
mwdchang, line 232: You need expressionML

shawnyama: What is that - do you mean I should put a mathml expression in there?

PR #1282, opened 2023-06-26
feat: update workflow menu items and context menu to match design

mj3cheun:
# Description

* Update workflow menu items list to differentiate between deterministic & probabilistic sim/calibrate.
    * Use disabled menu items for not yet built probabilistic options
* Update simulate drilldown panel header to `Simulate (deterministic)`

Screenshot 2023-06-26 at 3 19 12 PM

Note: since `primevue/contextmenu` does not support headers in the context menu (which is required in the design), a workaround was used to allow the `primevue/menu` component to be used as a contextmenu.

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-context-menu.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-context-menu.vue
@@ -1,0 +1,126 @@
+<template>
+	<Portal :appendTo="props.appendTo">
+		<div class="tera-context-menu" ref="container">
+			<Menu v-if="isVisible" :model="props.model" />
+		</div>
+	</Portal>
+</template>
+
+<script setup lang="ts">
+import { ref, onUnmounted, PropType } from 'vue';
+import Portal from 'primevue/portal';
+import Menu from 'primevue/menu';
+import { MenuItem } from 'primevue/menuitem';
+import { DomHandler, ZIndexUtils } from 'primevue/utils';
+
+const props = defineProps({
+	model: {
+		type: Array as PropType<MenuItem[]>,
+		default: null
+	},
+	appendTo: {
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/widgets/tera-context-menu.vue
```diff
@@ -0,0 +1,126 @@
...
+		type: String,
+		required: false,
+		default: 'body'
+	},
+	baseZIndex: {
+		type: Number,
+		required: false,
+		default: 0
+	}
+});
```
mwdchang, line 31: We have quite a few z-index issues already. 
Is this really a good idea to give a control here when you don't really have control? 
e.g. z-index is not absolutely comparative with each other....z-index of 10 in the side-panel is different than z-index of 10 in the header.

Why do we need a wrapper component anyway?

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -354,6 +374,7 @@ function onDrop(event) {
 }
 
 function toggleContextMenu(event) {
+	console.log('hi');
```
mj3cheun, line 377: remove

packages/client/hmi-client/src/components/widgets/tera-context-menu.vue
mj3cheun: we need a wrapper component because the `primevue/contextmenu` does as shown below and it doesnt seem to have an option to use headers instead of nested submenus

Screenshot 2023-06-26 at 12 08 57 PM

mwdchang (approved):

PR #1295, opened 2023-06-27
feat: display extractions to describe a dataset or model

YohannParis:
# Description

* Display extractions better.


packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -53,10 +51,13 @@
 					<section>{{ dataset?.username || '-' }}</section>
 				</section>
 			</section>
-			<section class="metadata data-row" v-if="!metadata">
+			<section class="metadata data-row">
 				<section>
 					<header>Source Name</header>
-					<section>{{ dataset?.source || '-' }}</section>
+					<section v-if="dataset.url === 'https://github.com/reichlab/covid19-forecast-hub/'">
```
dvince2, line 57: I'm assuming so, but, to confirm we want to hardcode this github url and title into `tera-dataset.vue`?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/DKGResource.java
```diff
@@ -0,0 +1,38 @@
...
+
+	@GET
+	@Path("/{curies}")
+	public List<DKG> searchConcept(
+		@PathParam("curies") final String curies
+	) {
+		try {
+			return  proxy.getEntities(curies);
+		} catch (RuntimeException e) {
+			log.error("Unable to fetch DKG", e);
```
dvince2, line 34: Has this code path been tested at all? How does the front end react to recieving `null` - is it safer to return a 500 response here?

dvince2 (approved):

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
YohannParis: that should not be here for some resons.

PR #1298, opened 2023-06-27
fix: better method of displaying model config table

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/models/tera-model-configuration.vue
+++ b/packages/client/hmi-client/src/components/models/tera-model-configuration.vue
@@ -1,9 +1,9 @@
 <template>
 	<DataTable
-		v-if="configurations?.[0]?.initials && configurations?.[0]?.parameters"
+		v-if="!isEmpty(amrConfigurations) && !isEmpty(tableHeaders)"
 		class="model-configuration"
 		v-model:selection="selectedModelConfig"
-		:value="modelConfigurationTable"
+		:value="amrConfigurations"
 		editMode="cell"
 		showGridlines
 		scrollable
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -1,99 +1,153 @@
...
-		</ColumnGroup> -->
-	</DataTable>
+		<div class="p-datatable-wrapper">
+			<table class="p-datatable-table p-datatable-scrollable-table editable-cells-table">
+				<thead class="p-datatable-thead">
+					<tr v-if="isEditable">
+						<th class="p-frozen-column"></th>
+						<th class="p-frozen-column second-frozen"></th>
+						<th v-for="({ name, colspan }, i) in tableHeaders" :colspan="colspan" :key="i">
+							{{ capitalize(name) }}
```
YohannParis, line 14: You can simply use CSS for that with [`text-transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform)
```css
.p-datatable-thead th {
    text-transform: capitalize;
}
```

packages/client/hmi-client/src/components/models/tera-model-configuration.vue
```diff
@@ -1,99 +1,153 @@
...
+								<span>{{ rate.expression }}</span>
+								<Button
+									class="p-button-icon-only p-button-text p-button-rounded p-button-icon-only-small cell-menu"
+									icon="pi pi-ellipsis-v"
+									@click.stop="openValueModal('rates', 'expression', i, j)"
+								/>
+							</section>
+							<InputText
+								v-else
+								v-model="modelConfigs[i].configuration.semantics.ode.rates[j].expression"
```
YohannParis, line 88: We talked about not using `v-model` to avoid issues

PR #1300, opened 2023-06-27
Ts/1225 calibration updates

Tom-Szendrey:
# Description
-Updating calibration to work with new API 
-Grabbing simulation.csv as result (from simulation) rather than running a new simulation based off of calibration's results
-Trying to reduce the use of the node store for calibration for future work around this node
-Removing some currently unused tabs

*Still todo*: 
Properly sync up the graphs on the node + drill down
Probably should have the runCalibrate use a poller instead but can worry about that in future
cleanup extra  

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/051bdb30-e000-446e-b152-874b4497f2fd


packages/client/hmi-client/src/stores/opened-workflow-node.ts
```diff
@@ -21,12 +20,7 @@ export const useOpenedWorkflowNodeStore = defineStore('opened-workflow-node', {
 		numCharts: 1,
 		chartConfigs: [] as ChartConfig[],
 		tspanUnit: TspanUnits[0],
-		tspan: [0, 100],
-		// calibrate node
-		calibrateNumCharts: 1,
-		calibrateRunIdList: [] as string[],
-		calibrateRunResults: {},
-		readOnlyMapping: null as any[] | null
```
mwdchang, line 29: Yes...burn it to the ground 🔥 👍

packages/client/hmi-client/src/components/workflow/tera-calibration-node.vue
```diff
@@ -1,71 +1,252 @@
...
+const startedRunId = ref<string>();
+const completedRunId = ref<string>();
+
+const datasetColumnNames = ref<string[]>();
+const modelColumnNames = ref<string[] | undefined>();
+const calibrateNumCharts = ref<number>(1);
+const runResults = ref<RunResults>({});
+const simulationIds: ComputedRef<any | undefined> = computed(
+	<any | undefined>(() => props.node.outputs[0]?.value)
+);
```
mwdchang, line 101: 1. Don't need to address it now, but there are a few things here that do not need to be `refs` 
2. I don't understand why your `simulationIds` is so complicated ..., should it just be a `string[]`?

Tom-Szendrey: Still cleaning up some stuff for sure
As for simulationIds it needs to be a ref in order for the simulation-chart to work properly - I was told by @mj3cheun 
Aside from this very fair it should not be of type any

PR #1301, opened 2023-06-28
feat: removing dependencies on pinia store

mwdchang:
### Summary
This is an overhaul of how messages and states are passed around with regards to a workflow. The current state is a mishmash of different things that were built on a pinia-store that is never intended to work end-to-end. In this PR:

- Unify and normalize the messages passing mechanisms so they are comprehendible, this means introducing proxy-components in favour of giving multiple meanings to individual IDs.
- Remove the store-cache and replace it with an EventBus
- Add a "state" section for each workflow operations to track internal states
- Remove chart dependency on pinia-store
- Fix amrConfiguration vs configuration
- Use new simulation-service
- Fix modelConfiguration generation bugs, we no longer generate a config every time the modelOperation node is open
- Make output/input port format a bit more uniform across the operations
- Clean up extra console log
- Fix serialization issue with expressionMathml


dvince2 (approved):
lgtm

PR #1315, opened 2023-07-05
feat: update createModel to take a model

Tom-Szendrey:
# Description
Add duplicate model
Update createModel to create a model now that we are using AMRs

Using the isEditable tag to see if this should have this drop down
@shawnyama please confirm this is the correct use of this? The model.vue is used in many locations I want to confirm this is the correct flag to check

image
image


shawnyama (approved):
Yeah you're using `isEditable` properly

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -25,6 +25,13 @@
 					@click="modelView = ModelView.MODEL"
 					:active="modelView === ModelView.MODEL"
 				/>
+				<Button
+					v-if="isEditable"
```
YohannParis, line 29: I don't think you need to edit a model to duplicate it.

YohannParis (approved):

packages/client/hmi-client/src/components/models/tera-model.vue
Tom-Szendrey: The tera-model.vue is being used for the dataexplorer, the workflow and the project page.

I believe that this flag is to denote that it is in the project page or the workflow. This allows me to basically say dont give this option if we are in the explorer because that doesnt make sense to do. 

If theres a more clear way to do this I would be happy to change it. I know the model.vue page has a lot going on

shawnyama: Ya that's what the flag was meant for, perhaps a name change like `isInProject` would be better?

YohannParis: Okay, yes indeed, a proper name is needed.

shawnyama: This renaming is quite the chore in itself as this prop is used everywhere. I want to change how this prop is fed into the components as well for easier dev use. I'll merge this and do the rename in another PR.

Issue #1242, opened 2023-06-19
[TASK]: Document page PDF viewer needs header so we can switch back to the Extractions tab


jamiewaese-uncharted:
The Document viewer has a header with a content switcher that lets you switch between the Extractions and PDF view

![Monosnap Terarium 2023-06-19 12-07-00](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/5892c6f2-8573-4d9c-804d-f4ff12f78de1)

Unfortunately, when you click on the PDF button to see the PDF view, this header disappears. So you can't click back.

![Monosnap Terarium 2023-06-19 12-08-24](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/4c2d052d-b896-4149-8199-d6de375c30d7)

When we switch to the PDF view, we should insert this "sticky header" version above the Adobe Accrobat toolbar. It shouldn't have the glassy semi-opaque background though, since nothing scrolls beneath it. 
![Monosnap Terarium 2023-06-19 12-09-55](https://github.com/DARPA-ASKEM/Terarium/assets/120480244/ec220146-a81e-4162-b68a-268c39c87917)

PR #1316, opened 2023-07-05
fix: improve documents header visibility

blanchco:
# Description
1. We needed the ability to navigate between the extractions and PDF tabs seamlessly, this fix allows the header to be visible while within the PDF view so that you can navigate between them.
2. We have also moved the green info box just below the the header tabs as requested for an improved design

Project View

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/56d6c1b0-fb74-405d-9b8a-1b644d0f70a9

Explorer View

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/2d4bbf87-41f9-4aa5-8a32-a0a5422d7d34

Resolves #(issue)
https://github.com/DARPA-ASKEM/Terarium/issues/1242


packages/client/hmi-client/src/components/documents/tera-document.vue
```diff
@@ -10,6 +10,7 @@
 		@close-preview="emit('close-preview')"
 		:hide-header="documentView === DocumentView.PDF"
 		:stretch-content="documentView === DocumentView.PDF"
+		:show-sticky-header="documentView === DocumentView.PDF"
```
YohannParis, line 13: You could have removed the `:hide-header` prop that was causing your issue.

packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -13,14 +13,17 @@
...
+			<header id="asset-top" ref="headerRef">
 				<section>
 					<!-- put the buttons above the title if there is an overline -->
 					<div v-if="overline" class="vertically-center">
 						<span class="overline">{{ overline }}</span>
 						<slot name="edit-buttons" />
 					</div>
+					<div>
+						<slot name="info-bar" />
+					</div>
```
YohannParis, line 26: Does it need to be inside a `<div>`?

blanchco: @YohannParis  you're right this isn't needed removed

YohannParis (approved):

packages/client/hmi-client/src/components/documents/tera-document.vue
blanchco: @YohannParis There's 2 notions of a header on the tera-asset component,
This is what the hide-header "hides" 
Screenshot 2023-07-06 at 8 58 23 AM

The below is the sticky header which shows when you scroll down and has slightly different layout
Screenshot 2023-07-06 at 8 58 59 AM

For viewing the PDF page we just want the sticky header to be visible at all times rather than the entire header section

YohannParis: damn, then could you rename `hide-header` to `hide-intro`? to be clearer

blanchco: @YohannParis good idea, this has been addressed!

YohannParis (approved):

PR #1319, opened 2023-07-06
feat: add skema amr endpoints

YohannParis:
# Description

Add Skema AMR endpoints to do:
* MathML -> AMR ----> skema-rs with [documentation](https://skema-rs.staging.terarium.ai/docs/#/skema%3A%3Aservices%3A%3Amathml/get_amr) 

Okay, so the model page needs a whole refactoring, and I think this needs to be done on its own PR that I will be starting.
This is merely the back-end with a basic trial to render it accessible to the front-end.


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TransformResource.java
```diff
@@ -32,6 +37,42 @@ public Response mathML2ACSet(List<String> list) {
...
+			// https://skema-rs.terarium.ai/docs/#/skema%3A%3Aservices%3A%3Amathml/get_amr
+			ObjectMapper mapper = new ObjectMapper();
+			ObjectNode request = mapper.createObjectNode();
+			request.put("model", framework);
+			request.put("mathml", mapper.valueToTree(mathml));
+			Response response = skemaProxy.convertMathML2AMR(request);
+
+			if (response.getStatus() != Response.Status.OK.ordinal()) {
+				Response.status(Response.Status.BAD_REQUEST).build();
+			}
```
mwdchang, line 67: If you did the input validation above, then this is arguably a sever error and not a Bad-Request.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TransformResource.java
```diff
@@ -32,6 +37,42 @@ public Response mathML2ACSet(List<String> list) {
...
+	@POST
+	@Path("/mathml-to-amr/{framework}")
+	@APIResponses({
+		@APIResponse(responseCode = "500", description = "An error occurred retrieving the AMR"),
+		@APIResponse(responseCode = "400", description = "Query must contain a valid MathML and a Model framework")
+	})
+	public Response mathML2AMR(
+		final List<String> mathml,
+		@PathParam("framework") final String framework
+	) {
```
mwdchang, line 49: It is a little bit weird that parts of the payload is on PathParam and part of the payload is in the POST-body, is there a particular reason why we want to structure the API like this? Why not just dump everything into the POST-body?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TransformResource.java
YohannParis: I agree, but for now, because we keep sending bad MathML to the service, and they have poor error handling, I decided to say it was a bad request and precised in the error message above that the MathML needs to be valid and mathematically sound.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TransformResource.java
YohannParis: That's what I had originally, this way it's a bit easier to duplicate the endpoints. But I guess it doesn't matter.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/skema/SkemaRustProxy.java
```diff
@@ -1,5 +1,8 @@
 package software.uncharted.terarium.hmiserver.proxies.skema;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
```
dvince2, line 5: Are these imports being used?

PR #1327, opened 2023-07-07
fix: separate model diagram and model equations

Jami159:
# Description

* splits up the model page's model diagram and model equation components into 2 separate accordion tabs instead of keeping them side by side
* Video demo: https://capture.dropbox.com/lyE3NspRLCKKmtDR

### DIFF
--- a/packages/client/hmi-client/src/components/models/tera-model-diagram.vue
+++ b/packages/client/hmi-client/src/components/models/tera-model-diagram.vue
@@ -1,87 +1,77 @@
 <template>
 	<main>
-		<TeraResizablePanel v-if="!nodePreview">
-			<div ref="splitterContainer" class="splitter-container">
-				<Splitter :gutterSize="5" :layout="layout">
-					<SplitterPanel
-						class="tera-split-panel"
-						:size="equationPanelSize"
-						:minSize="equationPanelMinSize"
-						:maxSize="equationPanelMaxSize"
-					>
-						<section class="graph-element">
-							<Toolbar>
-								<template #start>
+		<Accordion v-if="!nodePreview" multiple :activeIndex="[0, 1]" class="accordion-group">
+			<AccordionTab header="Model diagram">
+				<TeraResizablePanel class="diagram-container">
+					<section class="graph-element">
+						<Toolbar>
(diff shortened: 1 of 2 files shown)


YohannParis:
You should add a screenshot or video of your work on the PR.

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -1,87 +1,77 @@
...
-						:size="equationPanelSize"
-						:minSize="equationPanelMinSize"
-						:maxSize="equationPanelMaxSize"
-					>
-						<section class="graph-element">
-							<Toolbar>
-								<template #start>
+		<Accordion v-if="!nodePreview" multiple :activeIndex="[0, 1]" class="accordion-group">
+			<AccordionTab header="Model diagram">
+				<TeraResizablePanel class="diagram-container">
```
YohannParis, line 5: Do you need those `<TeraResizablePanel>`?

Jami159:
> You should add a screenshot or video of your work on the PR.

Just added a video demo. Wasn't able to show the model in "Workspace" because that's currently not working for me.

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
Jami159: Yes, we still want to be able to change the size of the diagram and equations containers

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -463,11 +445,19 @@ const addTransition = async () => {
 
 <style scoped>
 main {
-	border: 1px solid var(--surface-border-light);
-	border-radius: var(--border-radius);
 	overflow: auto;
 }
 
+.accordion-group {
```
shawnyama, line 451: I think you can just do `.p-accordion` the component has this class by default

Jami159: Thanks for the suggestion, just updated!

YohannParis (approved):

PR #1353, opened 2023-07-11
fix: ensure we can specify a limit for maximum number of rows in a csv download

chris-dickson:
This fixes the timeouts (and Java OOMs) when downloading csvs.  This ensures by default the hmi-client only gets the first 10 lines.  We can specify more in the `limit` query parameter on the resource. Pass in a value of -1 for no limit


dvince2 (approved):
Thanks!

PR #1357, opened 2023-07-11
feat: add ciemss simulate workflow node

### LINKED ISSUE #1268: [FEAT]: Add CIEMSS Simulate component
- [x] Modify Workflow menu (from toolbar and right-click) to distinguish Deterministic and Probabilistic components
image
- [x] Modify the header of the Simulate box/panel to differentiate probabilistic from deterministic
image
The input tab for Simulate (probabilistic) is the same as the existing deterministic version except for:
image
- [x] 'Add Uncertainty' toggle
- [ ] Other options section at the bottom with 'Number of stochastic samples (default to 100) and Method (default to what is in mockup)
The output tab is the same as the existing deterministic version, except that:
image
- [x] The chart can only show one variable at a time and one chart plots all samples for that variable in grey.
- [x] We need to compute the mean and display it on top as a thicker dark green line

mj3cheun:
# Description

Screenshot 2023-07-11 at 11 22 59 AM

Resolves #1268


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -91,4 +99,31 @@ public Simulation makeCalibrateJob(
...
+	) {
+		final JobResponse res = simulationServiceProxy.makeCalibrateJob(Converter.convertObjectToSnakeCaseJsonNode(request));
+
+		Simulation sim = new Simulation();
+		sim.setId(res.getSimulationId());
+		sim.setType("simulation");
+
+		sim.setExecutionPayload(request);
+
+		// // FIXME: These fiels are arguable unnecessary
```
dvince2, line 119: If they're not needed do we want to remove these, or should they stick around?

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
```diff
@@ -363,6 +371,7 @@ function appendOutputPort(node: WorkflowNode, port: { type: string; label?: stri
 
 // Run testOperation
 const testNode = (node: WorkflowNode) => {
+	console.log(node);
```
dvince2, line 374: debug print

dvince2 (approved):
There are a lot of debug/fixme/etc comments that could be cleaned up... but also we have a lot of cleanup to do after the hackathon

packages/client/hmi-client/src/components/workflow/simulate-ciemss-operation.ts
```diff
@@ -0,0 +1,30 @@
+import { TimeSpan } from '@/types/Types';
+import { Operation, WorkflowOperationTypes } from '@/types/workflow';
+import { ChartConfig } from '@/types/SimulateConfig';
+
+export interface SimulateOperationState {
+	chartConfigs: ChartConfig[];
+	currentTimespan: TimeSpan;
+}
```
mwdchang, line 8: I think we need to rename this, cause the same named interface is being exported and it might make things a tad confusing.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
mj3cheun: im not sure what those fields do, just copied and pasted them from the other endpoints. i hear @YohannParis might know a bit more what they do and if they should be removed or not

packages/client/hmi-client/src/components/workflow/tera-simulation-workflow.vue
mj3cheun: thanks removed.

packages/client/hmi-client/src/components/workflow/simulate-ciemss-operation.ts
mj3cheun: this type is a duplicate of the julia version, do we want to use the same type for both or have one type for ciemss and julia each?

mj3cheun: im going to have one type for each, we can change it down the line if we want

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
mwdchang: TDS should really handle this portion, but there is some confusion with with @YohannParis want terarium to do??
We will sort it out after the hackathon.

mwdchang:
Per discussion.
- Let's remove the modal and editing stuff from tera-model-configuation-ciemss
- Let's rename tera-model-configuration-ciemss => tera-model-configuration-distribution

PR #1362, opened 2023-07-11
fix: resource panel refreshes properly when assets are being added/removed

shawnyama:
# Description

Updating resource store within the project service whenever the project in postgres is modified. I also refetch the project once a resource is uploaded through the overview panel. `project` in App.vue is updated to match the resource store whenever resource store is updated. 

Additionally an asset doesn't open when you want to remove it.

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/e09703b9-dd86-43c9-bed2-ba11f5111199


YohannParis (commented):
Why all the FIXME, could they be fixed on this PR.

shawnyama:
> Why all the FIXME, could they be fixed on this PR.

These FIXMEs are not a product of the changes I made here, they have always been a problem. I am just highlighting them.

YohannParis (approved):

PR #1365, opened 2023-07-11
feat: added 2 more options to the workflow menu

jamiewaese-uncharted:
Added these:
image


Tom-Szendrey:
Can we leave this out right now?
I am working on a branch to add the Simulate Ensemble option but it involves a few more added files ect to make it a valid option 

If having these as disabled options helps the short term (today) then no complaints

jamiewaese-uncharted:
Also adjusted the menu so the options don't line wrap:

image

YohannParis (approved):

Tom-Szendrey (approved):

Issue #1220, opened 2023-06-13
[TASK]: Second pass at 'Search by Example'


pascaleproulx:
**Describe the task**
The 'Search by Example' feature is incomplete. Now that we have more control with access to ES, let's do a second pass to remove some rough edges.

**Task Items**
- [x] Update the search bar to reflect which search was issued e.g. Similar content to: [name of resource]
- [x] Update the 'Showing...' message to reflect which search was issued. e.g. Showing xxx results for similar content to: [name of resources]
- [ ] Get bigger set of results and populate the facets like any other search

**Current state**
- search context is from previous search, not the search by example one
- no facets
- only 10 results
<img width="1792" alt="image" src="https://github.com/DARPA-ASKEM/Terarium/assets/42645294/1a84544b-b595-41e5-9a39-ef58fa0ed9ae">

PR #1367, opened 2023-07-11
Search by example improvements

Jami159:
# Description

* updates the search bar to reflect which search was issued e.g. Similar content to: [name of resource]
* updates the 'Showing...' message to reflect which search was issued. e.g. Showing xxx results for similar content to: [name of resources]
* Video demo: https://capture.dropbox.com/d1jf0MmThQ6Wr6qv

Resolves the first 2 asks for #1220


packages/client/hmi-client/src/page/data-explorer/search-by-example.ts
```diff
@@ -8,6 +8,15 @@ const searchByExampleOptions = ref<SearchByExampleOptions>({
...
+	return { searchByExampleOptions, searchByExampleItem, searchByExampleAssetCardProp };
+}
+
+export function extractResourceName(resource): string {
+	if (resource.name) {
+		return resource.name;
+	}
+
+	return resource.title;
 }
```
YohannParis, line 22:
```suggestion
export function extractResourceName(resource): string {
	return resource?.name ?? resource.title;
}
```

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -145,13 +156,39 @@ const resultsCount = computed(() => {
...
+					s += ', ';
+				}
+				s += optionBoolstoStrs[key];
+			}
+		});
+	}
+
+	return s;
+});
+
```
YohannParis, line 183: If you are making a string that contains the same value every time, why not just a string literal?

Issue #1351, opened 2023-07-11
[BUG]: Graph default zoom is not correct


mwdchang:
The default zoom should try to encompass as much of the graph as you can, right now this is a bit messed up.

E.g. the graph's bottom portion is cut off

<img width="1133" alt="image" src="https://github.com/DARPA-ASKEM/Terarium/assets/1220927/4f7715dd-112e-493e-9ddf-b900f15c1823">

The code to handle this is  under `client/graph-scaffolder/src/core/renderer.ts`. Talk to @mwdchang if there are any questions.

```
	setToDefaultZoom() {
		const svg = d3.select(this.svgEl);
		let scale = 1;
		if (this.chart) {
			scale = this.chartSize.width / (this.graph.width ?? 1);
		}

		svg.call(
			this.zoom?.transform as any,
			d3.zoomIdentity.translate(1, 100).scale(scale).translate(0, 0)
		);
	}
```

PR #1369, opened 2023-07-11
fix: graph zoom and showing transition equations

blanchco:
# Description
- fix graph zoom so that it now centers and showing the entire model diagram
- adding equations above the transition node for better visibility

Showing Transition Nodes

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/866aa412-0c71-4b2d-8719-72bc06405c5e

Showing zoom fix

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/0ca04f23-5c48-4786-afac-c4a471df1ef7

Resolves #(issue)
https://github.com/DARPA-ASKEM/Terarium/issues/1351

### DIFF
--- a/packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
+++ b/packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
@@ -10,6 +10,7 @@ import * as petrinetService from '@/model-representation/petrinet/petrinet-servi
 export interface NodeData {
 	type: string;
 	strataType?: string;
+	expression?: string;
 }
 
 export interface EdgeData {
@@ -152,7 +153,7 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 			.on('mouseover', handleMouseOver)
 			.on('mouseout', handleMouseOut);
 
-		// transitions text
+		// transitions label text
 		transitions
 			.append('text')
 			.attr('y', () => 5)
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
```diff
@@ -167,6 +168,20 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
...
+			.style('text-anchor', 'middle')
+			.style('paint-order', 'stroke')
+			.style('stroke', (d) =>
+				d.data.strataType ? strataTypeColors[strataTypes.indexOf(d.data.strataType)] : '#FFF'
+			)
+			.style('stroke-width', '3px')
+			.style('stroke-linecap', 'butt')
+			.style('fill', 'var(--text-color-primary')
+			.style('pointer-events', 'none')
+			.html((d) => d.data.expression ?? null);
```
mwdchang, line 184: - Why do you care about strataTypes?

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -94,16 +95,19 @@ export const convertToIGraph = (amr: Model) => {
 		const typeMap = amr.semantics?.typing?.type_map.find(
 			(map) => map.length === 2 && transition.id === map[0]
 		);
+
+		const targetNode = amr.semantics?.ode.rates.find((node) => transition.id === node.target);
+
```
mwdchang, line 100: You are not trying to find a "node", would be clearer to call it `targetRate`.

packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
blanchco: @mwdchang copy pasted this from the previous label text, its not needed.  I've removed this and given the default which was '#FFF'

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
blanchco: @mwdchang makes sense - i've renamed to 'targetRate' and 'rate' respectively

mwdchang (approved):

PR #1371, opened 2023-07-11
feat: allow for non-local websocket configuration

chris-dickson:
This does lazy loading of the jupyter session and allows for a paramterized configuration of chatt/TGPT

### DIFF
--- a/packages/client/hmi-client/src/services/jupyter.ts
+++ b/packages/client/hmi-client/src/services/jupyter.ts
@@ -14,7 +14,6 @@ import { JSONObject } from '@lumino/coreutils';
 import * as messages from '@jupyterlab/services/lib/kernel/messages';
 import * as kernel from '@jupyterlab/services/lib/kernel/kernel';
 import { KernelConnection as JupyterKernelConnection } from '@jupyterlab/services/lib/kernel';
-import { onMounted } from 'vue';
 import API from '@/api/api';
 
 declare module '@jupyterlab/services/lib/kernel/messages' {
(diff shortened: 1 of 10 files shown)


mattprintz (approved):
Looks good

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TGPTResource.java
```diff
@@ -0,0 +1,44 @@
...
+	@ConfigProperty(name = "tgpt.baseUrl", defaultValue = "/chatty/")
+	String baseUrl;
+
+	@ConfigProperty(name = "tgpt.appUrl", defaultValue = "http://localhost:8078/chatty/")
+	String appUrl;
+
+	@ConfigProperty(name = "tgpt.wsUrl", defaultValue = "ws://localhost:8078/chatty_ws/")
+	String wsUrl;
+
+	@ConfigProperty(name = "tgpt.token", defaultValue = "89f73481102c46c0bc13b2998f9a4fce")
```
dvince2, line 30: Should this be in the clear?

packages/client/hmi-client/src/services/jupyter.ts
```diff
@@ -121,28 +122,40 @@ export interface IJupyterMessage<T extends JupyterMessageType = JupyterMessageTy
 export declare type JupyterMessage = IJupyterMessage | messages.Message;
 
 // TODO: These settings should be pulled from the environment variables or appropriate config setup.
```
dvince2, line 124: I think this TODO is now resolved

dvince2 (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TGPTResource.java
chris-dickson: This is only for local dev.  In production we will have a secret keypair.

PR #1373, opened 2023-07-11
feature: tgpt implementation

mattprintz:
# Description

squashed set of changes for TGPT implementation

Resolves #(issue)


dvince2 (approved):

docker-compose.yaml
```diff
@@ -44,21 +49,51 @@ services:
     image: ghcr.io/darpa-askem/data-service:latest
     environment:
       DATASET_STORAGE_BASE_URL: s3://datasets/
+      STORAGE_HOST: http://minio:9000
```
dvince2, line 52: minio should be gone now with everything going through s3

docker-compose.yaml
```diff
@@ -44,21 +49,51 @@ services:
...
+      S3_BUCKET: askem
+      S3_DATASET_PATH: datasets
+      S3_RESULT_PATH: simulations
```
dvince2, line 55: Will now also need `S3_ARTIFACT_PATH: artifacts`

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TGPTResource.java
```diff
@@ -0,0 +1,44 @@
...
+	@ConfigProperty(name = "tgpt.baseUrl", defaultValue = "/chatty/")
+	String baseUrl;
+
+	@ConfigProperty(name = "tgpt.appUrl", defaultValue = "http://localhost:8078/chatty/")
+	String appUrl;
+
+	@ConfigProperty(name = "tgpt.wsUrl", defaultValue = "ws://localhost:8078/chatty_ws/")
+	String wsUrl;
+
+	@ConfigProperty(name = "tgpt.token", defaultValue = "89f73481102c46c0bc13b2998f9a4fce")
```
dvince2, line 30: Should this token be in the clear?

dvince2 (changes requested):
Minio should be gone now as far as i know :o

docker-compose.yaml
chris-dickson: This is for matt's docker-compose.  It's fine.

docker-compose.yaml
chris-dickson: Don't worry about the docker-compose

docker-compose.yaml
mattprintz: orchestration uses s3, but docker-compose doesn't. I don't have your S3 keys.
For now, at least, keeping minio in the docker compose.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/TGPTResource.java
chris-dickson: It's fine.  For local dev only.  There will a secret key pair in production that's encrypted.

docker-compose.yaml
mattprintz: Probably true, but not worrying about it for this commit. Can add that later once this is done and we are doing something with artifacts via docker compose.

dvince2 (approved):

PR #1385, opened 2023-07-12
feat: stratify drilldown step 2 - type base model

echl:


packages/client/hmi-client/src/utils/color-schemes.ts
```diff
@@ -13,3 +14,33 @@ export const strataTypeColors = [
...
+	if (!id) {
+		return;
+	}
+	nodeTypeColorMap.value[id] = nodeTypeColors[count.value];
+	if (count.value >= nodeTypeColors.length) {
+		count.value = 0;
+	} else {
+		count.value++;
+	}
+}
```
mwdchang, line 42: I don't know why you want to have reactive stuff here,  but it seems like you are trying to create a categorical scale here. I think there is an easier way.

```
const scale = d3.scaleOrdinal(nodeTypeColors).domain(['a', 'b', 'c', 'd', 'e'])

scale('a') // '#a6cee3'
scale('c') // '#b2df8a'
scale('d') // '#33a02c'

```

packages/client/hmi-client/src/components/models/tera-strata-model-diagram.vue
```diff
@@ -119,16 +118,18 @@ const emit = defineEmits([
...
 	model: Model;
 	showTypingToolbar: boolean;
 	typeSystem?: TypeSystem;
+	showReflexivesToolbar: boolean;
 }>();
 
-const typedModel = ref<Model>();
+const typedModel = ref<Model>(props.model);
 
 const newModelName = ref('New Model');
```
mwdchang, line 134: Unrelated, but since you are in the area, it doesn't seem like newModelName is used, can you remove?

mwdchang:
Looks good, I don't completely follow some of the inference logic so may be we should do a code walkthrough some time later to spread out the knowledge a bit and maybe catch a few edge cases that are missed.

Left a comment about the color scale thing, let me know if you want me to elaborate on that.

mwdchang (approved):

PR #1389, opened 2023-07-12
fix: issue where model transitions expression where not showing

YohannParis:
# Description

Resolves #(issue)


packages/client/hmi-client/src/types/Types.ts
```diff
@@ -339,7 +339,7 @@ export interface XDDFacetBucket {
 export interface Rate {
     target: string;
     expression: string;
-    expression_mathml: string;
+    expression_mathml?: string;
```
mwdchang, line 342: Why are we making expression_mathml optional? 
If you have expression then you need to have a corresponding expression_mathml.

YohannParis: Because TA1 is sometimes giving me AMR with no `expression_mathml` and I need this to get working.

I agree with you we should enforce the AMR. But hackathon it is!

PR #1391, opened 2023-07-12
feat: increase search result list and add facets for Search by Example

### LINKED ISSUE #1366: [TASK]: "Search by Example" - get bigger set of results and populate the facets like any other search
Describe the task
The 'Search by Example' feature is incomplete. Now that we have more control with access to ES, let's do a second pass to remove some rough edges.

@iross has given us this sample URL: https://xdddev.chtc.io/api/v2/articles?docid=5e7e0724998e17af8269b284&similar_docs=true

Swagger docs: https://xdd.wisc.edu/api-docs/

**Task Items**
- [ ] In `application.properties` you can find the name of the xDD developer instance.  Change `DocumentProxy.java` to use this developer instance instead of the production resource.
- [ ] In `DocumentResource.java`, add a new `@QueryParam` to the `getDocuments` method for the new `similar_docs` query parameter. You will also need to modify the interface `DocumentProxy.java`
- [ ] In `getDocuments` add a bit of safety code that requires a docid to be set (and not a term, doi or github_url). Consider returning an error if the search does contain those.

If you click on the link above, you'll see the raw JSON returned. Inside of `success>data>0` is a document object which in our back end is represented by `Document.java`. 

- [ ] Create a new java class called SimilarDocuments.java in the same directory as Document.java.
- [ ] Add two fields to this list - one should be a list of `Document`s annotated with `@JsonAlias(bibjsons)`, and the other should be a map of facets like `private Map<String, XDDFacetsItemResponse> facets;`
- [ ] Now on `Document.java` add a simlarDocument field (annotated with `@JsonAlias(simlar_docs)`).
- [ ] Now, modify our front end to not make a call to the related documents end point, but instead, it should call into the getDocuments endpoint.

Then... hook up facets :)

**Current State**
- only 10 results

**Note**
This is a follow up task from task #1220

Jami159:
# Description

* Add a  new "similar_to" query param to the `/documents` route
* Replace the existing endpoint call to get similar docs with the new endpoint call to `/documents?similar_to=<id>`
* Search by Example now shows a list of 100 results and also populates the facets sidebar
* Facets filtering works correctly for documents search by example
* Video demo: https://capture.dropbox.com/t75H8hoz9XQlJspJ

Resolves #1366


dvince2 (commented):
This looks good so far. Waiting on giving it a ✅ until we have the new facet filtering from xdd

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/documentservice/DocumentResource.java
```diff
@@ -60,11 +58,12 @@ public Response getDocuments(
...
 		@QueryParam("match") String match,
 		@QueryParam("known_entities") String known_entities,
-		@QueryParam("github_url") String github_url
+		@QueryParam("github_url") String github_url,
+		@QueryParam("similar_to") String similar_to
 	) {
 
 		// only go ahead with the query if at least one param is present
-		if (docid != null || doi != null || term != null || github_url != null) {
+		if (docid != null || doi != null || term != null || github_url != null || similar_to != null) {
```
dvince2, line 66: Not for this PR but I have some strong feelings on how to cleanup and refactor this resource class as part of the next PR.  If you're doing back end work here let me know!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/documentservice/DocumentProxy.java
```diff
@@ -9,7 +9,7 @@
 import javax.ws.rs.*;
 import javax.ws.rs.core.MediaType;
 
-@RegisterRestClient(configKey = "xdd-document-service")
+@RegisterRestClient(configKey = "xdd-dev-service")
```
dvince2, line 12: Once this PR is merged you should post in the askem-hmi channel advising people that we're temporarily moving back to the dev instance of xdd and that search results may change as a result. You can also let people know that on their dev instance they've made performance improvements to the unified endpoint so they'll see that too

packages/client/hmi-client/src/services/data.ts
```diff
@@ -568,17 +575,18 @@ const fetchData = async (
...
-					};
-					finalResponse.allData.push(similarDocumentsSearchResults);
-					finalResponse.allDataFilteredWithFacets.push(similarDocumentsSearchResults);
+				if (searchParam?.xdd.similar_search_enabled && searchParam?.xdd.related_search_id) {
+					const response = await fetchResource('', resourceType as ResourceType, {
+						...searchParamWithFacetFilters,
+						xdd: {
+							...searchParamWithFacetFilters?.xdd,
+							similar_to: searchParam.xdd.related_search_id as string,
+							max: 100,
```
dvince2, line 584: Is this returning the search results to 100? I believe we had reduced the amount to 20 for performance reasons unless you're finding the unified endpoint now fast enough with Matt's changes

dvince2 (approved):
Looks good! just confirm the search count or performance and we're good to go

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/documentservice/DocumentProxy.java
Jami159: Will do :)

packages/client/hmi-client/src/services/data.ts
Jami159: Sounds good!

PR #1396, opened 2023-07-12
Preview pdf artifacts

dvince2:
# Description

Updating our PDF viewer to accept arraybuffers in addition to URLs. With this we can grab the bytes from s3 and show our PDFs that users have uploaded.

Resolves #noissue


packages/services/hmi-server/src/main/resources/application-secrets.properties.encrypted
dvince2, line 1: This is an intentional commit updating the aws key

dgauldie (approved):
LGTM

PR #1401, opened 2023-07-13
feat: stratify step 3

echl:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/914536c0-276c-44cf-b0e6-c0cedcb75c04

- Allow user to add reflexives to both base and strata model

Remaining work for step 3:
 - Add appropriate rates and values to generated strata model and to new transitions added during step 3.


YohannParis (approved):

mwdchang:
The graph is a bit jump, I think we can wrap the renderer init in tera-strat-model-diagram with

```
	if (!renderer) {
		// Create renderer
		renderer = new PetrinetRenderer({
			el: graphElement.value as HTMLDivElement,
			useAStarRouting: false,
			useStableZoomPan: true,
			runLayout: runDagreLayout,
			dragSelector: 'no-drag'
		});
		renderer.on('add-edge', (_evtName, _evt, _selection, d) => {
			renderer?.addEdge(d.source, d.target);
		});
	}
```

PR #1412, opened 2023-07-13
fix: added extension check to file-type check

dgauldie:
# Description

Resolves #(issue)


dvince2 (approved):
Thanks!

PR #1414, opened 2023-07-13
Allowing source code upload to artifacts

dvince2:
# Description

* Allowing source code upload to artifacts.
* This will be part of a larger task that also allows users to create and modify code directly in the hmi... but for now its all just coming from disk.

Resolves #noissue


YohannParis (approved):
Why not add `.jl` while at it?

YohannParis (approved):

PR #1429, opened 2023-07-14
stable zoom/pan, clean up

mwdchang:
### Summary
Fix graph's pan/zoom resetting after each graph colouring command. Minor clean up of unused styles.


echl (approved):
Thanks for the fix, I added it to `tera-strata-model-diagram` as well

PR #1434, opened 2023-07-14
Jamie english labels for workflow node headers

jamiewaese-uncharted:
The workflow node headers didn't match up with the associated drilldown panel headers. This is a quick fix, and easy enough to update as additional items get added.

BEFORE
image

AFTER
Monosnap Terarium 2023-07-14 10-29-16


packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
YohannParis, line 1: @mwdchang I'll let you decide, but I think it's a legit solution

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow-node.vue
YohannParis: @Tom-Szendrey is your latest PR fixing this? If yes, could you close it please?

PR #1436, opened 2023-07-14
Jamie subtle transition on mouseover of workflow nodes

jamiewaese-uncharted:
I added a 75ms transition effect to the box-shadow and node header when hovering over the workflow nodes.

For good performance, I created an ::after element with the box-shadow and then animate the opacity of that instead of animating the box-shadow properties directly. If you're interested, I found the technique here: [https://tobiasahlin.com/blog/how-to-animate-box-shadow/](https://tobiasahlin.com/blog/how-to-animate-box-shadow).

BEFORE
workflow nodes shadow no transition

AFTER
It's subtle but it feels more "complete".
workflow nodes shadow transition


YohannParis (approved):
Nice classic technic

PR #1455, opened 2023-07-14
Cole adding timeseries variables to model configuration

blanchco:
# Description

* Ability to switch between variable types
* Added time series variables
* Time series cells are uneditable from the table view
* Time series variable stored in `configuration.metadata.timeseries` (for now), as there is no other way to store this in the current AMR structure
Each key in the timeseries object references a parameter id for the corresponding model configuration
* Unsure, if we will need to hook this up down the line in regards for simulating

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/6fcaf021-b94d-4933-887b-1994e5f6074b


blanchco:
closing for now...might want to discuss how the metadata is being read

YohannParis (approved):

PR #1458, opened 2023-07-14
feat: add renderer for nested petrinets

mj3cheun:
# Description

* adds renderer for nested petrinets

Note: heres a code snipping demonstrating its use with rendering the amr
```
		if (props.model === null || graphElement.value === null) return;
		const graphData: IGraph<NodeData, EdgeData> = convertToIGraph(props.model.semantics?.span?.[0].system);
		const nestedMap = props.model.semantics?.span?.[0].map.reduce((childMap, [stratNode, baseNode]) => {
			if(!childMap[baseNode]) {
				childMap[baseNode] = [];
			}
			childMap[baseNode].push(stratNode);
			return childMap;
		}, {})

		// Create renderer
		renderer = new NestedPetrinetRenderer({
			el: graphElement.value as HTMLDivElement,
			useAStarRouting: false,
			useStableZoomPan: true,
			runLayout: runDagreLayout,
			dragSelector: 'no-drag',
			nestedMap
		});
```


packages/client/graph-scaffolder/src/core/renderer.ts
```diff
@@ -41,7 +41,7 @@ export const pathFn = d3
 export abstract class Renderer<V, E> extends EventEmitter {
 	options: Options;
 
-	parentMap: Map<string, INode<V>>;
```
mj3cheun, line 44: @mwdchang is this change fine?

mwdchang: `parentMap` is used for grouping nodes together - what problem is this causing?

packages/client/hmi-client/src/model-representation/petrinet/nested-petrinet-renderer.ts
```diff
@@ -0,0 +1,136 @@
...
+	// This is getting around algorithms that do not provide stand-alone routing capabilities, in
+	// which case we can internally route using A-star
+	useAStarRouting?: boolean;
+
+	// Whether to show grid
+	useGrid?: boolean;
+
+	// The css class for custom dragging event
+	dragSelector?: string;
+}
```
mwdchang, line 33: Slightly prefer if we can just extend Options interface .... or maybe something like a setNestedMap function.

We can export the Options interface in graph-scaffolder if that makes things easier

YohannParis (approved):

PR #1478, opened 2023-07-17
fix: reroute to url tab

blanchco:
# Description

* Route to the proper tab on refresh rather than it always going to the overview tab

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/ee12e02e-ab50-490f-a837-18353c62ba43

Resolves #(issue)


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -235,6 +228,16 @@ watch(
 		} else if (openedAssetRoute.value && openedAssetRoute.value.assetName) {
 			tabStore.addTab(projectContext.value, openedAssetRoute.value);
 		}
+
+		const overviewResource = {
+			assetName: 'Overview',
+			pageType: ProjectPages.OVERVIEW,
+			assetId: ''
```
YohannParis, line 235: I think this needs to be `underfined`

YohannParis (approved):

packages/client/hmi-client/src/stores/tabs.ts
```diff
@@ -17,6 +19,11 @@ export const useTabStore = defineStore('tabs', {
 			// Create a new tab set for this context if there is none, otherwise add a new tab
 			if (!this.getTabs(context)) {
 				this.setTabs(context, [newTab]);
+			} else if (
+				isEqual(newTab, { assetName: 'Overview', pageType: ProjectPages.OVERVIEW, assetId: '' })
```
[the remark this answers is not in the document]
blanchco: @YohannParis made this a simpler logical expression here - isEqual was not needed

packages/client/hmi-client/src/page/project/tera-project.vue
blanchco: @YohannParis unfortunately undefined causes mutliple 'Overview' tabs to be open when added with the addTabs function.  This worked with the router - however if they are being added through .addTabs it needs to be an empty string to passes the if on L249

YohannParis (approved):

PR #1484, opened 2023-07-17
feat: ensemble-simulate for ciemss

Tom-Szendrey:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/6e464e64-2867-4282-9b1f-c4395388d436


Tom-Szendrey:
@YohannParis 
I do not see /{id}/add-result-as-dataset-to-project/{projectId}
on my local

Is there a way to set the filename? For simulate (ensemble or otherwise) I am getting simulation.csv not result.csv
Will this break it?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -151,5 +152,17 @@ public JobResponse makeCalibrateJobCiemss(
 		return res;
 	}
 
+	@POST
+	@Path("ciemss/ensemble-simulate")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Consumes(MediaType.APPLICATION_JSON)
```
YohannParis, line 158: I don't think you need those annotations if they are on the class itself

YohannParis (approved):

PR #1499, opened 2023-07-17
feat: make model parameters editable

Jami159:
# Description

* Make the model parameters editable
* Update rate expression and expression_ml if parameter id is updated
* Video demo: https://capture.dropbox.com/d7XKtPi6J5KMfIwF


blanchco:
Testing:
1. Changing a _parameter_ ID
Propagates changes to default config, does not propagate to other configs (is this needed)?
`semantics.ode.rates.expression`, `semantics.ode.rates.expression_mathml`, `semantics.ode.parameters.id`

2. Change a _parameter_ value
Changes `semantics.ode.parameters.value` ✅ 

3. Changing a _parameter_ ID which is a substring of another _parameter_ ID
Propagates changes to default config ✅

YohannParis:
> 1. Changing a _parameter_ ID
>    Propagates changes to default config, does not propagate to other configs (is this needed)?
>    `semantics.ode.rates.expression`, `semantics.ode.rates.expression_mathml`, `semantics.ode.parameters.id`

Yes this is needed to update all configs.

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -866,20 +913,89 @@ async function updateModelName() {
...
+}
+
+function cancelEdit() {
+	isRowEditable.value = null;
+	transientTableValue.value = {};
+}
+
+function updateTable(tableType, idx, key, value) {
+	transientTableValue.value = {
+		...transientTableValue.value,
```
mwdchang, line 991: What `transientTableValue` contains is a bit confusing. Can it be statically typed? if not can you add one or two examples of what the data structure looks like?

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -866,20 +913,89 @@ async function updateModelName() {
...
+					if (key !== 'unit') {
+						// TODO: remove this condition when we proper editing of unit
+						modelClone.model.states[idx][key] = value;
+					}
+					// TODO: update all of the properties affected by state id
+				});
+				break;
+			default:
+				logger.info(`${tableType} not recognized`);
+		}
```
mwdchang, line 976: Might be worthwhile to see if the update logic (e.g. update parameter) can be moved into petrinet-service alongside other update functions. That way the logic isn't scattered across the code base.

Jami159: created some helper functions in `petrinet-service.ts`, let me know if it looks okay to you

packages/client/hmi-client/src/components/models/tera-model.vue
Jami159: added the type `ModelTableTypes` above

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -416,6 +418,114 @@ export const updateTransition = (
...
+			rate.expression = replaceValuesInExpression(rate.expression, id, newId);
+			if (rate.expression_mathml) {
+				rate.expression_mathml = replaceValuesInMathML(rate.expression_mathml, id, newId);
+			}
+		});
+
+		// if there's a timeseries field with the old parameter id then update it to the new id
+		if (amr.metadata?.timeseries) {
+			delete Object.assign(amr.metadata.timeseries, { [newId]: amr.metadata.timeseries[id] })[id];
+		}
```
mwdchang, line 514: Doesn't this introduce garbage if `timeseries` exist but`timeseries[id]` does not exist?

Would also prefer if we don't assign and delete on the same line, just to make things a bit more easier to understand.

blanchco:
Qa round 2:
* Successfully updates parameter ID for **all** model configurations:
`semantics.ode.rates.expression`, `semantics.ode.rates.expression_mathml`, ` semantics.ode.parameters.id`, `metadata.timeseries` ✅ 

* Successfully updates changes the parameter value - this does not propagate to other configuration as expected ✅ 

* Naming a parameter this is the substring of another is does not affect other parameters ✅ 

* naming a parameter name the same as another parameter will cause issues.  If this is done you will not be able to separate the ids again (perhaps we can just keep note of this) 😵‍💫 

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/54123081-e06f-46d1-aea6-06e0a6d9d07a

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
Jami159: fixed!

mwdchang (approved):

PR #1512, opened 2023-07-18
fix: bug observables equations no saving name

tangopapatime:
# Description

Fixes the observable saving problem.  You **have to** make sure your equation contains an `=` sign in order for it to be valid.  It then assigns the left hand variables to the `ID` & `Name` for the `Observable`. 

Also includes other bug fixes such as double click to edit only enabled  in model editing mode and transitions equations not showing up after changing the transition id.


tangopapatime:
[saveobs.webm](https://github.com/DARPA-ASKEM/Terarium/assets/98172806/d299a6ae-f44a-4a34-936d-d353c3721233)

YohannParis (approved):

PR #1517, opened 2023-07-18
feat: stratify step 3 output

echl:
# Description

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/a8153c87-a11b-45b1-bb2a-c8380dc028ae

* View stratified output after stratification with collapsed view + switch between collapsed/expanded view
* Refactor some components to remove unused code
* Open model tab with stratified model from stratify drilldown
* Add model configuration node to workflow from stratify drilldown

Resolves #(issue)


packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
```diff
@@ -0,0 +1,205 @@
...
+				el: graphElement.value as HTMLDivElement,
+				useAStarRouting: false,
+				useStableZoomPan: true,
+				runLayout: runDagreLayout,
+				dragSelector: 'no-drag',
+				nestedMap
+			});
+		} else {
+			renderer.isGraphDirty = true;
+		}
```
mwdchang, line 119: This else section makes no sense - renderer is null or undefined here.

packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
```diff
@@ -0,0 +1,205 @@
...
+
+		// Render graph
+		await renderer?.setData(graphData);
+		await renderer?.render();
+	},
+	{ deep: true }
```
mwdchang, line 125: Do we really need deep watchers here?  
Slightly concerned because `props.model` is not something you want to run recursive watchers on.

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -38,6 +38,7 @@ export const addNode = (
 	wf: Workflow,
 	op: Operation,
 	pos: Position,
+	state: any = {},
 	size: Size = { width: 180, height: 220 }
```
mwdchang, line 42: You can't just add a new parameter in the middle of a functional signature, it breaks other instances. e.g.

```
workflowService.addNode(wf.value, SimulateJuliaOperation, newNodePosition, {
	width: 420,
	height: 220
});

```

If you have more than one optional parameters, I would suggest you encapsulate them inside some kind of option object. Otherwise it becomes confusing as you have a situation where you need 4 arguments of different types.

packages/client/hmi-client/src/components/workflow/tera-stratify.vue
```diff
@@ -130,17 +131,52 @@
...
+					iconPos="right"
+					text
+					@click="
+						emit('open-asset', {
+							assetName: `${stratifiedModel?.name}`,
+							pageType: ProjectAssetTypes.MODELS,
+							assetId: stratifiedModel?.id
+						})
+					"
+				/>
```
mwdchang, line 163: I don't understand, are we:
- saving it as a new model? or
- opening the new model in a new tabl?

packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
echl: renderer is null or undefined in the `if` block, and has a value in the `else` block

packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
echl: This is the same for `tera-model-diagram` and all its derivatives. However, for this component we don't need to watch `props.model` at all as it shouldn't change.

packages/client/hmi-client/src/services/workflow.ts
echl: Just forgot to check other uses of `addNode`

packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
mwdchang: `tera-model-diagram` is not a good component to model after, it has a lot unnecessary baggages from a incomplete refactor so nearly all it's states, reactive elements, watchers and computed values are questionable.

packages/client/hmi-client/src/components/workflow/tera-stratify.vue
mwdchang: Then the title/label should not say `Save as` - basically this is false advertising.

packages/client/hmi-client/src/components/models/tera-stratify-output-model-diagram.vue
mwdchang: My bad, missed the fact this is in a watch-block.

mwdchang (approved):

PR #1532, opened 2023-07-18
Clean up LLM thoughts and simplify/smooth auto scrolling

mattprintz:
# Description

Clean up LLM thoughts and simplify/smooth auto scrolling


YohannParis (approved):
Damn, that was a gutting of all your nice front-end work!

PR #1533, opened 2023-07-19
feat: datasets limit increase

YohannParis:
# Description

We might have an issue or we are limited to 100 datasets at the time, and it might be because of `page_size` call to `data-service`.  It wasn't the case, but why not raise it to 1000 and call it a day.


mwdchang (changes requested):
It is generally not a good idea to return a lot of data from ElasticSearch all at once, or from any RDBMS in general.
Performance, network throughput, and rendering will all degrade. 

Is there a real need to raise the limit now? If there is no immediate need it is far better to spend sometime and do proper pagination, or through a scroll/scan API if you need high volume of data.

PR #1552, opened 2023-07-19
feat: multiple chart lines

shawnyama:
# Description

Screenshot 2023-07-19 at 5 42 54 PM


packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
```diff
@@ -23,20 +23,72 @@
...
 	data: number[];
 	label: string;
 	fill: boolean;
 	tension: number;
 };
 
+const renderedRuns = computed<RunResults>(() => {
+	if (!props.hasMeanLine) return props.runResults;
+
+	const runResult: RunResults = JSON.parse(JSON.stringify(props.runResults));
```
YohannParis, line 50: You could use `_.cloneDeep()` has lodash is already imported on the page.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
Tom-Szendrey: Very fair point will change that

PR #1558, opened 2023-07-20
refactor: replace `is-editable` instances with `feature-config` prop

shawnyama:
# Description

Replace the prop `is-editable` with a `feature-config` object that is clearer and more customizable for configuring the features we want to use depending on the context.

### DIFF
--- a/packages/client/hmi-client/src/components/asset/tera-asset.vue
+++ b/packages/client/hmi-client/src/components/asset/tera-asset.vue
@@ -73,7 +73,6 @@ defineExpose({
 	assetContainer
 });
 
-// Booleans default to false if not specified
 const props = defineProps({
 	name: {
 		type: String,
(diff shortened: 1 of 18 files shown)


mwdchang:
It is likely better to pass in a config-object rather than a boolean scalar-value. 
1. It reads a little better, `is-in-project` reads a little weird when you don't have context
2. Allows for a little more flexibility and future proofing

Let me know what you think.

packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -64,27 +64,47 @@
...
+	},
+	featureConfig: {
+		type: Object as PropType<FeatureConfig>,
+		default: { isPreview: false } as FeatureConfig
+	},
+	// Booleans default to false if not specified
+	isNamingAsset: Boolean,
+	hideIntro: Boolean,
+	showStickyHeader: Boolean,
+	stretchContent: Boolean
```
YohannParis, line 106: Those are perfect example of things that can be moved to the `FeatureConfig`

YohannParis (approved):

PR #1561, opened 2023-07-21
style: remove some !importants

shawnyama:
# Description

A pass on purging `!important`s.

I removed some - not all. Here are my findings:

- I ignored the ones in the math editor
- I left `TODOs` around secondary button classes - those should be removed and just be a part of the prime vue theme
- The `.mapping-table` classes have `!important`s but I think we might want to rethink how we display those in the first place (I believe the designs want something more detailed)
- `.navigation-menu` uses `!important` but it kinda makes sense and I am not sure if applying that as an inline style would be any better


mwdchang (approved):
Didn't test, changes looks alright.

PR #1562, opened 2023-07-24
Jamie petrinet diagram arrows fix

jamiewaese-uncharted:
Minor tweak to how the arrows at the end of links are drawn in the petrinet renderer.

**BEFORE**

Monosnap Terarium 2023-07-24 11-27-39

**AFTER**
Monosnap Terarium 2023-07-24 11-24-07

**TO DO**
I'd like to adjust the start and end point of the lines so they don't overlap the actual nodes but I don't know how to do that. I suspect it would happen here, but I'll leave that for @mwdchang:
image


mwdchang:
@jamiewaese-uncharted we have multiple renderers now, I suspect you probably want to apply the same changes to both.
- model-representation/petrinet/petrinet-renderer 
- model-representation/petrinet/nested-petrinet-renderer


In response to changing the start/end point - you can do that by artificially inflating the boundary-conditions in convertToIGraph and deflate in rendering (or vice versa). If you you want to do that in the pathFn function you probably want to convert XY to polar-coordinates to calculate the offsets.

jamiewaese-uncharted:
@mwdchang thanks for the heads up. I'll add the arrowhead color change to `nested-petrinet-renderer`.

Can I leave adjusting the start/end points of the lines to you? I can't find "convertToGraph" and I don't really understand what you mean by converting to polar coordinates.

jamiewaese-uncharted:
Update: Actually @mwdchang I can't find any arrowhead definitions in `nested-petrinet-renderer`.

YohannParis:
@mwdchang could you have a last look at it, and either merge it or put it back on draft

PR #1579, opened 2023-07-25
chore: clean save dataset used in a few nodes

Tom-Szendrey:
# Description
Replacing repeated copy pasting by throwing it into a service

Please ensure you agree with having this function in the dataset service


packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -202,6 +204,14 @@ async function createDatasetFromSimulationResult(
 	}
 }
 
+export const saveDataset = async (projectId: string, simulationId: string | undefined) => {
+	if (!simulationId) return;
+	if (await createDatasetFromSimulationResult(projectId, simulationId)) {
+		// TODO: See about getting rid of this - this refresh should preferably be within a service
```
mwdchang, line 210: TODO still apply?

Tom-Szendrey: Thanks good catch

packages/client/hmi-client/src/components/workflow/tera-simulate-ciemss.vue
```diff
@@ -212,6 +208,8 @@ const renderedRuns = ref<RunResults>({});
 const selectedCols = ref<string[]>([]);
 const paginatorRows = ref(10);
 const paginatorFirst = ref(0);
+const completedRunId = computed<string | undefined>(() => props?.node?.outputs?.[0]?.value?.[0]);
+const projectId = ref<string>(props.project.id);
```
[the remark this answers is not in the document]
Tom-Szendrey: True
I will fix that in another PR just to make it more legible because I will need to add that fix for many nodes I believe

mwdchang (approved):

PR #1580, opened 2023-07-25
Jamie hide unnecessary run buttons in workflow

jamiewaese-uncharted:
Just cleaning up the workflow nodes a bit. For instance:
- hide run buttons that are disabled 
- make help message type settings consistent

**BEFORE**
hide disabled run buttons

**AFTER**
hide disabled run buttons - fixed

help message type settings

Here are all the nodes hooked up properly. The 'Run' buttons appear as they should.
Monosnap Terarium 2023-07-25 11-34-29


packages/client/hmi-client/src/components/workflow/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div>
+	<div v-if="!disableRunButton">
```
Tom-Szendrey, line 2: Is there a reason this is on the div and not the button?

packages/client/hmi-client/src/components/workflow/tera-simulate-ensemble-node-ciemss.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div>
+	<div v-if="!disableRunButton">
```
Tom-Szendrey, line 2: Is there a reason this is on div and not button?
Im not sure why this div exists? Im not certain

packages/client/hmi-client/src/components/workflow/tera-calibrate-ensemble-node-ciemss.vue
jamiewaese-uncharted: It could go on the run button... but this works, and I'm not really sure what's staying or going

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
```diff
@@ -89,6 +89,7 @@
 			</p>
 		</section>
 		<Button
+			v-if="datasetColumnNames && modelColumnNames"
```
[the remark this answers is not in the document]
jamiewaese-uncharted: Maybe... I'm not sure how this is supposed to work. If we do as you suggest, we won't see the button until everything is configured. This shows partial progress.

packages/client/hmi-client/src/components/workflow/tera-simulate-ensemble-node-ciemss.vue
jamiewaese-uncharted: do you have something against possibly unnecessary divs?

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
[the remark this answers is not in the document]
Tom-Szendrey: Maybe im not understanding the goal here. Are we intending to hide the buttons until they are ready to be used? Thats what I gathered in the pictures above

Is the idea to hide them until the ports are connected?

Im not sure

packages/client/hmi-client/src/components/workflow/tera-simulate-ensemble-node-ciemss.vue
Tom-Szendrey: If we are cleaning things up in the area it seems to make sense to remove them
Its not a hill I will die on though

YohannParis: I have something against unnecessary div, as it get's hard to understand what's happening.

Tom-Szendrey:
I have changed from using modelColumnNames in these v-ifs to using modelConfigId and datasetId
If we are having logic to check if we have inputs on the nodes then we should use the values directly computed from the node's inputs rather than the column names that are derived later

Tom-Szendrey (approved):

Issue #1566, opened 2023-07-24
[TASK]: Side panel filter functionality


blanchco:
**Describe the task**
Extension to this PR: https://github.com/DARPA-ASKEM/Terarium/pull/1565
A few things are missing from the PR:
- The dropdown menu should also have links for "Model" and "Workflow". Possibly also "Upload resources". There are quicklink buttons on the overview panel for these, but I'm not sure how to trigger them from the side panel.
- The other thing is the Find filter. I think it should return results in a flat list, but if that's complicated we can keep the accordions.

**Task Items**
- [x] Filter side panel results
- [x] Dropdown menu should have "model" and "workflow" links

**Goal**
You should be able to filter the side panel results.
You should be able to open new model and workflows from the dropdown menu

**Screenshots**
![image](https://github.com/DARPA-ASKEM/Terarium/assets/33158416/d4dbbe8e-5673-4961-815d-dce93563df57)
![image](https://github.com/DARPA-ASKEM/Terarium/assets/33158416/14adb7de-a1f5-4563-a7d6-05533e41b70e)

PR #1583, opened 2023-07-25
task: add resource filtering and new asset buttons

blanchco:
# Description

* Added side panel filter functionality by text
* Added New Model and New Workflow to the sidepanel

**I've moved the new model and workflow up to the `tera-project.vue` component and there was no changes to the functionality.  Since the functionality to open a new model and workflow is the same between the side panel and the project overview page there needed to be an easier way to call the new asset functions from each component.  The event to open a new asset it now bubbled up from the `tera-project-overview-page` and `tera-resource-sidebar` where it is then handled in the shared `tera-project` component

Filtering:

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/07ea9756-1d51-45ef-aa07-c078854e87b7

Side panel and overview new asset buttons:

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/fe7491b5-9dd2-4c07-aa3f-b3f989462c80

Resolves https://github.com/DARPA-ASKEM/Terarium/issues/1566


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -121,6 +123,38 @@
...
+					</template>
+					<template #footer>
+						<Button @click="createNewModel">Create model</Button>
+						<Button class="p-button-secondary" @click="isNewModelModalVisible = false">
+							Cancel
+						</Button>
+					</template>
+				</tera-modal>
+			</Teleport>
+		</template>
```
YohannParis, line 157: Make this its own component, please.

packages/client/hmi-client/src/page/project/components/tera-model-modal.vue
```diff
@@ -0,0 +1,103 @@
...
+		router.push({
+			name: RouteName.ProjectRoute,
+			params: {
+				assetName: 'Model',
+				pageType: ProjectAssetTypes.MODELS,
+				assetId: modelId
+			}
+		});
+	}
+};
```
YohannParis, line 96: This should be part of Model service under a `addNewModelToProject()` has this might be used somewhere else, and will keep this component simpler.

packages/client/hmi-client/src/page/project/components/tera-model-modal.vue
```diff
@@ -0,0 +1,103 @@
...
+</script>
+
+<style>
+.modal main {
+	width: 50rem;
+}
+</style>
```
YohannParis, line 103: This should not be there, we should remove it, so that when we fixe the `tera-modal` we don't have to find all those single use code.

packages/client/hmi-client/src/page/project/components/tera-resource-sidebar.vue
```diff
@@ -11,7 +11,27 @@
...
+			<Menu ref="optionsMenu" :model="optionsMenuItems" :popup="true">
+				<!-- A way to use vue feather icons in the MenuItem component, it might be better to try to use 1 icon library for easier integration -->
+				<template #item="slotProps">
+					<a class="p-menuitem-link">
+						<vue-feather
+							v-if="typeof getAssetIcon(slotProps.item.key ?? null) === 'string'"
+							class="p-button-icon-left icon"
+							:type="getAssetIcon(slotProps.item.key ?? null)"
+							size="1rem"
+						/>
```
YohannParis, line 23: We don't use those icons in the project.

packages/client/hmi-client/src/page/project/components/tera-model-modal.vue
blanchco: done, this has been removed

packages/client/hmi-client/src/page/project/components/tera-model-modal.vue
blanchco: I've this function `addNewModelToProject`  to the model service, but due to circular dependency issues I also moved the `newAMR` function to the model service as well.  This was the only instance where the `newAMR` function is but used

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
YohannParis, line 1: This is better to left it here, because its a `PetriNet` AMR that's being built for now, in the near future we will have PDE AMR, etc.

YohannParis (approved):

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
blanchco: moved this back!

packages/client/hmi-client/src/page/project/components/tera-model-modal.vue
blanchco: moved newAMR back to petrinet service

PR #1584, opened 2023-07-25
chore: clean use of simulate chart

Tom-Szendrey:
# Description
lets default the widths and colours and use those defaults instead of all the copy paste for same result


Tom-Szendrey:
This is being done by Manfred on another PR 
https://github.com/DARPA-ASKEM/Terarium/pull/1552

PR #1588, opened 2023-07-25
Jamie dataset preview column toggler

jamiewaese-uncharted:
I added a column toggler feature to the dataset preview so users can select which columns to display. 

**DESIGN**
Monosnap Terarium - Workflow   etc 2023-07-25 15-12-43

**AS IMPLEMENTED**
column toggler


mwdchang:
Duplicate of https://github.com/DARPA-ASKEM/Terarium/pull/1589

PR #1589, opened 2023-07-25
Jamie dataset node column toggler

jamiewaese-uncharted:
I added a column toggler feature to the workflow dataset node. Here's a gif preview:

column toggler on workflow node


mwdchang:
@jamiewaese-uncharted is this something we want to save into the the dataset node's "state"? For example if I filter down from 5-columns down to 3-columns, the next time I come back to the workflow (assuming you closed it at some point) do you expect to see 5 or 3 columns?

PR #1594, opened 2023-07-26
feat: editable matrices for stratified model configs

shawnyama:
# Description

The values in the stratified matrices are editable. Something we should see about fixing is that the order of the configurations change sometimes - I assume this is elastic search related.

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/9e9bc457-4fa1-42ac-b910-0562d80f671c


blanchco (commented):
Left some notes to think about - but looks good!

packages/client/hmi-client/src/components/models/tera-stratified-model-configuration.vue
```diff
@@ -153,6 +180,16 @@ const tableHeaders = computed<{ name: string; colspan: number }[]>(() => [
...
+// Makes cell inputs focus once they appear
+const vFocus = {
+	mounted: (el) => el.focus()
+};
+
+function updateModelConfigName(configIndex: number) {
+	modelConfigurations.value[configIndex].name = modelConfigInputValue.value;
+	updateModelConfiguration(modelConfigurations.value[configIndex]);
+}
+
```
blanchco, line 192: Could be worth putting this in a separate util/service? Might be out of scope for this PR but I know the `tera-model-configuration` uses a similar function.  I imagine both these components use similar function to edit names/values as well.  Just a note

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -79,10 +79,35 @@ const pivotAxes = (data: any[], rowDimensions: string[], colDimensions: string[]
 	return { colAxis, rowAxis };
 };
 
+// Creates a M x 1 matrix where
+// M =  cardinality(rowDimensions[0]) * cardinality(rowDimensions[1]) * ... * cardinality(rowDimensions[m])
+export const createStateMatrix = (data: any[]) => {
+	const rows: any[] = [];
+
```
blanchco, line 86: Are we able to type this (no any)?

mwdchang: I think something is a little off here, we should be able to just use `createMatrx` to create a 1D matrix. 

Also utilities should not know anything about "states" or "transitions" - this is supposed to be a generalized pivot-table function.

packages/client/hmi-client/src/components/models/tera-stratified-model-configuration.vue
```diff
@@ -175,6 +212,15 @@ function openValueModal(id: string, configIndex: number) {
 	}
 }
 
+function resetCellEditing() {
+	const row = { name: false };
```
mwdchang, line 216: This data structure is a little weird, "name" is typically thought of as a string no? - what are we trying to convey?

packages/client/hmi-client/src/components/models/tera-stratified-model-configuration.vue
shawnyama: Yeah later I want to look into the viability of merging the two config components in some way. This component is made from a copy of the other config component - had to quickly get this version working during the hackathon.

packages/client/hmi-client/src/utils/pivot.ts
shawnyama: `createMatrix` ends up giving a 2D matrix for states, I'll double check why that is.

mwdchang: Or rather it should create a 2D-matrix with one row in it.

packages/client/hmi-client/src/components/models/tera-stratified-model-configuration.vue
```diff
@@ -175,6 +212,15 @@ function openValueModal(id: string, configIndex: number) {
...
+
+	// Can't use fill here because the same row object would be referenced throughout the array
+	const cellEditStatesArr = new Array(modelConfigurations.value.length);
```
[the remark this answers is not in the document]
shawnyama: will keep note of this when I merge the configs components later

mwdchang (approved):

PR #1595, opened 2023-07-26
fix: removed home resources

tangopapatime:
# Description
Removed the home resources endpoint and removed the call to the home resources endpoint

Resolves #(issue)


packages/client/hmi-client/src/page/Home.vue
```diff
@@ -218,7 +218,12 @@ onMounted(async () => {
 	resourcesStore.reset(); // Project related resources saved.
 	queryStore.reset(); // Facets queries.
 
-	projects.value = ((await ProjectService.home()) ?? []).slice().reverse();
+	projects.value = ((await ProjectService.getAll()) ?? []).slice().reverse();
+
+	projects.value.forEach(async (project) => {
+		project.assets = await ProjectService.getAssets(project.id);
+		project.relatedDocuments = await getRelatedDocuments(project.id, 'xdd-covid-19');
```
YohannParis, line 225: Could you hardcode this `xdd-covid=19` within the `getRelatedDocuments()` function directly?

YohannParis (approved):
Works for me.

packages/client/hmi-client/src/page/Home.vue
tangopapatime: yup

dvince2: just a heads up here @Jami159

dvince2 (approved):

PR #1598, opened 2023-07-27
Jamie histogram design fix

jamiewaese-uncharted:
A tiny PR in which I removed the gaps between bars in the histograms. Histograms don't have gaps between their bars.

image

**BEFORE**
Monosnap Terarium 2023-07-27 13-12-42

**AFTER**
Monosnap Terarium 2023-07-27 13-11-22


mwdchang:
@jamiewaese-uncharted the chart data flipped in the before and after screenshot, but the min/max labels did not. So it looks like all the data got inversed? Not sure what is going on?

mwdchang:
> @jamiewaese-uncharted the chart data flipped in the before and after screenshot, but the min/max labels did not. So it looks like all the data got inversed? Not sure what is going on?

Seemed to be a problem with hot-reload state?

PR #1600, opened 2023-07-27
fix: 1599 task updated related resources modal

tangopapatime:
# Description
Updated related resource modal

image


packages/client/hmi-client/src/services/data.ts
```diff
@@ -446,15 +446,15 @@ const getAssets = async (params: GetAssetsParams) => {
  * the "/documents?similar_to=<id>" endpoint as an alternative
  */
 const getRelatedDocuments = async (docid: string, dataset: string | null) => {
```
shawnyama, line 448: Rename `docid` to `projectId`?

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -256,14 +256,18 @@
 								<span>Suggested value</span>
 								<div>
 									<div class="suggested-value-source">
-										<i class="pi pi-file" />{{ dataset.metadata?.documents[0].title }}
+										<i class="pi pi-file" />{{
+											dataset.metadata?.documents ? dataset.metadata?.documents[0].title : ''
```
shawnyama, line 260: Nitpick: you can make it shorter - `dataset.metadata?.documents?.[0].title ?? ''`

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -277,7 +281,7 @@
 						</div>
 					</div>
 				</AccordionTab>
-				<AccordionTab v-if="pd.length > 0">
+				<AccordionTab v-if="pd && pd.length > 0">
```
shawnyama, line 284: Nitpick: `!isEmpty(pd)` could check for whether it is exists and if length > 0

packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
```diff
@@ -31,6 +32,17 @@
 					<Column field="name" sortable header="Name"></Column>
 					<Column field="authors" sortable header="Authors"></Column>
 				</DataTable>
+				<div v-else class="no-artifacts">
+					<div class="no-artifacts">
```
shawnyama, line 36: Repeated div

packages/client/hmi-client/src/services/data.ts
tangopapatime: docid matches the api call for now.

shawnyama (approved):

packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
```diff
@@ -147,4 +146,24 @@ const sendForEnrichments = async (/* _selectedResources */) => {
...
+	flex-direction: column;
+	align-items: center;
+}
+
+.no-artifacts-img {
+	width: 30%;
+	padding: 10px;
+}
+
+.no-artifacts-text {
```
YohannParis, line 161: You don't need specialize class when the parent as one.
`.no-artifacts img` would suffice

tangopapatime: @YohannParis I know.  I personally like to use a dedicated class vs doing something like that.  honestly don't know what's actually the preferred way.

PR #1606, opened 2023-07-28
fix: persist search by example on page refresh

Jami159:
# Description

* Make search by example persist when user refreshes the page or if they navigate back and forth using <- and -> buttons
* Do some cleanup around how search by example was executed


packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -440,7 +425,7 @@ const onSearchByExample = async (searchOptions: SearchByExampleOptions) => {
 
 		await executeSearch();
 
-		searchByExampleItem.value = null;
+		// searchByExampleItem.value = null;
```
YohannParis, line 428: Just remove it

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -2,12 +2,12 @@
...
 		<span class="result-count">
 			<template v-if="isLoading">Loading...</template>
-			<template v-else-if="props.searchTerm">
+			<template v-else-if="props.searchTerm || searchByExampleOptionsStr">
 				{{ resultsText }}
 				<span v-if="searchByExampleOptionsStr.length === 0"> "{{ props.searchTerm }}" </span>
 				<div v-else-if="searchByExampleOptionsStr.length > 0" class="search-by-example-card">
 					<tera-asset-card
-						:asset="searchByExampleAssetCardProp"
+						:asset="searchByExampleItem as Document & Model & Dataset"
```
YohannParis, line 10: Shouldn't it be `Document | Model | Dataset`?

YohannParis (approved):
I love the huge cleanup

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
Jami159: done!

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
Jami159: You're right, it should. For whatever reason the expected type for the `asset` prop was `Document & Model & Dataset` instead of `ResultType` (which is an alias for `Document | Model | Dataset`. I've gone ahead an updated the expected `asset` prop type as well as its usage everywhere else in code.

PR #1645, opened 2023-07-31
feat: use catlab extraction when showing matrices

shawnyama:
# Description

Proper dimensions appear for stratified models generated by catlab. Doubly stratified models also work.

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/52d3353a-6f53-4c4e-ba4b-506871709b25


packages/client/hmi-client/src/components/models/tera-stratified-model-configuration.vue
```diff
@@ -162,12 +162,24 @@ const configurations = computed<Model[]>(
...
+	while (c < MAX_DEPTH) {
+		c++;
+		if (level.semantics.span) {
+			level = level.semantics.span[0].system;
+		} else {
+			return level.model;
+		}
+	}
+	return level.model;
+});
```
mwdchang, line 179: I think we should move this into petrinet-service as `getBaseAMR` or something - mostly so the recursion stuff happens in one place.

mwdchang:
You have a few console.log here - are they still needed to debug or can we remove?

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -157,44 +157,31 @@ function updateModelConfigValue(variableName: string) {
...
+	console.log(matrixData);
+
+	if (isEmpty(matrixData)) return;
+
+	// Grab dimension names from the first matrix row
+	const dimensions = [cloneDeep(matrixData)[0]].map((d) => {
+		delete d.id;
+		delete d['@base'];
+		return Object.keys(d);
+	})[0];
```
mwdchang, line 181: By the way, I am not crazy about using the term `@base` ... only using it because the keys are generated and I wanted to avoid collisions if possible. If you have better suggestions I am all ears.

mwdchang (approved):

PR #1646, opened 2023-07-31
fix: correctly update the project after renaming a model

tangopapatime:
# Description

Update the renaming of a model so that it updates the resources on the side bar as well.


shawnyama:
I was about to do this myself but @mwdchang told me that we need to implement a proper asset manager. The way we have it now doesn't support eventual consistency - which means if we refetch the entire project we won't necessarily get the new data because the database may have not update yet.

mwdchang:
What @shawnyama said, this here is more likelihood it will do what we want, but there is no guarantee that it will.

We almost always need a project-context, so I don't think it is a bad idea to keep a proper copy of it in Pinia that is used across the components as computed values. We can then do some kind of staggered/debounced update heuristics to better ensure that the data we have in Pinia is up-to-date. The difference been that we handle this in one place rather than each individual component doing their own update-loops. Let's maybe do a quick chat about it.

CC @YohannParis

tangopapatime:
@mwdchang @shawnyama yeah when I was going through this portion of the code and also realize that this could be an issue.  Found another issue describing that there are operations that could take several seconds. to update.  Your general suggestion of a single service doing all of this make sense.  Any operations that isn't instant should be using some kind of polling.  This is tech debt should be addressed sooner than later imo.

tangopapatime:
@YohannParis This discussion here: https://github.com/DARPA-ASKEM/data-service/issues/302#issuecomment-1662671393 is relevant to addressing this.

YohannParis (approved):

PR #1647, opened 2023-07-31
feat: extract strata dimensions from mira model

shawnyama:
# Description

Mira matrices appear based on their base model variables.
We are temporarily showing both the normal and the stratified configurations for when we view a stratified model:

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/c4e137d8-90f8-4053-9ef3-419ea2af2939

To make both configurations sync with each other I wrote `syncConfigs` in `tera-model.vue`. I then emit `syncConfigs` to the model configuration components. I don't think there is a great way to sync the two sibling components but it should do the temporary job.


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Model.java
```diff
@@ -42,6 +42,6 @@ public class Model implements Serializable {
 	private ModelSemantics semantics;
 
 	@TSOptional
-	private ModelMetadata metadata;
+	private Object metadata;
```
dvince2, line 45: Did `ModelMetadata` change? We should use the POJO where possible

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -1,81 +1,107 @@
...
+			v-if="matrix"
+			class="p-datatable p-component p-datatable-scrollable p-datatable-responsive-scroll p-datatable-gridlines p-datatable-grouped-header stratified-value-matrix"
+		>
+			<div class="p-datatable-wrapper">
+				<table class="p-datatable-table p-datatable-scrollable-table editable-cells-table">
+					<thead v-if="nodeType === NodeType.Transition" class="p-datatable-thead">
+						<tr>
+							<th class="choose-criteria"></th>
+							<th class="choose-criteria"></th>
+							<th v-for="i in matrix[0].length + 1" :key="i" class="choose-criteria">
```
dvince2, line 13: I absolutely could be wrong, but, in this block of code you have `v-for"i in matrix[0].length+1`.  Wouldn't that just be scoped to elements inside that `<th` tag and whats inside of it (`<Dropdown...`)?  Down below in another `<tr><th>` tree you reference i which seems wrong or if its not, I'd love a quick primer on how that works

dvince2 (changes requested):

dvince2 (approved):
Items resolved over slack. looks good!

packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -591,7 +596,8 @@ export const stratify = async (baseModel: Model, strataModel: Model) => {
 // Check if AMR is a stratified AMR
 export const isStratifiedAMR = (amr: Model) => {
 	// Catlab stratification: this will have "semantics.span" field
-	if (amr.semantics?.span && amr.semantics.span.length > 1) return true;
+	if (amr.semantics?.span && amr.semantics.span.length > 1) return StratifiedModelType.Catlab;
+	if (!isEmpty(amr.model.states[0].grounding.modifiers)) return StratifiedModelType.Mira;
 	return false;
```
mwdchang, line 601: This function is now returning mixed types

mwdchang:
Putting this back into draft - we can't merge this because it hardwires everything as a static mira-model.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
```diff
@@ -15,6 +15,7 @@
 
 @Data
 @Accessors(chain = true)
+@Deprecated
```
dvince2, line 18: If we're marking this as deprecated would you mind throwing a javadoc with a note of what people should be using now?

dvince2 (commented):
Back end looks good - I'll let others weigh in on the front end.  Just make a note of what should be used with `ModelMetadata` being deprecated

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
mwdchang: I think that came from me, ModelMetadata should be removed.

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -197,7 +203,10 @@ function configureMatrix() {
 	chosenCol.value = colDimensions[0];
 	chosenRow.value = rowDimensions[0];
 
-	console.log('!!!!!', props.id, result, matrix.value);
+	// Matrix for editable cell states
+	for (let i = 0; i < matrix.value.length; i++) {
+		editableCellStates.value.push(Array(matrix.value[i].length).fill(false));
+	}
```
YohannParis, line 209: Why not ?

```JS
matrix.value.forEach(m => editableCellStates.value.push(Array(m.length).fill(false)));
```

YohannParis (approved):

PR #1651, opened 2023-08-01
fix: preserve stratify drilldown state when navigating forward/backward 

echl:
# Description

* Preserve the state of each step of the stratify drilldown when you navigate forward or backward within the drilldown view. 
* Make generated strata ids same as names
* Add type legend and collapsed view toggle button to tera-model-diagram


packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -188,9 +185,6 @@ export const addState = (amr: Model, id: string, name: string) => {
 };
 
 export const addTransition = (amr: Model, id: string, name: string, value?: number) => {
-	if (amr.model.transitions.find((t) => t.id === id)) {
-		return;
-	}
```
mwdchang, line 193: Why are we changing "addState" and "addTransition" logic? We don't want to allow states or transitions with the same IDs

echl: Yes, I originally added this condition for that reason, however I thought that it might be confusing that a function named `addTransition` may or may not actually add a transition, and that perhaps we should leave the checking of whether the id is a duplicate up to the invoker of the function. Thoughts?

mwdchang: Okay, can you double check the invokers are doing the right things then?

Also in the future it might be better to do cleaning-up/refactoring as a separate PR, it is hard to follow when there are feature engineering and other things going on at the same time.

mwdchang:
@echl can you briefly comment on what to look for, and how/where this notion of state is kept? The PR is pretty large and it is hard to follow.

packages/client/hmi-client/src/components/models/tera-model-type-legend.vue
```diff
@@ -0,0 +1,86 @@
...
+		<ul>
+			<li v-for="(type, i) in stateTypes" :key="i">
+				<div class="legend-key-circle" :style="getLegendKeyStyle(type ?? '')" />
+				{{ type }}
+			</li>
+		</ul>
+		<ul>
+			<li v-for="(type, i) in transitionTypes" :key="i">
+				<div class="legend-key-square" :style="getLegendKeyStyle(type ?? '')" />
+				{{ type }}
```
mwdchang, line 12: I don't understand why you need to do `type ?? ''` , what are you guarding against? why would it be undefined? Can we drop it in favour of cleaner intention/syntax?

packages/client/hmi-client/src/components/models/tera-strata-model-diagram.vue
```diff
@@ -4,30 +4,18 @@
 			<div ref="splitterContainer" class="splitter-container">
 				<section class="graph-element">
 					<tera-reflexives-toolbar
-						v-if="showReflexivesToolbar && strataModel && baseModel"
+						v-if="strataModel && baseModel"
+						v-show="showReflexivesToolbar"
```
echl, line 8: Separated `populateReflexiveOptions` into `v-show` instead of `v-if`. Since `showReflexivesToolbar` will switch between true and false  as user navigates forward or backward through the stratify drilldown (while the condition in v-if is unlikely to switch between true and false), this avoids some issues where `tera-reflexives-toolbar` would not be in the correct state after being unmounted and mounted again

packages/client/hmi-client/src/components/models/tera-strata-model-diagram.vue
```diff
@@ -66,20 +55,7 @@ const splitterContainer = ref<HTMLElement | null>(null);
 const graphElement = ref<HTMLDivElement | null>(null);
 let renderer: PetrinetRenderer | null = null;
 
-const stateTypes = computed(() =>
```
echl, line 69: Moved type legend related stuff into own component

packages/client/hmi-client/src/components/models/tera-typed-model-diagram.vue
```diff
@@ -144,12 +133,6 @@ const graphElement = ref<HTMLDivElement | null>(null);
 let renderer: PetrinetRenderer | null = null;
 
 const typedModel = ref<Model>(props.model);
-const stateTypes = computed(() =>
```
echl, line 147: Move type legend related stuff into own component

packages/client/hmi-client/src/components/models/tera-typed-model-diagram.vue
```diff
@@ -60,7 +60,8 @@
 						/>
 					</section>
 					<tera-reflexives-toolbar
-						v-if="showReflexivesToolbar && model && strataModel"
+						v-if="model && strataModel"
+						v-show="showReflexivesToolbar"
```
echl, line 64: Separated `populateReflexiveOptions` into `v-show` instead of `v-if`. Since `showReflexivesToolbar` will switch between true and false  as user navigates forward or backward through the stratify drilldown (while the condition in v-if is unlikely to switch between true and false), this avoids some issues where `tera-reflexives-toolbar` would not be in the correct state after being unmounted and mounted again

packages/client/hmi-client/src/components/models/tera-typed-model-diagram.vue
```diff
@@ -248,23 +235,61 @@ watch(
...
-				{
-					nodeType: 'Variable',
-					typeName: props.typeSystem?.states[0].name
-				},
-				{
-					nodeType: 'Transition',
-					typeName: props.typeSystem?.transitions[0].properties?.name
-				}
-			);
+		if (props.showTypingToolbar) {
```
echl, line 240: Prepopulate the selections for typing in Step 2 if the model is already typed

packages/client/hmi-client/src/components/models/tera-typed-model-diagram.vue
```diff
@@ -364,8 +386,10 @@ watch(
 watch(
 	numberTypedNodes,
 	() => {
-		if (numberTypedNodes.value === numberNodes.value) {
+		if (allNodesTyped.value) {
 			emit('all-nodes-typed', typedModel.value);
+		} else {
+			emit('not-all-nodes-typed', numberNodes.value - numberTypedNodes.value);
```
echl, line 392: Event emitted so that `tera-stratify` can keep track of whether all nodes have have been typed. Just `'all-nodes-typed'` was not sufficient because if the user fully types the model, but then unchecks some of their selections, `tera-stratify` would not know about it.

packages/client/hmi-client/src/components/models/tera-reflexives-toolbar.vue
```diff
@@ -72,96 +71,120 @@ const stateId2NameMap = computed<{ [id: string]: string }>(() => {
 	return map;
 });
```
echl, line 73: Issue 1: when the user adds reflexives in Step 3, then navigates away from Step 3 (either going back, or hitting stratify) and then they return to Step 3, the menu that prompts the user to add reflexives disappeared, because the logic to populate it (in `populateReflexiveOptions`) is based on comparing two models (`modelToUpdate` and `modelToCompare`) and checking to see if there are any missing types between them. However, once you've added a reflexive, the types between base and strata models will be the same, so the prompts won't appear and you won't be able to edit your selections, e.g. add/remove additional reflexives. 

Solution 1: `typedModel`, which is the amr that is being updated, is now a copy of `modelToUpdate` (previously they referenced the same object), so that the reflexives added to `typedModel` don't affect the logic in `populateReflexiveOptions`

Issue 2: the user was unable to remove reflexives after adding them (e.g. if you made a mistake with your selection, you couldn't undo it)

Solution 2: to my knowledge, the MultiSelect component (used for the dropdown to select which states to add reflexives to) doesn't know about past selections or what the user has selected or deselected, it only knows about the current selected options. So, to allow the user to remove added reflexives when they uncheck an option, every time the selection is updated, I overwrite `typedModel` with the reflexives added based on the user's current selection. 

Issue 3: there can be multiple MultiSelect components, one for each reflexive type that you want to add (e.g. Infect, Recover) that would overwrite each other. 

Solution  3: combine all the selections made from each MultiSelect in the `addedReflexivesRows`, and apply them all at once in `updateStatesToAddReflexives`

mwdchang (approved):

Issue #1634, opened 2023-07-28
Add progress info for CIEMSS components.


YohannParis:
Review and merge the GitHub pull request for the PyCiemss service at https://github.com/DARPA-ASKEM/pyciemss-service/pull/9.

YohannParis:
@blanchco Please bug @Tom-Szendrey and @mwdchang for guidance on where to find the information needed to implement this feature. It's an easy task, but a complexed one.

dgauldie:
duplicate

PR #1653, opened 2023-08-01
task: persist simulation states in the UI

blanchco:
# Description

* This PR has a few assumptions included, but its a basis to integrate the intermediate results into the HMI
* https://github.com/DARPA-ASKEM/pyciemss-service/pull/9 Looking at this PR it looks like the progress that is going to be a float between 0 and 1 AND for now only on the calibrate functionality in pyciemss.
* For now I'm assuming we will get a progress attribute when we get the job status, this can be changed once this is more fleshed out
* We can now persist the simulation/calibration state when we leave and come back to a workflow.  This is done by adding a new attribute called simulationsInProgress on the node state which is an array of simulationIds.  When queuing/running these simulations they will be added to the simulationsInProgress node. On mounting the component it will poll for the jobs.  When the simulations are complete, they will be deleted from the simulationsInProgress array.

As @mwdchang mentioned this polling will be changed for an event sourcing architecture - will be looking into websockets.

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/664edd07-99c2-41b9-ba8e-ddc9810d709a

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/3e79a4cb-5eb2-4662-8d8c-a4e51796e596

References #(issue)
https://github.com/DARPA-ASKEM/Terarium/issues/1634


packages/client/hmi-client/src/components/widgets/tera-progress-bar.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<div>
+		<i class="pi pi-spin pi-spinner"></i>
+		<span>{{ props.status }}...{{ '(' + props.value + '%)' }}</span>
+		<ProgressBar class="progress-bar" :value="props.value"></ProgressBar>
+	</div>
```
blanchco, line 6: hiding progress for now - until we get progress values

packages/client/hmi-client/src/components/widgets/tera-progress-bar.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<div>
+		<i class="pi pi-spin pi-spinner"></i>
+		<span>{{ props.status }}...{{ props.value ? '(' + props.value + '%)' : null }}</span>
+		<ProgressBar v-if="props.value" class="progress-bar" :value="props.value"></ProgressBar>
```
blanchco, line 5: hiding progress for now - until we get progress values

blanchco:
Just to clarify this PR a bit - it was initially an initiative to add intermediate results.  Since the BE isn't prepared yet the added changes are just a refactor around polling as well as persisting simulation results in the UI.

YohannParis:
I do believe we should use SSE for the this, as multiple window/user could look at this.

blanchco:
> I do believe we should use SSE for the this, as multiple window/user could look at this.

I discussed with @mwdchang a little earlier today - I'm looking into this PR https://github.dev/DARPA-ASKEM/Terarium/blob/yohannparis/sse-filtering/packages/client/hmi-client/src/temp/sse.vue and how it can be used for the intermediate results here

blanchco:
> > I do believe we should use SSE for the this, as multiple window/user could look at this.
> 
> I discussed with @mwdchang a little earlier today - I'm looking into this PR [this file specifically](https://github.dev/DARPA-ASKEM/Terarium/blob/yohannparis/sse-filtering/packages/client/hmi-client/src/temp/sse.vue) and how it can be used for the intermediate results here

But just an FYI this PR is not for the intermediate results but instead the persisting of simulation jobs.  Intermediate results will follow

packages/client/hmi-client/src/components/widgets/tera-progress-bar.vue
```diff
@@ -0,0 +1,23 @@
...
+
+<script setup lang="ts">
+import { ProgressState } from '@/types/workflow';
+import ProgressBar from 'primevue/progressbar';
+
+const props = defineProps<{
+	status: ProgressState;
+	value: number;
+}>();
+</script>
```
mwdchang, line 17: This notion of `queued/running/complete` is rather specific to the workflow processes, it might make more sense to place the component there instead of under "widget".

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -195,3 +197,134 @@ export async function makeEnsembleCiemssCalibration(params: EnsembleCalibrationC
...
+			return addSimulationInProgress(node, runId);
+		case SimulationStateOperation.DELETE:
+			return deleteSimulationInProgress(node, runId);
+		case SimulationStateOperation.QUERY:
+			return querySimulationInProgress(node);
+		default:
+			break;
+	}
+	return null;
+}
```
mwdchang, line 281: What is the purpose of this function? It seems like it is just a pass-thru facade, why wouldn't we just directly call each `addSimulationInProgress/deleteSimulationInProgress/querySimulationInProgress` directly given that you have to specify the operation anyway?


Also we wouldn't generally want to return `null` as the new state, because it will mess up other things. We didn't do this part particularly well due to lack of time, but each "operation" does have a statically defined state definition. of what it would expect. For example, see

workflow/calibrate-operation-julia.ts#CalibrationOperationStateJulia and the initState function

mwdchang:
Two things

1. You have some cases where the functional parameters have variable types plus defaults. e.g. `string | string[] = ''` ... personally I think this makes things a lot more difficult to comprehend than necessary - I prefer functions to be more restrictive than being flexible, makes it a lot easier to use, to test and to maintain. In these cases I think we should either reduce it down to `string`, or `string[]` ... let me know your thoughts.

2. Your return patterns are weird.  ` handleSimulationsInProgress` returns "state", but then the return is fed into "getStatus" which expect a list of runIds. These two things are not compatible, am I missing something here?

packages/client/hmi-client/src/components/widgets/tera-progress-bar.vue
blanchco: @mwdchang Ill move this over to workflow folder

blanchco: @mwdchang this has been moved

packages/client/hmi-client/src/services/models/simulation-service.ts
blanchco: @mwdchang You're right, I've got rid of this pass-thru function entirely it wasn't necessary.  As for the add/delete query they are no longer returning null but always return a state.  When this state is return it is checked against the older state to see if we should be emitting a state change or not.

I have also added the `simulationsInProgress` key to all the node states interfaces.

blanchco:
> Two things
> 
> 1. You have some cases where the functional parameters have variable types plus defaults. e.g. `string | string[] = ''` ... personally I think this makes things a lot more difficult to comprehend than necessary - I prefer functions to be more restrictive than being flexible, makes it a lot easier to use, to test and to maintain. In these cases I think we should either reduce it down to `string`, or `string[]` ... let me know your thoughts.
> 2. Your return patterns are weird.  ` handleSimulationsInProgress` returns "state", but then the return is fed into "getStatus" which expect a list of runIds. These two things are not compatible, am I missing something here?

@mwdchang
1. Makes sense - It will now always take in a string[] (run ids) and have simplified the add/delete function to only accommodate this.  
2. Agreed.  It worked but that pass-thru function was a little too flexible to make any sense.  I've fixed this up so I hope it makes more sense now.  Add/Delete return a state, Query returns a list of run ids

blanchco:
I've also stopped the poller when unmounting the component

packages/client/hmi-client/src/components/workflow/tera-progress-bar.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<div>
+		<i class="pi pi-spin pi-spinner"></i>
+		<span>{{ props.status }}...{{ props.value ? '(' + props.value + '%)' : null }}</span>
+		<ProgressBar v-if="props.value" class="progress-bar" :value="props.value"></ProgressBar>
+	</div>
```
YohannParis, line 6:
```suggestion
		<i class="pi pi-spin pi-spinner" />
		<p>{{ status }}...<template v-if="value">{{ `(${value}%)` }}</template></p>
		<ProgressBar v-if="value" class="progress-bar" :value="value" />
```

YohannParis: To ben honest you should create a computed for that Status message, instead of doing complex HTML stuff.
```
<p>{{ message }}</p>
...
const message = computed(() => `${props.status}...${ props.value ? `(${value}%)`}` );
```
or something like that.

packages/client/hmi-client/src/components/workflow/tera-simulate-ciemss-node.vue
```diff
@@ -33,33 +33,36 @@
 		</Accordion>
 	</section>
 	<section v-else>
-		<div>loading...</div>
+		<tera-progress-bar :value="progress.value" :status="progress.status"></tera-progress-bar>
```
YohannParis, line 36:
```suggestion
		<tera-progress-bar :value="progress.value" :status="progress.status" />
```

YohannParis (commented):
All those node having the same progress-bar and code is in big need of refactoring.
Why not leveraging the `tera-worflow-node.vue`

packages/client/hmi-client/src/components/workflow/tera-progress-bar.vue
blanchco: good idea - ill leverage computed here

blanchco:
> All those node having the same progress-bar and code is in big need of refactoring. Why not leveraging the `tera-worflow-node.vue`

@YohannParis I agree that this is in need of a refactor here, but I think we would need a new component to encapsulate all of the simulation/calibration/ensemble nodes since a lot of the code is shared between those components only.  Workflow nodes such as model & dataset do not need this progress/polling functionality within them so Im not sure it would make sense to put it on the `tera-workflow-node`.  

Do you think I should include this refactor as a part of this PR?  I'm a little worried of doing too much in this PR alone.  (Persisting simulation status + Intermediate results + refactor of simulation/calibration/ensemble nodes)

packages/client/hmi-client/src/components/workflow/tera-progress-bar.vue
blanchco: done

packages/client/hmi-client/src/components/workflow/tera-simulate-ciemss-node.vue
blanchco: done

mwdchang:
@YohannParis @blanchco Let's leave progress-bar refactor for another time, arguably it might have been introduced a tad too early 😄 
- It isn't clear that all the engined (julia/ciemss), or even individual operations will have the same behaviour when it comes to relaying progress information
- Given some of the conversations we have had in the last 2 days or so, I think there are still some confusion over the notions of `progress`, 'simulation-poll', and `intermediates`.  Both intermediate and simulations-polls can denote some sense of "progress" and we should sync up on what we want to convey before undertaking large refactoring efforts.

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -195,3 +197,94 @@ export async function makeEnsembleCiemssCalibration(params: EnsembleCalibrationC
...
+	}
+
+	// return an empty array if no run ids are present
+	return [];
+};
+
+export async function simulationPollAction(
+	simulationIds: string[],
+	node: WorkflowNode,
+	progress,
```
mwdchang, line 248: I think you can do

```
progress: Ref<ProgressState> 
```

here to type the function param

mwdchang:
Didn't test all the cases, but looks fine in general, thanks

mwdchang (approved):

PR #1658, opened 2023-08-03
Fix simulation dataset creation

dvince2:
# Description

* fixed a toast issue that was showing up as an error when there was no error. This is because the response code was 201 not 200.
* Users now must name their dataset when saving.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-simulate-julia.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-simulate-julia.vue
@@ -42,13 +42,29 @@
 				icon="pi pi-plus"
 			/>
 			<Button
-				class="add-chart"
-				text
-				:outlined="true"
-				@click="saveDataset(projectId, completedRunId)"
-				label="Save as Dataset"
-				icon="pi pi-save"
-			/>
+				class="save-button p-button p-button-secondary p-button-sm"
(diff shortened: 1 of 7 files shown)


packages/client/hmi-client/src/components/workflow/tera-simulate-julia.vue
```diff
@@ -43,12 +43,28 @@
...
-				label="Save as Dataset"
-				icon="pi pi-save"
-			/>
+				title="Saves the current version of the model as a new Terarium asset"
+				@click="showSaveInput = !showSaveInput"
+			>
+				<span class="pi pi-save p-button-icon p-button-icon-left"></span>
+				<span class="p-button-text">Save as</span>
+			</Button>
+			<span v-if="showSaveInput" style="padding-left: 1em; padding-right: 2em">
```
tangopapatime, line 52: if we can classes/ids over inline styles

tangopapatime (approved):

packages/client/hmi-client/src/components/workflow/tera-simulate-julia.vue
dvince2: Noted. I've lifted this code from the chatty component so once we have that it can take it too.

PR #1662, opened 2023-08-04
feat: logic to parse Mira AMR petrinet + reorg

mwdchang:
### Summary
Add in logic to parse out data for displaying compact graphs and matrix configurations
- compacted model
- state matrix data
- state nesting table
- transition data

Also split out the mirroring Catlab functions into own service to make things a little better organized.
- catlab-petri, and
- mira-petri

Also added an AMR debugging UI

image

### Testing
- Goto `http://localhost:8078/app/#/amr-petri-test`
- Paste in an AMR e.g. from `https://github.com/liunelson/mira/blob/nliu/experiment/notebooks/nliu/`
- Should be able see AMR graph, can collapse/expand if the model has been stratified


packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -159,27 +156,28 @@ function updateModelConfigValue(variableName: string) {
...
-		delete d['@base'];
+		delete d._id;
+		delete d._base;
 		return Object.keys(d);
 	})[0];
 
+	// FIXME: resolve id vs _id
+	matrixData.forEach((d) => {
+		d.id = d._id;
+	});
```
mwdchang, line 179: I will resolve the _id vs id in a subsequent PR,  as the scope is already larger than what I would've liked.

Jami159 (approved):

PR #1680, opened 2023-08-09
Feat(hmi): PDF extractions

### LINKED ISSUE #1679: [FEAT]: Send documents for extraction TA1
When a user uploads a PDF document (or imports one from github) we need to send the pdf away for extractions. 

When a file is uploaded, we need to make two calls to extraction service. One is to extract text from the pdf. The second, after the text is extracted, is to do pdf extractions.

dvince2:
Extracting pdf files as they're uploaded by the user either on disk or github. Fire and forget, the artifacts are silently updated in the background.

Resolves #1679

### DIFF
--- a/packages/client/hmi-client/src/services/models/extractions.ts
+++ b/packages/client/hmi-client/src/services/models/extractions.ts
@@ -106,12 +106,34 @@ export const profileDataset = async (datasetId: string, artifactId: string | nul
 	return response.data.id;
 };
 
-const extractTextFromPDFArtifact = async (artifactId: string) => {
-	const response = await API.post(`/extract/pdf-to-text?artifact_id=${artifactId}`);
-	return response.data.id;
+const extractTextFromPDFArtifact = async (artifactId: string): Promise<string | null> => {
+	try {
+		const response = await API.post(`/extract/pdf-to-text?artifact_id=${artifactId}`);
+		if (response?.status === 200 && response?.data?.id) return response.data.id;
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/services/models/extractions.ts
```diff
@@ -106,4 +106,34 @@ export const profileDataset = async (datasetId: string, artifactId: string | nul
 	return response.data.id;
 };
 
+const extractTextFromPDFArtifact = async (artifactId: string) => {
+	const response = await API.post(`/extract/pdf-to-text?artifact_id=${artifactId}`);
+	return response.data.id;
+};
```
YohannParis, line 112: you should add some error handling

packages/client/hmi-client/src/services/models/extractions.ts
```diff
@@ -106,4 +106,34 @@ export const profileDataset = async (datasetId: string, artifactId: string | nul
...
+
+	let url = `/extract/pdf-extractions?artifact_id=${artifactId}`;
+	if (pdfName) {
+		url += `&pdf_name=${pdfName}`;
+	}
+	if (description) {
+		url += `&description=${description}`;
+	}
+	const response = await API.post(url);
+	return response.data.id;
```
YohannParis, line 126: you should add some error handling

YohannParis (approved):

YohannParis (approved):

PR #1682, opened 2023-08-10
Spike/move to quarkus 3

dvince2:
This change moves us to quarkus 3.  It is largely a straightforward update that will require devs only to install the latest quarkus on their machine (if they don't have it already). 

*THIS DOES NOT MOVE US TO REACTIVE*
- The move to quarkus reactive resulted in a lot of errors which need to be resolved first.
- I want to move us to 3.X to iron out any unknown issues with that migration before updating us to quarkus reactive.

Of note in this PR:
* Moving from Quarkus 2.16.6.Final to Quarkus 3.2.3.Final
* As part of the dependency changes required for that, moving from `javax.*` to `jakarta.*`
* Previously we had a double colon in our application.properties file. This is unallowed and is a reserved character combination for a secrets handler. There will be a corresponding orchestration fix for this.
* Fixing deprecation warning in `Converter.java`

Note: This PR requires https://github.com/DARPA-ASKEM/orchestration/pull/258 to also be ✅ for running from orchestration to work.


YohannParis (approved):
Works on my machine

PR #1683, opened 2023-08-10
feat: adding event sourcing for progress and intermediate results

blanchco:
# Description

* This PR will handle the intermediate results.  The way this is being done is through an `EventSourceManager` class which will handle the connections of all event sources.  When polling for a status on a simulation job an EventSource connection will be opened that job where we can listen to the incoming events.  When it is completed/error the connection will be closed
* EventSourceManager class can now be used when we want want to use this paradigm

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/6297c3e2-efe6-40cb-b065-527914166ad4

Resolves #(issue)


packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -302,3 +326,56 @@ export async function simulationPollAction(
 		error: true
 	};
 }
+
+type MessageHandler = (message: any) => void;
+
+export class EventSourceManager {
```
Tom-Szendrey, line 332: Feels like this class deserves its own file in `/utils` folder?

blanchco: agreed - this has been done

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -250,12 +242,46 @@ export async function simulationPollAction(
...
+	const unhandledStateSimulationIds = response.filter(
+		(simulation) =>
+			simulation?.status !== ProgressState.QUEUED &&
+			simulation?.status !== ProgressState.RUNNING &&
+			simulation?.status !== ProgressState.COMPLETE
+	);
+
+	// there are unhandled states - we will return an error
+	if (unhandledStateSimulationIds.length > 0) {
 		const newState = deleteSimulationInProgress(node, simulationIds);
```
blanchco, line 269: did some cleaning here to make the if conditions hopefully more readable understandable

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -250,12 +242,48 @@ export async function simulationPollAction(
...
+
 	simulationIds.forEach((id) => {
 		requestList.push(getSimulation(id));
 	});
 	const response = await Promise.all(requestList);
-	if (response.every((simulation) => simulation!.status === ProgressState.COMPLETE)) {
+
+	const completedSimulationIds = response
+		.filter((simulation) => simulation?.status === ProgressState.COMPLETE)
+		.map((simulation) => simulation!.id);
```
YohannParis, line 253: I think this is a typo? `simulation!.id` instead of `simulation?.id` or is this `!.` a new thing?

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
```diff
@@ -288,13 +264,30 @@ const getStatus = async (simulationId: string) => {
...
+
+	// open a connection for each run id and handle the messages
+	runIds.forEach((id) => {
+		eventSourceManager.openConnection(id, `/api/simulations/${id}/partial-result`);
+		eventSourceManager.setMessageHandler(id, (message) => {
+			const parsedMessage = JSON.parse(message);
+			if (parsedMessage.progress) {
+				progress.value.value = Math.round(parsedMessage.progress * 100);
+			}
+		});
```
YohannParis, line 276: The function would be clearer if it was clearer defined with a named and then passed as a parameter. In my opinion.
```js
function handlingProgress(message) {...}
[...]
eventSourceManager.setMessageHandler(id, handlingProgress);
```

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
```diff
@@ -288,13 +264,30 @@ const getStatus = async (simulationId: string) => {
...
 		.setInterval(3000)
 		.setThreshold(300)
 		.setPollAction(async () => simulationPollAction(runIds, props.node, progress, emit));
 	console.log('Poller defined');
 	const pollerResults = await poller.start();
 
+	// closing event source connections
+	runIds.forEach((id) => {
+		eventSourceManager.closeConnection(id);
+	});
```
YohannParis, line 289: You can simplify `forEach` by just calling it: `runIds.forEach(eventSourceManager.closeConnection);`

packages/client/hmi-client/src/utils/event-source-manager.ts
YohannParis, line 1: I think this should be under the `api/` folder and not `utils`

packages/client/hmi-client/src/services/models/simulation-service.ts
blanchco: Not a typo here -` !.` is the Non-null assertion operator in typescript: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator.

In this context it asserts the simulation variable is non-null and non-undefined

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
blanchco: since `closeConnection` isn't a callback function I don't believe the `id` is implicitly passed to the function.  I tried it out and it seems to be giving an error that the connection is undefined.  I have simplified the syntax a little and got rid of the `runIds.forEach( id => eventSourceManager.closeConnection(id))`

YohannParis: Don't worry then.

PR #1686, opened 2023-08-10
feat: save stratification workflow state

echl:
# Description

* Save state of stratify drilldown nodes within and across workflows.


packages/client/hmi-client/src/components/models/tera-typed-model-diagram.vue
```diff
@@ -224,70 +225,86 @@ async function toggleCollapsedView() {
...
+		typedRows.value = [];
+		populateTypingRows();
 	},
 	{ immediate: true }
 );
 
 watch(
-	() => [props.typeSystem, props.showTypingToolbar],
-	() => {
-		if (props.showTypingToolbar) {
```
echl, line 240: Just extracted this into a function.

packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -80,6 +80,7 @@
 				<tera-stratify
 					v-if="workflowNode && workflowNode.operationType === WorkflowOperationTypes.STRATIFY"
 					:node="workflowNode"
+					:key="workflowNode.id"
```
echl, line 83: Maybe kind of a "cheat", but this is to make the tera-stratify component fully re-render everytime a different stratify drilldown is opened, because how Vue's internal logic handles reactively updating a component causes a lot of difficult to handle edge cases.

packages/client/hmi-client/src/components/workflow/tera-stratify.vue
```diff
@@ -248,8 +248,33 @@ const modelConfiguration = ref<ModelConfiguration>();
...
+	stratifyStep.value = state.stratifyStep;
+	strataModel.value = state.strataModel;
+	strataType.value = state.strataType;
+	labels.value = state.labels;
+	model.value = state.model;
+	typedBaseModel.value = state.typedBaseModel;
+	typedStrataModel.value = state.typedStrataModel;
+	stratifiedModel.value = state.stratifiedModel;
+	numUntypedNodes.value = state.numUntypedNodes;
+}
```
mwdchang, line 277: Probably `restoreState` is a better name here.

Also note generally speaking this is not how "state" is treated in the workflow, they actually have typing information and are initialized in a specific manner. I don't think it is worthwhile to change to a proper state usage now since we are wrapping up the catlab stuff, but please make sure you understand the convention and paradigm we are using.

mwdchang (approved):

Issue #1688, opened 2023-08-10
[BUG]: Selecting text in an input text field in any modal closes the modal


pascaleproulx:
**Describe the bug**
Selecting text in an input text field in any modal closes the modal. This happens in many places in the app and it is very aggravating paper cut that we saw over and over during the eval. 

**To Reproduce**
1. From the Overview page, click on 'New Model'
2. On the Model tab, click on 'Edit model' button
3. In the Add/Edit state modal, type in a string for model ID
4. Select the string
=> modal closes.

**Expected behavior**
Do not close the modal on text selection.

PR #1702, opened 2023-08-14
fix: modal closing on text select

blanchco:
# Description

* Modals were closing when selecting text.  This was because the modal was listening to the click option in the mask.  When they would go to select text, they would select and drag outside the modal and mouseup which would cause the modal to close.
* This has been changed so that the main and aside are siblings so that the main mouse events are not bubbling up to the aside

BEFORE:

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/07ff9cde-0662-42c9-952e-3c3703c496f0

AFTER:

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/4f0c1d33-9c13-46cf-8fd8-86ff3bf2ab8b

Resolves #(issue)
https://github.com/DARPA-ASKEM/Terarium/issues/1688


YohannParis (approved):

YohannParis (changes requested):
See @mwdchang comment

mwdchang:
@blanchco @YohannParis I think part of the problem is that we are interchanging parent-child relations vs z-indices but they are different things.

Here, `<aside>` is the parent of `<main>` so the events are bubbled up and we have some problems telling them apart.

Maybe we can try something like this (make them siblings) that will prevent further issues and not only clicks?

```
<transition>
  <div>
     <aside></aside>
     <main></main>
  </div>
 </transitin>
```

We'd need to adjust the CSS though

```
main {
  position: relative;
  z-index: 1005;
  ...
}
```


Also we did something similar before in World-Modeler with trying to disambiguate events using the raw mouse-events themselves, here is the implementation if interested.
https://github.com/uncharted-causemos/causemos/blob/master/client/src/components/modals/modal.vue

blanchco:
root of the problem should be addressed now - videos are behaving the same way as before

YohannParis (changes requested):

packages/client/hmi-client/src/components/widgets/tera-modal.vue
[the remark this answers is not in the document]
blanchco: @YohannParis This poses the same issue as before:

The user starts their click inside the modal and finishes their click outside the modal.  When they finish their click the HTML target will be in the backdrop (aside) despite the user starting their click inside the modal.

What my original solution essentially did was whenever the user started their click on the backdrop (aside), the modal would close so that if they started the click inside the modal the backdrop (aside) would not listen to the finishing of the click if they were so drag their mouse outside the modal. 

What @mwdchang solution is doing is separating the backdrop (aside) and modal so that the backdrop can listen to it's own events and the event couldn't bubble up to the backdrop.

Please let me know if I'm interpreting anything incorrectly!

packages/client/hmi-client/src/components/widgets/tera-modal.vue
```diff
@@ -12,29 +12,43 @@
...
+		default: 'var(--z-index-modal)'
+	}
+});
 </script>
 
 <template>
 	<Transition name="modal">
-		<aside @click.self="$emit('modalMaskClicked')">
-			<main>
+		<main :style="{ 'z-index': zIndex }">
```
YohannParis, line 26:
```suggestion
		<main :style="{ '--z-index': zIndex }">
```

packages/client/hmi-client/src/components/widgets/tera-modal.vue
```diff
@@ -12,29 +12,43 @@
...
+			</section>
+			<aside @click.self="$emit('modalMaskClicked')" />
+		</main>
 	</Transition>
 </template>
 
 <style scoped>
+main {
+	isolation: isolate;
+}
```
YohannParis, line 45: This is how you do default in CSS, not by using JS, but by using custom properties

```suggestion
main {
  isolation: isolate;
  z-index: var(--z-index, var(--z-index-modal));
}
```

YohannParis (changes requested):

packages/client/hmi-client/src/components/widgets/tera-modal.vue
blanchco: added

packages/client/hmi-client/src/components/widgets/tera-modal.vue
blanchco: added

YohannParis (approved):

PR #1703, opened 2023-08-14
1681 improve dataset upload metadata

### LINKED ISSUE #1681: Improve dataset upload metadata
Between the upload of dataset and sending them to TDS we need to:
- ✅ Add the column names
- ✅ Add the rows numbers
- ✅ The Source URLs if from GitHub
- ✅ the user uploading the dataset

Send all of this as metadata to TDS when creating a new dataset (cf. https://data-service.staging.terarium.ai/#/Dataset/create_dataset_datasets_post)

dvince2:
# Description

* Add the column names
* Add the rows numbers
* The Source URLs if from GitHub
* the user uploading the dataset

Resolves #1681


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
```diff
@@ -37,7 +36,7 @@ public Response getProjects(
 
 		projects.forEach(project -> {
 			Assets assets = proxy.getAssets(project.getProjectID(), Arrays.asList("datasets", "models", "publications"));
-			Map metadata = new HashMap();
+			Map<String, String> metadata = new HashMap<>();
```
dvince2, line 39: Raw types shouldn't have been used here.  This is unrelated to this PR at all but was like nails on a chalkboard to me ;)

packages/client/hmi-client/src/components/widgets/tera-import-github-file.vue
```diff
@@ -334,8 +335,9 @@ async function importDataFiles(githubFiles: GithubFile[]) {
 		await createNewDatasetFromGithubFile(
 			repoOwnerAndName.value,
 			githubFile.path,
-			props.project?.username ?? '',
-			props.project?.id ?? ''
+			auth.name ?? '',
```
YohannParis, line 338: I think those should be done on the HMI-server, I don't see the need to send that information from the front-end.
What do you think?

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -138,7 +139,7 @@ export interface Dataset {
     description?: string;
     dataSourceDate?: string;
     fileNames?: string[];
-    url?: string;
+    datasetUrl?: string;
```
YohannParis, line 142: Why `datasetUrl` instead of `url`? for a need of clarity with S3 signed URLs?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/ProjectResource.java
YohannParis: Make sense, thanks!

YohannParis (approved):

PR #1705, opened 2023-08-14
fix: prevent opening of  Publications & Documents asset from breaking tabs

### LINKED ISSUE #1669: [BUG]: Opening an asset of type 'Publications & Documents' breaks the tabs
**Describe the bug**
Tested on staging on Aug 8. Open a project with some publications in it. Open a publication by clicking on it from the resource panel. => The tabs are gone. Opening any other kind of asset (before opening a publication) works fine.

Jami159:
# Description

* When clicking on a "Publication & Documents" asset from the Resource panel it would break the tabs because it tried accessing `asset.name` instead of `asset.title` for `DocumentAsset` types (`DocumentAsset` types don't have a `name` property)
* This code change now correctly checks for the type and returns the appropriate title/name

Resolves #1669


packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
YohannParis, line 1: Should it be better to fix the source of why `ProjectAssetTypes.DOCUMENTS` as an  `asset.title` instead of a `name`. This sounds arbitrary.

Creating a proper `Asset` type would fix this and help us remove all the `any` in this code.

packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
```diff
@@ -63,8 +63,10 @@ const getTabName = (tab: Tab) => {
 	const assets = resourceStore.activeProjectAssets;
 
 	if (assets) {
-		const asset: any = assets[tab.pageType as string].find((d: any) => d.id === tab.assetId);
-		return asset.name ?? 'n/a';
+		const asset: any = assets[tab.pageType as string].find(
+			(d: any) => d.id?.toString() === tab.assetId?.toString()
```
YohannParis, line 67: Can you rename `d` to `asset` for clarity sake?

YohannParis (changes requested):

packages/client/hmi-client/src/components/widgets/tera-tab-group.vue
Jami159: So [ProjectAsset](https://github.com/DARPA-ASKEM/Terarium/blob/c99aa4a31634a0aa754cd41011c1aed65b1a6159/packages/client/hmi-client/src/types/Project.ts#L26) has the property `[ProjectAssetTypes.DOCUMENTS]: DocumentAsset[]`, and this `DocumentAsset` has the `title` property. Meaning that we would have to update the [DocumentAsset](https://github.com/DARPA-ASKEM/Terarium/blob/c99aa4a31634a0aa754cd41011c1aed65b1a6159/packages/client/hmi-client/src/types/Types.ts#L72) type here. This change isn't really trivial because then we would have to update everything that is of type `DocumentAsset`. So I think for now it might make sense to leave it as it is and do the explicit check to figure out if we should return `title` or `name`

YohannParis: Or update `DocumentAsset` to have a `name`, which by default is the `title` of the document. But in the near futur we might want to let the user change the name of a Document.
For example the Document as a Title `The endothelial function Provides early prognostic Information in patients with COVID-19: A cohort study` to use a `name` like `Endothelial function` for ease of management.

YohannParis (approved):

PR #1721, opened 2023-08-15
feat: test pyodide

mwdchang:
### Summary
This is an experiment to incorporate python (in. WASM) into Terarium to handle the parsing of symbolics and expressions.

- Expression parsing and return different format (e.g. latex and mathml)
- Evaluation
- Variable extraction

### Test
http://localhost:8078/app/#/pyodide-test

image


YohannParis (approved):
Really neat!
Works on my machine, but clearly a prototype.

PR #1724, opened 2023-08-16
fix: send artifact id when enriching dataset

blanchco:
# Description

* send an artifact id when enriching a dataset/model 
* single selection until we can support sending multiple endpoints

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/d19a6cf5-71d4-4fdc-801c-91acf545730b

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
@@ -27,8 +27,9 @@
 					:value="allResources"
 					v-model:selection="selectedResources"
 					tableStyle="min-width: 50rem"
+					selection-mode="single"
 				>
-					<Column selectionMode="multiple" headerStyle="width: 3rem"></Column>
+					<Column selectionMode="single" headerStyle="width: 3rem"></Column>
 					<Column field="name" sortable header="Name"></Column>
 					<Column field="authors" sortable header="Authors"></Column>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
```diff
@@ -119,8 +120,7 @@ const addResources = () => {
 
 const sendForEnrichments = async (/* _selectedResources */) => {
 	// 1. Send dataset profile request
-	/* TODO: send selected resources along with dataset to backend for enrichment */
-	const resp = await profileDataset(props.assetId);
+	const resp = await profileDataset(props.assetId, selectedResources.value.id);
```
YohannParis, line 123: Be careful that `selectedResources` can be null, a user is not obliged to select one.

YohannParis (changes requested):

packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
blanchco: i've surrounded it in a if to check if it is null now

blanchco: scratch that ^.  Now defaults to null if there is no selection

YohannParis (approved):

PR #1725, opened 2023-08-16
use xy-scatter to get around uneven/irregular intervals

mwdchang:
### Summary
Fix scaling issues: We have been stretching and squishing the simulation results incorrectly because the intervals are not necessarily regular in the simulation result.

ChartJS doesn't seem to support this kind of rendering, but we can leverage XY-scatter to achieve the desired effect.

Before

image

After

image


packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
```diff
@@ -214,11 +214,16 @@ const renderGraph = () => {
 			.forEach((run, runIdx) => {
 				const dataset = {
 					data: run.map(
-						(datum: { [key: string]: number }) => datum[variable] // - runResults[selectedRun.value][timeIdx][code]
+						// - runResults[selectedRun.value][timeIdx][code]
+						(datum: { [key: string]: number }) =>
+							// return datum[variable];
+							({
+								y: +datum[variable],
```
echl, line 221: Why the plus operator?

echl (approved):

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
mwdchang: coerce numbers, they are strings for some reason.

Issue #1719, opened 2023-08-15
Click "enter" on create new project/assets should submit the form and not close the modal


blanchco:

blanchco:
@YohannParis @pascaleproulx 
Could you confirm the expected behaviour here:
1. Open a modal to create project or asset
2. fill information in modal
3. click enter, modal does not close
4. project or asset is created in the background
6. modal is still open, the modal information clears and you can then create another project or asset

YohannParis:
Nope, right now when you click enter it closes the modal without submitting the form.
You should make sure the form is submitted when you click enter like if you had clicked on the `create`  button, and then close the modal.

blanchco:
@YohannParis one more question here - should this behaviour for all modals or just the create project and create model modals?

PR #1730, opened 2023-08-17
fix: enter on form modal submit

blanchco:
# Description

* fix an issue for adding a project/asset where the modal would close and not submit the form
* modal will now emit an event when enter is press
* modals themselves can define the action they want to do when enter is pressed
* I've added enter action to modals where it makes sense

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/7e25ffe0-1be6-4a78-8e5b-65788596b76b

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/2b7284ac-1149-45eb-b354-d4feb61e585a

Resolves #(issue)
https://github.com/DARPA-ASKEM/Terarium/issues/1719


packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -623,6 +624,9 @@ const prepareTransitionEdit = () => {
 const addNode = async () => {
 	if (!renderer) return;
 	const node = editNodeObj.value;
+	if (!node?.id) {
+		return;
+	}
 	if (props.model?.model.states.find((s) => s.id === node.id)) {
```
blanchco, line 630: ive added this since the button is disabled when editNodeObj.id is empty, so i have added this check for the enter action to capture it

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -134,12 +134,13 @@
...
 				class="modal"
 				@modal-mask-clicked="isNewProjectModalVisible = false"
+				@modal-enter-press="createNewProject"
 			>
 				<template #header>
 					<h4>Create project</h4>
 				</template>
 				<template #default>
-					<form>
+					<form @submit.prevent>
```
blanchco, line 143: when there is a form element - i have prevent the default behavior that which refreshes the page

YohannParis (approved):

mwdchang (approved):

PR #1735, opened 2023-08-18
1167 feat implement new code asset type in back end

### LINKED ISSUE #1167: [FEAT]: Implement new Code asset type in back end
Currently code files cannot be saved in a project.  This issue is to allow users to save their code files in TDS for a project, and also be able to import them as described in issue #1142 from the github file browser

Code files should persist in a project

**Additional context**
See issue 1142 for more information

dvince2:
# Description

Adding a new code type asset in the back end, and hooking the front up as-is. This right now is very similar to the old code artifact.

Resolves #1167

### DIFF
--- a/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/CodeProxy.java
+++ b/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/CodeProxy.java
@@ -1,0 +1,56 @@
+package software.uncharted.terarium.hmiserver.proxies.dataservice;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+import software.uncharted.terarium.hmiserver.annotations.LogRestClientTime;
+import software.uncharted.terarium.hmiserver.models.dataservice.code.Code;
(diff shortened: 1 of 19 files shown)


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/CodeProxy.java
```diff
@@ -0,0 +1,56 @@
...
+		@POST
+		@LogRestClientTime
+		Code createCode(JsonNode code);
+
+		@GET
+		@Path("/{id}")
+		@LogRestClientTime
+		Code getCode(@PathParam("id") String codeId);
+
+		@PUT
```
YohannParis, line 35: We should be consistent, sometimes it's PATCH, sometimes PUT.

YohannParis (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/proxies/dataservice/CodeProxy.java
toddroper: We are using `PUTs` for every endpoint set.  The `PATCH` endpoint caused problems so we only implemented it on the datasets.

PR #1736, opened 2023-08-18
update default config also updates model

mwdchang:
### Summary
This PR does two things
- Allows default configuration to be changed
- When changing default config, the change is propagated into the model. This introduces a refreshing bug where model's description page is not refreshed properly. I think rather than trying to force a workaround here, we should just wait for the model page refactor/restructure that we plan to solve this, and other issues.

### Testing
Open up a model
- Go to the model tab
- Change a parameter's value for Default config
- Refresh the page, note that a) The change is persisting, and b) the Description page reflects the same change

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -36,6 +36,7 @@
 		v-if="isAboutModalVisible"
 		class="modal"
 		@modal-mask-clicked="isAboutModalVisible = false"
+		@modal-enter-press="isAboutModalVisible = false"
 	>
 		<template #header>
 			<h4>About Terarium</h4>
--- a/packages/client/hmi-client/src/components/models/tera-model-configuration.vue
+++ b/packages/client/hmi-client/src/components/models/tera-model-configuration.vue
@@ -162,7 +162,11 @@
 		:model="configItems"
 	></SplitButton>
 	<Teleport to="body">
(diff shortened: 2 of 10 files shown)


packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -44,6 +44,12 @@ export const updateModelConfiguration = async (config: ModelConfiguration) => {
 		});
 	}
 
+	// If the "default" config is updated we want to update the model as well
+	// because the model as a copy of the data
+	if (config.name === 'Default config') {
+		API.put(`/models/${config.configuration.id}`, config.configuration);
+	}
+
```
YohannParis, line 52: This should be done on `hmi-server` behind the PUT `/model_configurations/${config.id}` call.

shawnyama (approved):

packages/client/hmi-client/src/services/model-configurations.ts
mwdchang: This may be a good point for a small discussion.

Putting model-update behind the API-layer makes the interfaces cleaner, but you lose a bit of control because we wouldn't necessarily known when the model itself is updated. 

So the overall question is whether we want to refresh the model independently of the model-configuration updates, or we couple them together. e.g. by calling update model-config, I would known that model-config has updated, but there is also a chance that model is updated ... I need to know if it is necessary to refresh the model-sections because they might be stale.

PR #1740, opened 2023-08-21
fix: update simulation.csv to result.csv

Jami159:
# Description

* Verified with Five that the `simulation.csv` filename should indeed always be `result.csv` (apparently the change was made about a month ago).
* Also updated some strings to use sentence case as per Jamie's request

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-calibrate-ensemble-ciemss.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-calibrate-ensemble-ciemss.vue
@@ -33,7 +33,7 @@
 				text
 				:outlined="true"
 				@click="addChart"
-				label="Add Chart"
+				label="Add chart"
 				icon="pi pi-plus"
 			/>
 			<Button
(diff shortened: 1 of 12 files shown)


YohannParis (approved):
If this `results.csv` is something that cannot be changed from the front-end, then we should hard-code it on the `hmi-server` endpoints and remove any reference from it on the `hmi-client`.

Jami159:
> If this `results.csv` is something that cannot be changed from the front-end, then we should hard-code it on the `hmi-server` endpoints and remove any reference from it on the `hmi-client`.

There are a couple of cases where it's called with `visualization.json`. I should have been more clear in my description in terms of specifying that all instances of it being called with `simulation.csv` should be replaced with `result.csv`, but not necessarily all cases in general should be replaced with `result.csv`.

YohannParis:
> 
> 
> 
> 
> 
> 
> > If this `results.csv` is something that cannot be changed from the front-end, then we should hard-code it on the `hmi-server` endpoints and remove any reference from it on the `hmi-client`.
> 
> 
> 
> There are a couple of cases where it's called with `visualization.json`. I should have been more clear in my description in terms of specifying that all instances of it being called with `simulation.csv` should be replaced with `result.csv`, but not necessarily all cases in general should be replaced with `result.csv`.

Even so, if those values are not changed or updated on the front-end they should be hardcoded on the backend

PR #1749, opened 2023-08-22
fix: issue where duplicate and blank publication appears when adding for the first time

blanchco:
# Description

* Fixed an issue where added publications to a project would cause duplicate an empty publication and a blank publication
* removed line of code which added the publications to the project locally but this is no necessary since the project page is reloaded when we navigate to the page

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/7f2a8aed-4add-4445-9c73-57ebef2d5d35


YohannParis (approved):
That's comic!

PR #1750, opened 2023-08-22
refactor: start tera-model rewrite

shawnyama:
# Description

Sets up the `tera-model` basis. All the new components are in the `/model` folder old version is in `/models` folder. You can test the new `tera-model` here: `http://localhost:8078/app/#/new-tera-model`

Child components started:
- `tera-model-description`
- `tera-model-diagram`

`tera-model-type-legend` is not changed just moved.

The default `assetId` is for SIR. You can change the model you want to view by changing the default value of the `assetId` prop.

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/42426099-b5c4-4aa3-83ea-93605365b2e2


mwdchang:
I would suggest you have better names than `models` and `models` to accommodate different modelling frameworks.

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -0,0 +1,168 @@
...
+					<tera-model-diagram
+						:model="model"
+						:is-editable="!featureConfig.isPreview"
+						@update-model="updateModelContent"
+					/>
+				</AccordionTab>
+				<AccordionTab header="Model equations"> </AccordionTab>
+				<AccordionTab header="Model observables"></AccordionTab>
+				<AccordionTab header="Model configurations"></AccordionTab>
+			</Accordion>
```
YohannParis, line 66: This should be a component itself.
Not on this PR, but the next one.

packages/client/hmi-client/src/components/model/tera-model-type-legend.vue
```diff
@@ -0,0 +1,86 @@
...
+		backgroundColor: getNodeTypeColor(id)
+	};
+}
+</script>
+
+<style scoped>
+.legend {
+	position: absolute;
+	bottom: 0;
+	z-index: 1;
```
YohannParis, line 52: Why this is needed, this should be set by its parent

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -0,0 +1,168 @@
...
+	}
+});
+
+const model = ref<Model | null>(null);
+const modelView = ref(ModelView.DESCRIPTION);
+
+const newModelName = ref('New Model');
+const isRenamingModel = ref(false);
+
+const isNamingModel = computed(() => isEmpty(props.assetId) || isRenamingModel.value);
```
YohannParis, line 118: The word `model` can be removed from all those `const` it feel heavy.

YohannParis (approved):

packages/client/hmi-client/src/components/model/tera-model.vue
shawnyama: That makes sense thanks

PR #1751, opened 2023-08-22
feat: plot input dataset values in calibrate chart

### LINKED ISSUE #1659: The calibrate boxes should plot the data mapped to the model variables

Jami159:
# Description

* Right now only plots the input dataset values for the pyciemss calibrate
* Will be implementing for julia calibrate and calibrate ensemble as part of #1752 

How it looks now:
Screenshot 2023-08-23 at 4 49 34 PM

How it looks previously:

Screenshot 2023-08-22 at 12 04 33 PM

Resolves #1659


packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
YohannParis, line 1: Everything in the `selectedVariable.value.forEach((variable) => {})` should be in an util

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -56,8 +56,12 @@ async function getBulkDatasets(datasetIDs: string[]) {
...
  * @return Array<string>|null - the dataset raw content, or null if none returned by API
  */
-async function downloadRawFile(datasetId: string, filename: string): Promise<CsvAsset | null> {
-	const URL = `/datasets/${datasetId}/downloadCSV?filename=${filename}`;
+async function downloadRawFile(
+	datasetId: string,
+	filename: string,
+	limit: number = 100
+): Promise<CsvAsset | null> {
+	const URL = `/datasets/${datasetId}/downloadCSV?filename=${filename}&limit=${limit}`;
```
YohannParis, line 64: This is what I was talking about with the `result.csv`. You can create a new endpoints on `DatasetResource.java` to do what you need:  no limits, and only filename = 'result.csv'. instead of adding capability to `downloadRawFile()`

Jami159: Is it worth it to create a new endpoint for this in `DatasetResource.java` right now? Just asking because from our conversation with Daniel the other day, it sounds like this might just be a temporary solution.

On a separate note, the `filename` here refers to the name of the dataset that's being used as the input dataset for the calibrate. It has nothing to do with `result.csv` which is the filename for the `/simulations/{id}/result` endpoint in `SimulationResource.java`

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
YohannParis: No, you imported `renderGraph`, what I'm talking about it's all the 20+ lines of code you wrote within your `electedVariable.value.forEach` should be in an util function. Where, maybe a `utils.ts` next to the componet

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
```diff
@@ -217,6 +217,17 @@ const runCalibrate = async () => {
...
 
+	let start = state.timeSpan.start;
+	let end = state.timeSpan.end;
+	// If we have the min/max timestamp available from the csv asset use it
+	if (csvAsset.value) {
+		const tIndex = csvAsset.value.headers.indexOf('timestamp');
+		if (tIndex !== -1) {
+			start = csvAsset.value.stats?.[tIndex].minValue;
+			end = csvAsset.value.stats?.[tIndex].maxValue;
+		}
```
mwdchang, line 228: Do we have guarantee that the that whatever "timestamp" is mapped to is a numeric value? Can they be, for example, date strings like "2012-02-02"? 

@Tom-Szendrey do you know?

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
```diff
@@ -228,8 +231,47 @@ const renderGraph = () => {
...
 				datasets.push(dataset);
-			})
-	);
+			});
+
+		if (props.initialData) {
+			let v = variable;
+			// get the dataset variable from the mapping of model variable to dataset variable if there's a mapping
+			if (props.mapping) {
+				if (!(props.mapping.length === 1 && Object.values(props.mapping[0]).some((val) => !val))) {
```
mwdchang, line 240: I don't understand what you are checking in this predicate expression, especially the 2nd part
```
Object.values(props.mapping[0]).some((val) => !val)
```

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
```diff
@@ -228,8 +231,47 @@ const renderGraph = () => {
...
+				// ignore the first row, it's the header
+				data: props.initialData.csv.slice(1).map((datum: string[]) => ({
+					x: +datum[tIndex].trim(),
+					y: +datum[colIdx].trim()
+				})),
+				label: `${variable} - dataset`,
+				fill: false,
+				borderColor: '#000000',
+				borderWidth: 3
+			};
```
mwdchang, line 270: It is difficult to tell the lines apart, can we try something like a dashed style for training-data? 

Also feel free to post previews to slack channel to get folks' thoughts and comments

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
```diff
@@ -228,8 +231,47 @@ const renderGraph = () => {
...
+					colIdx = i;
+					break;
+				}
+			}
+
+			const dataset = {
+				// ignore the first row, it's the header
+				data: props.initialData.csv.slice(1).map((datum: string[]) => ({
+					x: +datum[tIndex].trim(),
+					y: +datum[colIdx].trim()
```
mwdchang, line 264: I don't think you need `trim()` with coercion ?

```
> let x = "   22.123  "
< undefined
> x
< '   22.123  '
> +x
< 22.123
```

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
Tom-Szendrey: the `timestamp` mapping is just a column in a .csv
This technically can be any type

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
Jami159: Sorry, I'm not fully sure what you mean by imported `renderGraph`. [renderGraph](https://github.com/DARPA-ASKEM/Terarium/blob/2bb1a081172efa8786ca92322c97c7c88e715b24/packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue#L203-L239) is defined in this file and it contains the `selectedVariable.value.forEach` loop. Since the lines of code I added and the existing code in `selectedVariable.value.forEach` are closely related, it would probably make sense to keep all of the logic in that loop together. I can definitely make a util function for the full `selectedVariable.value.forEach` loop. Just want to confirm that's what you're suggesting.

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
Jami159: Checking for the case that where `props.mapping = [{datasetVariable: '', modelVariable: ''}]`. This is the case where a user has not provided an explicit mapping.

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
Jami159: Oh, does it automatically get rid of whitespaces and newlines and such? I initially added this trim because some of the values were like `"123.456\r"` so I wanted to make sure we got rid of the `\r`. I didn't realize this was redundant. I will get rid of the `trim()` then.

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
Jami159: Yeah I agree that it's difficult to distinguish the lines. I'll give dashed lines a shot.

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
mwdchang: Ugh...if we are getting `\r` or `\n` or some combination of them, it means we didn't parse the file properly 😢

packages/client/hmi-client/src/components/workflow/tera-simulate-chart.vue
mwdchang: Okay, can you add a code comment? because it isn't entirely obvious.

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
mwdchang: @Jami159 @Tom-Szendrey this is marked as as resolved, do we already have a resolution here? Or do we know what to do for the next step?

mwdchang: Talked offline - it depends on the stats parsing logic, which we will need to amend/fix 
https://github.com/DARPA-ASKEM/Terarium/blob/e98ed653cf560db9afd0b5af7f9a24faff3802a3/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/DatasetResource.java#L305-L337

mwdchang (approved):

PR #1755, opened 2023-08-23
feat: pyodide service support

mwdchang:
### Summary
There are no new features, this reorganize the experiments into a more service-friendly manner, and get it into a state where it can be more readily used by the rest of the application.

image

### Testing
For now just in the test page
http://localhost:8078/app/#/pyodide-test


packages/client/hmi-client/src/temp/Pyodide.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+// Bootstrap
+const keys = Object.keys(variableMap);
+pyodide.runPython(`
+    ${keys.join(',')} = sympy.symbols('${keys.join(' ')}')
+`);
+
+const endLoad = Date.now();
+
+console.log(endLoad - startLoad);
+postMessage(true);
```
mwdchang, line 58: Probably makes sense the return the loading time (endload - startLoad)

YohannParis (approved):

PR #1756, opened 2023-08-23
Fix: show rate expression instead of rate name

blanchco:
# Description

* Show rate expression instead of rate name
* Did some post processing on the transition in the ACset before we send it to the acset-to-latex endpoint.
* expressions will show within brackets

Screenshot 2023-08-23 at 2 06 13 PM
Screenshot 2023-08-23 at 2 02 00 PM

Resolves #(issue)


packages/client/hmi-client/src/utils/math.ts
```diff
@@ -229,6 +229,7 @@ export const cleanLatexEquations = (equations: Array<string>): Array<string> =>
 				.replaceAll('\\left', '')
 				.replaceAll('{align}', '')
 				.replaceAll('=&', '=')
+				.replaceAll('*', ' *') // this fixes some parsing issues with variables such as '\beta' and '\gamma' with a '*' right after
```
mwdchang, line 232: Please be more specific with comments and intents: `this fixes some parsing issues...` is rather vague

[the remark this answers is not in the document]
mwdchang: This is not related to TA1

mwdchang:
So I noticed there is a slight discrepancy here, for example we have
- susceptible, versus
- susceptible(t)

Is this something we should address @liunelson

packages/client/hmi-client/src/utils/math.ts
blanchco: i made the comment clearer hopefully, essentially scientific variable such as \beta and \gamma would not be parsed if there was just '*' character following them, but it was fine with other math operators.

liunelson:
I have a few comments:
1. Why are we still using an ACSet-to-LaTeX converter, as opposed to an AMR-to-LaTeX converter? It took me a little to even find a ACSet to even test that model-service endpoint!
2. `(t)` in the returned LaTeX should probably be always be not shown since it is effectively redundant. The fact that an entity has a differential equation `d X / d t = f(X)` associated with it is sufficient to determine that it is time-dependent. Furthermore, I tested TA1's equation-to-AMR converter and it strips the `(t)` anyway. Showing the `(t)` would make the user type extra LaTeX than is necessary.
3. It is becoming increasingly untenable to permit parameters and variables to have IDs that are words (i.e. multi-character strings). Justin (developer for the TA1 equation-to-AMR converter) says that we should be only allowing IDs of the form `X_{abc}` (not `Xabc`, `abc`, `X_abc`, etc.). Else, we risk interconverting AMRs and equations that interpret a variable/parameter ID such as `poParam` into the product of `p, o, P, a, r, a, m`.
4. It is somewhat mentally taxing to read rendered equations and map them to un-rendered ones in the graph. Ideally, the equations in the graph should be rendered as well. Same goes with the labels of the state variables.
5. Generally, when typesetting math equations, multiplication is implicit - i.e. no `*`, maybe a non-breaking space. The asterisk or star symbol `*` is usually reserved for special operators like convolution or  complex conjugation. If we really want to show multiplication, the standard is to use `\cdot` (i.e. `·`) symbol.

blanchco:
Updates:
- Now we can no longer edit equations when a model is already present

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
```diff
@@ -73,7 +73,7 @@
 			<!-- Model equations -->
 			<AccordionTab header="Model equations">
 				<section :class="isEditingEQ ? `diagram-container-editing` : `diagram-container`">
-					<section v-if="props.isEditable" class="controls">
+					<section v-if="props.isEditable && props.isEquationsEditable" class="controls">
 						<Button
```
blanchco, line 77: the `isEditable` prop seems to encompass a wide variety of editability within the componenet.  We need a specific prop for when just equations themselves can be edited

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -301,8 +304,23 @@ watch(
...
 );
 
+function modelIsEmpty(model2: Model): boolean {
+	// model does not have any states or transitions
+	if (model2.model?.states?.length === 0 && model2.model?.transitions?.length === 0) {
+		return true;
+	}
+
+	return false;
+}
```
YohannParis:
```suggestion
function modelIsEmpty(model2: Model): boolean {
	// model does not have any states or transitions
	return model2.model?.states?.length === 0 && model2.model?.transitions?.length === 0;
}
```

packages/client/hmi-client/src/components/models/tera-model.vue
```diff
@@ -301,8 +304,23 @@ watch(
...
 async function fetchModel() {
 	model.value = await getModel(props.assetId);
+
+	// we can only edit equations when the AMR is empty
+	if (model.value && !modelIsEmpty(model.value)) {
+		canEditEquations.value = false;
+	} else {
+		canEditEquations.value = true;
+	}
```
YohannParis, line 323: Why not just reversing the test?
```suggestion
	// we can only edit equations when the AMR is empty
	canEditEquations.value = model.value && modelIsEmpty(model.value);
```

YohannParis (approved):

packages/client/hmi-client/src/components/models/tera-model-diagram.vue
mwdchang: Food for thoughts for @shawnyama

Issue #1757, opened 2023-08-23
[TASK]: Look into pyciemss-service and Rabbit MQ to sort out typing


Tom-Szendrey:
When Pyciemss is writing to RabbitMQ it is only receiving it as a btye[] 
This is different as to when we send it we can write it as an object. We can then read it back as a string deserialized into our object again

Tom-Szendrey:
You will need to update pyciemss-service. You will have some rabbit MQ library to read up on and there should likely be a flag for them to be writing as String rather than Byte[]

Once the pyciemss-service is updated you will need to update our hmi-server

https://github.com/DARPA-ASKEM/Terarium/blob/0b1249484ceb507fe9c196a136801c91c373736d/packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java#L168

1) This channel needs its type updated
partialSimulationStream
2) the stream function does too
3) the try catch stuff in the stream function can be removed as all of that typing will happen as we are reading it in instead

Tom-Szendrey:
Now that we are moving to Spring this seems irrelevant due to deserializing with ObjectMapper anyways

PR #1758, opened 2023-08-23
[Task]: Ts/1024 calibrate tool read from mq

Tom-Szendrey:
I have created an issue https://github.com/DARPA-ASKEM/Terarium/issues/1757 to check into the whole byte[] thing going on in this PR due to the pyciemss-service's output to MQ

In the meantime this functionally works:
image

Next steps will be to add a quick helper .ts file on hmi-client so using this poler can be more repeatable
as well as adding the progress bar itself


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -162,25 +164,43 @@ public Response createFromSimulationResult(
...
 	) {
-		return Multi.createFrom().publisher(partialSimulationResults).select().where(event -> event.getJobId().equals(jobId));
+		return Multi.createFrom().publisher(partialSimulationStream).filter(event -> {
+			try{ 
+				//TODO: https://github.com/DARPA-ASKEM/Terarium/issues/1757
+				String jsonString = new String(event);
+				jsonString = jsonString.replace(" ","");
+
+				ObjectMapper mapper = new ObjectMapper();
+				SimulationIntermediateResults interResult = mapper.readValue(jsonString, SimulationIntermediateResults.class);
```
mwdchang, line 177: Can we declare the mapper outside and reuse it to be more efficient? 
Or does it need to be a new instantiation for every single message?

packages/client/hmi-client/src/components/workflow/tera-calibration-node-ciemss.vue
```diff
@@ -237,20 +239,49 @@ const runCalibrate = async () => {
 	};
 	const response = await makeCalibrateJobCiemss(calibrationRequest);
 
+	// Run MQ Test
+	console.log(`Checking MQ for job Id ${response.simulationId}`);
+	const auth = useAuthStore();
+	const MQPoller = new EventSourcePolyfill(
+		`/api/simulations/${response.simulationId}/partial-result`,
```
mwdchang, line 246: Maybe rename `MQPoller` to something else to avoid confusion. We are not performing a polling-action or a request per se, it is sitting there waiting for messages to arrive from the sever-side.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Tom-Szendrey: Very fair point will do

mwdchang:
Didn't test. Code looks fine!

mwdchang (approved):

YohannParis (approved):
Looks good, but it would be better to tidy up this code into a util before merging and moving on to the next step

PR #1783, opened 2023-08-28
chore: event source changes

blanchco:
# Description

* A follow up PR to address some comments and clean up the code on this [issue]((https://github.com/DARPA-ASKEM/Terarium/pull/1683))

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/api/event-source-manager.ts
+++ b/packages/client/hmi-client/src/api/event-source-manager.ts
@@ -32,6 +32,12 @@ export class EventSourceManager {
 			};
 
 			eventSource.onerror = (error) => {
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
Lovely!

mwdchang:
Are we also handling the case that the connection will break when the auth expires? Or is that for another day?

blanchco:
> > Are we also handling the case that the connection will break when the auth expires? Or is that for another day?
> 
> We aren't right now, but I think I can make that change here, might not be too bad

@mwdchang addressed now with the latest commit, when we get an error with 401 status we close then reopen the connection

PR #1784, opened 2023-08-28
fix: simplify matrix modal

shawnyama:
# Description

- hide dropdowns for now as they don't work
- hide side panel
- fully qualified column/row headings

image


mwdchang (approved):
Looks fine, thanks!

PR #1785, opened 2023-08-28
chore: refactor services/project call out of components using composables

### LINKED ISSUE #1734: Refactor services/project call out of components into tera-project

echl:
# Description

* Use composable to pass project object to components
* Remove use of resources store to get project data

Resolves #1734


packages/client/hmi-client/src/composables/project.ts
```diff
@@ -0,0 +1,224 @@
...
+		if (projectId && !!projectId) {
+			activeProject.value = await ProjectService.get(projectId, true);
+		} else {
+			activeProject.value = null;
+		}
+		return activeProject.value;
+	}
+
+	async function getAllProjects(): Promise<IProject[]> {
+		allProjects.value = (await ProjectService.getAll()) as unknown as IProject[];
```
Tom-Szendrey, line 40: Seems like this should just be making the call rather than using the project service.
Should probably remove project service all together if im understanding correctly

This goes for all of these i think but this is a good example

Tom-Szendrey: Not sure if im making sense above. This section is just a duplication of the project service essentially.
Would be cleaner to have no duplication anywhere so if this getAll call needs to get updated we dont need to change it here as well as the service file

packages/client/hmi-client/src/composables/project.ts
```diff
@@ -0,0 +1,224 @@
...
+import * as ArtifactService from '@/services/artifact';
+import * as DatasetService from '@/services/dataset';
+import * as ModelService from '@/services/model';
+import { AssetType } from '@/types/Types';
+import DatasetIcon from '@/assets/svg/icons/dataset.svg?component';
+
+const TIMEOUT_MS = 1000;
+
+const activeProject = shallowRef<IProject | null>(null);
+const allProjects = shallowRef<IProject[] | null>(null);
```
mwdchang, line 14: What are the implications of using shallowRefs? Does this mean that any direct changes to the data structure (e.g. changing the description) are not reflected until we refetch the data again? 

It would be good if you can add some comments on the expected usage patterns to make the expectations more explicit.

packages/client/hmi-client/src/composables/project.ts
mwdchang: I have a somewhat similar thought here, it feels a little bit like in some cases this composable is over-stepping its boundaries, by accessing other services such as DataService, ModelService, CodeService etc etc. 

I am not really sure what to feel about that but definitely need to watch out for getting into situations of circular dependencies, and potentially overloading the composable with too many responsibilities. 

I guess I am somewhat ambivalent here but curious to hear other peoples' thoughts here.

echl: It's not exactly a duplication of ProjectService, ProjectService is (mostly) pure API calls whereas this composable contains stateful logic.

echl: @mwdchang regarding boundaries, fair point; the reason why I am calling other services is because every time we, for example, call `CodeService.uploadCodeToProject`, we are also adding the code asset to the project, so as a matter of convenience it might make sense to call one function that does that rather than calling two functions. But I can see how keeping them separate might be preferable.

packages/client/hmi-client/src/composables/project.ts
echl: Yes, that is correct, changes to the data description would not be reflected until the data is refreshed. Note that this is the existing paradigm as well, and in the previous implementation, `project` and `projects` in `App.vue` were also shallowRefs. 

However the idea here is that all the functions in `useProjects` guarantee that any changes to `activeProject` will be propagated to all of its users, so any changes to `activeProject` or its assets should be made through `useProjects` functions if you want to guarantee that those changes are propagated.

mwdchang: Okay, can you just add a brief comment to the code as such? Just so the readers know explicitly what are the intentions and what to expect.

echl: Done

YohannParis: You can mark as resolved yourself if it is done.

packages/client/hmi-client/src/composables/project.ts
```diff
@@ -0,0 +1,236 @@
...
+		getPublicationAssets,
+		getAssetIcon,
+		uploadCodeToProject,
+		uploadCodeToProjectFromGithub,
+		uploadArtifactToProject,
+		createNewArtifactFromGithubFile,
+		createNewDatasetFromCSV,
+		createNewDatasetFromGithubFile,
+		addNewModelToProject,
+		saveDatasetFromSimulationResultToProject
```
YohannParis, line 234: Some of those methods have the word `project` and some don't.
Could you uniform this by just removing it everywhere. 
I expect that we are going to use them with the pattern `useProject.active` for example. So it will be easier to read.

packages/client/hmi-client/src/composables/project.ts
YohannParis: I agree with @mwdchang, I don't think `useProject` should do that much.

i.e. If someone wants to save a Dataset and add it to a project, they should do:
```Javascript
const datasetId = useDataset.create();
useProject.addAsset(datasetId); // If the projectId is omitted, we use the active project
```

packages/client/hmi-client/src/services/model.ts
```diff
@@ -48,11 +46,11 @@ export async function getAllModelDescriptions(): Promise<Model[] | null> {
 	return response?.data ?? null;
 }
 
-export async function updateModel(model: Model) {
+export async function updateModel(model: Model, projectId?: IProject['id']) {
 	const response = await API.put(`/models/${model.id}`, model);
 	EventService.create(
 		EventType.PersistModel,
-		useResourcesStore().activeProject?.id,
+		projectId,
```
YohannParis, line 53: Why not using `useProject.active.value.id` in this case?

packages/client/hmi-client/src/services/model.ts
```diff
@@ -74,21 +72,13 @@ export async function reconstructAMR(amr: any) {
 }
 
 // function adds model to project, returns modelId if successful otherwise null
-export async function addNewModelToProject(
-	modelName: string,
-	project: IProject
-): Promise<string | null> {
+export async function addNewModelToProject(modelName: string): Promise<string | null> {
```
YohannParis, line 75: Where is the part which we add the model to the project?

packages/client/hmi-client/src/services/code.ts
```diff
@@ -118,10 +109,6 @@ async function uploadCodeToProjectFromGithub(
 		return null;
 	}
 
-	const resp = addAsset(projectId, AssetType.Code, newCode.id);
```
YohannParis, line 121: I think you removed a lot of those by making the assumptions that we want to always save an asset to a project, which might not be the case in the future.

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -203,14 +201,17 @@ import TabView from 'primevue/tabview';
...
 import useAuthStore from '@/stores/auth';
 import { RouteName } from '@/router/routes';
 import Skeleton from 'primevue/skeleton';
 import { isEmpty } from 'lodash';
 import TeraProjectCard from '@/components/home/tera-project-card.vue';
+import { useProjects } from '@/composables/project';
+import { IProject } from '@/types/Project';
 import Dropdown from 'primevue/dropdown';
 
+const { allProjects, getAllProjects, getPublicationAssets, create } = useProjects();
```
YohannParis, line 213: It willl be clearer to keep the `useProjects` in front of every methods call.

packages/client/hmi-client/src/composables/project.ts
```diff
@@ -0,0 +1,236 @@
...
+	async function getPublicationAssets(projectId: IProject['id']) {
+		return ProjectService.getPublicationAssets(projectId);
+	}
+
+	function getAssetIcon(type: AssetType | string | null): string | Component {
+		if (type && icons.has(type)) {
+			return icons.get(type) ?? 'circle';
+		}
+		return 'circle';
+	}
```
YohannParis, line 93: I'm not sure this needs to be in a composable, a util would do fine.

packages/client/hmi-client/src/components/widgets/tera-import-github-file.vue
```diff
@@ -188,18 +188,22 @@ import { getModeForPath } from 'ace-builds/src-noconflict/ext-modelist';
...
 	urlString: string;
 	showImportButton: boolean;
 	project?: IProject;
 }>();
 
+const {
+	uploadCodeToProjectFromGithub,
+	createNewArtifactFromGithubFile,
+	createNewDatasetFromGithubFile
+} = useProjects();
```
YohannParis, line 205: Those should not be in `useProject`, but in `useDataset`, `useArtifact`, `useCode` etc.

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -160,10 +161,10 @@ const optionsMenuItems = ref([
 ]);
 
 async function updateModelContent(updatedModel: Model) {
-	await updateModel(updatedModel);
+	await updateModel(updatedModel, activeProject.value?.id);
```
YohannParis, line 164: Why do you need to pass the project here?

packages/client/hmi-client/src/components/llm/tera-beaker-input.vue
```diff
@@ -72,18 +74,14 @@ const addCodeCellButton = ref<HTMLElement | null>(null);
 const fixedDivWidth = ref(0);
 
 const submitQuery = () => {
-	EventService.create(
-		EventType.TransformPrompt,
-		useResourcesStore().activeProject?.id,
-		queryString.value
-	);
+	EventService.create(EventType.TransformPrompt, activeProject.value?.id, queryString.value);
```
YohannParis, line 77: We should remove the active project from the call, and instead call `useProject.activeProject.value.id` in the `EventService.create()` method.

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -386,7 +388,7 @@ async function updateDatasetName() {
 		datasetClone.name = newDatasetName.value;
 		await updateDataset(datasetClone);
 		dataset.value = await getDataset(props.assetId);
-		useResourcesStore().setActiveProject(await ProjectService.get(props.project.id, true));
+		getActiveProject(props.project.id);
```
YohannParis, line 391: Do we need to do that? Should we just update our active project assets lost with the new name of the Dataset once we got the response that the name has changed? @mwdchang what do you think, for things like names, does it matter?

packages/client/hmi-client/src/composables/project.ts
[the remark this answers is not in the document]
echl: I considered this, but actually there is a difference between `[]` and `null`. `null` means the value of `allProjects` has not been loaded yet, whereas `[]` means the value has been loaded, but there are no projects.

packages/client/hmi-client/src/composables/project.ts
echl: I would suggest keeping `project` rather than getting rid of it. If we renamed `activeProject` to `active` for example, we would have

```const { active } = useProject()```

which would be more confusing in my opinion, since the context of `active` would not immediately obvious.

File: packages/client/hmi-client/src/page/data-explorer/components/tera-selected-resources-header-pane.vue [hunk 13]
```diff
@@ -62,57 +60,42 @@ const addResourcesToProject = async (projectId: string) => {
...
-	let projectId = '';
-	if (projectName !== undefined && typeof projectName.value === 'string') {
-		const project = projectsList.value.find((p) => p.name === projectName.value);
-		projectId = project?.id as string;
-	} else {
-		if (!resources.activeProject) return;
-		projectId = resources.activeProject.id;
-	}
-
+	const projectId = projectOption.id ?? activeProject.value?.id;
```
shawnyama, line 93: This wasn't working for me, you need to change it to `projectOption.value.id` for it to work

packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
```diff
@@ -408,14 +411,19 @@ async function processArtifact(file: File) {
...
+	let addedCSV: CsvAsset | null = null;
+	const addedDataset: Dataset | null = await createNewDatasetFromCSV(
 		progress,
 		file,
 		auth.name ?? '',
 		props.project.id,
 		description
 	);
 
+	if (addedDataset && addedDataset.id) {
```
shawnyama, line 423: `addedDataset?id` would be shorter

packages/client/hmi-client/src/services/model.ts
shawnyama: Found it in the composable `model.ts`. It calls this as its step 1

packages/client/hmi-client/src/composables/project.ts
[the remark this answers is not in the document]
YohannParis: You mean for the loading status?

packages/client/hmi-client/src/services/model.ts
YohannParis: This ain't pretty at all.

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
YohannParis: I'm saying, let's not make another call to refresh the whole project, and just update the local object. Which does the same job on the front-end, but only to what matter.

mwdchang: I think this has to do with the shallow-ref vs regular-ref discussion here https://github.com/DARPA-ASKEM/Terarium/pull/1785#discussion_r1324662310

I don't have a strong feeling towards one or the other, as long as the mechanism is well understood by all.

echl: @YohannParis We shouldn't update the local object because then it may get out of sync - precisely what we're trying to fix. In fact, there doesn't even need to be a local object, and I should just delete the `project` prop. `activeProject` needs to be refreshed so that any changes are propagated to all its users.

packages/client/hmi-client/src/components/llm/tera-beaker-input.vue
echl: I think we should keep it actually as there are some places where this is used where the project might not be the active project

packages/client/hmi-client/src/services/code.ts
echl: I removed this addAsset call because I wasn't sure if `code.ts` should be doing `useProjects().addAsset` inside `uploadCodeToProjectFromGithub`, or whether the caller of `uploadCodeToProjectFromGithub` should subsequently call `useProjects().addAsset`.

e.g.

```
CodeService.uploadCodeToProjectFromGithub() //method implicitly adds the asset
```
vs.

```
CodeService.uploadCodeToProjectFromGithub() //doesn't add the asset
useProjects().addAsset() 
```

packages/client/hmi-client/src/services/model.ts
echl: In my opinion this function should not even be in `model.ts`, it should be in `tera-model-modal` where it is being used.

Currently `tera-modal-modal` is calling `model.ts`, which calls 2 other services (petrinet service and project service) inside of it. I think `tera-modal-model` should do something like 

```
PetrinetService.newAmr();
ModelService.createModel();
useProjects().addAsset();
```
That way we limit the side effects hidden inside method calls.

packages/client/hmi-client/src/services/dataset.ts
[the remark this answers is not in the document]
echl: Which line is this for?

PR #1788, opened 2023-08-28
fix: update timestamp util for calibrate input dataset

Jami159:
Update timestamp util


packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -5,12 +5,26 @@ export const enum RunType {
...
 	let end = inputTimespan.end;
 	// If we have the min/max timestamp available from the csv asset use it
+	// Try to get `timestamp` first, then `t`
+	// TODO: figure out a more robust way to get the timestamp column from the csv asset
 	if (dataset) {
-		const tIndex = dataset.headers.indexOf('timestamp');
+		let tIndex = dataset.headers.indexOf('timestamp');
+		if (tIndex === -1) {
+			tIndex = dataset.headers.indexOf('t');
+		}
```
mwdchang, line 27: You should have and leverage a mappings data structure provided to you that tells you which column is the timestamp column.

packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -5,12 +5,26 @@ export const enum RunType {
...
 
-export const getTimespan = (inputTimespan: TimeSpan, dataset?: CsvAsset) => {
+export type DatasetType = {
+	data: { x: number; y: number }[];
+	label: string;
+	fill: boolean;
+	borderColor?: string;
+	borderWidth?: number;
+	borderDash?: number[];
+};
```
mwdchang, line 15: The naming does not reflect the purpose of this data structure. 

Probably more appropriate as `DataseriesConfig` or something?
Also consider moving the definition over to types/workflow.ts or types/SimulateConfig.ts

Jami159: name updated and moved to `SimulateConfig.ts`

packages/client/hmi-client/src/components/workflow/util.ts
Jami159: Done!

PR #1794, opened 2023-08-29
feat: model enrichment

shawnyama:
# Description

- Refactored model description into `tera-model-description` - it is compatible with the old and new `tera-models`
- Model enrichment endpoints work

(old video model refresh works and UNKNOWNS are now --)

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/b6608597-85e1-4b50-8580-cc1df294ddcc


packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
```diff
@@ -5,71 +5,616 @@
...
+						{{ card?.AUTHOR_AUTHOR }}
+						<template v-if="model?.metadata?.annotations?.authors">
+							, {{ model.metadata.annotations.authors.join(', ') }}
+						</template>
+					</td>
+					<td>{{ card?.AUTHOR_EMAIL }}</td>
 					<td>{{ model?.metadata?.processed_by }}</td>
+					<td>{{ card?.AUTHOR_INST }}</td>
+					<td>{{ card?.LICENSE }}</td>
+					<td>{{ card?.COMPLEXITY }}</td>
```
YohannParis, line 34: Don't use the `card` anymore as you are refreshing the model and reading those values from `model.metadata.card` now.

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -102,7 +117,7 @@ export const profileDataset = async (datasetId: string, artifactId: string | nul
 	} else {
 		response = await API.post(`/knowledge/profile-dataset/${datasetId}`);
 	}
-	console.log('data profile response', response.data);
+	console.log('data profile response', response);
```
YohannParis, line 120: Why did you remove the `data` here?

YohannParis (approved):

packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
shawnyama: `card` is just a computed version of `model.metadata.card` without the `UNKNOWN` strings

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/metadata/Card.java
```diff
@@ -0,0 +1,54 @@
...
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+import software.uncharted.terarium.hmiserver.annotations.TSOptional;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@Data
+@Accessors(chain = true)
+public class Card {
+    @JsonProperty("DESCRIPTION")
```
dvince2, line 11: Unless intentional, you'll probably want to use `JsonAlias` here. When re-serialized these will go out as captials, etc, instead of camel case

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -463,6 +464,20 @@ export interface Annotations {
     model_types?: string[];
 }
 
+export interface Card {
+    DESCRIPTION?: string;
```
dvince2, line 468: These are captials I believe because of jsonProperty

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -76,21 +76,36 @@ const mathmlToAMR = async (mathml: string[], framework = 'petrinet'): Promise<Mo
...
+ * Returns a runId used to poll for result
+ */
+export const profileModel = async (modelId: string, artifactId: string | null = null) => {
+	let response: any = null;
+	if (artifactId) {
+		response = await API.post(`/knowledge/profile-model/${modelId}?artifact_id=${artifactId}`);
+	} else {
+		response = await API.post(`/knowledge/profile-model/${modelId}`);
+	}
+	console.log('model profile response', response.data);
```
dvince2, line 105: debug log

packages/client/hmi-client/src/types/Types.ts
shawnyama: Yeah, they come in as capitals 
https://github.com/DARPA-ASKEM/knowledge-middleware/blob/main/tests/scenarios/basic/model_card.json

packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
YohannParis: The naming convention are not great here.

packages/client/hmi-client/src/types/Types.ts
YohannParis: This is the kind of things we don't have time to clean, but we should...
Maybe we will have some magic time to do so.

dvince2: These are capitals though because the wrong annotation is being used - im not sure if we're going to run into serialization issues on the back end because of it

PR #1798, opened 2023-08-29
feat: fix parameters over time

Tom-Szendrey:
When a model has a timeseries we should be using it in our ciemss simulation 

Sample timeseries:
image

image


packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -20,12 +22,19 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.jboss.logging.Logger;
 
 @Path("/api/simulation-request")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
 @Tag(name = "Simulation Service REST Endpoint")
 public class SimulationRequestResource {
+	private static final Logger log = Logger.getLogger(SimulationRequestResource.class);
```
Tom-Szendrey, line 37: Not sure why i need to add this?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -81,6 +93,7 @@ public Simulation makeForecastRun(
 	public Simulation makeForecastRunCiemss(
 		final SimulationRequest request
 	) {
+		request.setInterventions(getIntervention(request.getModelConfigId()));
```
Tom-Szendrey, line 96: Should probably check if client is sending this before overwriting it

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/SimulationRequest.java
```diff
@@ -22,4 +23,7 @@ public class SimulationRequest implements Serializable {
 	private Object extra;
 
 	private String engine;
+
+	@TSOptional
+	private List<Intervention> interventions;
```
Tom-Szendrey, line 28: Should this class be split into Julia and Ciemss? Im not sure if both take this parameter...

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -177,5 +192,42 @@ public JobResponse makeEnsembleCalibrateCiemssJob(
...
+			JsonNode configuration =  mapper.convertValue(modelConfig.getConfiguration(), JsonNode.class);
+			if (configuration.findValue("timeseries") != null){
+				JsonNode timeseries = mapper.convertValue(configuration.findValue("timeseries"), JsonNode.class);
+				List<String> fieldNames = new ArrayList<>();
+				timeseries.fieldNames().forEachRemaining(key -> fieldNames.add(key));
+				for (int i = 0; i < fieldNames.size(); i++){
+					// Eg) Beta
+					String interventionName = fieldNames.get(i).replaceAll("\"",",");
+					// Eg) "1:0.14, 10:0.1, 20:0.2, 30:0.3"
+					String tempString = timeseries.findValue(fieldNames.get(i)).toString().replaceAll("\"","").replaceAll(" ","");
```
mwdchang, line 213: what are we trying to do here? If you need to resort to regex here it usually means we are on the wrong path.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -177,5 +192,42 @@ public JobResponse makeEnsembleCalibrateCiemssJob(
...
+	//Check if it has timeseries in its metadata
+	//If it does for each element convert it to type Intervention and add it to this.interventions
+	//Schema: http://json-schema.org/draft-07/schema#
+	public List<Intervention> getInterventionFromId(String modelConfigId){
+		List<Intervention> interventionList = new ArrayList<>();
+		try {
+			ObjectMapper mapper = new ObjectMapper();
+			ModelConfiguration modelConfig = modelConfigProxy.getModelConfiguration(modelConfigId);
+			JsonNode configuration =  mapper.convertValue(modelConfig.getConfiguration(), JsonNode.class);
+			if (configuration.findValue("timeseries") != null){
```
mwdchang, line 205: Can we be more specific here? There are a number of things that can also be called "timeseries" in the overall structure of the model, since "groundings" and "modifiers" field are kind of freeform it might be the case that they can contain "timeseries" as well.

Tom-Szendrey: Good call I can go through the full path to get to the correct key

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
mwdchang: I thought we are using that `@slf4j `stuff

dvince2: ^Echoing @mwdchang here. That is *a* way to do that... but you should be able to just add with the annotation

Tom-Szendrey: Thanks guys

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
Tom-Szendrey: Parse this junk which at some point needs a redesign
image

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -177,5 +192,42 @@ public JobResponse makeEnsembleCalibrateCiemssJob(
 		return res;
 	}
 
-
+	//Get modelConfigId
+	//Check if it has timeseries in its metadata
+	//If it does for each element convert it to type Intervention and add it to this.interventions
+	//Schema: http://json-schema.org/draft-07/schema#
+	public List<Intervention> getInterventionFromId(String modelConfigId){
```
dvince2, line 199: Not a blocker to this PR at all but if this is only being used internally then convention would be to make this `private`

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
```diff
@@ -177,5 +192,42 @@ public JobResponse makeEnsembleCalibrateCiemssJob(
...
+		List<Intervention> interventionList = new ArrayList<>();
+		try {
+			ObjectMapper mapper = new ObjectMapper();
+			ModelConfiguration modelConfig = modelConfigProxy.getModelConfiguration(modelConfigId);
+			JsonNode configuration =  mapper.convertValue(modelConfig.getConfiguration(), JsonNode.class);
+			if (configuration.get("metadata").get("timeseries") != null){
+				JsonNode timeseries = mapper.convertValue(configuration.get("metadata").get("timeseries"), JsonNode.class);
```
dvince2, line 206: Because all of these operations are being done on just a `JsonNode` created from another `JsonNode` its a little hard to follow whats happening or what the replacements are doing/for. A comment or two may be good here.. but again not going to block the PR for it

dvince2 (approved):

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
Tom-Szendrey: Okay thats a very fair point

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/simulationservice/SimulationRequestResource.java
Tom-Szendrey: Fair ill leave some comments.
This will all be torn out post redesign as the underlying data structure will be fixed up

PR #1799, opened 2023-08-30
feat: dataset operator node

blanchco:
# Description

* adding a dataset operator node to use the LLM to perform dataset operations on multiple datasets
* notebook history is saved
 please see this [slack message](https://unchartedsoftware.slack.com/archives/C03S02YH3D5/p1693606895129989) for the video

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/tera-dataset-transformer.vue
```diff
@@ -0,0 +1,85 @@
...
+	node: WorkflowNode;
+	project: IProject;
+}>();
+const showKernels = ref(<boolean>false);
+const showChatThoughts = ref(<boolean>false);
+const assetIds = computed(() =>
+	props.node?.inputs
+		.filter((inputNode) => inputNode.status === WorkflowPortStatus.CONNECTED && inputNode.value)
+		.map((inputNode) => inputNode.value![0])
+);
```
mwdchang, line 40: Should leverage the notion of `WorkflowNode.State`

mattprintz (commented):
Looks fine to me. I still think that we should split out the preview data table widget from the other data table widget since they are definitely starting to diverge, but I'm fine with pushing that down the road.
Also, I think the ability to name/map the datasets to variables in the workflow cell will be important functionality, but this is fine as a first step.

mwdchang (approved):

PR #1805, opened 2023-08-31
Eval transitions

### LINKED ISSUE #1765: [TASK]: Evaluate and change expressions in configuration matrix
We should be able to replace the expression with:
- another expression, or
- a literal numerical value

We should find a way to display both the 
- expression, and
- the numerical equivalent 

in the matrix cell if applicable

This task requires first-class symbolic support from the Pyodide/SymPy experiments

mj3cheun:
# Description

Screenshot 2023-08-31 at 12 48 07 AM
Screenshot 2023-08-31 at 12 47 50 AM

Resolves #1765

### DIFF
--- a/packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
+++ b/packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
@@ -96,11 +96,10 @@ watch(
 				.map((row) =>
 					row.map(async (cell) => {
 						if (cell.value?.id) {
-							const matrixVal = await getMatrixValue(cell.value.id, props.shouldEval);
 							if (!output[cell.row]) {
 								output[cell.row] = [];
 							}
-							output[cell.row][cell.col] = matrixVal;
+							output[cell.row][cell.col] = await getMatrixValue(cell.value.id, props.shouldEval);
 						}
 					})
 				)
--- a/packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
+++ b/packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
@@ -154,6 +154,7 @@ function getMatrixExpression(variableName: string) {
 	return variableName;
 }
 
+// Returns the presentation mathml
 async function getMatrixValue(variableName: string, shouldEvaluate: boolean) {
 	const expressionBase = getMatrixExpression(variableName);
 
@@ -162,10 +163,10 @@ async function getMatrixValue(variableName: string, shouldEvaluate: boolean) {
 			expressionBase,
 			parametersValueList.value
 		);
-		return (await pythonInstance.parseExpression(expressionEval)).mathml;
+		return (await pythonInstance.parseExpression(expressionEval)).pmathml;
 	}
 
-	return (await pythonInstance.parseExpression(expressionBase)).mathml;
+	return (await pythonInstance.parseExpression(expressionBase)).pmathml;
 }
 
 async function updateModelConfigValue(variableName: string, rowIdx: number, colIdx: number) {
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -73,13 +78,45 @@ const chosenRow = ref('');
...
+							}
+							output[cell.row][cell.col] = matrixVal;
+						}
+					})
+				)
+			)
+		);
+		matrixExpressionsList.value = output;
+	}
+);
```
mwdchang, line 111: Slight preference if we don't have multiple layers of promise alls, is it possible to do something like?

```
matrix.value.map(row => {
  row.map(cell => {
    matrixExpressionList.value[row][col] = await getMatrixValue(...)  
  })
})
```

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -73,13 +78,45 @@ const chosenRow = ref('');
...
 const editableCellStates = ref<boolean[][]>([]);
 
+const matrixExpressionsList: any = ref([]);
+
+const parametersValueList = computed(() =>
+	props.modelConfiguration.configuration?.semantics.ode.parameters.reduce((acc, val) => {
+		acc[val.id] = val.value;
+		return acc;
+	}, {})
+);
```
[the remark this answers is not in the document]
mwdchang: never mind this, got confused with something else.

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
mj3cheun: fixed

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -73,13 +78,45 @@ const chosenRow = ref('');
...
+		await Promise.all(
+			matrix.value.map(async (row) =>
+				Promise.all(
+					row.map(async (cell) => {
+						if (cell.value?.id) {
+							const matrixVal = await getMatrixValue(cell.value.id, props.shouldEval);
+							if (!output[cell.row]) {
+								output[cell.row] = [];
+							}
+							output[cell.row][cell.col] = matrixVal;
```
YohannParis, line 103:
```suggestion
							if (!output[cell.row]) {
								output[cell.row] = [];
							}
							output[cell.row][cell.col] = await getMatrixValue(cell.value.id, props.shouldEval);
```

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
```diff
@@ -118,6 +155,28 @@ function getMatrixValue(variableName: string) {
...
+async function getMatrixValue(variableName: string, shouldEvaluate: boolean) {
+	const expressionBase = getMatrixExpression(variableName);
+
+	if (shouldEvaluate) {
+		const expressionEval = (await pythonInstance.evaluateExpression(
+			expressionBase,
+			parametersValueList.value
+		)) as string;
+		return (
+			(await pythonInstance.parseExpression(expressionEval)) as {
```
YohannParis: Why not update create a `parseExpressionToMathML()` that return a `string` directly, so you don't have to do this weird typing `( .... as { mathml: string; }).mathml;`

packages/client/hmi-client/src/python/PyodideWorker.ts
```diff
@@ -84,9 +84,16 @@ const parseExpression = (expr: string) => {
 	// function to convert expression to mathml
 	let result = pyodide.runPython(`
 		eq = sympy.S("${expr}", locals=_clash)
-		sympy.mathml(eq)
+		sympy.mathml(eq, printer="presentation")
```
YohannParis, line 87: Is this a parameters to choose between _presentation_ and _content_ MathML?
If yes, could you add a small comment like: 
```suggestion
		sympy.mathml(eq, printer="presentation") // printer refers to the type of MathML used "presentation" or "content"
```

mwdchang: oh it got changed to presentation mathML? That is going to have a problem when we save expressions, so I guess we need the function to return both.

packages/client/hmi-client/src/components/models/tera-stratified-value-matrix.vue
mj3cheun: this has been fixed by updating the `parseExpression` type

packages/client/hmi-client/src/python/PyodideWorker.ts
mwdchang: addressed this

mwdchang (approved):

PR #1808, opened 2023-08-31
chore: fix code to model

### LINKED ISSUE #1771: [TASK] code to model
**Describe the task**
- [x] Finish the code to model integration within the `knowledge-middleware`.
- [x] Make sure that the process is working at least on the test.
- [ ] Implement it back onto the HMI by cleaning up and fixing `tera-code-editor`.

echl:
# Description

* Implement new `tera-code` component that implements code-to-model feature
* Kept the `code-editor` component for displaying .pdf files, for now
* Saving model currently not working due to recent model changes

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/46c4c2a4-b569-44c2-8090-d747b56ab1fc

Resolves #1771


packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -179,4 +179,23 @@ export const extractPDF = async (artifact: Artifact) => {
...
+		}
+		if (status === 'finished') {
+			return response.data.result?.job_result.amr;
+		}
+	}
+	logger.error(`Code to AMR request failed`, { toastTitle: 'Error - ta1-service' });
+	return null;
+}
+
 export { mathmlToAMR, latexToAMR };
```
YohannParis, line 201: You should either add your function to this export, or remove this line.

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -179,4 +179,23 @@ export const extractPDF = async (artifact: Artifact) => {
...
+			if (extraction?.state === PollerState.Done) {
+				const data = extraction.data as any; // fix linting
+				return data?.job_result.amr;
+			}
+		}
+		if (status === 'finished') {
+			return response.data.result?.job_result.amr;
+		}
+	}
+	logger.error(`Code to AMR request failed`, { toastTitle: 'Error - ta1-service' });
```
YohannParis, line 197:
```suggestion
	logger.error(`Code to AMR request failed`, { toastTitle: 'Error - knowledge-middleware' });
```

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -179,4 +179,23 @@ export const extractPDF = async (artifact: Artifact) => {
 	}
 };
 
+export async function codeToAMR(codeId: string) {
+	const response = await API.post(`/extract/code-to-amr?code_id=${codeId}`);
```
YohannParis, line 183: if `name` and `description` are optional, why not remove them from our backend?
Should we let the user set them up when clicking on the call?
```suggestion
	const response = await API.post(`/knowledge/code-to-amr?code_id=${codeId}`);
```

packages/client/hmi-client/src/services/code.ts
```diff
@@ -4,6 +4,39 @@ import { Ref } from 'vue';
...
+		return null;
+	}
+
+	return response.data as Code;
+}
+
+async function updateCodeAsset(
+	code: Code,
+	file: File,
+	progress: Ref<number>
```
YohannParis, line 21: What is this `progress`?

packages/client/hmi-client/src/page/project/components/tera-project-page.vue
```diff
@@ -5,14 +5,12 @@
...
+	<tera-code
+		:project-id="project.id"
+		:asset-id="assetId ?? ''"
 		v-else-if="pageType === AssetType.Code"
-		:initial-code="code"
-		@vue:mounted="
-			emit('asset-loaded');
-			openCode();
-		"
+		@vue:mounted="() => emit('asset-loaded')"
```
YohannParis, line 12: Why is this necessary?

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -0,0 +1,264 @@
...
+				modelName.value = extractedAmr.header.name;
+				isModalVisible.value = true;
+			}
+		}
+	}
+}
+
+async function saveModel() {
+	const modelToSave = model.value;
+	const createdModel = await createModel(modelToSave);
```
YohannParis, line 166:
```suggestion
	const createdModel = await createModel(model.value);
```

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -0,0 +1,264 @@
+<template>
+	<main>
+		<header>
+			<section class="header">
+				<!-- <h4>{{ name }}</h4> -->
+				<section class="name">
+					<InputText v-model="codeName" class="name-input" />
+				</section>
```
YohannParis, line 8: This should be a conditional if the asset as a name.

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -0,0 +1,264 @@
...
+			<tera-modal
+				v-if="isModalVisible && model"
+				class="modal"
+				@modal-mask-clicked="isModalVisible = false"
+				@modal-enter-press="isModalVisible = false"
+			>
+				<template #header>
+					<InputText v-model="modelName" class="name-input" />
+				</template>
+				<template #default>
```
YohannParis, line 42: This is where you can add a `name` and `description` field. or even better re-use the create a new model modal.

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis, line 165: See comment in the resource endpoint about not needing to create a model

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/knowledgemiddleware/KnowledgeResource.java
```diff
@@ -106,16 +106,11 @@ public Model postLaTeXToAMR(
...
+		@QueryParam("name") String name,
+		@QueryParam("description") String description
 	) {
-		// Fetch the related artifact to fill potential missing name and description
-		final Artifact artifact = artifactProxy.getArtifact(artifactId);
-		if (name == null) {	name = artifact.getName(); }
-		if (description == null) { description = artifact.getDescription();	}
-
-		return knowledgeMiddlewareProxy.postCodeToAMR(artifactId, name, description);
+		return knowledgeMiddlewareProxy.postCodeToAMR(codeId, name, description);
```
YohannParis, line 113: [It returns more](https://github.com/DARPA-ASKEM/knowledge-middleware/blob/main/worker/operations.py#L509-L515) than just a `Model`.
This would be useful to maybe display the model directly by open it instead of returning the AMR?

packages/client/hmi-client/src/components/code/tera-code.vue
echl: I wanted to test something out here; I think the title (if editable) should always be an InputText. What's the point of having a separate edit button and having extra conditional logic? What do you think @jamiewaese-uncharted 

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/2af113c7-76db-4a57-95dd-e95ea8563c9c

packages/client/hmi-client/src/components/code/tera-code.vue
echl: Good catch, leftover from when I had to wrangle the value of modelToSave since it had `header` in it

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis: I don't think we should do that. If a user click on it by inadvertence and type keys that get saved automatically. This will be really error prone.

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis, line 1: This component should use `tera-asset`, and it is missing here.

packages/client/hmi-client/src/services/code.ts
echl: It's required in `addFileToCodeAsset` which is being called in this method

YohannParis: Then why are we not displaying it in the button?
If this would take too long, could you write an issue to add this feature please?

packages/client/hmi-client/src/components/code/tera-code.vue
echl: It seems very unlikely to me that a user would actually accidentally click it. If they do, anything they accidentally type is very noticeable, and the consequences are trivial and easy to undo (the name does not automatically save when you focus out of the element). 

Whereas having to click a small edit button is much more annoying, and the user will face this annoyance almost every time.

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -0,0 +1,264 @@
...
+		if (newCodeAsset && newCodeAsset.id) {
+			const extractedAmr = await codeToAMR(newCodeAsset.id);
+			if (extractedAmr) {
+				model.value = extractedAmr as Model;
+				modelName.value = extractedAmr.header.name;
+				isModalVisible.value = true;
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
echl: Already fixed, I just got rid of the condition. It should save every time you click extract in case content has changed (and to save user a click)

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/knowledgemiddleware/KnowledgeResource.java
echl: Not sure what you mean. The model already gets saved in tds as a result of calling code-to-amr?

packages/client/hmi-client/src/components/code/tera-code.vue
echl: What does `tera-asset` have in it that `tera-code` needs? I don't think `tera-code` is using any of the functionality inside `tera-asset`

YohannParis: `tera-asset` is the component that all asset uses to derive the same basic behaviour, so yes it needs it.

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -179,4 +185,21 @@ export const extractPDF = async (artifact: Artifact) => {
...
+			if (extraction?.state === PollerState.Done) {
+				const data = extraction.data as any; // fix linting
+				return data?.job_result.amr;
+			}
+		}
+		if (status === 'finished') {
+			return response.data.result?.job_result.amr;
+		}
+	}
+	logger.error(`Code to AMR request failed`, { toastTitle: 'Error - knowledge-middlewareõ' });
```
YohannParis, line 203:
```suggestion
	logger.error(`Code to AMR request failed`, { toastTitle: 'Error - knowledge-middleware' });
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/knowledgemiddleware/KnowledgeResource.java
YohannParis: Right now we use the response to display the AMR, but we still don't open it, there is a `createModel` call made after this.

packages/client/hmi-client/src/components/code/tera-code.vue
ngraham76: Is the title of an asset changed often? If not, I would think we should stick with the current design of clicking an edit button to enable the input field to avoid inadvertently re-naming the assets as Yohann mentioned. Thoughts @jamiewaese-uncharted and @pascaleproulx ?

jamiewaese-uncharted: Maybe we can use the PrimeVue `inplace` component here: https://primevue.org/inplace/#basic

It would look sorta like this, but we use a heading font size, and we add an 'X' button to cancel and a checkmark button to confirm.

vue inplace

packages/client/hmi-client/src/components/code/tera-code.vue
echl: Could we possibly address this in another pr? Looking at `tera-asset` I honestly don't think there is any important functionality that is missing without it in this particular case.

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis: Sure @echl if you could make a ticket of it.

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis: ... Okay write the issue, but we need to do it, even if for now it's not needing any functionality.
When we move to multiple tabs, SSE, windows, we will be working against `tera-asset` and not each sub-component.

YohannParis (approved):

PR #1810, opened 2023-09-01
chore: amr header fixes

mwdchang:
### Summary
This address various parts that need to be converted to use `amr.header` format.

Note the tabs/summary portions are rather janky, they need to be re-addressed by having a better unified summary endpoint instead of doing specialized parsing.

### DIFF
--- a/packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
+++ b/packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
@@ -551,13 +551,13 @@ export const updateExistingModelContent = (amr: Model, amrOld: Model): Model => 
 export const cloneModelWithExtendedTypeSystem = (amr: Model) => {
 	const amrCopy = cloneDeep(amr);
 	if (amrCopy.semantics?.typing) {
-		const { name, description, schema, semantics } = amrCopy;
+		const { name, description, schema } = amrCopy.header;
 		const typeSystem = {
 			name,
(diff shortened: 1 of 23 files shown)


packages/client/hmi-client/src/model-representation/petrinet/petrinet-service.ts
```diff
@@ -551,13 +551,13 @@ export const updateExistingModelContent = (amr: Model, amrOld: Model): Model =>
 export const cloneModelWithExtendedTypeSystem = (amr: Model) => {
 	const amrCopy = cloneDeep(amr);
 	if (amrCopy.semantics?.typing) {
-		const { name, description, schema, semantics } = amrCopy;
+		const { name, description, schema } = amrCopy.header;
 		const typeSystem = {
```
Jami159, line 555: @mwdchang do you know what this `typeSystem` is used for? Does it need to follow the structure of
```
{
  header: {...},
  model: semantics?.typing?.system
}
```

mwdchang: This is @echl 's work on Catlab stratification, so he may be better to answer here if it should be addressed.

We are kind pivoting away from using Catlab stratify in favour of using MIRA stratify so this is not high priority.

YohannParis (approved):
Thanks a lot !

dvince2 (approved):

PR #1812, opened 2023-09-01
Equations to model

### LINKED ISSUE #1770: [TASK] equations to model
**Describe the task**
- [ ] Finish the equations to model integration within the `knowledge-middleware`.
- [ ] Make sure that the process is working at least on the test.
- [ ] Implement it back onto the HMI.

Jami159:
# Description

* Update the api to correctly call `/equations-to-amr` endpoint

Resolves #1770 
Screenshot 2023-09-01 at 3 35 44 PM

### DIFF
--- a/packages/client/hmi-client/src/services/knowledge.ts
+++ b/packages/client/hmi-client/src/services/knowledge.ts
@@ -1,6 +1,6 @@
-import API, { Poller, PollerState, PollResponse } from '@/api/api';
+import API, { Poller, PollerState, PollResponse, PollerResult } from '@/api/api';
 import { AxiosError, AxiosResponse } from 'axios';
-import { Artifact, Model } from '@/types/Types';
+import { Artifact, ExtractionResponse, Model } from '@/types/Types';
 import { logger } from '@/utils/logger';
 
 /**
@@ -8,7 +8,7 @@ import { logger } from '@/utils/logger';
  * @param id
  * @return {Promise<PollerResult>}
  */
-export async function fetchExtraction(id: string) {
+export async function fetchExtraction(id: string): Promise<PollerResult<any>> {
 	const pollerResult: PollResponse<any> = { data: null, progress: null, error: null };
 	const poller = new Poller<object>()
 		.setPollAction(async () => {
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -1,14 +1,14 @@
...
+import { Artifact, ExtractionResponse, Model } from '@/types/Types';
 import { logger } from '@/utils/logger';
 
 /**
  * Fetch information from the extraction service via the Poller utility
  * @param id
  * @return {Promise<PollerResult>}
  */
-export async function fetchExtraction(id: string) {
+export async function fetchExtraction(id: string): Promise<PollerResult<any>> {
```
YohannParis, line 11: If you want to specifically add the type, please do so, but don't add `any`.

YohannParis (approved):

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -36,21 +36,35 @@ export async function fetchExtraction(id: string) {
...
+ * @param modelId string - the model id to use for the extraction
  * @param framework [string] - the framework to use for the extraction, default to 'petrinet'
  * @return {Promise<Model | null>}
  */
-const latexToAMR = async (latex: string[], framework = 'petrinet'): Promise<Model | null> => {
+const latexToAMR = async (
+	latex: string[],
+	modelId?: string,
+	framework = 'petrinet'
+): Promise<Model | null> => {
```
mwdchang, line 47: These parameters don't make a lot of sense, you don't want to have an optional parameter in the middle because you can never specify

```
latexToAMR(blah, 'ABM_framework')
```

packages/client/hmi-client/src/services/knowledge.ts
Jami159: It looks like [PollerResult](https://github.com/DARPA-ASKEM/Terarium/blob/fe352e45f229a30f748f381f6b6914e9d5438be9/packages/client/hmi-client/src/api/api.ts#L70-L73) is a generic interface with a placeholder `<T>` type parameter, where `PollerResult.data` is `<T>`. So I think the return type _should_ be `PollerResult<any>`? But also don't have the best understanding when generic types (`T`) are involved. Do you have a recommendation on what the correct return type would be?

packages/client/hmi-client/src/services/knowledge.ts
Jami159: from what I understand `modelId` and `framework` are both optional, no? `framework` just happens to have a default value. For example, if you look at the calls to `latexToAMR` in either `tera-model-equation.vue` or `tera-model-diagram.vue`, you'll see that the calls are just `latexToAMR(equations, model id)`

mwdchang: Judging by the way the API looks

```
`/knowledge/latex-to-amr/${framework}?modelId=${modelId}`
```

- framework is not optional, it just happens to have a default value, the value must be provided for the API end point to work
- modelId is optional, as it isn't needed for the API

These are not the same type of "optional" behaviour. The way you wrote up the interface makes it impossible to use a different framework.

mwdchang: I don't think we use latexToAMR much (only in one or two places?). So it might be easier just to be explicit and do away with these kind of optional paradigms.

packages/client/hmi-client/src/services/knowledge.ts
mwdchang: Poller isn't the easiest thing to use currently because it wants to be agnostic to errors/progress/data but needs to know the types, let's leave it and I can craft better examples or tightening up the API.

PR #1813, opened 2023-09-01
Update nested graph

### LINKED ISSUE #1753: [FEAT]: Nested graph colour update for model diagrams
In an effort to increase contrast and reduce clutter, will be changing our graphs from greyscale to colour for model diagrams and removing labels from nested nodes. In discussion with Jamie, Pascale and Nelson we have settled on a modified viridis colour palette for our nested nodes. We will reply on the legend as an alternate to nested node labels to avoid double encoding and allow the graph to be viewed clearly without the clutter of labels. 

Matrices will not be coloured, but grey and/or white with labels removed. If you have any questions please feel free to chat with me any time.

**Current:**
image
image

**Proposed:**

image
image
image

mj3cheun:
# Description

Screenshot 2023-09-01 at 5 28 38 PM

Resolves #1753

### DIFF
--- a/packages/client/hmi-client/src/utils/petrinet-color-palette.ts
+++ b/packages/client/hmi-client/src/utils/petrinet-color-palette.ts
@@ -15,11 +15,26 @@ export const nodeTypeColors = [
 	'#b15928'
 ];
 
-let scale: d3.ScaleOrdinal<string, string, never> = d3.scaleOrdinal(nodeTypeColors).domain([]);
+export const nestedTypeColors = [
+	'#c9e9a9',
+	'#aedc30',
+	'#53d25e',
+	'#03c386',
+	'#01a9a1',
+	'#0098b3',
+	'#007eb3',
+	'#00659f',
+	'#3e4d83'
+];
+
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/utils/petrinet-color-palette.ts
```diff
@@ -28,9 +43,23 @@ function setNodeTypeColor(ids: string[]): void {
...
+	return d3.piecewise(d3.interpolateRgb.gamma(2.2), nestedTypeColors)(scaleNestedType(id));
+}
+
+function setNestedTypeColor(setDomain: [number, number]): void {
+	scaleNestedType = d3.scaleLinear().domain(setDomain);
+}
+
+export function useNestedTypeColorPalette() {
+	return { getNestedTypeColor, setNestedTypeColor };
+}
```
mwdchang, line 65: I don't understand a couple of things here
- why are we interpolating colours, and
- why scaleNestedType a shared variable

mwdchang (approved):

PR #1830, opened 2023-09-06
fix: incorrect transition matrix generation

mwdchang:
### Summary
This address the problem where a transition can straddle multiple strata-terms, for example a transition of young-old contact straddles both [young, old] of the age strata.

The current algorithm selects the first one and ignores the rest, this always result in an diagonal matrix which is not correct.

To amend this, the entries of the matrixData are converted into lists, for example instead of

```
[
  { age: young, location: L1 }
]
```

We will now have

```
[
  { age: [young, old], location: [L1] }
]
```

To fill in the 2D matrix, we take the convention that the i-th row use the first element (e.g. young), while the j-th column uses the last element (e.g. old). Note while the matrixData is generic, because of this convention, the matrix generation works for our 1D/2D case but will not work beyond 2-dimensions.

image

image

### Testing
If using the staging database, see the models under `Eval - do not touch`


shawnyama (approved):

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -25,21 +25,29 @@ const pivotAxes = (data: any[], rowDimensions: string[], colDimensions: string[]
...
-			}
+			data[i][key].forEach((term: any) => {
+				if (!terms.includes(term)) {
+					terms.push(term);
+				}
+			});
+
+			// if (!terms.includes(data[i][key])) {
+			// 	terms.push(data[i][key]);
+			// }
```
YohannParis, line 45:
```suggestion
```

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -25,21 +25,29 @@ const pivotAxes = (data: any[], rowDimensions: string[], colDimensions: string[]
...
 			cardinality.set(key, terms);
 		});
 	}
 
-	// Expansion, basica this is a certesian product across the terms of the specified dimensions
+	// Expansion, basically this is a certesian product across the terms of the specified dimensions
```
YohannParis, line 50:
```suggestion
	// Expansion, basically this is a cartesian product across the terms of the specified dimensions
```

YohannParis (approved):

packages/client/hmi-client/src/utils/pivot.ts
mwdchang: fixed

PR #1831, opened 2023-09-06
feat: tera-code polish

echl:
# Description
Added some polish to the code editor but would like to get your input. I thought it should be similar to how one might expect a text editor to work (e.g. new text files don't automatically save + you can explicitly overwrite or save as a new file)

* Added a highlight on hover over the title. Automatically saves when you change the name. @jwaese as per [this](https://github.com/DARPA-ASKEM/Terarium/pull/1808#discussion_r1313537485) thread, the Inplace component can't really be used to add arbitrary buttons like you suggested. I still think it is unlikely that user will accidentally edit this field (and it seems equally likely that any button or field on the page may be accidentally clicked) but happy to go with whatever best practices are. 
* Dropdown to select language with corresponding change in syntax highlighting. Language type changes automatically based on file extension and it's added automatically if you save without a file extension. 
* Save new button to save as new code asset with file naming modal. 
* Ignore the sidebar and tab headers not updating correctly - this is an unrelated issue that will be fixed as part of a refactor. 

https://github.com/DARPA-ASKEM/Terarium/assets/16358728/105a475f-4a9d-4179-ad78-795767a8b1f2


jamiewaese-uncharted:
This is looking really good!

I like your solution for the rename-a-file flow. It looks really intuitive. I wish it didn't resize when focused on, which then moves the other buttons around though.

I'd like to make a tweak to the toolbar, since it's bad practice to have more than one primary button in a group.

CURRENT
image

Can we try something more like this?

PROPOSED
Monosnap Terarium - Workflow   etc 2023-09-06 12-38-09

mwdchang:
What is the expected behaviour of `Save as new`? From the video it looks you are just renaming the code-fragment instead of creating new code-fragment? I was expecting that you would have two code-fragments showing up in the resource-sidebar but it looks like you just replaced the existing one.

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -131,13 +192,27 @@ function onSelectedTextChange() {
...
 
+function setFileExtension(fileName: string) {
+	// check name for file extension
+	// if there is no extension, add the appropriate one based on the selected language
+	const name = fileName.split('.');
+	if (name.length > 0) {
+		return name[0].concat('.').concat(getFileExtension(programmingLanguage.value));
+	}
+	return fileName;
+}
```
mwdchang, line 203: I think it would be better off to have this function in utilities or services, would be easier to unit test than being inside a component.

echl:
> What is the expected behaviour of `Save as new`? From the video it looks you are just renaming the code-fragment instead of creating new code-fragment? I was expecting that you would have two code-fragments showing up in the resource-sidebar but it looks like you just replaced the existing one.

It opens in a new tab. It should also show in the resource sidebar but that's an existing bug with the sidebar not refreshing properly.

YohannParis:
Is this ready to be merged by now? are all questions been answered.

echl:
> Is this ready to be merged by now? are all questions been answered.

Ready to merge for now, we can re-open discussions when we come back to this feature.

YohannParis (approved):

PR #1834, opened 2023-09-08
feat: floating windows example

YohannParis:
### Description
Added a method to load Workflow node on their own floating window instead of a split panel.

### Clean up
* Rename `tera-simulation-worflow` to `tera-workflow` as this is more than simulation now.
* Made `project` an optional prop on some node wrapper: disable the `Save as Dataset` button if not provided.

### Work
* Made a util to display workflow-node into a _unique_ floating window instead of a split panel.
* New `WorkflowNode` route and page to display nodes on their own, not within a project.
* This does not work on full-screen, it will open a new tab in most browser.

### Video
https://github.com/DARPA-ASKEM/Terarium/assets/636801/ac124124-52ba-47de-8d3c-c96dbe232a25


Tom-Szendrey:
I think there is an issue here due to the refreshing rate of our workflow.
Our workflow only saves periodically. 

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/dc0b377f-3899-4eec-96a7-afd38003d363


If i create a node and try to open it, I will then only see the opened node if I opened it after that workflow save has happened:

Tom-Szendrey:
There are some state communication/saving issues going on here:

https://github.com/DARPA-ASKEM/Terarium/assets/17088680/420bc88a-a244-4fa3-8ee3-348a6cafa8ab

YohannParis:
@Tom-Szendrey both points are correct, but out of the scope of this PR, this is mostly to show that it is possible to display those nodes on a _floating window_. The synchronisation and refresh rate of saving a workflow is two complete beast on their own.

packages/client/hmi-client/src/components/workflow/tera-simulate-ciemss.vue
```diff
@@ -218,7 +219,7 @@ const selectedCols = ref<string[]>([]);
 const paginatorRows = ref(10);
 const paginatorFirst = ref(0);
 const completedRunId = computed<string | undefined>(() => props?.node?.outputs?.[0]?.value?.[0]);
-const projectId = ref<string>(props.project.id);
+const projectId = ref(props?.project?.id);
```
shawnyama, line 222: Why are `projectId`s  refs? I think it should be computed or even a normal variable

packages/client/hmi-client/src/page/WorkflowNode.vue
```diff
@@ -0,0 +1,63 @@
...
+		<tera-calibrate-ensemble-ciemss
+			v-else-if="isNodeType(OperationType.CALIBRATE_ENSEMBLE_CIEMSS)"
+			:node="node"
+		/>
+		<tera-simulate-ciemss v-else-if="isNodeType(OperationType.SIMULATE_CIEMSS)" :node="node" />
+		<tera-simulate-ensemble-ciemss
+			v-else-if="isNodeType(OperationType.SIMULATE_ENSEMBLE_CIEMSS)"
+			:node="node"
+		/>
+	</template>
```
shawnyama, line 25: Not sure if there is a clean way to use a dynamic `<component>` here

packages/client/hmi-client/src/page/WorkflowNode.vue
shawnyama, line 1: Maybe a better filename would be `WorkflowNodeDrilldown`?

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/tera-simulate-ciemss.vue
YohannParis: I completely agree, we need to clean up all those workflow-node. Let me work on that on this PR, a bit more clean up shouldn't be an issue.

PR #1839, opened 2023-09-08
refactor: add llm component and destroy old tera-model

shawnyama:
# Description

- integrate llm component into new `tera-model`
- looked through git history and matched a few changes within `/models` into `/model` (just `tera-model-diagram`)
- delete old model components and move ones we still use into `/model`
- reorganize diagram components into `/model-diagrams`
- renamed components related to model configuration to end with `configurations` instead of `configuration`
- remove `/new-tera-model` route from `index.ts`
- tweaked cell update in model config to update only if the edited value is new


YohannParis:
I don't know if we want to move the LLM part of the `tera-model` on this PR or wait after it's merged.
What do you think @shawnyama ?

mwdchang:
Did a few quick tests, seems good to me, nice work!

mwdchang (approved):

PR #1850, opened 2023-09-11
chore: reorganize workflow ops

mwdchang:
### Summary
- Reorganize by individual operations, this should make it a little easier to find things, as well as how to setup new operators.
- Adding a "mod.ts" for each operation to organize imports/exports
- Attempt to have a more consistent file names and component names, with the general format being: `tera-{operation}-[node]-{engine}`

There are still some cleanups I'd like to do on the project-page side (drilldowns). But I think this is large enough for a first go at reorganizing things.

### Testing
Workflow operations work, hopefully.


YohannParis (commented):
For the naming convention, know that it is organized by folder, why not renaming the files within a folder more simply. for example:
```
 /calibrate-ensemble-ciemss
   L calibrate-operations.ts
   L mod.ts
   L tera-calibrate-ensemble-ciemss.vue
   L tera-calibrate-ensemble-node-ciemss.vue
```
To:
```
 /calibrate-ensemble-ciemss
   L operations.ts
   L mod.ts
   L tera-page.vue
   L tera-node.vue
```

packages/client/hmi-client/src/workflow/tera-simulation-workflow.vue
```diff
@@ -82,38 +82,40 @@
 						:node="node"
 						@append-input-port="(event) => appendInputPort(node, event)"
 					/>
-					<tera-simulate-julia-node
+					<tera-simulate-node-julia
```
YohannParis, line 85: Why the `node` is not at the end of the name? This is really weird

YohannParis (approved):
Thank you for cleaning up all those imports without `@`.

packages/client/hmi-client/src/workflow/tera-simulation-workflow.vue
mwdchang: @Tom-Szendrey 's convention, just following it.

mwdchang:
> For the naming convention, know that it is organized by folder, why not renaming the files within a folder more simply. for example:

I like this idea of having more simple names. But we had tried this on the World Modelers project and there are some gripes about being able to find files (since we will have multiple tera-node and tera-page artifacts). Maybe we should just take it to a vote??

YohannParis:
> > For the naming convention, know that it is organized by folder, why not renaming the files within a folder more simply. for example:
> 
> I like this idea of having more simple names. But we had tried this on the World Modelers project and there are some gripes about being able to find files (since we will have multiple tera-node and tera-page artifacts). Maybe we should just take it to a vote??

Oh right, people who do not look at folders but just search for files... monsters!

shawnyama (approved):

PR #1857, opened 2023-09-12
feat: update project card design

shawnyama:
# Description
- Add Author and Creation Date to additional details on card mouseover
- Hide the kebab menu icon by default - display on card mouseover
- Re-arrange the order of the stat icons to: Participants, Documents, Datasets and Models
- Truncate text of title and description if exceeding allotted space
- Remove extra space below shorter title

Not sure what I did to make the footer move up along with the description - need some help on finding why that is. 

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/965d6b8d-525e-4a14-8d8c-c26d1fc59803


packages/client/hmi-client/src/components/home/tera-project-card.vue
```diff
@@ -1,27 +1,37 @@
...
+			</section>
+		</template>
+		<template #footer>
+			<!--FIXME: There is no 'last updated' property in project yet-->
+			<span>Last updated {{ formatDdMmmYyyy(project.timestamp) }}</span>
+			<Button
+				icon="pi pi-ellipsis-v"
+				class="p-button-icon-only p-button-text p-button-rounded"
+				@click.stop="showProjectMenu"
+			/>
```
YohannParis, line 34: Maybe because this button is not visible until on hover that its padding/margin are changing the footer and make it jump.

packages/client/hmi-client/src/components/home/tera-project-card.vue
```diff
@@ -78,6 +88,17 @@ const emit = defineEmits<{
...
+const titleRef = ref();
+const descriptionLines = computed(() => {
+	const titleHeight = titleRef.value?.clientHeight;
+	for (let i = 1; i < 3; i++) {
+		if (titleHeight <= 17 * i) {
+			return 6 + i;
+		}
+	}
+	return 6;
+});
```
YohannParis, line 100: I think you should limit it to 6 anyway and have the title limited as well, we should never have long title.

YohannParis (approved):
Great work

packages/client/hmi-client/src/components/home/tera-project-card.vue
shawnyama: Just made an adjustment. 3 lines for the title is the max that Neil gave me. A long title would require 7 lines for the description - a one line title would require 9 lines. (6 was my previous minimum)

packages/client/hmi-client/src/components/home/tera-project-card.vue
shawnyama: That wasn't it, I think the `-webkit-box` triggers some weird behaviour. I ended up giving the footer an absolute position relative to the card body.

PR #1858, opened 2023-09-12
chore: cleanups

mwdchang:
### Summary
Small cleanups
- remove unnecessary workflowId prop, workflowId is already embedded into each workflowNode
- remove uuid package in favour of crypto.randomUUID()

### Testing
No functional changes, workflow should work as before

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,27657 @@
+{
+	"name": "terarium",
+	"lockfileVersion": 2,
+	"requires": true,
+	"packages": {
+		"": {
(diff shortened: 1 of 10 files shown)


YohannParis (approved):
Nice clean up

chris-dickson:
Not a concern, but what's the issue with uuidv4?

mwdchang:
> Not a concern, but what's the issue with uuidv4?

@chris-dickson nothing with with uuidv4, just thought we can have one less package if the function is already supported natively.

PR #1859, opened 2023-09-12
Feat model transformer node

blanchco:
# Description

Resolves #(issue)


packages/client/hmi-client/src/workflow/ops/model-transformer/model-transformer-operation.ts
```diff
@@ -0,0 +1,24 @@
+import { Operation, WorkflowOperationTypes } from '@/types/workflow';
+
+interface ModelOperationState {
+	modelId: string | null;
+	modelConfigurationIds: any[];
```
mwdchang, line 5: Configuration ids are  `strings[]`

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+	<div class="background">
+		<Suspense>
+			<tera-model-jupyter-panel
+				asset-id="bcd32c48-1daa-4a52-8ed7-147c1dc4b0b0"
+				:model-configuration-id="modelConfigurationId"
```
mwdchang, line 6: Why do we have a hardwired `bcd32c48-1daa-4a52-8ed7-147c1dc4b0b0`

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
```diff
@@ -0,0 +1,114 @@
...
+			data: { history: [] },
+			timestamp: new Date().toISOString()
+		});
+		notebookSessionId = response?.id;
+
+		if (notebookSessionId) {
+			// update the node state with the notebook session id
+			const state = cloneDeep(props.node.state);
+			state.notebookSessionId = notebookSessionId;
+			workflowEventBus.emit('update-state', {
```
mwdchang, line 65: Your state setup is strange - I know that the typing information is lost after passing it through a prop. But if we type it out:

```
const state = cloneDeep(props.node.state) as ModelTransformState; 
```

And we don't even have a `ModelTransformState` here, neither is the notion of `state.notebookSessionId`

Incidentally if you know a good way to enforce types please let me know :)

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
```diff
@@ -0,0 +1,114 @@
...
+	const state = cloneDeep(props.node.state);
+	state.modelId = model?.id;
+	// update node state with the model id
+	workflowEventBus.emit('update-state', { node: props.node, state });
+
+	// set default configuration
+	await addDefaultConfiguration(model);
+
+	// setting timeout...elastic search might not update default config in time
+	await setTimeout(async () => {
```
mwdchang, line 91: This should be fixed for document creation in ES, can we verify that this is working without the timeout?

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer-node.vue
```diff
@@ -0,0 +1,38 @@
...
+const fetchModel = async () => {
+	if (!props.node?.state?.modelId) return;
+	model.value = await getModel(props.node?.state?.modelId);
+};
+
+onMounted(async () => {
+	await fetchModel();
+});
+watch(
+	() => props.node.state,
```
mwdchang, line 31: You may want to tighten this watcher a bit if the only thing you need is the modelId field,

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: sorry i believethe asset-id is not longer used and I was using it for testing

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: hmm - ill test again, I was having issues where I wasn't getting the default config back. Will test without

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: fair, I hadn't used the state type at all and will get into the habit of using it.  Change will be made.  

Im not too sure about enforcing types, will have to be something I can look into

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: did a few tests, unfortunately it looks like the timeout is needed!

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer-node.vue
blanchco: done!

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: i've typed the states, I've also done the some with the dataset transformer

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
blanchco: this has been removed...

packages/client/hmi-client/src/workflow/ops/model-transformer/tera-model-transformer.vue
mwdchang: okay ... we will come back to this later

mwdchang (approved):

PR #1861, opened 2023-09-13
Consume intermediate julia calibrate results using rabbitmq

### LINKED ISSUE #1766: [TASK]: Intermediate results for Julia calibrate operation
See if we can get intermediate results from Julia's calibrate. At a minimum we want to have
- accuracy
- current estimates on the parameters

Subtasks
- [ ] Check, and work with Julia team to see that intermediate results are written to MQ
- [ ] We can receive and interpret the results in the HMI-server
- [ ] The client and listen and graph/print the intermediate results in the calibrate operation

Jami159:
# Description

* This code change consumes the intermediate results from a julia calibrate operation that are published to the "sciml-queue"
* For now it simply console logs the intermediate results, they will be processed in a follow up task
* Right now we're using a `JsonObject` as the response type, this is because the message consumed from the queue is a JsonObject. We're not mapping it to a named class because the response object will have properties updated in the near future, and we don't know yet what that's going to look like.

Resolves #1766


packages/services/hmi-server/src/main/resources/application.properties
```diff
@@ -80,7 +80,11 @@ rabbitmq-username=${terarium-mq-username:terarium}
 rabbitmq-password=${terarium-mq-password:terarium123}
 mp.messaging.outgoing.user-event.connector=smallrye-rabbitmq
 mp.messaging.outgoing.user-event.exchange.name=user-event
+# PyCiemss
 mp.messaging.incoming.simulationStatus.connector=smallrye-rabbitmq
 mp.messaging.incoming.simulationStatus.queue.name=simulation-status
+# Julia
```
YohannParis, line 86: Could you add the name of the repo, or the url for those comments please:
```suggestion
# PyCiemss - https://github.com/DARPA-ASKEM/pyciemss-service
mp.messaging.incoming.simulationStatus.connector=smallrye-rabbitmq
mp.messaging.incoming.simulationStatus.queue.name=simulation-status
# Julia - https://github.com/DARPA-ASKEM/simulation-service
```

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
```diff
@@ -231,17 +233,32 @@ const runCalibrate = async () => {
...
+	runIds.forEach((id) => {
+		eventSourceManager.openConnection(id, `/simulations/${id}/sciml-intermediate-results`);
+		eventSourceManager.setMessageHandler(id, handlingProgress);
+	});
+
 	poller
 		.setInterval(3000)
 		.setThreshold(300)
 		.setPollAction(async () => simulationPollAction(runIds, props.node, progress, emit));
 	const pollerResults = await poller.start();
```
YohannParis, line 257: Do we still need this if you are implementing the SSE endpoints?

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
```diff
@@ -231,17 +233,32 @@ const runCalibrate = async () => {
...
 
+	// closing event source connections
+	runIds.forEach((id) => eventSourceManager.closeConnection(id));
+
```
YohannParis, line 261: Why are you closing them right after opening them?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -187,6 +188,19 @@ public Publisher<byte[]> stream(
 		});
 	}
 
+	@GET
+	@Path("/{jobId}/sciml-intermediate-results")
```
YohannParis, line 192: Better keep matching URLs
```suggestion
	@Path("/{jobId}/sciml/partial-result")
```

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -27,7 +25,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.smallrye.reactive.messaging.annotations.Broadcast;
 import org.eclipse.microprofile.reactive.messaging.Emitter;
-import java.util.Map;
+import io.vertx.core.json.JsonObject;
```
YohannParis, line 28: @dvince2 is this okay to use JsonObject instead of JsonNode?
I don't see why the former is better or necessary.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -169,7 +170,7 @@ public Publisher<byte[]> stream(
 	) {
 		ObjectMapper mapper = new ObjectMapper();
 		return Multi.createFrom().publisher(simulationStatusStream).filter(event -> {
-			try{ 
+			try{
```
YohannParis, line 173: You should also add a try catch to your endpoint

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
Jami159: This is necessary in order to know when to close the `eventSourceManager`. See below

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
Jami159: This doesn't close until the poller above finishes polling (i.e. knows if the job finished or not).

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -61,6 +59,9 @@ public class SimulationResource implements SnakeCaseResource {
 	@Channel("simulationStatus")
 	Emitter<SimulationIntermediateResultsCiemss> simulationStatusEmitter;
 
+	@Inject
+	@Channel("scimlQueue") Publisher<JsonObject> scimlQueueStream;
```
Tom-Szendrey, line 63: Is there a reason we are reading this as a JsonObject rather than a specified object like: `SimulationIntermediateResultsJulia`

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
```diff
@@ -187,6 +188,19 @@ public Publisher<byte[]> stream(
...
 
+	@GET
+	@Path("/{jobId}/sciml-intermediate-results")
+	@Produces(MediaType.SERVER_SENT_EVENTS)
+	@SseElementType(MediaType.APPLICATION_JSON)
+	@Tag(name = "sciml simulation intermediate results associated with run ID")
+	public Publisher<JsonObject> scimlResult(
+		@PathParam("jobId") final String jobId
+	) {
+		return Multi.createFrom().publisher(scimlQueueStream).filter(event -> {
```
Tom-Szendrey, line 199: (Related to my above comment)
We should specify the object type here to allow us to getId() rather than getValue("id")
This will help us work with this type as well rather than just being a generic object. 
This will also allow us to catch errors if they start sending a different object to us

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
dvince2: Working with @Jami159 yesterday, the object being returned from here seems to be coming in as a `JsonObject` so we went with that - admittedly unsure why a `JsonNode` isn't being returned or what the difference here is.
Once we move to Spring I imagine this will end up as a JsonNode

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
dvince2: 💯

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Jami159: Yeah, I mentioned it in the PR description!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Jami159: Same reason as above.

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Tom-Szendrey: I understand it will be changed in the near future but i think its worth having everything set up and just updating said class when it needs to be updated
What are your thoughts on this?

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Jami159: Done!

packages/services/hmi-server/src/main/java/software/uncharted/terarium/hmiserver/resources/dataservice/SimulationResource.java
Jami159: I have tried deserializing the response to a named class as you suggested but I couldn't get it working properly. So that was another reason I went to `JsonObject` for the time being. I don't want that to be a blocker to landing this PR so I'd like to set up `SimulationIntermediateResultsJulia` up in a separate branch if that's cool.
If you have a strong preference though I can try to make the change here!

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
mwdchang: We are using SSE/MQ to communicate `intermediate events`, that is a separate concept than `progress`  - the fact we have been calling it "progress" or "handleProgress" is not technically correct, and I think further exacerbates the the confusions we have here.

YohannParis: Yep, so this needs to be renamed to avoid confusion.
This comes back to my notifications system we need to setup to improve UX.

Jami159: Wait sorry, so what is the suggestion here? Should I be making any additional changes to the code here?

mwdchang: Rename `handleProgress` to `handleIntermediate` or something similar ..., the poller is polling for progress, but the existing handleProgress is not related to progress, so this was causing some confusion.

Jami159: done!

mwdchang (approved):

PR #1862, opened 2023-09-13
fix: clear facet filters on resource type change

blanchco:
# Description

* A fix to change clear the facet filters when changing the resource type

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
@@ -486,6 +486,7 @@ const toggleDataItemSelected = (dataItem: { item: ResultType; type?: string }) =
 
 const updateAssetType = async (newResourceType: ResourceType) => {
 	if (resourceType.value !== newResourceType) {
+		const oldResourceType = resourceType.value;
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -495,6 +496,15 @@ const updateAssetType = async (newResourceType: ResourceType) => {
 		const resList = dataItemsUnfiltered.value.find(
 			(res) => res.searchSubsystem === resourceType.value
 		);
+
+		/** clear filters if they exist, we when to set the old resource type to
+		 * have dirty results since now they will need to be refetched when the facets filters are cleared
+		 * */
+		if (clientFilters.value.clauses.length > 0) {
```
shawnyama, line 503: Get into the habit of using `isEmpty`

shawnyama (approved):

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
blanchco: will do! This has been added

PR #1863, opened 2023-09-13
feat: sort project cards

shawnyama:
# Description

- ~Search project cards by keyword~ (We are leaving search for another time pretend video doesn't have it)
- Sort project cards in specific orders

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/5b692eb0-3f34-4962-927f-0be12c07f261


mwdchang:
The search should be done at the API level in my opinion, otherwise you will not be able to find stuff.

For example, if we have a pagination limit of 20, no matter how you write the search function you will never be able to find the 23rd project.

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -190,8 +205,51 @@ import { RouteName } from '@/router/routes';
...
 import TeraProjectCard from '@/components/home/tera-project-card.vue';
+import Dropdown from 'primevue/dropdown';
+
+const selectedSort = ref('Last updated (descending)');
+const sortOptions = ref([
+	'Last updated (descending)',
+	'Last updated (ascending)',
+	'Creation date (descending)',
+	'Creation date (ascending)',
+	'Alphabetic'
```
YohannParis, line 216:
```suggestion
	'Alphabetical'
```

YohannParis (commented):
I agree with @mwdchang the search should all be done on the `hmi-server`.

mwdchang (approved):

PR #1868, opened 2023-09-15
feat: share project

echl:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/permissions/PermissionUser.java
dvince2, line 1: Switching to just use lombok

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/permissions/PermissionRelationships.java
```diff
@@ -25,16 +30,4 @@ public void addGroup(PermissionGroup permissionGroup, Schema.Relationship relati
 	public void addProject(String id, Schema.Relationship relationship) {
 		permissionProjects.add(new PermissionProject(id, relationship.toString()));
 	}
-
```
dvince2, line 28: This was the cause of all the issues. The getters/setters for these properties didn't use the expected java naming convention of `getPropertyName`. For example, `permissionGroups` should have an accessor of `getPermissionGroups` not `getGroups`. This threw our ts generator into a weird state.  As a fix, we're going to just use lombok like in all the other models

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/permissions/PermissionProject.java
dvince2, line 1: opting to use lombok

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/permissions/PermissionGroup.java
dvince2, line 1: opting to use lombok

dvince2 (approved):

YohannParis (approved):

PR #1869, opened 2023-09-15
feat: project table view

shawnyama:
# Description

- Add a segmented control to toggle between cards and table view
- Table can be filtered by adding or removing columns
- Description can be expanded
- Project titles are links that, when clicked, open the project
- When a table row is hovered, its kebab menu icon is displayed

Video of row expansion:

https://github.com/DARPA-ASKEM/Terarium/assets/28237258/33d312aa-1ecd-4804-909f-bd69c2a68512


shawnyama:
Need some help identifying why two dialog boxes appear here. This happens with the cards too.
https://github.com/DARPA-ASKEM/Terarium/assets/28237258/286dfe6e-0a49-4965-80bc-d00ba5167183

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -33,56 +33,179 @@
...
-								<p>Shared projects will be displayed on this page</p>
-							</template>
-						</div>
-						<div v-else class="carousel">
-							<div class="chevron-left" @click="scroll('left', $event)">
-								<i class="pi pi-chevron-left" />
+						<section class="list-of-projects">
+							<div v-if="!isLoadingProjects && isEmpty(tab.projects)" class="no-projects">
+								<img src="@assets/svg/seed.svg" alt="" />
+								<template v-if="tab.title === 'My projects'">
```
blanchco, line 66: perhaps making the tab title an enum would be good here? i.e.
```
enum TabTitles = {
  MY_PROJECTS = 'My projects',
  SHARED_PROJECTS = 'Shared Projects'
}
```
Might be better for the future if statements or if the text were to change

blanchco:
> Need some help identifying why two dialog boxes appear here. This happens with the cards too.
> 
>  Screen.Recording.2023-09-19.at.3.31.21.PM.mov 
> Also not sure why I am getting those linting errors because I am using those functions but they are read as unused...

moved the dialog a couple levels in the templating helps:
```
		</Teleport>
		<Dialog
			:header="`Remove ${selectedProjectMenu?.name}`"
			v-model:visible="isRemoveDialog"
		>
			<p>
				You are about to remove project <em>{{ selectedProjectMenu?.name }}</em
				>.
			</p>
			<p>Are you sure?</p>
			<template #footer>
				<Button label="Cancel" class="p-button-secondary" @click="closeRemoveDialog" />
				<Button label="Remove project" @click="removeProject" />
			</template>
		</Dialog>
	</main>
</template>
```

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -253,9 +396,68 @@ const myFilteredSortedProjects = computed(() => {
...
+]);
+
+// Table view
+const columns = ref([
+	{ field: 'name', header: 'Project title' },
+	{ field: 'description', header: 'Description' },
+	{ field: 'username', header: 'Author' },
+	{ field: 'stats', header: 'Stats' },
+	{ field: 'timestamp', header: 'Created' },
+	{ field: 'lastUpdated', header: 'Last updated' } // Last update property doesn't exist yet
```
mwdchang, line 410: Does columns need to be a ref or can they just be vanilla JS variable?

shawnyama: In this case it needs to be - changing the `selectedColumns` stops working as expected

blanchco (approved):
LGTM , just need to fix those linter issues...

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -33,56 +33,141 @@
...
-										:project="project"
-										@click="openProject(project.id)"
-										@removed="removeProject"
-									/>
-								</li>
-								<li>
-									<section class="new-project-card" @click="isNewProjectModalVisible = true">
-										<div>
-											<img src="@assets/svg/plus.svg" alt="" />
+							<DataTable
```
YohannParis, line 115: This has to be its own component, the home.vue is not supposed to be that complicated.

YohannParis (approved):

PR #1878, opened 2023-09-19
feat: fullscreen modals

blanchco:
# Description

* adding fullscreen modals to workflow

FIGMA:
Monosnap Terarium - New   Improved 2023-09-20 10-03-38

VIDEO (new sept 22):

https://github.com/DARPA-ASKEM/Terarium/assets/33158416/eb0f2cb5-0ae8-406d-821a-0b08ce40ddc0

Resolves #(issue)


blanchco:
the entire modal is now fully scrollable. There are definitely some minor performance issues when scrolling due to the amount of content we have on the screen, however this seems to be only prevalent when scrolling really fast.  Unsure about a sure fire way for a solution for this at the moment

YohannParis (approved):
Cheers

PR #1881, opened 2023-09-20
feat: parameter matrices

mwdchang:
# Description

Collected all unique inputs and outputs of the parameters to create a matrix. Then using the input output pairs found in the `transitions` attribute of the AMR I filled in the matrix.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/5d3223df-fba9-49f0-a5c0-6c8a1c48b2fd


packages/client/hmi-client/src/services/project.ts
```diff
@@ -93,6 +93,7 @@ async function getAll(): Promise<Project[] | null> {
  * @return ProjectAssets|null - the appropriate project, or null if none returned by API
  */
 async function getAssets(projectId: string, types?: string[]): Promise<ProjectAssets | null> {
+	if (projectId === '') return null;
```
mwdchang, line 96: This seems unrelated, what are you trying to do?

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -1,12 +1,14 @@
...
 
 export interface PivotMatrixCell {
 	row: number;
 	col: number;
 	rowCriteria: any;
 	colCriteria: any;
-	value: any;
+	controllers?: string[];
+	content: any;
 }
```
mwdchang, line 12: I don't like the idea of introducing domain/business specific logic into utils - ideally what I want for util is to be agnostic to both framework and specific domains, so they are true-utilities that can be reused with relative ease.

`controllers` is quite specific to these types of ODE-problems with topological contraints.

Let's maybe talk a bit about what we really need and see if we can make this a bit more flexible.

packages/client/hmi-client/src/services/project.ts
shawnyama: will remove, I was debugging the 404 error when projects weren't appearing

packages/client/hmi-client/src/model-representation/petrinet/mira-petri.ts
```diff
@@ -119,6 +119,9 @@ export const getTransitions = (amr: Model, lookup: Map<string, string>) => {
 		} else {
 			obj.base = existingTransition.id;
 		}
+		obj.input = transition.input;
+		obj.output = transition.output;
+
```
mwdchang, line 124: These are a bit dangerous because we don't know if there may be some strata called `input` or `output`, since you need an AMR instance in `createParameterMatrix` anyway ... it would be better to make the construction there instead.

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -97,14 +99,84 @@ export const createMatrix1D = (data: any[]) => {
...
 
 	return { matrix: rows };
 };
 
+export const createParameterMatrix = (
+	amr: Model,
+	transitionMatrixData: any[],
+	childParameterIds: string[]
+) => {
+	const rows: any[] = [];
```
mwdchang, line 115: I think we can improve the construction logic to be more efficient 
- remove the input/output from the transitionMatrixData array, this makes the algo. more general
- resolve and cache transition's input/output
- resolve and cache input/output indices

Then we just need to loop over the transitionMatrixData twice, once to build the cache, and one more to populate the matrix, as oppose to looping for as many times as we have number of matrix-cells. 

The rough pseudo code is below:

```
createParameterMatrix = ( amr, transitionMatrixData, childParameterIds) => {
  const inputMap = Map<string, string>();
  const outputMap = Map<string, string>();

  // 1. Do the controller/input/output stuff
  for (let i = 0; i < transitionMatrixData.length; i++) {
    const transition = findTransition(amr, transitionMatrixData[i].id);

    // input/output/controller stuff

    inputMap.set(transitionMatrixData[i].id, ...);
    outputMap.set(transitionMatrixData[i].id, ...);
  }

  // 2. Build grid
  const rows: any[] = [];
  for (let rowIdx = 0; rowIdx < inputs.length; rowIdx++) {
    const row: PivotMatrixCell[] = [];
    for (let colIdx = 0; colIdx < outputs.length; colIdx++) {
      row.push({
        ...
      });
    }
  }

  const rowIndexMap = new Map();
  for (let rowIdx = 0; rowIdx < inputs.length; rowIdx++) { rowIndexMap.set(inputs[i], rowIdx); }

  const colIndexMap = new Map();
  for (let rowIdx = 0; rowIdx < inputs.length; rowIdx++) { colIndexMap.set(inputs[i], rowIdx); }

  // 3. Fill in the grid
  for (let i = 0; i < transitionMatrixData.length; i++) {
    const t = transitionMatrixData[i];
    const tInput = inputMap.get(t.id);
    const tOUtput = outputMap.get(t.id);

    const rowIdx = rowIndexMap.get(tInput);
    const colIdx = rowIndexMap.get(tOutput);
    rows[rowIdx][colIdx].content = ....
  }
}
```

mwdchang:
Per discussion
- if we remove X from input we also need to remove X from the output to avoid asymmetric relations later on. Order need to be preserved
- input/output do not form product relations, e.g. they are like parallel arrays.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -86,25 +78,21 @@ ResponseEntity<Model> getModel(
...
 
-
 	@GetMapping("/{id}/model_configurations")
 	ResponseEntity<List<ModelConfiguration>> getModelConfigurations(
-		@PathVariable("id") String id,
-		@RequestParam("page_size") int pageSize
-	) {
+			@PathVariable("id") String id,
+			@RequestParam(value = "page_size", required = false, defaultValue = "100") int pageSize) {
 		return ResponseEntity.ok(proxy.getModelConfigurations(id, 100).getBody());
```
dvince2, line 96: Since we're asking for it, we may as well pass in pageSize rather than the hardcoded `100` here?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -86,25 +78,21 @@ ResponseEntity<Model> getModel(
...
 	}
 
-
 	@GetMapping("/{id}/model_configurations")
 	ResponseEntity<List<ModelConfiguration>> getModelConfigurations(
-		@PathVariable("id") String id,
-		@RequestParam("page_size") int pageSize
-	) {
+			@PathVariable("id") String id,
+			@RequestParam(value = "page_size", required = false, defaultValue = "100") int pageSize) {
```
dvince2, line 95: Since we're asking for it, we may as well pass in pageSize rather than the hardcoded 100 here?

shawnyama (approved):

Issue #1883, opened 2023-09-20
[BUG]: Typo in the parameter modal associated with a model's configuration table


liunelson:
**Describe the bug**
`A distibution`.

**To Reproduce**
1. Go to any model and its "Model" tab
2. Scroll down to the configuration table
3. Click on the hamburger of any parameter
4. Click on the "Type" dropdown
5. See the typo

**Expected behavior**
It should be `A distribution` (missing `r`).

**Screenshots**
![Screenshot 2023-09-20 at 16 57 31](https://github.com/DARPA-ASKEM/Terarium/assets/1030720/bee82ba0-fa80-43fc-9c67-e25ce3fcccc6)

**Additional context**
This was on staging and Jami's local.

liunelson:
@Jami159 I assigned it to you but feel free to have it assigned to whoever else is appropriate.

Jami159:
LOL yeah it's a very quick fix

Issue #1722, opened 2023-08-16
[FEAT]: For a dataset preview of a Simulate, append output data to same CSV if there are multiple configurations as input


Jami159:
**Is your feature request related to a problem? Please describe.**
This issue follows from task #1267 

**Describe the solution you'd like**
If there are multiple configurations piped into a simulate box, the output data from each run should be appended in the same csv. Note that we would need an extra column to store the run id.

PR #1896, opened 2023-09-22
Add dataset preview in the simulate output panel

### LINKED ISSUE #1267: [FEAT]: Add a dataset preview in the output tab of Simulate and ability to download the csv
Modelers like to be able to see the raw data used in the charts.

Add the same dataset preview component used in several parts of the app to display the raw simulation output data. Show also be able to download the corresponding csv.

If there are multiple configurations piped into a simulate box, the output data from each run should be appended in the same csv. That could be done in a second pass. Note that we would need an extra column to store the run id.

image

Jami159:
# Description

* This PR only adds the dataset table for a single simulation run
* In the future we will add the ability for a user to select the run they want to view via a dropdown

Resolves #1267 #1722 #1883 

Screenshot 2023-09-22 at 2 12 29 PM

Screenshot 2023-09-22 at 2 26 31 PM


packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -228,6 +230,92 @@ export const saveDataset = async (
...
+
+	return csvData;
+};
+
+/**
+ * This is a client side implementation of the getStats function from DatasetResource.java
+ * @param csvColumn
+ * @returns CsvColumnStats object
+ */
+const getCsvColumnStats = (csvColumn: number[]): CsvColumnStats => {
```
dvince2, line 288: We do already have this functionality in `DatasetResource` as a helper function - It may be more beneficial to just throw an endpoint on that to keep the functionality insync with each other and to offload the more expensive math operations to the server

Jami159: Oh I like that idea, thank you for the suggestion!

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -228,6 +230,92 @@ export const saveDataset = async (
...
+/**
+ * This is a client side function to create a CsvAsset similar to the getCsv function from DatasetResource.java
+ * The reason we reimplement this client side is because runResults already has all of the data we need, except
+ * it's not in the correct format necessary to be consumed by the TeraDatasetDatatable component.
+ * @param runResults
+ * @param runId
+ * @returns CsvAsset object with the data from the runResults
+ */
+const createCsvAssetFromRunResults = (runResults: RunResults, runId?: string): CsvAsset | null => {
+	const runResult: RunResults = _.cloneDeep(runResults);
```
YohannParis, line 242: just `import { cloneDeep } from lodash;`
```suggestion
	const runResult: RunResults = cloneDeep(runResults);
```

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -228,6 +230,92 @@ export const saveDataset = async (
...
+			csvColHeaders.forEach((header, index) => {
+				if (!csvColumns[header]) {
+					csvColumns[header] = [];
+				}
+				csvColumns[header].push(+rowValues[index]);
+			});
+		});
+	});
+
+	csvColHeaders.forEach((header) => {
```
dvince2, line 276: Now that I've said that the call should be made up here one level higher to avoid having a rest call inside a tight loop - perhaps an endpoint to receive `csvColHeaders` which uses the existing helper?

packages/client/hmi-client/src/services/dataset.ts
Jami159: For visibility: Derek and I had a discussion on slack and decided to leave this as it is for now. We can revisit it if we notice performance issues in the future.

dvince2 (approved):

PR #1909, opened 2023-09-26
feat: toaster on node failure 

Tom-Szendrey:
The error messages are not on TDS for us to pull but at least this lets the user know that the simulation failed

image


YohannParis (changes requested):
You are going to hate me, but we have a logging system that shows toast already and save the logs in the back end as well.

https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/utils/logger.ts

You can see an example of it here, specially the toast title is useful:
https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/services/knowledge.ts#L26-L29

Tom-Szendrey:
> You are going to hate me, but we have a logging system that shows toast already and save the logs in the back end as well.
> 
> https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/utils/logger.ts
> 
> You can see an example of it here, specially the toast title is useful: https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/services/knowledge.ts#L26-L29

That is handy!

Do you want me to move these over to logger as well?
image

YohannParis (approved):
🎉  Amazing!

I think has a bonus point you could remove the `console.error` above the call to `logger.error` as the logger will use the console as well.

YohannParis (approved):

PR #1911, opened 2023-09-26
fix: no longer remove state on error

Tom-Szendrey:
# Description
This is a bit of a bandaid fix. I think we should discuss how to better guardrail `WorkflowNode.state`
Maybe its type `CalibrationOperationStateCiemss | CalibrateEnsembleCiemssOperationState | ... | null`

Im not sure

image


blanchco (approved):
LGTM

PR #1912, opened 2023-09-27
feat: transition matrices

shawnyama:
# Description

- added transition matrices
- can open transition matrices from model diagram
- renamed some variables and created a `StratifiedMatrix` enum instead of `NodeType` to make things clearer

https://github.com/DARPA-ASKEM/terarium/assets/28237258/c0486d08-b810-4c13-ba3d-27dc55700448


mwdchang:
Discussed offline, but records keeping:
- Let's not have interfaces like `getAOrB`, it imprecise and hard to know what to expect. Interface should be very explicit to make it easy to infer intentions and also to test.
- addendum to above, would also recommend not using option parameters (e.g. childParameterIds) to determine the types of output, I think if we want something just make it explicit. 
- Instead of OdeSemantic enum which is kind of weird in the usage context, let's roll with a MatrixType enum instead

mwdchang (approved):
Thanks!

Issue #1899, opened 2023-09-25
[TASK]: If the same config for a simulation is run multiple times, add ability to choose which run


Jami159:
If a user runs the same config for a simulation multiple times, show a dropdown in the simulation node as well as the output panel with the names of the different runs and allow the user to choose which run to view.

PR #1913, opened 2023-09-27
Disable multiple inputs into same simulate box

### LINKED ISSUE #1898: [TASK]: Disable multiple configs going into the same simulation node
Only allow a single config for a simulation node.
If user wants to simulate another config, they need to do that via an additional simulation node.

Jami159:
Resolves #1898 #1899


mwdchang:
There are a decent number of list-processing logic in the simulate boxes. 

If indeed we don't need to accept multiple configurations anymore we might as well try to simplify them, as I recall they were rather quickly hacked together, so cleaning things up would make the code more manageable.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/simulationservice/SimulationCiemssServiceProxy.java
```diff
@@ -10,7 +10,7 @@
 import software.uncharted.terarium.hmiserver.models.simulationservice.JobResponse;
 
 
-@FeignClient(name = "ciemss-service", url = "${terarium.dataservice.url}", path = "/ciemss-service")
+@FeignClient(name = "ciemss-service", url = "${ciemss-service.url}", path = "")
```
dvince2, line 13: I believe here you can fully drop `path`

Jami159: done!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/simulationservice/SimulationServiceProxy.java
```diff
@@ -9,7 +9,7 @@
 import org.springframework.web.bind.annotation.RequestBody;
 import software.uncharted.terarium.hmiserver.models.simulationservice.JobResponse;
 
-@FeignClient(name = "simulation-service", url = "${terarium.dataservice.url}", path = "/simulation-service")
+@FeignClient(name = "simulation-service", url = "${simulation-service.url}", path = "")
```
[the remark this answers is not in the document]
Jami159: done!

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
```diff
@@ -161,33 +191,42 @@ watch(
...
-});
+	const configId = props.node.inputs[0].value?.[0];
+	if (configId) {
+		const modelConfig = await getModelConfigurationById(configId);
+		const parameters = modelConfig.configuration.semantics.ode.parameters;
+		csvData.forEach((row) =>
+			parameters.forEach((parameter) => {
+				row[parameter.id] = parameter.value;
+			})
+		);
```
mwdchang, line 208: What makes you assume that `csvData` and `ode.parameters` will be in the same order?

mwdchang (changes requested):
There are a few confusing points that should be addressed
- You removed the `v-for" context for the tera-simulate-chart, but you still have a `:key` assignment, so what exactly do you want to do?
- I think you might be mangling the `:run-results` assignment in tera-simulate-chart, can you clarify what is been passed in? Seems like you are passing in the value but the chart is expecting an object-map?

mwdchang:
How does "add-chart" work under this new logic?

mwdchang (approved):

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
Jami159: marking as resolved for now as we'll take a deeper look at this in the future as per our conversation earlier

PR #1914, opened 2023-09-27
feat: add document asset

blanchco:
# Description

* Adding document asset template to front end

Some context - right now there is no way to add/delete/get document assets on a project so this will have to be added at some point.  The template is a stripped down version of what we have for external publications since we only get extractions for tables, figures, and equations right now.

An external publication and document asset have vastly different data structures - so we may want to address combining these 2 data types in a future PR.

https://github.com/DARPA-ASKEM/terarium/assets/33158416/74d295bd-1ef9-4bd9-bcc9-3b7de5aa6277

Resolves #(issue)


blanchco:
waiting for spring changes - types need to be updated

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
```diff
@@ -26,7 +26,7 @@ public class DocumentAsset {
 	private String description;
 
 	@TSOptional
-	private Instant timestamp;
+	private String timestamp;
```
blanchco, line 29: timestamp we receive from tds is not serializable with instant

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/DocumentExtraction.java
```diff
@@ -0,0 +1,17 @@
...
+import java.io.Serializable;
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+@Data
+@Accessors(chain = true)
+public class DocumentExtraction implements Serializable {
+
+	private String file_name;
```
dvince2, line 12: Because the parent object is passed through the snake case converter, you should be able to write this variables as
```
@JsonAlias("file_name")
	private String fileName;
```

to have it be a bit more java/javascript like.  This is true of `asset_type` as well.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DocumentController.java
```diff
@@ -98,8 +99,8 @@ public ResponseEntity<byte[]> downloadDocument(
 			.build()) {
 
 			final PresignedURL presignedURL = proxy.getDownloadUrl(id, filename).getBody();
-			final HttpPut put = new HttpPut(presignedURL.getUrl());
-			final HttpResponse response = httpclient.execute(put);
+			final HttpGet get = new HttpGet(presignedURL.getUrl());
```
dvince2, line 102: FANTASTIC CATCH :D

This was 100% a copy-paste error setting up the spring server and is everywhere in our code... I'll fix the other ones. thanks!

packages/client/hmi-client/src/services/project.ts
```diff
@@ -102,7 +102,8 @@ async function getAssets(projectId: string, types?: string[]): Promise<ProjectAs
 			});
 		} else {
 			Object.values(AssetType).forEach((type, indx) => {
-				url += `${indx === 0 ? '?' : '&'}types=${type}`;
+				// FIX: right now the we cannot get a documents assets, this condition is a temporary measure so that this endpoint doesnt break until we can get documents assets
```
dvince2, line 105: Do you mind opening an issue in github so we remember to change this once TDS makes the change in their end?

dvince2 (approved):

packages/client/hmi-client/src/services/project.ts
blanchco: Yup! Opened this last week still in progress: https://github.com/DARPA-ASKEM/data-service/issues/333

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/DocumentExtraction.java
blanchco: done!

packages/client/hmi-client/src/services/project.ts
blanchco: Refer to this issue for this change on our side: https://github.com/DARPA-ASKEM/terarium/issues/1954

PR #1916, opened 2023-09-28
feat: align model

blanchco:
# Description

* align model feature
* separate enrich and align functionality into 2 buttons
* use the `pdf_extractions` and `link_amr` endpoints in knowledge middleware to align a model with a document, do this after selecting a resource.
* select `documents` instead of `artifacts` when add resources to a model

https://github.com/DARPA-ASKEM/terarium/assets/33158416/09f687b9-acb2-4566-8c9b-2dab601d5e56

Resolves #(issue)


blanchco:
There seems like there may be an issue with the `link_amr` endpoint right now from this [discussion](https://unchartedsoftware.slack.com/archives/C049E6PA16V/p1696524178255779).  Going to open this for review for now to thoughts on how we call `link_amr`

packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
```diff
@@ -61,10 +61,12 @@ import Dialog from 'primevue/dialog';
...
 import { ResourceType } from '@/types/common';
-import { profileDataset, profileModel, fetchExtraction } from '@/services/knowledge';
+import { profileDataset, profileModel, fetchExtraction, alignModel } from '@/services/knowledge';
+import { PollerResult } from '@/api/api';
+import { isEmpty } from 'lodash';
 
 const props = defineProps<{
-	publications?: Array<{ name: string; id: string | undefined }>;
+	publications?: Array<{ name: string | undefined; id: string | undefined }>;
 	relatedPublications?: Array<{ name: string; id: string | undefined }>;
```
YohannParis, line 70: Are those `Documents` or `Publications`. If it's the former, we need to rename those props.

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-related-publications.vue
blanchco: I guess we were using Artifacts before, but I agree the naming is confusing regardless.  Going to rename not only the props but the entire modal itself.

YohannParis (approved):

PR #1920, opened 2023-10-02
chore: enforce workflownode state

mwdchang:
### Summary
- Put in some enforcement of workflow-node state typings, so it need to be declared at the props declaration stage.
- Cleaned up test-operation


Tom-Szendrey (approved):
Thanks for this!!!!

YohannParis (commented):
Amazing

PR #1921, opened 2023-10-02
feat: provenance to model to doc

blanchco:
# Description

* A working-ish version of adding and reading provenance for model/dataset enrichments.  Will need to do some updates when changes have been made to the data-service around provenance to add document assets.  For now I used the EXTRACTED_FROM relation and PUBLICATION for the linking.  I'm assuming a relation like ENRICHED_FROM and asset type DOCUMENT will be added in some time.
* provenance linkage happens on `profile-model` and `profile-dataset` calls in the backend
* added provenance service to FE
* fixed up getRelatedArtifacts function to retrieve artifacts

https://github.com/DARPA-ASKEM/terarium/assets/33158416/b532e7ff-e0f1-45b4-8e94-f678d3600c96

Resolves #(issue)


packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
```diff
@@ -635,6 +638,29 @@ function updateTable(tableType: string, idx: number, key: string, value: string)
...
+	]);
+
+	relatedDocuments.value =
+		(provenanceNodes.filter((res) => isDocumentAsset(res)) as DocumentAsset[]).map(
+			(documentAsset) => ({
+				name: documentAsset.name,
+				id: documentAsset.id
+			})
+		) ?? [];
+}
```
blanchco, line 654: list all related document assets

packages/client/hmi-client/src/services/provenance.ts
```diff
@@ -21,6 +23,14 @@ import { getBulkModels } from './model';
...
 const MAX_RELATED_ARTIFACT_COUNT = 5;
 
+// Allowed types to search by in provenance so far...
+const ALLOWED_PROVENANCE_TYPES = [
+	ProvenanceType.Concept,
+	ProvenanceType.Dataset,
+	ProvenanceType.Model,
+	ProvenanceType.Project,
+	ProvenanceType.Publication
+];
```
blanchco, line 33: it looks like so far only a subset of provenance types are allowed to be search types

blanchco:
@YohannParis Unsure if we should merge this now or after the provenance updates...
There are some changes needed in provenance like an updated relation type such as ENRICHED_FROM and the DOCUMENT asset type.  The provenance linking works albeit the relation and asset type being incorrect.  We would simply need to make a few changes in a separate issue update it once the provenance work is updated if we wanted to merge this now.

cc. @brandomr

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
```diff
@@ -43,4 +44,8 @@ public class ModelMetadata {
 	@JsonProperty("card")
 	@TSOptional
 	private Card card;
+
+	@JsonProperty("provenance")
```
dvince2, line 48: The `@JsonProperty` here isn't needed as the field seems to match the incoming JSON name... but leaving it isn't wrong either. just a FYI!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -129,6 +139,10 @@ public ResponseEntity<JsonNode> postProfileModel(
 		@PathVariable("model_id") String modelId,
 		@RequestParam("document_id") String documentId
 	) {
+
+		Provenance provenancePayload = new Provenance(ProvenanceRelationType.EXTRACTED_FROM, modelId, ProvenanceType.MODEL, documentId, ProvenanceType.PUBLICATION);
+		provenanceProxy.createProvenance(provenancePayload);
```
dvince2, line 144: Right now this and the other call to createProvenance is sort of shouting into the void - we should probably make sure that this succeeds and if not write something to the log (also if this call fails I believe it will throw an exception which will end wit this not returning properly.  You may want to do something like...

```

try{
     ResponseEntity<Provenance> r = provenanceProxy.createProvenance(provenancePayload);
     if (!r.getStatusCode().is2xxSuccessful())
        log.error("unable to create provenance");
} catch (Exception e) {
     log.error("unable to create provenance", e);
}

```

blanchco: Good point! I was thinking at the time I wouldn't want the provenance link to block the enrichment job from happening but thinking over it now, its probably better make sure the link happens.  Will make the change

dvince2: I think the way it is right now would end up blocking if there was any issues with the provenance call

blanchco: Fix this up! Too much javascript brain 😆

packages/client/hmi-client/src/services/provenance.ts
YohannParis, line 1: this needs to be reviewed at some point, it's a mess right now.

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -5,7 +5,9 @@
 		</p>
 		<ul>
 			<li v-for="document in relatedDocuments" :key="document.id">
-				{{ document.name }}
+				<a @click="openAsset({ assetId: document.id!, pageType: AssetType.Documents })">{{
+					document.name
```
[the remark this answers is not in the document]
blanchco: @YohannParis I've made the tera-asset-link component for this.  I've also made a tera-asset-icon component since the way we were retrieving component was pretty tedious to be doing everwhere.
Screenshot 2023-10-20 at 12 35 59 PM

packages/client/hmi-client/src/services/provenance.ts
blanchco: Ya we can take a look at this when the provenance change have been made 👀

packages/client/hmi-client/src/page/project/components/tera-resource-sidebar.vue
```diff
@@ -15,17 +15,7 @@
 				<!-- A way to use vue feather icons in the MenuItem component, it might be better to try to use 1 icon library for easier integration -->
 				<template #item="slotProps">
 					<a class="p-menuitem-link">
```
YohannParis, line 17: Why not using your new `tera-asset-link`?

blanchco: The `tera-asset-link` just has the functionality to open an existing asset and that's it (not to mention the styles are different as well).  For this in particular, it's within the PrimeVue Menu component which handles it own actions to create _new_ assets.

Screenshot 2023-10-20 at 1 26 32 PM

packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -153,8 +161,52 @@ const sendToAlignModel = async () => {
...
+
+async function getRelatedDocuments() {
+	if (!props.assetType) return;
+	let provenanceType;
+	if (props.assetType === ResourceType.MODEL) {
+		provenanceType = ProvenanceType.Model;
+	}
+	if (props.assetType === ResourceType.DATASET) {
+		provenanceType = ProvenanceType.Dataset;
+	}
```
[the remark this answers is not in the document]
blanchco: I think we still want to pass the ProvenanceType here since we could have much more than just ResourceTypes being passed into the `getRelatedArtifacts` function.  I have instead made this better by making a sepatate `mapResourceTypeToProvenanceType` function to make this more readable

packages/client/hmi-client/src/page/project/components/tera-resource-sidebar.vue
YohannParis: I see what you mean...

packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -153,8 +161,39 @@ const sendToAlignModel = async () => {
...
+	const provenanceType = mapResourceTypeToProvenanceType(props.assetType);
+
+	if (!provenanceType) return;
+
+	const provenanceNodes = await getRelatedArtifacts(props.assetId, provenanceType, [
+		ProvenanceType.Publication
+	]);
+
+	relatedDocuments.value =
+		(provenanceNodes.filter((res) => isDocumentAsset(res)) as DocumentAsset[]).map(
```
YohannParis, line 190:
```suggestion
		(provenanceNodes.filter((node) => isDocumentAsset(node)) as DocumentAsset[]).map(
```

YohannParis (approved):

PR #1931, opened 2023-10-03
fix: minor fix

Tom-Szendrey:
# Description
hmiserver was not able to use currentUserService in this file as it was not set up (was just null)


dvince2:
@dgauldie or @YohannParis can you force merge this?

PR #1937, opened 2023-10-03
task: add better intellij support for local devlopment

YohannParis:
# Description

* Remove the old `packages/services`
* Added a `start-server-intellij` function to `./hmiServerDev.sh` to use with Intellij
* Refactor `./hmiServerDev.sh`


hmiServerDev.sh
```diff
@@ -12,42 +42,50 @@ case ${1} in
 	start)
 		COMMAND="start"
 		;;
+	start-server-intellij)
```
dvince2, line 45: Thinking more on this, I think we should avoid using `intellij` here, because this could be used with any ide or also just run seperately to up the containers?

YohannParis: changed to `start-server-ide`.

hmiServerDev.sh
```diff
@@ -1,9 +1,39 @@
+#!/bin/bash
```
mwdchang, line 1: Nitpick: might want to consider `#!/usr/bin/env bash` shebang

https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash

dvince2 (approved):

PR #1946, opened 2023-10-04
feat: add regnet model definition

mwdchang:
### Summary
This adds the RegNet schema definition, it is used to model `AMR.model` field which is a placeholder for different syntactical models.

Note this is just a first pass, there are a handful of things that may need to be worked out with TA2, for example while PetriNet has a semantics section, it appears to be lacking/unneeded for a RegNet?

See reference schema definition:
- https://github.com/DARPA-ASKEM/Model-Representations/blob/main/regnet/regnet_schema.json


dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/regnet/RegNetBaseProperties.java
```diff
@@ -0,0 +1,23 @@
...
+import software.uncharted.terarium.hmiserver.models.dataservice.modelparts.ModelGrounding;
+
+import java.util.List;
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class RegNetBaseProperties {
+	@TSOptional
+	private String name;
```
YohannParis, line 16: There is no `id` because of the conclusion from today's TA2 WG?

mwdchang: There is no `id` because I don't think base-properties are (or should be) uniquely identifiable, and the schema itself has no notion of id of this type of object.

But I think think the TSOptional annotations here are correct so I will fix that.

PR #1955, opened 2023-10-04
feat: editable parameter matrices

shawnyama:
# Description

Algorithm update: Put controllers in proper cell instead of assuming every input has a combo with every controller

Editing parameter value works: 

https://github.com/DARPA-ASKEM/terarium/assets/28237258/72045063-1e69-49a5-86a4-492b0cccda9b


packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -104,36 +103,53 @@ export const createMatrix1D = (data: any[]) => {
...
+			}
+		}
+
+		if (!_.isEmpty(newControllers)) {
+			// Map controllers unique to these input/output combos
+			for (let j = 0; j < newInputs.length; j++) {
+				for (let k = 0; k < newOutputs.length; k++) {
+					controllerIndexMap.set(newInputs[j].concat('|', newOutputs[k]), newControllers);
+				}
+			}
```
mwdchang, line 145: I don't thinks this is a cartesian product kind of situation. Probably more correct is soemthing like this

```
for (let j = 0; j < newInputs.length; j++) {
  const key = newInputs[j].concat('|', newOutputs[j])
  controllerIndexMap.set(...)
}
```

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-value-matrix.vue
```diff
@@ -62,33 +63,24 @@
...
 import { pythonInstance } from '@/python/PyodideController';
 
 const props = defineProps<{
 	modelConfiguration: ModelConfiguration;
 	id: string;
+	configIndex: number;
 	stratifiedModelType: StratifiedModelType;
 	nodeType: NodeType;
 	shouldEval: boolean;
 }>();
```
mwdchang, line 79: Not a fan of the addition of a `configIndex`, it makes it hard to hard to reason with the component, and more difficult to reuse because you must have some kind of array context....which is weird. Are there any blockers to just use "modelConfiguration.id" ?

Also what is props.id suppose to do?

packages/client/hmi-client/src/utils/pivot.ts
```diff
@@ -104,36 +103,53 @@ export const createMatrix1D = (data: any[]) => {
 		rows.push(row);
 	}
 
-	return { matrix: rows };
+	return rows;
```
mwdchang, line 106: ??? I don't understand the change here - are you even using this function??

shawnyama: Thought it made it less complicated when assigning a value to `createMatrix1D`. I'll revert it to keep things consistent

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-value-matrix.vue
shawnyama: Decided to update the emit to take the `configIndex` by default. Updating configuration is done ultimately in `tera-model.vue` so if we want to update a config by id that'd be best done in a seperate PR where we can prepare all of its children for that. 

`props.id` is the name of the base parameter. eg. `beta`. In a transition matrix the id would just be the transition id like `t1`

shawnyama: Looking at where configIndex is used (`tera-model.vue`). it's used in `updateConfiguration` to help update the frontend:

```
async function updateConfiguration(updatedConfiguration: ModelConfiguration, index: number) {
	await updateModelConfiguration(updatedConfiguration);
	setTimeout(async () => {
		emit('update-model-configuration');
		modelConfigurations.value[index] = updatedConfiguration; // Below line would be ideal but the order of the configs change after the refetch
		// await fetchConfigurations(); // elastic search might still not update in time
	}, 800);
}
```

Worth removing in a separate PR as we can just find the id

mwdchang: okay - let's do a clean up pass!

mwdchang (approved):

PR #1964, opened 2023-10-05
fix: return type for acset-to-latex

mwdchang:
# Summary
Fix return type, should be string and not JsonNode


dvince2:
Why was this closed off?

dvince2:
oh i see now!

PR #1969, opened 2023-10-06
fix documents

blanchco:
# Description

* documents array is still broken since we can't get the empty object yet, adding a null checker here!

Resolves #(issue)


mwdchang (approved):
Thanks!

PR #1977, opened 2023-10-10
1889 funman 1 determine current state and implement api end points

tangopapatime:
# Description

First attempt at creating proxy to FUNMAN.  I started to add typing information but I do think that it's still relatively too fluid to do add anything in stone.  As of right this writing, FUNMAN does not like the new AMR format so that will need to be resolved on their end.   Not sure if we want to merge this now or keep this branch until all FUNMAN related tasks are completed.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/funman/FunmanProxy.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+
+import software.uncharted.terarium.hmiserver.proxies.funman.FunmanProxy;
+
+@FeignClient(name = "funman-api", url = "${funman-service.url}")
```
dvince2, line 14: In the end its all the same, but for the url path we had been following the convention where in application.properties you would have `http://funman.staging.terarium.ai/api` and here in the FeignClient the path would be `/queries

dvince2 (changes requested):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
```diff
@@ -0,0 +1,42 @@
...
+public class FunmanController {
+
+    @Autowired
+    private FunmanProxy funmanProxy;
+
+    @GetMapping("/{query_id}/halt")
+    public ResponseEntity<JsonNode> halt(@PathVariable String query_id) {
+        ResponseEntity<JsonNode> response = funmanProxy.halt(query_id);
+
+        return response;
```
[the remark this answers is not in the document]
tangopapatime: good to know. thanks.

packages/server/src/main/resources/application.properties
```diff
@@ -66,6 +66,7 @@ skema-rs.url=https://skema-rs.staging.terarium.ai
 skema-unified.url=https://api.askem.lum.ai
 terarium.dataservice.url=https://data-service.staging.terarium.ai
 xdd-dev-service.url=https://xdddev.chtc.io
+funman-service.url=http://funman.staging.terarium.ai/api
```
dvince2, line 69: Just to confirm is there a corresponding addition to the orchestration project adding this for staging/prod?
Also @dgauldie another entry for local development here

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/funman/config/FunmanConfig.java
```diff
@@ -0,0 +1,32 @@
...
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+
+@Data
+@Accessors(chain = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
dvince2, line 11: I wont put a comment on each, but, technically this is unneeded as we have `spring.jackson.default-property-inclusion=NON_NULL` in our application.properties

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/funman/config/FunmanConfig.java
tangopapatime: yeah I was curious if we needed this.  I'll remove it if we don't.  as always, appreciate the explanation.

Tom-Szendrey:
https://github.com/DARPA-ASKEM/terarium/pull/2107

PR #1983, opened 2023-10-10
fix: jsdelivr.url var read properly from app.props

shawnyama:
# Description

All I changed was line 8 on JsDelivrProxy.java. My formatter shifted stuff

Code is fetched properly now:
image

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/code/CodeController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/code/CodeController.java
@@ -39,9 +39,8 @@ public class CodeController {
 	 * Stores a model from a code snippit
 	 *
 	 * @param code the python code snippit
-	 * @return a {@link StoredModel} instance containing the model id, inputs, and
-	 *         outputs of the model
-	 *         derived from the code input
+	 * @return a {@link StoredModel} instance containing the model id, inputs, and outputs of the model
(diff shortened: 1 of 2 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/jsdelivr/JsDelivrProxy.java
```diff
@@ -5,12 +5,10 @@
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 
-
-@FeignClient(name = "jsdeliver", url = "jsdelivr.url")
+@FeignClient(name = "jsdelivr", url = "${jsdelivr.url}")
```
dvince2, line 8: Good catch on the missing `${...}`.  Is this the only fix in this PR and the rest is just formatting?

shawnyama: ya thats it

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/code/CodeController.java
dvince2, line 1: I think this can be reverted - just whitespace changes?

PR #1986, opened 2023-10-10
chore: move from artifacts -> documents where applicable

Tom-Szendrey:
1) Adding documents back to valid types that we pull and can have in our Projects type
2) changing github + upload to use documents instead of artifacts
image

To test:
upload a pdf that is less than or equal to 1mb:
image


packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
```diff
@@ -364,12 +364,13 @@ async function processFiles(files: File[], csvDescription: string) {
...
 			case AcceptedTypes.TXT:
 			case AcceptedTypes.MD:
-				return processArtifact(file);
+				return processDocument(file);
 			case AcceptedTypes.PY:
 			case AcceptedTypes.R:
 			case AcceptedTypes.JL:
 				return processCode(file);
 			default:
+				// TODO: Should this be process Artifact?
```
Tom-Szendrey, line 373: @YohannParis @dvince2  please let me know here

packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
```diff
@@ -364,12 +364,13 @@ async function processFiles(files: File[], csvDescription: string) {
 			case AcceptedTypes.PDF:
 			case AcceptedTypes.TXT:
```
Tom-Szendrey, line 365: Should .txt files still be considered artifacts? 
The document viewer is not designed or .txt

image

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
dvince2, line 1: can revert

packages/client/hmi-client/src/services/artifact.ts
```diff
@@ -82,7 +82,7 @@ async function uploadArtifactToProject(
  * @param artifact the artifact to create
  */
 async function createNewArtifact(artifact: Artifact): Promise<Artifact | null> {
-	const response = await API.put('/artifacts', artifact);
+	const response = await API.post('/artifacts', artifact);
```
dvince2, line 85: nice

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
Tom-Szendrey: but its so ugly!

packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
dvince2: There is a bit of weirdness here around coding for "where we are now" vs "where the puck is going".  I think in theory right now we can only have PDF `Document`s, and giving a .txt or .md would potentially blow things up.  So right now we may want to have those extensions be `Artifact`.

However eventually I would assume txt, md and even docx would go in as a `Document`.  In this case, `Artifact` becomes the catch all "other" category... but then there are real security/usability concerns around letting people just arbitrarily upload "stuff".

This is a long winded way of saying... I don't know :D.  I lean in favour of option 1 here assuming txt and md blow up documents, but thats not what we'll want in the end?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
dvince2: oh! if it was a purposeful change then keep it! I wasn't sure if it was an accidental formatting comit

packages/client/hmi-client/src/page/project/components/tera-project-overview.vue
Tom-Szendrey: I will throw md and txt to artifacts and leave the default as an a future issue open for discussion

PR #1988, opened 2023-10-10
feat: remove project page tabs

shawnyama:
# Description

- removed tabs from `tera-project.vue` and related props/emits that are passed to its children components
- removed tab storage
- created an `AssetRoute` type that contains `assetId: string, pageType: AssetType | ProjectPages` (the only params needed for a the routing)
- note that the `Tab` type used in `tera-resource-sidebar.vue` is still useful for listing the assets. ~Not sure if it's worth changing its name.~ (Changed name to `AssetItem`).

In `tera-project.vue` `pageType` and `assetId` are extracted from the route and stored into a computed variable called `openedAssetRoute`. It's values are passed down as props to the child components. Not sure if it's worth extracting the values from the route from the children instead. The way it is right now is a bit more convenient since they just have to be extracted from the route and typed once.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/9a3607d7-df31-4c21-b03f-cb4607594724


packages/client/hmi-client/src/types/common.ts
```diff
@@ -106,9 +106,14 @@ export type SidePanelTab = {
 // Tabs
 export type Tab = {
 	icon?: string;
-	assetId?: string;
 	assetName?: string;
-	pageType?: AssetType | ProjectPages;
+	assetId: string;
+	pageType: AssetType | ProjectPages;
+};
```
YohannParis, line 112: Why not extending `Tab` with `AssetRoute`?
Also why do we need `Tab`?

shawnyama: I mentioned the use of `Tab` in the description - though I just renamed it to `AssetItem` to make things more clear. Let me know your thoughts.

YohannParis (approved):
Magnificient!

PR #1989, opened 2023-10-10
fix: do not change route if name/path exists

mwdchang:
### Summary
This address a problem where refresh a page that is not the home-page results in a redirection back home.

### Testing
- Login
- Goto any project page
- Refresh the page, should remain in the same project page with login/credentials intact


YohannParis (approved):
Simple, and yet

PR #1992, opened 2023-10-10
chore: project assets for text files are now documents (not artifacts)

Tom-Szendrey:
# Description
Txt is now a document not an artifact:

Utilizing the code editor for txt files

image

image

### DIFF
--- a/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
+++ b/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
@@ -20,12 +20,22 @@
 					:active="documentView === DocumentView.EXRACTIONS"
 				/>
 				<Button
+					v-if="documentType === DocumentView.PDF"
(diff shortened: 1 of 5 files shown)


Tom-Szendrey:
Setting as draft because i still need to test, im not familiar with the `code-editor` yet so i need to make sure its okay for text files and that this will save back correctly

dvince2:
If we're making .txt's documents, we should also move .md's in as well as I know that Pascale wants to be able to use README.md's to enrich.

Tom-Szendrey:
> If we're making .txt's documents, we should also move .md's in as well as I know that Pascale wants to be able to use README.md's to enrich.

https://github.com/DARPA-ASKEM/terarium/pull/1992/files#diff-8a3c209b372c58b5dd6788202c6fd60c3c59db9ca6c5b7e6ab77da6025c45ee3R366

YohannParis (approved):

PR #1994, opened 2023-10-11
1993 feat limit xdd extraction previews for performance

### LINKED ISSUE #1993: [FEAT]: Limit xDD extraction previews for performance
When a user does a document search, all of the extractions for a document are returned as well. If our search contains ~20 documents with ~20-100 extractions we're now looking at 400-2000 full res images being sent over the wire.

There is a new api from xdd to limit the number of extractions returned.  We should...
* Limit the number of extractions previewed to 5.
* Not limit the number of extractions displayed in the actual document.
* Update the UI to to look something like.... 1 2 *3* 4 5 (50+) for more than 5, and then just 1 *2* 3 for <5 results

dvince2:
# Description

* Leveraging new xDD api to limit returned extractions
* Thumbnails are now limited to 5 with total extractions displayed in brackets
* Extractions in document are not affected.
* New Test

Resolves #1993

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/components/tera-asset-card.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/components/tera-asset-card.vue
@@ -98,13 +98,22 @@
 				</template>
 				<div class="asset-nav-arrows">
 					<span class="asset-pages" v-if="!isEmpty(extractions)">
-						<i class="pi pi-arrow-left" @click.stop="previewMovement(-1)"></i>
-						<span class="asset-count">
-							{{ chosenExtractionFilter }}
-							<span class="asset-count-text">{{ relatedAssetPage + 1 }}</span> of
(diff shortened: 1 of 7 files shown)


packages/client/hmi-client/src/page/data-explorer/components/tera-asset-card.vue
```diff
@@ -215,11 +234,10 @@ watch(
 );
 
 function previewMovement(movement: number) {
-	const newPage = relatedAssetPage.value + movement;
+	const newPage = movement;
 	if (newPage > -1 && newPage < extractions.value.length) {
 		relatedAssetPage.value = newPage;
 	}
```
YohannParis, line 240:
```suggestion
	if (movement > -1 && movement < extractions.value.length) {
		relatedAssetPage.value = movement;
	}
```

YohannParis (approved):

PR #1999, opened 2023-10-11
fix(hmi-server): Increase max upload size for files

### LINKED ISSUE #1995: [BUG]: Increase max upload size for files
**Describe the bug**
Update the server/client to accept reasonably large files. Currently it looks like there is a limit of ~1MB

dvince2:
* Increase the limit in our nginx instance to 100mb as well.

Resolves #1995


dgauldie (approved):
LGTM

YohannParis (approved):

PR #2001, opened 2023-10-11
style: simplify overview page - make banner fit in the actual header

shawnyama:
# Description

Note the background fills the bottom with white. Before the banner background was used as the main background. The banner background is moved to the header of `tera-asset.vue` when Overview is chosen.

image

### DIFF
--- a/packages/client/hmi-client/src/components/asset/tera-asset.vue
+++ b/packages/client/hmi-client/src/components/asset/tera-asset.vue
@@ -14,7 +14,7 @@
 			</aside>
 		</header>
 		<template v-if="!hideIntro">
-			<header id="asset-top" ref="headerRef">
+			<header id="asset-top" :class="isOverview && 'overview-banner'" ref="headerRef">
 				<section>
 					<!-- put the buttons above the title if there is an overline -->
 					<div v-if="overline" class="vertically-center">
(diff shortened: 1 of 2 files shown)


mwdchang:
Not sure if it is possible or how difficult it is, can we leverage pageType/assetType instead of an isOverView predicate? It makes tera-asset a little more idiomatic in terms of reading.

YohannParis:
I don't understand the need to do this task?

shawnyama:
> I don't understand the need to do this task?

Overview page wasn't fitting the space properly. Just got annoyed looking at it while removing tabs.

packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -14,7 +14,11 @@
 			</aside>
 		</header>
 		<template v-if="!hideIntro">
-			<header id="asset-top" ref="headerRef">
+			<header
+				id="asset-top"
+				:class="pageType === ProjectPages.OVERVIEW && 'overview-banner'"
```
YohannParis, line 19: Use [class binding](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes) instead.
```suggestion
				:class="{ 'overview-banner': pageType === ProjectPages.OVERVIEW }"
```

YohannParis (approved):

PR #2006, opened 2023-10-11
fix: equation to model and default configurations

blanchco:
# Description

Fixes for equations to amr:
- when saving equations to amr, the editing state will close.

Fixes for default config:
Before:
- when creating a model and there were no default configurations, an empty default config would be made and never updated.
- specifically when called `equations_to_amr`, the knowledge-middleware would create a model configuration that was no the 'Default config', we then ended up with an empty 'Default Config' and a new model configuration from knowledge-middleware. 

After:
- an empty model config will not be made automatically when a new model is create (there is an empty check for states and transitions)
- when the model is updated and there are no configurations a default config will be created if it does no already exist.

This PR will need a follow-up [PR](https://github.com/DARPA-ASKEM/knowledge-middleware/issues/96) for it to fully be completed.  In the knowledge-middleware the model configuration through `equations_to_amr` will need to be called 'Default config'.

https://github.com/DARPA-ASKEM/terarium/assets/33158416/ac1d2522-6e1d-4df4-93a8-c21526510f54


packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -181,11 +181,17 @@ async function fetchConfigurations() {
...
+			!isModelEmpty(model.value)
 		) {
 			await addDefaultConfiguration(model.value);
-			tempConfigurations = await getModelConfigurations(model.value.id);
+			setTimeout(async () => {
+				// elastic search might still not update in time
+				tempConfigurations = await getModelConfigurations(model.value?.id!);
+				modelConfigurations.value = tempConfigurations;
+			}, 800);
+			return;
```
blanchco, line 194: wrapping this in a timeout...it seems that model configuration aren't updated immediately

YohannParis (approved):

PR #2018, opened 2023-10-13
1960 bug new users are unable to see projects after they get set a user role

bigglesandginger:
# Description

Resolves #(issue)


packages/server/src/main/resources/application.properties
```diff
@@ -32,8 +32,13 @@ terarium.clientConfig.clientLogShippingIntervalMillis=3000
 ########################################################################################################################
 terarium.keycloak.url=https://keycloak.staging.terarium.ai
 spring.security.oauth2.client.provider.keycloak.issuer-uri=${terarium.keycloak.url:localhost}/realms/${terarium.keycloak.realm:Terarium}
+#spring.security.oauth2.client.provider.keycloak.user-name-attribute=preferred_username
```
dvince2, line 35: If none of these are used, should we leave out?

bigglesandginger: great question, I had commented them out because I worried they would be needed - and then forgot - will delete

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/httputil/IoUtil.java
```diff
@@ -0,0 +1,247 @@
...
+                        AclEntryPermission.APPEND_DATA, AclEntryPermission.READ_NAMED_ATTRS,
+                        AclEntryPermission.WRITE_NAMED_ATTRS, AclEntryPermission.EXECUTE,
+                        AclEntryPermission.READ_ATTRIBUTES, AclEntryPermission.WRITE_ATTRIBUTES,
+                        AclEntryPermission.DELETE, AclEntryPermission.READ_ACL, AclEntryPermission.SYNCHRONIZE)
+                .build();
+        acl.add(entry);
+        view.setAcl(acl);
+    }
+
+    public static void printOut(String msg) {
```
dvince2, line 216: To be captured correctly in the logger this should be annotated with `@Slf4J` and then use the application logger.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/httputil/HttpUtil.java
```diff
@@ -0,0 +1,498 @@
...
+                .build();
+        sslsf = new SSLConnectionSocketFactory(theContext);
+    }
+
+    public static void setSkipCertificateValidation() {
+        if (!tlsWarningEmitted.getAndSet(true)) {
+            // Since this is a static util, it may happen that TLS is setup many times in one command
+            // invocation (e.g. when a command requires logging in). However, we would like to
+            // prevent this warning from appearing multiple times. That's why we need to guard it with a boolean.
+            System.err.println("The server is configured to use TLS but there is no truststore specified.");
```
dvince2, line 276: These should also go through application logger.

dvince2 (approved):

PR #2043, opened 2023-10-17
fix: fix equations to amr

### LINKED ISSUE #2032: [BUG]: Configuration created from eqs to amr produced garbled parameter names
To reproduce:
1. Create a new model
2. Enter the equations and click save
=> The generated configuration has weird parameter names that do not correspond to the names of the parameters in the equations.

image

echl:
# Description

* As per [this slack thread](https://askemgroup.slack.com/archives/C03THCGK2DU/p1697563475264709), state and transition ids not be interpreted as indicating semantic interpretations
* Transition names are not outputted as a result of equation-to-amr 
* **This is a hack** as a proper fix will require more time - will probably have unintended consequences

Resolves #2032

### DIFF
--- a/packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
+++ b/packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
@@ -160,7 +160,7 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 			.style('paint-order', 'stroke')
 			.style('fill', 'var(--text-color-primary')
 			.style('pointer-events', 'none')
-			.html((d) => d.id);
+			.html((d) => d.label);
 
 		// transitions expression text
 		transitions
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
```diff
@@ -219,7 +219,7 @@ export class PetrinetRenderer extends BasicRenderer<NodeData, EdgeData> {
 			.style('paint-order', 'stroke')
 			.style('fill', 'var(--text-color-primary')
 			.style('pointer-events', 'none')
-			.text((d) => d.id);
+			.text((d) => d.label);
```
mwdchang, line 222: I would suggest that we pass-in a custom label functor, rather than assuming that everyone wants this change.

E.g. we could either pass it in the constructor, or have something like `renderer.setLabelFormatter` before the render call invocation.

YohannParis:
@mwdchang should we close this?

YohannParis:
This has been fixed by TA1 by going back to normal

PR #2059, opened 2023-10-18
fix: extraction and other concepts

blanchco:
# Description

* Fix to show extractions when clicked
* Fix to show Other Concepts (unaligned variables)

Video shows:
 1) clicking on the extractions tag will show the extractions
 2) other concepts are now properly mapped

https://github.com/DARPA-ASKEM/terarium/assets/33158416/f381243a-f86f-4493-af8d-647f881752bf

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
@@ -87,11 +87,11 @@
 						v-for="(parameter, i) in parameters"
 						:key="parameter.id"
 						:class="[
-							{ active: isRowEditable === `parameter-${parameter.id}` },
-							`parameter-${parameter.id}`
+							{ active: isRowEditable === `${VariableTypes.PARAMETER}-${parameter.id}` },
+							`${VariableTypes.PARAMETER}-${parameter.id}`
 						]"
 					>
-						<template v-if="isRowEditable === `parameter-${parameter.id}`">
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -43,6 +43,7 @@
 			:highlight="highlight"
 			@update-model="updateModelContent"
 			@fetch-model="fetchModel"
+			:key="model?.id"
```
blanchco, line 46: needed to add a key here to properly refresh this component when navigating to and from different models

packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
```diff
@@ -531,11 +577,18 @@ const otherConcepts = computed(() => {
...
+	const unalignedKeys = Object.keys(extractions.value).filter((k) => !ids.includes(k));
+
+	let unalignedExtractions: Dictionary<any>[] = [];
+	unalignedKeys.forEach((key) => {
+		unalignedExtractions = unalignedExtractions.concat(
+			extractions.value[key.toString()].filter((e) => e.type === 'anchored_extraction')
+		);
+	});
+
+	return unalignedExtractions ?? [];
```
blanchco, line 591: unaligned keys can be the string "null" or "undefined" so needed a way to pass the falsy check.  This will also map _all_ unaligned key instead of just the first one found.

YohannParis (approved):

Issue #2027, opened 2023-10-16
[BUG]: selecting an asset in project view breaks


shawnyama:
**Describe the bug**
Choosing an asset breaks when done in a certain order (steps below).

**To Reproduce**
1. Open a project with more than one model and at least one other asset type
2. Select a model then select another model you have
3. Then select an asset of another type
4. See how the asset isn't chosen

Example video (note asset selection breaks as soon as I go from one model to another):

https://github.com/DARPA-ASKEM/terarium/assets/28237258/1904d70f-1276-415a-9676-ab45ea7d7ed9

shawnyama:
Can't reproduce this anywhere anymore

PR #2061, opened 2023-10-18
fix: asset selection in project page

shawnyama:
# Description

We should avoid conditionally rendering `<AccordionTab>`s. The `active-tab` prop in `<Accordion>` expects all the tabs specified in it to be rendered. If they are not rendered the component breaks - which breaks switching assets. See the [video](https://github.com/DARPA-ASKEM/terarium/issues/2027).

In order to hide empty accordion tabs I added a css rule which applies `display: none` if the tab empty. The rule uses the `has` property which is properly supported in Chromium based browsers. Firefox users may still see an AccordionTab without any content.

Feel free to use `v-if` inside the scope `<AccordionTab>` but not on the component itself: 
`<AccordionTab v-if="condition that ruins things">`


mwdchang:
Tagging @Tom-Szendrey here, since he was fixing a problem along the same areas where the collapsed accordions created an update problem.

shawnyama:
Noticing that `tera-external-publications.vue` has the warn error related to the Accordion. Can't narrow this one down. Asset switching doesn't break in the video but it did happen to me once when switching through publications

https://github.com/DARPA-ASKEM/terarium/assets/28237258/48b27a61-3ab0-4a4e-838d-e616137d83e2

Tom-Szendrey:
I had a similar issue that was to do with the accordion not correctly updating upon opening an accordion tab
`tera-model-editor` would generate a renderer based off of a closed accordion tab and have a width + height = 0. When the tab was reopened it would not retrigger this. 
On the Accordion level I added `v-bind:lazy="true"`

could a similar solution be used here?

If v-if is an issue inside accordiontab we should raise this in hmi chat so it wont be as commonly done

shawnyama:
Update:

After chatting with @Tom-Szendrey I want to try out rerendering the component. For example when switching from one `tera-model.vue` to another it should reload the component instead of populating the same one with different data. A rerender may fix this.

 The CSS rule forces us to work with AccordionTabs in a specific way which is not really sustainable. 
 
 Also I am not coming across these Accordion related errors that is breaking the asset switching anymore... did something change?

PR #2064, opened 2023-10-19
feat: implement new text viewer/editor

### LINKED ISSUE #1996: [TASK]: Document Text Viewer should be prettier and more functional
The code-editor currently being used needs a bit of fixing up to make it more compatible for how its being used

echl:
# Description

Resolves #1996

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,5 +38,8 @@
 		"lint-staged": "13.2.3",
 		"prettier": "2.8.8",
 		"typescript": "5.0.4"
+	},
+	"dependencies": {
+		"@vueup/vue-quill": "latest"
 	}
 }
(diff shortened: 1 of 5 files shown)


Tom-Szendrey:
It looks like youre replacing `code-editor.vue` with your new component is that correct?
Is there a need to not remove the `code-editor.vue` file in that case?

Tom-Szendrey:
> It looks like youre replacing `code-editor.vue` with your new component is that correct? Is there a need to not remove the `code-editor.vue` file in that case? I believe it was previously only used in the 2 files that youre removing it from I may be mistaken

This was answered in standup, please disregard

PR #2065, opened 2023-10-20
Persist chart configs when switching between different runs

### LINKED ISSUE #2034: [TASK]: Preserve chart settings when rerunning a model and select latest output on node drilldown
**Describe the task**
Preserve chart settings when rerunning a simulation.

1. On a workflow, add a model and pipe it into simulate probabilistic
2. Run it and chart the I variable
3. Open the model and change the config (e.g. make one of the parameter a distribution) 
4. Rerun the simulation
=> output 2 is created but the chart settings are lost which forces the user to reselect the I variable in order to see the new simulation results. The chart settings should be preserved.

Also, when you drilldown on the simulation node and go to the output tab, the dropdown is showing output1 instead of output2. It should show the latest run by default.

Jami159:
Demo video: https://capture.dropbox.com/mGz7OQen0wshrOty

Resolves #2034

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
@@ -210,7 +210,7 @@ const handleSelectedRunChange = () => {
 	lazyLoadSimulationData(selectedRun.value.runId);
 
 	const state = _.cloneDeep(props.node.state);
-	// set the active status for the selected run in the chart configs
+	// set the active status for the selected run in the run configs
 	Object.keys(state.simConfigs.runConfigs).forEach((runId) => {
 		state.simConfigs.runConfigs[runId].active = runId === selectedRun.value?.runId;
 	});
--- a/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
@@ -22,40 +22,52 @@
 				placeholder="Select a simulation run"
 				@update:model-value="handleSelectedRunChange"
 			/>
-			<tera-simulate-chart
-				v-if="runResults[selectedRun?.runId]"
-				:run-results="runResults[selectedRun.runId]"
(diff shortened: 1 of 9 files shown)


YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -429,29 +433,34 @@ function appendOutputPort(
...
+		const state = node.state as SimulateJuliaOperationState;
+		if (state.simConfigs.chartConfigs.length === 0) {
+			state.simConfigs.chartConfigs.push([]);
+		}
+		state.simConfigs.runConfigs[port.value[0]] = {
+			runId: port.value[0],
+			active: true
+		};
+	}
+
```
mwdchang, line 463: This section feels a little strange, as per the FIXME comment on the big if-block. Ideally we don't want to mess with state logic at this level, and now that state-logic is even more complex than before only exacerbates the situation. 

Let have a chat and see if we can work around this.

packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -22,40 +22,52 @@
...
-			</span>
+			<template v-if="runResults[selectedRun?.runId]">
+				<tera-simulate-chart
+					v-for="(cfg, idx) in node.state.simConfigs.chartConfigs"
+					:key="idx"
+					:run-results="runResults[selectedRun.runId]"
+					:chartConfig="{ selectedRun: selectedRun.runId, selectedVariable: cfg }"
+					has-mean-line
+					@configuration-change="configurationChange(idx, $event)"
+				/>
```
mwdchang, line 33: We will want to refactor this tera-simulate-chart as a subsequent task, it can not decide if it wants to operate on an array or operate on an object, and in turn that is confusing to whoever is using the component.

packages/client/hmi-client/src/workflow/tera-workflow.vue
Jami159: moved this logic out into `tera-simulate-node-julia.vue` and `tera-simulate-node-ciemss.vue`

mwdchang (approved):

PR #2072, opened 2023-10-20
feat: impl tera-text-editor

### LINKED ISSUE #1996: [TASK]: Document Text Viewer should be prettier and more functional

echl:
- Implement barebones `tera-text-editor` component to replace remaining uses of `code-editor`
- Contains pdf viewer for pdf files and rich text editor for text files
- No ability to save edits to text files yet

https://github.com/DARPA-ASKEM/terarium/assets/16358728/3263ad89-07bb-4c20-81cc-fcdc2684e00b

Resolves #1996


mwdchang:
@echl can you provide a few brief descriptions describing what the PR is trying to accomplish, and how to go about testing it? The issue you are pointing to isn't very helpful in terms of understanding the context.

echl:
> @echl can you provide a few brief descriptions describing what the PR is trying to accomplish, and how to go about testing it? The issue you are pointing to isn't very helpful in terms of understanding the context.

Yes, will do when pr is finalized :)

packages/client/hmi-client/package.json
```diff
@@ -37,6 +37,7 @@
 		"@uncharted.software/facets-core": "3.2.6",
 		"@uncharted.software/facets-plugins": "3.2.5",
 		"@vue/runtime-dom": "3.3.4",
+		"@vueup/vue-quill": "latest",
```
YohannParis, line 40: please use `yarn add --exact @vueup/vue-quill` to have the a precise package number.

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -140,8 +140,16 @@ const props = defineProps<{
...
-const code = ref<string>();
+const viewOptions = computed(() => {
+	if (doc.value?.fileNames?.at(0)?.endsWith('.pdf')) {
+		return [extractionsOption, pdfOption];
+	}
+	return [extractionsOption, txtOption];
+});
+const extractionsOption = { value: DocumentView.EXTRACTIONS, icon: 'pi pi-list' };
+const pdfOption = { value: DocumentView.PDF, icon: 'pi pi-file-pdf' };
+const txtOption = { value: DocumentView.TXT, icon: 'pi pi-file' };
```
YohannParis, line 151: Please add markdown support as well using the text-editor.

YohannParis (changes requested):

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -171,25 +179,27 @@ async function openTextDocument() {
...
+			doc.value = document;
+			if (doc.value?.fileNames?.at(0)?.endsWith('.pdf')) {
+				if (view.value === DocumentView.TXT) {
+					view.value = DocumentView.PDF;
+				}
+			} else {
+				await openTextDocument();
+				if (view.value === DocumentView.PDF) {
+					view.value = DocumentView.TXT;
 				}
```
mwdchang, line 202: I don't quite understand the if-block logic here.

If we open up a file that ends with "pdf", why do we need to check what is the current view-value instead of just directly changing it to PDF?  If we have more view-types like markdown or code, wouldn't this make the logic quite complicated?

echl: There are three DocumentView values - PDF, TXT, and EXTRACTIONS. The logic here is to ensure the switcher stays on the same "side" when you switch between documents. I.e. if it was on extractions, it stays on extractions. If it was on PDF/TXT, it stays on PDF/TXT, depending on file type.

mwdchang: If the idea is to avoid "resetting" `view.value` then this pattern might be less verbose.

```
if (file_is_pdf) {
  if (view.value !== PDF) view.value = PDF
} else if (file_is_text) {
  if (veiw.value !== TEXT) view.value = TEXT
}
```

packages/client/hmi-client/src/components/documents/tera-text-editor.vue
```diff
@@ -0,0 +1,25 @@
...
+</template>
+
+<script setup lang="ts">
+import { ref, watch } from 'vue';
+import { QuillEditor } from '@vueup/vue-quill';
+import '@vueup/vue-quill/dist/vue-quill.snow.css';
+
+const props = defineProps<{
+	initialText: string;
+}>();
```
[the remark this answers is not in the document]
mwdchang: No you misunderstood. I am saying why ware we calling it `initialText`, it gives the impression that the text is going to go under some kind of transformation.

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
echl: That doesn't quite work because if `view.value === DocumentView.EXTRACTIONS`, the view will be switched to pdf or text view rather than remaining on extractions. And it's not any less verbose other than not having any braces and personally I prefer using braces even for single line if statements :P

mwdchang: Okay - I think there are some convoluted business logic here that I am not aware of. We can take this on later if needed.

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
echl: vue quill doesn't natively handle markdown but there are some vue quill modules I can try.

YohannParis: No worries, then we should display Mardown as simple text for now.

packages/client/hmi-client/src/components/documents/tera-text-editor.vue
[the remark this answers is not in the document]
echl: Ah. Open to suggestions for a different name, if I could I would name the prop and the ref text. `initialText` makes sense to me as the "initial value of `text`", which is true. And technically `text` could be transformed.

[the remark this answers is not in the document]
echl: In fact `text` *is* transformed as user can change the content displayed in the editor (though nothing is being saved currently)

YohannParis (approved):

packages/client/hmi-client/src/components/documents/tera-text-editor.vue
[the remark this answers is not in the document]
mwdchang: Erm... how about `textContent`? 

If we are allowed to changed the text, then we need to be careful here because changing props is generally an anti-pattern in Vue, as it can have unforeseen impact upstream if the ancestor components are also depending on the same field. I lack the full context here so maybe @YohannParis and yourself can decide the best route to follow.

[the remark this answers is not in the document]
echl: `textContent` sounds good

We won't be changing the props since even though `text` is a ref of `prop.text`, changes to `text` will not change `prop.text` and any upstream components since `prop.text` is a primitive type.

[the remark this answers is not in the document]
mwdchang: Okay thanks - I think I misread the props handling code. Let's just make the name change and merge 👍

PR #2078, opened 2023-10-23
feat: update code-to-amr endpoint params

shawnyama:
# Description

- Update params for `code-to-amr`
- Add loading spinner to Create model from code button

https://github.com/DARPA-ASKEM/terarium/assets/28237258/143dd50a-7a73-411e-9aa1-ba0b3a3bd62c

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/knowledge/KnowledgeMiddlewareProxy.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/knowledge/KnowledgeMiddlewareProxy.java
@@ -101,16 +101,18 @@ public interface KnowledgeMiddlewareProxy {
 	/**
 	 * Transform source code to AMR
 	 *
-	 * @param codeId      (String): id of the code artifact
-	 * @param name        (String): the name to set on the newly created model
-	 * @param description (String): the description to set on the newly created model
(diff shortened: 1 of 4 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/knowledge/KnowledgeMiddlewareProxy.java
```diff
@@ -101,16 +101,18 @@ ResponseEntity<JsonNode> postProfileDataset(
...
+	 * @param description  (String): the description to set on the newly created model
+	 * @param dynamicsOnly (Boolean): whether to only run the amr extraction over specified dynamics from the code object in TDS
 	 * @return (ExtractionResponse)
 	 */
 	@PostMapping("/code_to_amr")
 	ResponseEntity<ExtractionResponse> postCodeToAMR(
 		@RequestParam("code_id") String codeId,
 		@RequestParam("name") String name,
-		@RequestParam("description") String description
+		@RequestParam("description") String description,
```
dvince2, line 114: I think there may be an issue here with the controller having these marked as optional but the proxy having them marked as required.  The Controller should also define a default value (part of the @RequestParam), or the proxy should also be updated to mark `name`, `description` and `dynamicsOnly` as not required

dvince2 (approved):

PR #2079, opened 2023-10-23
Show input configs in simulate drilldown

### LINKED ISSUE #1900: [TASK]: In the simulation drill down, show the input configs
Screenshot 2023-09-20 at 4 50 29 PM

Jami159:
Resolves #1900

Screenshot 2023-10-23 at 11 55 10 AM

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
@@ -23,6 +23,24 @@
 				@update:model-value="handleSelectedRunChange"
 			/>
 			<template v-if="runResults[selectedRun?.runId]">
+				<div class="metadata-container">
+					<Text><b>Run ID:</b> {{ selectedRun.runId }}</Text>
+					<Text
+						><b>Configuration name:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].configName }}</Text
+					>
+					<Text
+						><b>Method selected:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].method }}</Text
+					>
+					<Text><b>Number of samples:</b> {{ node.state.numSamples }}</Text>
+					<Text
+						><b>Start step:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].timeSpan?.start }}
+						<b>End step:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].timeSpan?.end }}</Text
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
```diff
@@ -159,21 +159,24 @@ const watchCompletedRunList = async (runIdList: string[]) => {
...
 	const state = _.cloneDeep(props.node.state);
 	if (state.simConfigs.chartConfigs.length === 0) {
 		state.simConfigs.chartConfigs.push([]);
 	}
 	state.simConfigs.runConfigs[runIdList[0]] = {
 		runId: runIdList[0],
-		active: true
+		active: true,
+		configName: port.label,
+		timeSpan: state.currentTimespan
```
mwdchang, line 172: You changed timeSpan from string => Timespan, so when you are making assignment it is by reference and not by value. Can you confirm if this is okay? Or if we need to clone a copy of it?

Ditto for the CIEMSS side of things.

Jami159: Just did some testing with updating the timespan in the UI and it seems like the initial assignment is preserved. But should I clone it just to be extra safe?

packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -23,6 +23,24 @@
...
+						><b>Method selected:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].method }}</Text
+					>
+					<Text><b>Number of samples:</b> {{ node.state.numSamples }}</Text>
+					<Text
+						><b>Start step:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].timeSpan?.start }}
+						<b>End step:</b>
+						{{ node.state.simConfigs.runConfigs[selectedRun.runId].timeSpan?.end }}</Text
+					>
```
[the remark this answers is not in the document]
Jami159: Updated to use `ul`, `li`, and `span`

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
Jami159: Updated to assign a copy of `state.currentTimespan` instead of direct assignment.

mwdchang (approved):

PR #2083, opened 2023-10-23
task: vue-router refresh issue

YohannParis:
# Description

* Using `keycloack-js` with `vue-router` brings multiple source of conflict: https://github.com/keycloak/keycloak/issues/14742
* Tried the first solution to move the login/auth before the mounting of the router
* Changed `vue-router` to use `createWebHistory` instead

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -4,7 +4,7 @@
 	<Toast position="top-right" group="warn" />
 	<Toast position="bottom-right" group="info" />
 	<Toast position="bottom-right" group="success" />
-	<tera-navbar class="header" :active="!isErrorState" :show-suggestions="showSuggestions" />
+	<tera-navbar class="header" :active="displayNavBar" :show-suggestions="showSuggestions" />
 	<main>
 		<router-view v-slot="{ Component }">
 			<component class="page" ref="pageRef" :is="Component" />
@@ -97,7 +97,7 @@ const route = useRoute();
 const router = useRouter();
 const currentRoute = useCurrentRoute();
 
-const isErrorState = computed(() => currentRoute.value.name === 'unauthorized');
+const displayNavBar = computed(() => currentRoute.value.name !== 'unauthorized');
 
 // This pageRef is used to grab the assetType being searched for in data-explorer.vue, it is accessed using defineExpose
 const pageRef = ref();
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -15,15 +15,16 @@
 		<div class="footer-group">
 			<a target="_blank" rel="noopener noreferrer" @click="isAboutModalVisible = true">About</a>
 			<a target="_blank" rel="noopener noreferrer" :href="documentation">Documentation</a>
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/router/index.ts
```diff
@@ -61,9 +61,7 @@ const routes = [
 
 const router = createRouter({
 	// 4. Provide the history implementation to use. We are using the hash history for simplicity here.
-	history: createWebHashHistory(),
-
-	// short for `routes: routes`
+	history: createWebHistory(),
```
dvince2, line 64: comment will need updating ^

packages/client/hmi-client/src/main.ts
```diff
@@ -15,77 +15,77 @@ import { EventType } from '@/types/Types';
...
-			window.location.reload();
-		} else {
-			await authStore.init();
-			logger.info('Authenticated');
+// If the authentication failed, reload the page
+function failedAuth(e: unknown) {
+	console.error(e);
+	logger.error('Authentication Failed, reloading a the page');
+	window.location.reload();
+}
```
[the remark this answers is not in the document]
YohannParis: Send the user to the homepage for now.

packages/client/hmi-client/src/main.ts
```diff
@@ -15,77 +15,77 @@ import { EventType } from '@/types/Types';
...
+app
+	.use(router)
+	.use(ToastService)
+	.use(ConfirmationService)
+	.use(PrimeVue, { ripple: true })
+	.use(VueKatex)
+	.directive('tooltip', Tooltip);
+
+// Configure Google Analytics
+const GTAG = await axios.get('/api/configuration/ga');
```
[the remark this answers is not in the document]
YohannParis: Done

packages/client/hmi-client/src/main.ts
```diff
@@ -14,78 +13,81 @@ import VueKatex from '@hsorby/vue3-katex';
...
+	await keycloak.updateToken(70);
+}, 6000);
 
-			app.use(router);
-			app.mount('body');
-			logger.info('Application Mounted', { showToast: false, silent: true });
+// Set the hash value of the window.location to null
+// This is to prevent the Keycloak from redirecting to the hash value
+// after the authentication
+window.location.hash = '';
```
YohannParis, line 64: Ugly hack to clean up the URL

packages/client/hmi-client/src/App.vue
```diff
@@ -126,12 +129,12 @@ API.interceptors.response.use(
...
 
+// Update the project when the projectId changes
 watch(
 	() => route.params.projectId,
 	async (projectId) => {
-		useProjects().get(projectId as IProject['id']);
-		// Refetch the list of all projects
-		useProjects().getAll();
+		await useProjects().get(projectId as IProject['id']);
+		await useProjects().getAll();
```
chris-dickson, line 137: If the project id is defined, why would we bother fetching all the projects in here? Isn't that slow and not needed? Why not defer that until it's necessary?

YohannParis: Good point, there is no need to do this here.

dvince2 (approved):

PR #2084, opened 2023-10-23
feat: save dynamic part of a code file

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/code/tera-code.vue
+++ b/packages/client/hmi-client/src/components/code/tera-code.vue
@@ -27,7 +27,6 @@
 						mode="basic"
 						auto
 						chooseLabel="Load file"
-						class="p-button-secondary"
 					/>
 					<Button label="Save" @click="saveCode" />
 					<Button label="Save as new" @click="isCodeNamingModalVisible = true" />
@@ -36,6 +35,7 @@
 						@click="isModelNamingModalVisible = true"
 						:loading="isCodeToModelLoading"
 					/>
+					<Button label="Make highlight dynamic" @click="createDynamic" />
 				</section>
 			</section>
 		</template>
@@ -204,7 +204,7 @@ async function saveCode() {
 		codeName.value = setFileExtension(codeName.value, programmingLanguage.value);
 		const file = new File([codeText.value], codeName.value);
 		const updatedCode = await updateCodeAsset(
-			{ ...existingCode, name: codeName.value },
+			{ ...existingCode, name: codeName.value, files: {} },
 			file,
 			progress
 		);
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -193,44 +257,111 @@ async function initialize(editorInstance) {
...
+			if (fileName === name) {
+				const { block } = files[fileName].dynamics;
+				// Loop through every highlighted block
+				for (let i = 0; i < block.length; i++) {
+					// Avoids rehighlighting
+					if (!existingMarkers.has(block[i])) {
+						// Extract start and end rows and highlight them in the editor
+						const match = block[i].match(/L(\d+)-L(\d+)/) || [];
+						const startRow = parseInt(match[1], 10) - 1;
+						const endRow = parseInt(match[2], 10) - 1;
```
mwdchang, line 278: I think we can move this parsing into a helper function. 

Do you expect this to ever _not_ match the pattern? Because your fallback condition is not going to work for startRow/endRow

```
> const match = []
< undefined

> parseInt(match[2], 10)
< NaN
```

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -158,26 +209,39 @@ const emit = defineEmits(['asset-loaded']);
...
 	ProgrammingLanguage.Julia,
 	ProgrammingLanguage.Python,
 	ProgrammingLanguage.R
 ];
 
+const selectedRangeToString = computed(() =>
+	selectionRange.value
+		? `L${selectionRange.value.start.row + 1}-L${selectionRange.value.end.row + 1}`
+		: ''
+);
```
YohannParis, line 243: @Sorrento110 I remember you mentioning that we are moving away from GitHub style of line selection.
Could you point us to the new way to write those please?

packages/client/hmi-client/src/components/code/tera-code.vue
YohannParis: I agree that this needs to be written/refactor, but could we just merge it for the demo, then create a new GitHub issue about it?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
```diff
@@ -32,6 +34,9 @@ public class Code {
 	/* The name of the file in this code*/
 	private String filename;
 
+	@TSOptional
+	private Map<String, CodeFile> files;
```
YohannParis, line 38: Could you add a comment like the other property of the object

YohannParis (approved):
Thank you, that's an excellent second pass

PR #2085, opened 2023-10-23
task: update MIRAProxy.java

YohannParis:
# Description

* small issue where MIRA endpoints needs the `/api/` in their calls: http://34.230.33.149:8771/docs#/

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MIRAProxy.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MIRAProxy.java
@@ -13,26 +13,26 @@ import software.uncharted.terarium.hmiserver.models.mira.DKG;
 import java.util.List;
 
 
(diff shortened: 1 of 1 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MIRAProxy.java
```diff
@@ -15,24 +15,24 @@
 
 @FeignClient(name = "mira-api", url = "${mira-api.url}")
```
dvince2, line 16: In the end its all the same, but other proxies have opted to add `path = "/api"` to the FeignClient declaration, but, in the end its the same

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/mira/MIRAProxy.java
YohannParis: how so?

Issue #2069, opened 2023-10-20
[FEAT]: Notifications: Update `hmi-server` to interact with Job Schedulers


YohannParis:
### Description
We need to enhance our `hmi-server` to interact directly with all the different job schedulers (`knowledge-middleware`, and simulation services) in order to create and update deferred tasks. This will allow us to efficiently manage and schedule tasks for execution. The integration should be modular, allowing us to support multiple job schedulers in the future.

### Tasks
- [ ] Implement a modular integration service that supports multiple job schedulers.
- [ ] Implement logic to create and update deferred tasks using the selected job scheduler(s).
- [ ] Ensure that the server can monitor the status of scheduled tasks and update the database accordingly.
- [ ] Implement error handling and retries for scheduled tasks.
- [ ] Create documentation for configuring and using the job scheduler integration.

### Acceptance Criteria
- The server is able to interact with the selected job scheduler(s) for creating, updating, and monitoring deferred tasks.
- The integration is modular and can be extended to support additional job schedulers in the future.
- Task scheduling and execution are reliable, with error handling and retries in place.
- Documentation is available for configuring and using the job scheduler integration.

### Additional Information
- Related Issues: depend on #2068

PR #2101, opened 2023-10-26
task: spinner for documents

blanchco:
# Description

* Adding a spinner when navigating to a document, this endpoint does take a while to show - so giving a sense of progress here.
* I figure it wouldn't hurt to add a simple spinner component here and we can define its size.
* Added a snippet of text for the scenario when pdf-extractions haven't completed.  This can probably be refactored when we work on [this](https://github.com/DARPA-ASKEM/terarium/issues/2069) and can listen to these async job properly

https://github.com/DARPA-ASKEM/terarium/assets/33158416/3d58f563-3b74-46eb-9fb9-cb050789a265

### DIFF
--- a/packages/client/hmi-client/src/components/asset/tera-asset.vue
+++ b/packages/client/hmi-client/src/components/asset/tera-asset.vue
@@ -1,5 +1,5 @@
 <template>
-	<main @scroll="updateScrollPosition">
+	<main v-if="!isLoading" @scroll="updateScrollPosition">
 		<slot name="nav" />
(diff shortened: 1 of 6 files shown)


YohannParis (approved):
Neat

blanchco:
> Neat, but this could be added to Tera-Asset as a loading prop!

good idea, i was trying to figure out a way to do this better...I might look at adding this to every asset type tomorrow morning

blanchco:
@YohannParis loading has been added for all assets on their initial load, leveraging the `tera-asset`.

PR #2107, opened 2023-10-27
Task: Adding FUNMAN basic APIs into hmiserver.

Tom-Szendrey:
- Replacement for the other 1889 PR this aims to simplify by not trying to keep up with the "fluid" responses and just use jsonnode

- Changing config -> parts and using base folder for request type to follow convention

- Keeping other PR available rather than overwriting is as im currently having trouble connecting to and testing Funman. Safety is handy


dvince2 (approved):
fun, man!

PR #2113, opened 2023-10-30
task: fix issue where a dataset could not be downloaded

YohannParis:
# Description

* Apparently when opening some `Dataset` asset they might not download properly because of the hash added by keycloak.js.
* This is a second attempt to resolve this issue.

@chris-dickson and @bigglesandginger I added you for awarness of this issue with keycloak.

### DIFF
--- a/packages/client/hmi-client/src/main.ts
+++ b/packages/client/hmi-client/src/main.ts
@@ -21,6 +21,13 @@ import App from '@/App.vue';
 import { useProjects } from '@/composables/project';
 import '@/assets/css/style.scss';
 
+function cleanLocationHash(): void {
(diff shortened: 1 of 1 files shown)


chris-dickson:
Have you considered adding this in the router?

```javascript
router.afterEach(() => {
  window.location.hash = '';
});
```

YohannParis:
> Have you considered adding this in the router?
> 
> ```js
> router.afterEach(() => {
>   window.location.hash = '';
> });
> ```

This is a smart idea!
Do you mind adding this to your PR, if it is necessary, as `keycloack.js` should do it automatically, and I will close mine now that is irrelevant.

PR #2114, opened 2023-10-30
feat: explorer topic selector

blanchco:
# Description

* A way to select topics in the explorer
* additions the filter-bar will be made in an additional PR (eg. document type, number of documents...)

https://github.com/DARPA-ASKEM/terarium/assets/33158416/e7dd4a6b-f09d-4e0d-b694-77efa58c191b

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -7,7 +7,7 @@
 	<tera-navbar class="header" :active="displayNavBar" :show-suggestions="showSuggestions" />
 	<main>
 		<router-view v-slot="{ Component }">
-			<component class="page" ref="pageRef" :is="Component" :key="route.fullPath" />
+			<component class="page" ref="pageRef" :is="Component" :key="route.path" />
 		</router-view>
 	</main>
 	<footer class="footer">
--- a/packages/client/hmi-client/src/assets/css/theme/_variables.scss
+++ b/packages/client/hmi-client/src/assets/css/theme/_variables.scss
@@ -947,4 +947,5 @@ $petri-inputBox: var(--surface-0);
 	--node-header-hover: var(--text-color-subdued);
 	--z-index-modal: 1000;
 	--keyboard-zindex: calc(var(--z-index-modal) + 100);
+	--explorer-background-colour: #F4F7FA;
 }
--- a/packages/client/hmi-client/src/assets/css/theme/_variables.scss
+++ b/packages/client/hmi-client/src/assets/css/theme/_variables.scss
@@ -947,5 +947,4 @@ $petri-inputBox: var(--surface-0);
 	--node-header-hover: var(--text-color-subdued);
 	--z-index-modal: 1000;
 	--keyboard-zindex: calc(var(--z-index-modal) + 100);
-	--explorer-background-colour: #F4F7FA;
 }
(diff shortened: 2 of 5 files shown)


packages/client/hmi-client/src/App.vue
```diff
@@ -7,7 +7,7 @@
 	<tera-navbar class="header" :active="displayNavBar" :show-suggestions="showSuggestions" />
 	<main>
 		<router-view v-slot="{ Component }">
-			<component class="page" ref="pageRef" :is="Component" :key="route.fullPath" />
```
blanchco, line 10: chaning this to path, the explorer would be remounted for every query change, we only need to have a unique key for the path without params

packages/client/hmi-client/src/assets/css/theme/_variables.scss
```diff
@@ -947,4 +947,5 @@ $petri-inputBox: var(--surface-0);
 	--node-header-hover: var(--text-color-subdued);
 	--z-index-modal: 1000;
 	--keyboard-zindex: calc(var(--z-index-modal) + 100);
+	--explorer-background-colour: #F4F7FA;
```
YohannParis, line 950: Could you check which `--surface` (or `--grey`) value is the closest to it and use that instead?

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -1,104 +1,112 @@
...
-				:model-value="resourceType"
-				@change="if ($event.value) resourceType = $event.value;"
-				:options="assetOptions"
-				option-value="value"
+	<main class="data-explorer-container flex flex-column">
+		<section class="filter-bar">
+			<tera-filter-bar
+				:topic-options="topicOptions"
+				@filter-changed="executeNewQuery"
+			></tera-filter-bar>
```
YohannParis, line 7:
```suggestion
			<tera-filter-bar
				:topic-options="topicOptions"
				@filter-changed="executeNewQuery"
			/>
```

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -161,16 +169,18 @@ const isLoading = ref<boolean>(false);
...
 
 const assetOptions = ref([
 	{ label: 'Documents', value: ResourceType.XDD, icon: 'pi pi-file' },
 	{ label: 'Models', value: ResourceType.MODEL, icon: 'pi pi-share-alt' },
 	{ label: 'Datasets', value: ResourceType.DATASET, icon: 'pi pi-database' }
 ]);
 
+const topicOptions = ref([
+	{ label: 'Covid-19', value: 'xdd-covid-19' },
+	{ label: 'Climate Change', value: 'climate-change-modeling' }
```
YohannParis, line 181:
```suggestion
	{ label: 'Climate Weather', value: 'climate-change-modeling' }
```

Sorry, because it is an U.S.A. project, we cannot use the words _Climate Change_ in the app.

YohannParis (approved):
Thank you for the quick turnaround.

packages/client/hmi-client/src/assets/css/theme/_variables.scss
blanchco: Ya I can change it to one of those - I was going with the Figma design colours.  Should we use these defaults instead?

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
blanchco: fair enough!

packages/client/hmi-client/src/assets/css/theme/_variables.scss
blanchco: Went with `surface-100` in the meantime

PR #2116, opened 2023-10-30
if a CSV row has null values, we set them as "" instead

dgauldie:
# Description

Resolves #(issue)


mwdchang:
Out of curiosity, what are our thoughts of using `val`?

https://projectlombok.org/features/val

dvince2 (approved):
We should move soon to a library to handle csv parsing

Issue #2117, opened 2023-10-30
[TASK]: MIRA Stratify Pt 2


Tom-Szendrey:
- [ ] Add library/code samples needs to be added to the beaker repo.
- [ ] need to hook up the run button.
- [ ] make the connection from the wizard and notebook.
- [ ] Weird state management when deleting from the forms

Tom-Szendrey:
Dan closed off

PR #2118, opened 2023-10-30
Ts/1871 task stratify a model workflow

Tom-Szendrey:
# Description
This includes some first step front end work on mira stratify
The following will need to be done before this node is functional: https://github.com/DARPA-ASKEM/terarium/issues/2117 

image


packages/client/hmi-client/src/workflow/ops/stratify-mira/stratify-mira-operation.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface StratifyOperationStateMira {
+	strataGroups: StratifyGroup[];
+}
+
+export const StratifyMiraOperation: Operation = {
+	name: WorkflowOperationTypes.STRATIFY_MIRA,
+	displayName: 'Stratify MIRA',
+	description: 'Stratify a model',
+	inputs: [{ type: 'modelConfigId', label: 'Model configuration', acceptMultiple: false }],
```
mwdchang, line 19: Should probably be model-id instead of model-configuration-id

Tom-Szendrey: I think this should be model-id when our model selection node is changed as well? 
Will be hard to demo if it has a disconnect with all other nodes

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -0,0 +1,200 @@
...
+			size="small"
+			:active="activeTab === StratifyTabs.notebook"
+			@click="activeTab = StratifyTabs.notebook"
+		/>
+	</div>
+	<div v-if="activeTab === StratifyTabs.wizard" class="container">
+		<div class="left-side">
+			<h4>Stratify Model <i class="pi pi-info-circle" /></h4>
+			<p>The model will be stratified with the following settings.</p>
+			<teraStratificationGroupForm
```
shawnyama, line 24: Use kebab-case 🔎

mwdchang (approved):

PR #2119, opened 2023-10-30
feat: group/user admin fixes and polish

echl:
# Description

* Implement service methods
* Loading skeletons/indicators where applicable
* Refresh users list when changing permissions in group and user view
* And and delete users to groups

https://github.com/DARPA-ASKEM/terarium/assets/16358728/61afa75f-9185-4201-a879-f86496cad280

### DIFF
--- a/packages/client/hmi-client/src/services/groups.ts
+++ b/packages/client/hmi-client/src/services/groups.ts
@@ -17,7 +17,10 @@ export async function addGroupUserPermissions(
 	relationship: string
 ) {
 	const response = await API.post(`/groups/${groupId}/permissions/user/${userId}/${relationship}`);
-	return response;
+	if (response.status === 200) {
+		return true;
+	}
+	return false;
 }
 
 export async function updateGroupUserPermissions(
--- a/packages/client/hmi-client/src/services/groups.ts
+++ b/packages/client/hmi-client/src/services/groups.ts
@@ -29,10 +29,13 @@ export async function updateGroupUserPermissions(
 	const response = await API.put(
 		`/groups/${groupId}/permissions/user/${userId}/${oldRelationship}?to=${newRelationship}`
 	);
-	return response;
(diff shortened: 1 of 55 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/permissions/GroupsController.java
```diff
@@ -40,9 +40,9 @@ public ResponseEntity<List<PermissionGroup>> getGroups(
 		@RequestParam(name = "page", defaultValue = "0") Integer page
 	) {
 		List<PermissionGroup> groups = reBACService.getGroups();
-		groups.forEach(group -> {
-			group.setPermissionRelationships((new PermissionRelationships()));
-		});
+		// groups.forEach(group -> {
```
dvince2, line 43: may as well remove?

packages/client/hmi-client/src/services/groups.ts
```diff
@@ -17,7 +17,10 @@ export async function addGroupUserPermissions(
 	relationship: string
 ) {
 	const response = await API.post(`/groups/${groupId}/permissions/user/${userId}/${relationship}`);
-	return response?.data ?? null;
+	if (response.status === 200) {
+		return true;
+	}
+	return false;
```
YohannParis, line 23:
```suggestion
	return response.status === 200;
```

Why the if/else?

packages/client/hmi-client/src/services/roles.ts
YohannParis, line 1: Missing some try/catch and `logger` to notify the user about errors.

packages/client/hmi-client/src/services/user.ts
YohannParis, line 1: Same things about error handling

YohannParis (approved):

PR #2130, opened 2023-10-31
2127 bug missing role checking for api calls

bigglesandginger:
Adds Role checking for api calls (except the google analytics and the like)


dvince2:
I'm worried by the requirement to have the `@Secured(Roles.USER)` on seemingly every single endpoint. That feels like a easy point of failure when adding new endpoints if we forget.  Without that, are these endpoints not secured? Is it possible to make the default state for endpoints be this, and then we add annotations if we specifically want something unsecured (or with a different kind of security?)

bigglesandginger:
> I'm worried by the requirement to have the `@Secured(Roles.USER)` on seemingly every single endpoint. That feels like a easy point of failure when adding new endpoints if we forget. Without that, are these endpoints not secured? Is it possible to make the default state for endpoints be this, and then we add annotations if we specifically want something unsecured (or with a different kind of security?)

I am not sure if it is possible to remove a blanket policy (or add additional requirements) - I will look into that option.  There is the flip side argument that is if it is explicitly stated, then one knows why the user can or cannot access a given endpoint.

bigglesandginger:
I investigated the security, and due to how our endpoints are currently configured we cannot add blanket permissions in `SecurityConfig` that would be any less complex that adding `@Security` but would require developers to know that they would have to edit that file in addition to the controller they are writing.  If we wanted to have a blanket method we would have to modify all API endpoints to fall into at least one of two categories and maybe more - ie `/api/public/...`, `/api/protected/...`, and maybe even `/api/admin/...`

dvince2 (approved):
From our discussion, we'll go ahead with the annotations for now but if it becomes an ongoing issue of things being forgotten we'll revisit the idea of a blanket policy on renamed endpoints.
✅ 'ed so you can merge in as soon as tests are working and main merged in

PR #2131, opened 2023-10-31
task: save assets from github url

blanchco:
# Description

* Add a way to download a github repo to code asset (zip)
* Fix: use `files` key instead of `name` to determine file name and type


YohannParis:
@blanchco I'm sorry, I wan't clear on the task.
The goal is to add an extra button which import the WHOLE repo into the code asset.
The code asset can handle multiple files, they do not need to be separate assets.

blanchco:
> @blanchco I'm sorry, I wan't clear on the task. The goal is to add an extra button which import the WHOLE repo into the code asset. The code asset can handle multiple files, they do not need to be separate assets.

Oh interesting, ill add this functionality then!

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -15,36 +15,41 @@
...
+						/>
+						<Button
+							label="Remove all dynamics"
+							:disabled="!codeAsset?.files"
+							@click="removeAllDynamics"
+						/>
+					</template>
+					<template v-else>
+						<Button label="Download Zip" />
+					</template>
```
blanchco, line 52: Add a separate view for zip types for now...until we can get a new view for a file tree

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -429,15 +442,21 @@ watch(
...
+
+				const filename = Object.keys(code.files)[0];
+
+				const text = await getCodeFileAsText(props.assetId, filename);
 				if (text) {
 					codeText.value = text;
 				}
-				programmingLanguage.value = getProgrammingLanguage(codeName.value);
+				programmingLanguage.value = code.files[filename].language;
 			} else {
```
blanchco, line 460: use first file found from code asset

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/TDSCodeController.java
```diff
@@ -148,6 +183,18 @@ private ResponseEntity<Integer> uploadCodeHelper(String codeId, String fileName,
...
+			codeFile.setProgrammingLanguageFromFileName(fileName);
+
+			Map<String, CodeFile> fileMap = code.getFiles();
+
+			if(fileMap == null){
+				fileMap = new HashMap<>();
+			}
+			fileMap.put(fileName, codeFile);
+			code.setFiles(fileMap);
+			codeProxy.updateAsset(codeId, convertObjectToSnakeCaseJsonNode(code)).getBody();
```
blanchco, line 197: need to update the files object when we upload a file, we append to the map and then update the asset

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
```diff
@@ -54,7 +48,8 @@ public class Code {
 	public enum ProgrammingLanguage {
 		PYTHON("python"),
 		R("r"),
-		Julia("julia");
+		Julia("julia"),
+		ZIP("zip");
```
blanchco, line 52: I know zip is not a language, but i couldn't find a better way to represent zips for now...Im sure this will be fixed when we add the file tree

YohannParis (approved):

PR #2134, opened 2023-10-31
fix: keycloack user config not being downloaded

YohannParis:
Solve the problem where the user was trying to be initialized before keycloak was.


dgauldie (approved):
LGTM

PR #2139, opened 2023-11-01
task: update variables extractions

YohannParis:
# Description

* Update how extraction variables are called and added Provenance call.
* This is most likely going to be removed from the front-end and moved to the back-end.
* Mostly tidying up for no reasons.

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
@@ -15,7 +15,7 @@
 		<Button text label="Extract variables" :loading="isLoading" @click="dialogForExtraction" />
 		<Button
 			text
-			:disabled="props.assetType != ResourceType.MODEL"
+			:disabled="props.assetType != AssetType.Models"
 			:label="`Align extractions to ${assetType}`"
 			:loading="isLoading"
 			@click="dialogForAlignment"
@@ -52,7 +52,7 @@
 					</div>
 				</div>
 			</div>
-			<aside v-if="dialogType === DialogType.EXTRACT && assetType === ResourceType.MODEL">
+			<aside v-if="dialogType === DialogType.EXTRACT && assetType === AssetType.Models">
 				<p>Which extraction service would you like to use?</p>
 				<RadioButton
 					v-model="extractionService"
@@ -84,7 +84,6 @@ import Dialog from 'primevue/dialog';
 import DataTable from 'primevue/datatable';
 import Column from 'primevue/column';
 import RadioButton from 'primevue/radiobutton';
-import { ResourceType } from '@/types/common';
 import {
 	alignModel,
 	Extractor,
@@ -176,10 +175,10 @@ const sendForEnrichment = async () => {
 
 	isLoading.value = true;
 	// Build enrichment job ids list (profile asset, align model, etc...)
-	if (props.assetType === ResourceType.MODEL) {
+	if (props.assetType === AssetType.Models) {
 		const profileModelJobId = await profileModel(props.assetId, selectedResourceId);
 		jobIds.push(profileModelJobId);
-	} else if (props.assetType === ResourceType.DATASET) {
+	} else if (props.assetType === AssetType.Datasets) {
 		const profileDatasetJobId = await profileDataset(props.assetId, selectedResourceId);
 		jobIds.push(profileDatasetJobId);
 	}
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -84,25 +84,29 @@ import Dialog from 'primevue/dialog';
...
+	mapAssetTypeToProvenanceType,
+	RelationshipType
+} from '@/services/provenance';
 import { isDocumentAsset } from '@/utils/data-util';
 import TeraAssetLink from './tera-asset-link.vue';
 
 const props = defineProps<{
 	documents?: Array<{ name: string | undefined; id: string | undefined }>;
-	assetType: ResourceType;
+	assetType: AssetType;
```
blanchco, line 109: 👍

blanchco (approved):
looks good!

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -967,6 +967,7 @@ export enum ProvenanceType {
     Artifact = "Artifact",
     Code = "Code",
     Document = "Document",
+    Workflow = "Workflow",
```
shawnyama, line 970: I think this change is supposed to be done in `AssetType.java` instead. That'll fix the linting error.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -80,6 +97,28 @@ ResponseEntity<Model> getModel(
...
+		body.setTypes(List.of(ProvenanceType.DOCUMENT));
+		final JsonNode provenanceResults = provenanceProxy.search(body, "models_from_document").getBody();
+		final JsonNode resultsNode = provenanceResults.get("result");
+
+		// If there are results, fetch the Document Assets, gather their extractions
+		if (resultsNode != null && resultsNode.isArray() && !resultsNode.isEmpty()) {
+			final List<String> documentIds = mapper.convertValue(resultsNode, new TypeReference<List<String>>() {});
+			final List<JsonNode> extractions = new ArrayList<>();
+			documentIds.forEach(documentId -> {
+				final DocumentAsset document = documentProxy.getAsset(documentId).getBody();
```
dvince2, line 113: You may need to put this in a try/catch, because if this call throws an exception (or if any of these calls do) then i believe you'll be kicked out of the `forEach` loop. If you catch the exception (and just log it?) then you can try again for the next documentid

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -80,6 +97,28 @@ ResponseEntity<Model> getModel(
...
+				final JsonNode documentExtractions = mapper.convertValue(document.getMetadata().get("attributes"), JsonNode.class);
+				final List<JsonNode> documentExtractionsAsList = mapper.convertValue(documentExtractions, new TypeReference<List<JsonNode>>() {});
+				extractions.addAll(documentExtractionsAsList);
+			});
+
+			model.setMetadata(model.getMetadata().setAttributes(extractions));
```
dvince2, line 119: You should be able to just do
`model.getMetadata().setAttributes(extractions);`

dvince2: And just to be clear... this will replace the list of attributes.  If you're looking to add to a list that may or may not be there it would be something like

```
if(model.getMetadata().getAttributes() == null)
    model.getMetadata().setAttributes(new ArrayList<>());
model.getMetadata().getAttributes().addAll(extractions)
```

dvince2 (approved):
Two small notes on back end but looks good. Front end not reviewed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
YohannParis: Interesting that an `getXXX().addAll()` will append, kind of becoming a `setXXX()`.

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
YohannParis: I just find it weird, doesn't kind of negate the purpose of a setter then?

Issue #1925, opened 2023-10-03
FUNMAN 5: Workflow and drilldown UI


mwdchang:
Create workflow operation, funman node, and funman drilldown to support user interactions. Note this does not include the charts and utilities to manipulate funman outputs - these are covered in other tasks. Please check with Jami and Pascale as there may be some subtle design changes.

Design Reference:
- https://www.figma.com/file/GkVfEqpItMg67pNgh5oUAd/Terarium---New-%26-Improved?type=design&node-id=15-47982&mode=design&t=lC7kKMEXvYFKCizP-0

pascaleproulx:
@jamiewaese-uncharted @ngraham76 could you update the figma link to point to the latest version as seen in the prototype please? e.g. placing number of samples and tolerance into the 'more options' section, etc. Thanks.

ngraham76:
> @jamiewaese-uncharted @ngraham76 could you update the figma link to point to the latest version as seen in the prototype please? e.g. placing number of samples and tolerance into the 'more options' section, etc. Thanks.

This Figma link points to the Sept 28th version of the file, is there a newer version? If so, I can't find it - @jamiewaese-uncharted will have to update this link as he worked on this.

Issue #1923, opened 2023-10-03
FUNMAN 3: Develop tools/algorithms to slice and dice output for the user interface.


mwdchang:
We need to have algorithms and utilities to slice-n-dice FUNMAN results to provide as input for the various charts and graphs.

Note we need an example of a successful FUNMAN output, @liunelson will provide.

Things needed
- We need to have 2D pairwise slice of the feasible/infeasible regions of 2 parameters.
- We may need to aggregate the output to produce  timeseries
- We need to filter above timeseries by their respective parameters criteria

Things to consider
- Effectively this is a hyber-cube problem, there may be JS-libraries we can leverage to do these types of aggregate/filter/project operations
- The output size is heavy, we need to keep an eye out for performance and memory issues

Design reference:
- https://www.figma.com/file/GkVfEqpItMg67pNgh5oUAd/Terarium---New-%26-Improved?type=design&node-id=15-47982&mode=design&t=EN04kVOYA1MWXcXI-0

Issue #1890, opened 2023-09-22
FUNMAN 2: Additional plumbing and storage (maybe?)


dgauldie:
FUNMAN API is already asynchronous and provides an end-point to fetch query results. However looking at the deployment configurations of funman, it doesn't look like it has any persistent storage capabilities - which means we could lose all the run-results if we redeploy/update.

We should investigate if there are storage options we can turn on/off, and more importantly whether this is something we want to pursue, which would involve storing the results into S3 as we do other simulation-type results, or whether we want to leave things as-is and live with the fact data can be lost.

Tom-Szendrey:
Im marking this as Ready as i do not believe we need this for the hackathon demo which is first prio

All progress that was previously on this ticket has been lost between transition

PR #2143, opened 2023-11-02
Ts/1925 funman workflow

Tom-Szendrey:
Initial funman exploration. Sorry for the massive PR
A few things will be done in future PRs such as:

-Poller to follow regular format for getStatus #2196
-cleanup of UI #1925
-Storing this in TDS as object #1890
-Update renderer to display false boxes for current and future timesteps to more accurately display to the user #1923


packages/client/hmi-client/src/components/funman/tera-constraint-group-form.vue
YohannParis, line 1: Who wrote this abomination!?

Tom-Szendrey: Ill never tell

mwdchang: Are there any actual change requests or constructive comments here?

YohannParis: I'm waiting for the PR to be off Draft to write constructive comments.

dvince2:
back end 👍

packages/client/hmi-client/src/workflow/ops/funman/funman-operation.ts
```diff
@@ -0,0 +1,37 @@
...
+	weights?: number[]; // 1 to 1 mapping with variables
+	timepoints?: FunmanInterval;
+	interval?: FunmanInterval;
+}
+
+export interface FunmanOperationState {
+	currentTimespan: TimeSpan;
+	numSteps: number;
+	tolerance: number;
+	constraintGroups: ConstraintGroup[]; // This should be optional not required.
```
Tom-Szendrey, line 17: FYI I am fixing this at the moment

packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
```diff
@@ -0,0 +1,436 @@
...
+// Grab startTime, endTime, numberOfSteps and create list.
+function getStepList() {
+	const aList = [startTime.value];
+	const stepSize = (endTime.value - startTime.value) / numberOfSteps.value;
+	for (let i = 1; i < numberOfSteps.value; i++) {
+		aList[i] = i * stepSize;
+	}
+	aList.push(endTime.value);
+	return aList;
+}
```
mwdchang, line 302: This algorithm will  allow for steps that are not integers, is that okay with Funman or will we need to round/floor/ceil the numbers?

Tom-Szendrey: Good catch! I will look into this

packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
```diff
@@ -0,0 +1,436 @@
...
+
+						return {
+							name: ele.id,
+							interval: {
+								lb: ele.value,
+								ub: ele.value
+							},
+							label: 'any'
+						};
+					});
```
mwdchang, line 344: - Have we checked that Funman can even process observables?
- I would recommend that we move the parsing out into its own function, makes things easier to read and to do tests.

packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
Tom-Szendrey: timestep will be integer only

startTime, numberOfSteps, endTime are restricted to integers by the UI however the type is `number`
adding floor to stepSize

packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
Tom-Szendrey: good catch on observables.

Good call to throw this into helpers I have made
setModelOptions, setRequestParameters

mwdchang (approved):

YohannParis:
Okay, I will make my PR to fix most of the front-end issues

PR #2144, opened 2023-11-02
fix: clear cookies before keycloak is initialized

shawnyama:
# Description

Potential fix for those who were having trouble with production on Chrome. Also threw in a color tweak that Neil wanted.

Inspiration: https://stackoverflow.com/questions/179355/clearing-all-cookies-with-javascript

### DIFF
--- a/packages/client/hmi-client/index.html
+++ b/packages/client/hmi-client/index.html
@@ -3,7 +3,6 @@
 <head>
 	<meta charset="UTF-8" />
 	<link rel="icon" type="image/png" href="/src/assets/images/terarium-icon.png" />
-	<link rel="manifest" href="./manifest.json">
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/index.html
```diff
@@ -3,10 +3,16 @@
 <head>
 	<meta charset="UTF-8" />
 	<link rel="icon" type="image/png" href="/src/assets/images/terarium-icon.png" />
+	<link rel="manifest" href="./manifest.json">
```
YohannParis, line 6: We do not have a `manifest.json`
```suggestion
```

YohannParis (approved):

packages/client/hmi-client/index.html
shawnyama: oops I was trying something out

PR #2147, opened 2023-11-02
Directory Structure in Code Asset

blanchco:
# Description

* Small fixes and improvements to github code asset
* Added support for a directory structures (i.e multiple files on a code asset). Following this Figma design 
Screenshot 2023-11-06 at 4 14 00 PM
* The redesign will be done in parts for more digestable PRs.
* This will not break code assets with single files for now..

I haven't uploaded any files currently so we are just getting s3 keys not found, but you can see the search changes when clicking on a file.

https://github.com/DARPA-ASKEM/terarium/assets/33158416/9919b785-ce2f-4fc0-aab8-84859f2ef02f


packages/client/hmi-client/src/components/code/tera-directory.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+	<aside class="min-w-10rem max-w-20rem">
```
YohannParis, line 2: Please use CSS. This is an abomination from hell, and should never be in the front-end with a ten foot pole.

packages/client/hmi-client/src/components/code/tera-directory.vue
```diff
@@ -0,0 +1,87 @@
...
+		<Tree
+			v-model:selectionKeys="selectedKey"
+			:value="directoryTree"
+			selectionMode="single"
+			:filter="true"
+			filterMode="lenient"
+			filter-placeholder="Find"
+			@node-select="onNodeSelect"
+		>
+		</Tree>
```
YohannParis, line 12:
```suggestion
		/>
```

packages/client/hmi-client/src/components/code/tera-directory.vue
```diff
@@ -0,0 +1,87 @@
...
+
+const emit = defineEmits(['file-clicked']);
+
+const selectedKey = ref<TreeSelectionKeys>();
+onMounted(() => {
+	directoryTree.value = buildDirectoryTree(props.files);
+});
+
+function buildDirectoryTree(files) {
+	const root: TreeNode = {
```
YohannParis, line 34: Could you make sure the root node is always open by default?
Even better if we can hide it, we do not need it on the interface.

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -48,20 +48,31 @@
...
-			style="height: 100%; width: 100%"
-			class="ace-editor"
-			v-if="programmingLanguage !== ProgrammingLanguage.Zip"
-		/>
+		<div v-if="programmingLanguage !== ProgrammingLanguage.Zip" class="flex h-full">
+			<tera-directory
+				v-if="fileNames.length > 1"
+				:files="fileNames"
+				@fileClicked="onFileSelect"
+			></tera-directory>
```
YohannParis, line 66:
```suggestion
			/>
```

packages/client/hmi-client/src/components/code/tera-code.vue
```diff
@@ -48,20 +48,31 @@
...
-		<v-ace-editor
-			v-model:value="codeText"
-			@init="initialize"
-			:lang="programmingLanguage"
-			theme="chrome"
-			style="height: 100%; width: 100%"
-			class="ace-editor"
-			v-if="programmingLanguage !== ProgrammingLanguage.Zip"
-		/>
+		<div v-if="programmingLanguage !== ProgrammingLanguage.Zip" class="flex h-full">
```
YohannParis, line 61: My eyes, it burns!!! Those classes needs to go... in hell!!!

YohannParis (changes requested):

packages/client/hmi-client/src/components/code/tera-directory.vue
blanchco: 😆 I didn't know we were so against these classes!!

packages/client/hmi-client/src/components/code/tera-directory.vue
```diff
@@ -0,0 +1,87 @@
...
+		children: []
+	};
+
+	files.forEach((filePath) => {
+		const parts = filePath.split('/');
+		let currentNode: TreeNode = root;
+
+		parts.forEach((part, i) => {
+			const isDirectory = i < parts.length - 1;
+			const existingNode = currentNode.children?.find((node) => node.label === part);
```
mwdchang, line 51: This `find(...)` is not particularly efficient. Do we have some notion of what is an average-sized code-asset and how deep the paths go?

packages/client/hmi-client/src/components/code/tera-directory.vue
YohannParis: Those classes are awful, it's like coding in the 90's. It's saying that CSS is a useless technology.

packages/client/hmi-client/src/components/code/tera-directory.vue
YohannParis: Nope, but those should be done on the back-end eventually, so for now, just to have the feature in is good enough. This should be done once on the repo upload and saved as a JSON or something.

blanchco: I can't say I know how deep the paths go on average...but its a downloaded git repo - so it can be any size.  I'll look to see if I can use a hashmap here to improve the performance instead of using `find(...)`

packages/client/hmi-client/src/components/code/tera-directory.vue
blanchco: i see i see, I sort of liked them cause it made it easy to add those simple styles instead of making a new class everytime(i.e flex).  Made the change though to use a class instead

packages/client/hmi-client/src/components/code/tera-code.vue
blanchco: changed!

packages/client/hmi-client/src/components/code/tera-directory.vue
mwdchang: Let's leave it as is then, only Java repos will have very large number of varieties in paths and files - and Java isn't supported by the program currently.

packages/client/hmi-client/src/components/code/tera-directory.vue
blanchco: Done!

https://github.com/DARPA-ASKEM/terarium/assets/33158416/899786d3-7b4a-4ab8-a329-621dfffe4db2

YohannParis (approved):

PR #2152, opened 2023-11-03
chore: move app footer into its own component

shawnyama:
# Description

Put the app footer in its own component so that App.vue is easier to look at.

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -10,81 +10,20 @@
 			<component class="page" ref="pageRef" :is="Component" :key="route.path" />
 		</router-view>
 	</main>
-	<footer class="footer">
-		<img src="@assets/svg/uncharted-logo-dark.svg" alt="logo" class="ml-2" />
-		<div class="footer-group">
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/App.vue
```diff
@@ -10,81 +10,20 @@
...
-			</div>
-		</template>
-		<template #footer>
-			<div class="modal-footer">
-				<p>&copy; Copyright Uncharted Software {{ new Date().getFullYear() }}</p>
-				<Button class="p-button" @click="isAboutModalVisible = false">Close</Button>
-			</div>
-		</template>
-	</tera-modal>
+	<tera-footer class="footer" />
```
YohannParis, line 13:
```suggestion
	<footer><tera-footer /></footer>
```
The fact that we place it in a `<footer>` in the `App.vue` doesn't mean that `<terar-footer>` should be aware of that. It should be an `<aside>` or `<section>` component.

YohannParis (approved):

YohannParis (approved):
Gorgeous 🎉

PR #2153, opened 2023-11-03
350 save xdd extractions when importing publications as documents

### LINKED ISSUE #350: [TASK] Refactor matrix basic configurations
Currently we are passing in a rather large listing of properties to control such thing like
- margins
- row / column labels
- row / column formatter
- ...

This makes the logic handling somewhat complicated for watchers and other reactivity based interactins. We should consolidating these properties into a configuration object

dvince2:
# Description

* When importing an xdd document, grab the extractions and place them in s3
* Assign the correct metadata to the document object.

Resolves #350


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentExtraction.java
```diff
@@ -12,10 +12,16 @@
 @Accessors(chain = true)
 public class DocumentExtraction implements Serializable {
 
-    @JsonAlias("file_name")
+	public static final String TABLE_ASSETTYPE = "table";
+	public static final String EQUATION_ASSETTYPE = "equation";
+	public static final String FIGURE_ASSETTYPE = "figure";
```
YohannParis, line 17: Why are they lowercase here and Capital case in front-end?

dvince2: Thanks! back end isn't caring about case, but, they should be the same. fixed.

YohannParis (approved):

PR #2166, opened 2023-11-06
fix: sse reconnecting too often

YohannParis:
# Description

* For some reasons the `hmi-server` might answer a 503 to the `hmi-client` to the `/api/client-event` endpoint.
* I updated the error status code test to include all `5XX`.
* I added a couple of console.log to explain the issue with the retry times.
* I removed the `init()` call in `main.ts` as the *reconnecting* heartbeat does it anyway, which might produce some conflict.

### DIFF
--- a/packages/client/hmi-client/src/services/ClientEventService.ts
+++ b/packages/client/hmi-client/src/services/ClientEventService.ts
@@ -20,11 +20,6 @@ let lastHeartbeat = new Date().valueOf();
 let backoffMs = 1000;
 
 /**
- * Whether we are currently reconnecting to the SSE endpoint
- */
-let reconnecting = false;
-
-/**
  * An error that can be retried
  */
 class RetriableError extends Error {}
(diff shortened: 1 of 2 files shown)


YohannParis:
@chris-dickson in the [following code](https://github.com/DARPA-ASKEM/terarium/pull/2166/files#diff-127d9dc3521493458894d343c7707df3a8c65472a4e774e3aad4350fde3a2763R85-R99)
I don't understand the need of `reconnecting` has it is never set anywhere else. By default it's at `false`, and after passing the test, it set again at `false`, so the condition will always be true.

```TypeScript
/**
 * Periodically checks if we have received a heartbeat within the configured interval
 * and reconnects if not
 */
setInterval(async () => {
	if (!reconnecting) {
		const config = await getConfiguration();
		const heartbeatIntervalMillis = config?.sseHeartbeatIntervalMillis ?? 10000;
		if (new Date().valueOf() - lastHeartbeat > heartbeatIntervalMillis) {
			reconnecting = true;
			await init();
			reconnecting = false;
		}
	}
}, 1000);
```

packages/client/hmi-client/src/services/ClientEventService.ts
```diff
@@ -84,14 +82,10 @@ export async function init(): Promise<void> {
...
 setInterval(async () => {
-	if (!reconnecting) {
-		const config = await getConfiguration();
-		const heartbeatIntervalMillis = config?.sseHeartbeatIntervalMillis ?? 10000;
-		if (new Date().valueOf() - lastHeartbeat > heartbeatIntervalMillis) {
-			reconnecting = true;
-			await init();
-			reconnecting = false;
-		}
+	const config = await getConfiguration();
```
dvince2, line 85: Without `reconnecting` here, don't we run the risk of calling `init()` while we're still waiting on a previous call to `init()`?

dvince2 (approved):

PR #2174, opened 2023-11-07
feat: edit/create project modal

shawnyama:
# Description

- Shared modal for creating and editing project configuration
- UI for domain and image select is done but is disabled until we have backend support for those
- Remove usage `modal` class where `tera-modal` component is used

https://github.com/DARPA-ASKEM/terarium/assets/28237258/878f1f85-9bc4-429b-b145-571ba467ea3c


YohannParis:
Based on the video, does this PR also closes #2159 ?

shawnyama:
> Based on the video, does this PR also closes #2159 ?

Good catch - I meant to do that but the dialog boxes for those only appear if you're on the home page - will fix that.

shawnyama:
@YohannParis I just migrated the share and remove modals properly. I am not leveraging the `project-menu.ts` composable though. If we were to use that I think we would have to put these modals in `App.vue` instead. Then from `tera-project-menu.vue` you'd trigger the states of the modals using the composable. Not sure what we want to go for.

YohannParis (approved):

YohannParis:
> @YohannParis I just migrated the share and remove modals properly. I am not leveraging the `project-menu.ts` composable though. If we were to use that I think we would have to put the modals in `App.vue` instead. Then from `tera-project-menu.vue` you'd trigger the states of the modals using the composable. Not sure what we want to go for.

I think that how this should be behave honestly. Let me know how hard it would be to do so.

YohannParis (approved):
Thank you, that's much tidier

PR #2175, opened 2023-11-07
Apparently the RoleTypes need to be in the UI

bigglesandginger:
colour me baffled

# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/types/Types.ts
+++ b/packages/client/hmi-client/src/types/Types.ts
@@ -931,6 +931,9 @@ export enum AuthorityType {
 export enum RoleType {
     Admin = "ADMIN",
     User = "USER",
+    Group = "GROUP",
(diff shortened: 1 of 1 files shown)


dvince2:
@shawnyama already has a fix in for this so we're good!

@bigglesandginger with the `@TSModel` annotation at the top of `RoleType.java` you're marking the class as "TypeScript Model", which means the front end will expect these values to show up in Types.ts

bigglesandginger:
> @shawnyama already has a fix in for this so we're good!
> 
> @bigglesandginger with the `@TSModel` annotation at the top of `RoleType.java` you're marking the class as "TypeScript Model", which means the front end will expect these values to show up in Types.ts

You say "you're" but that was someone else.  Also I am going by how this class should be used... and I am surprised it is desired in the front end - oh well

PR #2177, opened 2023-11-07
Stratification workflow 2

mwdchang:
### Summary
Simple stratification through the use of wizard. This demonstrate ability to stratify via beaker-kernel, plus saving the new model back into the project's resources. 

This is still very much an engineering interface and additional development is required:
- Run python code fragment, this will require some thoughts regarding state-management because the running code does not translate back into the wizard interface.
- Make the modal look and behave more like other modals
- Save code fragment into stratify operation's state
- The kernel message passing is hard to follow, consider wrapping it up some kind of closure and expose a fluent interface.

image

### Testing
- Select model from model-operation
- Pipe into stratify-mira
- Run stratify, and save new model if desired


packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -126,41 +149,155 @@ const deleteStratifyGroupForm = (data) => {
...
+				concepts_to_stratify: strataOption.selectedVariables,
+				cartesian_control: strataOption.cartesianProduct
+			}
+		},
+		msgType: 'stratify_request',
+		msgId: createMessageId('stratify_request')
+	};
+	const contextMessage: JupyterMessage = createMessage(messageBody);
+	kernel.sendJupyterMessage(contextMessage);
+};
```
Tom-Szendrey, line 192: As far as keeping history are we just going to throw the context message into strataCodeHistory here?

Tom-Szendrey (approved):

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
mwdchang: Yes I think so - I will do that in the next PR.

PR #2199, opened 2023-11-09
change when/how the projects a user can see are collected.

bigglesandginger:
# Description

This alters what projects are fetched from TDS before determining if the user can read the project.  Does this perform better (faster to get all projects) is up to data volumn.

Next steps would be to provide a `getProjects(projectIds)` function so that we can fetch multiple (not all) projects and not individually.

Resolves #(issue)


bigglesandginger:
Logged in as Adam - old way took 5666ms, new way takes 4019ms (to return 11 projects)
Logged in as Charles - old way took 3509ms, new way takes 1807ms (to return 4 projects)

dvince2 (approved):

PR #2206, opened 2023-11-10
fix: xdd extractions not displaying

blanchco:
# Description

* issue where extractions are not showing from extraction job 
* XDD extractions like 'Figure', KM extraction like 'figure'.

### DIFF
--- a/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
+++ b/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
@@ -146,6 +146,11 @@ enum DocumentView {
 	NOT_FOUND = 'Not found'
 }
 
+enum AssetType {
+	FIGURE = 'figure',
+	TABLE = 'table',
+	EQUATION = 'equation'
+}
 const props = defineProps<{
 	assetId: string;
 	highlight?: string;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -146,6 +146,11 @@ enum DocumentView {
 	NOT_FOUND = 'Not found'
 }
 
+enum AssetType {
+	FIGURE = 'figure',
+	TABLE = 'table',
+	EQUATION = 'equation'
+}
```
YohannParis, line 153: I think those should come from the back-end with the Type Generator:
https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentExtraction.java#L15-L17

YohannParis (approved):

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
blanchco: sure ill look into it!

blanchco: should be good now

YohannParis (approved):

PR #2209, opened 2023-11-13
chore: move workflow drilldown from project to workflow page

mwdchang:
### Summary
This is a refactoring change to move the workflow drilldowns from the project page into the workflow page itself. This will gave us better control over sending and receiving events as everything is encapsulated under the tera-workflow umbrella. 

Note: Since we are transitioning to use MIRA stratify I have commented the Julia stratify options.

Note 2: The eventbus is still being used even though now the drilldowns and workflows are in the same place. Will revisit after we have audited the operations for TA1-TA3.


YohannParis (approved):

shawnyama (approved):

shawnyama:
Maybe this would be a good time to rename `tera-fullscreen-modal` to `tera-drilldown`

mwdchang:
> Maybe this would be a good time to rename `tera-fullscreen-modal` to `tera-drilldown`

I have a follow up PR immediately after this for small cleanups - I will do this then.

PR #2220, opened 2023-11-13
[fix: hmi-server] Use askem_object_limit instead of entity_limit with xdd

### LINKED ISSUE #2218: Use `askem_object_limit` instead of `entity_limit` with xdd

dvince2:
# Description

* Parameter Name Change

Resolves #2218


dvince2:
This test will fail right now as the change has not been made yet by @iross on xdd

YohannParis (approved):

PR #2221, opened 2023-11-13
refactor: move asset components to tera project

shawnyama:
# Description

Resolves #(issue)


packages/client/hmi-client/src/page/project/tera-project.vue
```diff
@@ -131,6 +186,18 @@ const openNewAsset = (assetType: AssetType) => {
 	}
 };
 
+// This conversion should maybe be done in the document component - tera-preview-panel.vue does this conversion differently though...
+const getXDDuri = (docAssetId: string): string =>
```
mwdchang, line 190: Nitpick: please add a FIXME or TODO and more affirmative observations if you think we should do something here, otherwise the comment is a bit frivolous as it isn't clear what follow up actions should be.

mwdchang (approved):

PR #2229, opened 2023-11-14
Plot intermediate loss, states, params

### LINKED ISSUE #1860: [TASK]: Process the intermediate results of Julia calibrate operation
This task follow from #1766 
In #1766 we simply get the intermediate results and log it to the console.
The intermediate results currently returned don't have all the info we want. There's an issue to fix this: https://github.com/DARPA-ASKEM/simulation-service/issues/112
Once this issue is completed and we're getting the info we want, we probably want to display it to the user in some way (TBD).

Jami159:
This PR makes different runs selectable via a dropdown only in the drilldown view - in the workflow node view it shows the latest run, unless the selected run is modified via the drilldown.
It also plots the intermediate loss, params, and state values - only in the workflow node view. In the drilldown view, only the final loss, params, and state values can be viewed.

demo: https://capture.dropbox.com/FSd3e1vlGxVOF0mt

Resolves #1860


packages/client/hmi-client/src/services/calibrate-workflow.ts
mwdchang, line 1: 1. We known the data coming in, so I would suggest we replace the `any` or `any[]` with concrete types
2. Why is width optional but height is mandatory? Can we have both width/height be mandatory? (this makes the chart itself a little easier to reason with).
3. D3 provides several utility functions that are a little more terse and avoid reinventing wheels

```
// e.g. getting the min/max
const [min1, max1] = d3.extent(data.map(d => d.iter));
const [min2, max2] = d3.extent(data.map(d => d.loss));

```

packages/client/hmi-client/src/types/SimulateConfig.ts
```diff
@@ -15,6 +15,18 @@ export type ChartConfig = {
...
+	chartConfigs: string[][];
+};
+
+export type CalibrateStore = {
+	runId: string;
+	active: boolean;
+	loss: { [key: string]: number }[];
+	params: { [key: string]: number };
+};
+
```
mwdchang, line 29: I don't quite understand this - did we get the files mixed up? This is `simulate` but you are adding `calibrate` related definitions, also "CalibrateStore" doesn't seem to be used.

packages/client/hmi-client/src/workflow/ops/calibrate-julia/calibrate-operation.ts
```diff
@@ -79,3 +82,18 @@ export const CalibrationOperationJulia: Operation = {
...
+const showSpinner = ref(false);
+const runInProgress = ref<string>();
+export const useJuliaCalibrateOptions = () => ({
+	runResults,
+	selectedRun,
+	currentIntermediateVals,
+	parameterResult,
+	showSpinner,
+	runInProgress
+});
```
mwdchang, line 99: I don't this will work if you have several Julia-Calibrate operations open, the variables might clobber each other.

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -172,15 +203,34 @@ const modelConfig = ref<ModelConfiguration>();
...
+		runId
+	}))
+);
+const drilldownLossPlot = ref<HTMLElement>();
+
+const currentParams = computed(() =>
+	showSpinner.value
+		? parameterResult.value
+		: props.node.state.calibrateConfigs.runConfigs[selectedRun.value?.runId]?.params
+);
```
mwdchang, line 228: We know when the calibration finishes, so we don't need this kind of logic - also please avoid predicating logic on the state of the UI.

packages/client/hmi-client/src/types/SimulateConfig.ts
Jami159: moved to its own file now

packages/client/hmi-client/src/workflow/ops/calibrate-julia/calibrate-operation.ts
Jami159: removed

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
Jami159: removed

mwdchang (approved):

PR #2240, opened 2023-11-15
[fix: hmi-server] Add safety code to sciml message reading

### LINKED ISSUE #2239: Add safety code to sciml message reading. Begin back end work for individual queues

dvince2:
This is some back end legwork.
* Adding safety to message parsing
* Better error message for problems parsing
* Sciml and pyciemss now use individual queues
* Some safety code in case someone's hostname is staging or production. This isn't an exhaustive safety plan but enough to stop an accident ;)
* Note that this will break sciml messages from coming in in the current form as the sciml service is hard coded against a specific queue name. However since thats all broken anyway, its not a big loss ;)
@Jami159 you'll need to hard code your queue name locally though until part two of this is in

Resolves #2239

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/SimulationEventService.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/SimulationEventService.java
@@ -10,6 +10,7 @@ import org.springframework.amqp.core.Message;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.rabbit.annotation.RabbitListener;
 import org.springframework.amqp.rabbit.core.RabbitAdmin;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import software.uncharted.terarium.hmiserver.configuration.Config;
 import software.uncharted.terarium.hmiserver.models.ClientEvent;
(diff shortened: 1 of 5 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/SimulationEventService.java
```diff
@@ -95,21 +106,23 @@ private void onScimlSendToUserEvent(final Message message, final Channel channel
...
      */
     @RabbitListener(
-            queues = {PYCIEMSS_QUEUE},
+            queues = "${terarium.simulation-status}${terarium.queue.suffix:${terarium.userqueue.suffix}}",
             concurrency = "1")
     private void onPyciemssSendToUserEvent(final Message message, final Channel channel) throws IOException {
-        // TODO this should mirror the implementation of onScimlSendToUserEvent
+
+			/*
+				//TODO implement PyciemssStatusUpdate
```
Tom-Szendrey, line 114: Is there an issue we can link to this todo?

dvince2: There is no issue... logging one would be good but I don't know who or what it would be or what the status/holdup is. @mwdchang ?

mwdchang: I will create an issue if none exist - the hold up is ciemss is doing a big refactor over their codebase and we are not expecting to have a usable intermediate result in the next short little while.

dvince2: In theory once they're done their refactor then the only backend work will be to implement the `PyciemssStatusUpdate` model. Everything else should automagically work.

mwdchang (approved):

PR #2254, opened 2023-11-16
fix: redirect to overview page if you removed an asset that's currently opened

shawnyama:
# Description

If you remove an asset that is currently opened it will now redirect to the overview page instead of staying at the deleted asset.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/1c2d0e68-1ac9-4939-b595-fe9b92794b6f


YohannParis (approved):
Look good

PR #2256, opened 2023-11-16
feat: operator inputs

shawnyama:
# Description

- Put operator inputs into `tera-operator-inputs.vue`
- Edges can be removed by clicking the Unlink button when an input port is hovered
- Added `isOptional` property to `WorkflowPort` to show users if an input is optional
- Simplified css related to input/output ports

Things that are a bit off:
- There is no easy way to reset the name of the input label as it gets overwritten from what was connected to it before. I was thinking it may be worth adding a `defaultLabel` property to the different operation `input.label`s.
- I am not sure if we want to make an issue related to resetting `rem`s to be 16px. The sooner the better because at some point we'd have to change all the px to rem. Not sure if it'll be more annoying to reset the base rem here or in the Figma.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/5654c550-c51b-48af-9414-954c12ca3509


YohannParis:
> There is no easy way to reset the name of the input label as it gets overwritten from what was connected to it before. I was thinking it may be worth adding a defaultLabel property to the different operation input.labels.

I agree, you should do that, we need to update the operators information about `inputs` to add a `defaullLabel`.

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -850,6 +852,12 @@ function createNewEdge(node: WorkflowNode<any>, port: WorkflowPort, direction: W
 	}
 }
 
+function removeEdge(portId: string) {
+	const edge = wf.value.edges.find(({ targetPortId }) => targetPortId === portId);
+	if (edge) workflowService.removeEdge(wf.value, edge.id);
+	else logger.error('Edge not found.');
```
YohannParis, line 858: Please add the `portId` to the error message.

YohannParis (approved):

PR #2257, opened 2023-11-17
Kernel dispatch model from code

### LINKED ISSUE #2247: Apply kernel dispatcher to model-from-code operation
Currently each workflow operation that leverages the kernel hand rolls their own communication mechanism for sending and receiving messages. This is a bit messy and more difficult to build complicated features: because the request/response is not one-to-one, and the response types are not unique. For example:

Request A => { Response X, Response Y, Response Z }
Request B => { Response X, Response Z }
Because the responses are asynchronous it is somewhat difficult to have a cohesive logic around the requests.

The recent change https://github.com/DARPA-ASKEM/terarium/pull/2234 tries to wrap these boiler plate code and have a more systematic way of dealing with message passing.

Apply the kernel dispatch pattern to model-from-code. An example of the kernel dispatch can be found under /temp/JupyterTest.vue to be used as a guideline.

Jami159:
Resolves #2247

### DIFF
--- a/packages/client/hmi-client/src/services/jupyter.ts
+++ b/packages/client/hmi-client/src/services/jupyter.ts
@@ -299,4 +299,10 @@ export class KernelSessionManager {
 	disposeMessage(msgId: string) {
 		this.map.delete(msgId);
 	}
+
+	shutdown() {
+		if (this.jupyterSession) {
+			this.jupyterSession.shutdown();
+		}
+	}
 }
--- a/packages/client/hmi-client/src/services/jupyter.ts
+++ b/packages/client/hmi-client/src/services/jupyter.ts
@@ -301,8 +301,6 @@ export class KernelSessionManager {
 	}
 
 	shutdown() {
-		if (this.jupyterSession) {
-			this.jupyterSession.shutdown();
-		}
+		this.jupyterSession?.shutdown();
 	}
 }
--- a/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
@@ -128,9 +128,7 @@ import 'ace-builds/src-noconflict/mode-r';
 import { ProgrammingLanguage } from '@/types/Types';
 import { WorkflowNode } from '@/types/workflow';
 import { cloneDeep, isEmpty } from 'lodash';
-import { newSession, JupyterMessage, createMessageId } from '@/services/jupyter';
-import { SessionContext } from '@jupyterlab/apputils/lib/sessioncontext';
-import { createMessage } from '@jupyterlab/services/lib/kernel/messages';
+import { KernelSessionManager } from '@/services/jupyter';
 import { JSONObject } from '@lumino/coreutils';
 import { ModelFromCodeState } from './model-from-code-operation';
 
@@ -156,7 +154,7 @@ const selectedModelFramework = ref(
 		: props.node.state.modelFramework
 );
 const modelValid = ref(false);
-const jupyterSession = ref<SessionContext | null>(null);
+const kernelManager = new KernelSessionManager();
 const previewHTML = ref('');
 
 const codeBlock = {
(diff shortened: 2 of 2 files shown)


packages/client/hmi-client/src/services/jupyter.ts
```diff
@@ -299,4 +299,10 @@ export class KernelSessionManager {
 	disposeMessage(msgId: string) {
 		this.map.delete(msgId);
 	}
+
+	shutdown() {
+		if (this.jupyterSession) {
+			this.jupyterSession.shutdown();
+		}
```
YohannParis, line 306: Wouldn't `this.jupyterSession?.shutdoww();` do the same thing?

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -168,145 +166,118 @@ const codeBlock = {
...
-			session.iopubMessage.connect(iopubMessageHandler);
-			setKernelContext();
+	try {
+		const jupyterContext = buildJupyterContext();
+		if (jupyterContext) {
+			await kernelManager.init('beaker', 'Beaker', jupyterContext);
 		}
-	});
+	} catch (error) {
+		console.error('Error initializing Jupyter session:', error);
```
YohannParis, line 175: Please use `logger.error()` for this, and display a toast with a title and message explaining that the notebook cannot be initialized.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -168,145 +166,118 @@ const codeBlock = {
...
-	if (!kernel) {
-		return;
-	}
+function buildJupyterContext() {
 	const contextName =
 		selectedModelFramework.value === ModelFramework.Decapodes ? 'decapodes_creation' : null;
 	const languageName =
 		selectedProgrammingLanguage.value === ProgrammingLanguage.Julia ? 'julia-1.9' : null;
 	if (contextName === null || languageName === null) {
 		console.log("Can't work with current language. Do nothing.");
```
YohannParis, line 191: Please use `logger.info`

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -168,145 +166,118 @@ const codeBlock = {
...
-function setKernelContext() {
-	const kernel = jupyterSession.value?.session?.kernel;
-	if (!kernel) {
-		return;
-	}
+function buildJupyterContext() {
 	const contextName =
 		selectedModelFramework.value === ModelFramework.Decapodes ? 'decapodes_creation' : null;
 	const languageName =
 		selectedProgrammingLanguage.value === ProgrammingLanguage.Julia ? 'julia-1.9' : null;
```
YohannParis, line 189: Because those ENUM are set properly, those test should be done in `kernelManager()`

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -168,145 +166,118 @@ const codeBlock = {
...
-		msgId: createMessageId('context_setup')
+	const messageContent = <JSONObject>{
+		declaration: code
 	};
-	const compileExprMessage: JupyterMessage = createMessage(messageBody);
 
 	modelValid.value = false;
-	kernel.sendJupyterMessage(compileExprMessage);
+
+	const compileExprMessage = kernelManager.sendMessage('compile_expr_request', messageContent);
```
YohannParis, line 219: Are those message define somewhere as an ENUM? If not could we do so?

mwdchang: They are not. It remains to be seen whether enum makes sense or not, because so far these are very specific request that will only get called once.  E.g. you will unlikely have another instance of `compile_expr_request` anywhere in Terarium.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
mwdchang: Kernel manage helps send the context, it does not create the context. The context creation is up to each individual component.

packages/client/hmi-client/src/services/jupyter.ts
Jami159: changed as suggested

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
Jami159: done

mwdchang (approved):

PR #2270, opened 2023-11-17
fix: get default port label by port type

shawnyama:
# Description

Wrote `getPortLabel()` to help return default port labels

https://github.com/DARPA-ASKEM/terarium/assets/28237258/34a46b80-0618-49f0-b91f-b10b3f702074

### DIFF
--- a/packages/client/hmi-client/src/workflow/operator/tera-operator-inputs.vue
+++ b/packages/client/hmi-client/src/workflow/operator/tera-operator-inputs.vue
@@ -15,7 +15,7 @@
 			</div>
 			<span>
 				<label>
-					{{ input.label ?? input.defaultLabel ?? input.type ?? '' }}
+					{{ getPortLabel(input) }}
 					{{ input.isOptional ? '(optional)' : '' }}
 				</label>
 				<!--TODO: label is a string type not an array consider adding this back in if we support an array of labels-->
@@ -31,6 +31,7 @@
 <script setup lang="ts">
 import { PropType } from 'vue';
 import { WorkflowPort, WorkflowPortStatus, WorkflowDirection } from '@/types/workflow';
+import { getPortLabel } from '@/services/workflow';
 import { isEmpty } from 'lodash';
 import Button from 'primevue/button';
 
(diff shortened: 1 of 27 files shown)


packages/client/hmi-client/src/workflow/operator/tera-operator-inputs.vue
```diff
@@ -15,7 +15,7 @@
 			</div>
 			<span>
 				<label>
-					{{ input.label ?? input.type ?? '' }}
+					{{ input.label ?? input.defaultLabel ?? input.type ?? '' }}
```
mwdchang, line 18: This kind of syntax is hard to read, maybe use a computed value or something.

mwdchang:
I am not really a fan of having both `label` and `defaultLabel` as concrete - I think mostly it is the naming than anything else. Maybe the terms "label" and "itemName/assetName" would be better since we are effectively carrying forth the name of the dataset, model, configuration ... and what not.


But thinking about this a bit more - we enforce (or should enforce if we are not) `type` information, and it might be better to derive your default-labels from a typing-system rather than hardwiring them into individual operations which seems a little redundant. So for example, instead of this:

```
{{ input.label ?? input.defaultLabel ?? input.type ?? '' }}
```

We should be able to do something like:

```
{{ getLabel(input) }}
...
const getLabel = (input) => {
  if (input.label) return input.label;
 
  if (input.type === 'modelConfigId') return 'Model Configuration';      
  if (input.type === 'datasetId') return 'Dataset';
  if (input.type === 'modelId) return 'Model';
}
```

packages/client/hmi-client/src/workflow/operator/tera-operator-inputs.vue
YohannParis: Isn't a computed just hiding away the same syntax?

```JS
<label>{{ label }}</label>
...
const label = computed(() => input.label ?? input.defaultLabel ?? input.type ?? '');
```

mwdchang: I was complaining about null-coalesce chain being harder to read, but this comment is made rather redundant by my big blurb about leveraging typing systems.

mwdchang (approved):

PR #2276, opened 2023-11-20
feat: operation outputs

shawnyama:
# Description

- Put operator outputs into tera-operator-outputs.vue
- Edges can be removed by clicking the Unlink button when an output port is hovered
- Fixed dataset node so that it doesn't reinitialize every time the workflow mounts
- Organized css related to input/output ports

https://github.com/DARPA-ASKEM/terarium/assets/28237258/58c3c15a-f417-4b85-b662-c78ddbb57d0c


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -880,7 +880,9 @@ function createNewEdge(node: WorkflowNode<any>, port: WorkflowPort, direction: W
 }
 
 function removeEdge(portId: string) {
-	const edge = wf.value.edges.find(({ targetPortId }) => targetPortId === portId);
+	const edge = wf.value.edges.find(
+		({ targetPortId, sourcePortId }) => targetPortId === portId || sourcePortId === portId
+	);
 	if (edge) workflowService.removeEdge(wf.value, edge.id);
 	else logger.error(`Edge with port id:${portId} not found.`);
 }
```
mwdchang, line 888: This function appears to have a split-brain, your logics would seeming indicate you want to remove multiple edges, but your execution logic only removes a single edge. What exactly do you want to do?

Imagine a Y-junction and you click "unlink" at "A"
```
         /----  X
A ==  
        \ ----- Y
```

shawnyama: When this is called from `operator-inputs` we want to match the `portId` with `targetPortId`. With `operator-outputs` we looking to match `sourcePortId`. I'll split those up more.

mwdchang: I meant you seemingly want to match multiple edges, but your code

```
if (edge) workflowService.removeEdge(wf.value, edge.id);
```

removes only one, I am not sure if you want to remove a single-edge? or all-edges attached to a particular port-id.

shawnyama: I thought for inputs I'd remove one and for outputs I'd remove all. I assumed `removeEdge` in `services/workflow.ts` removes all edges specifically for outputs but I read that wrong. Do you think it's fine to keep it removing one edge for now or should I edit `removeEdge` to remove all edges related to the one port id?

mwdchang: I see, yeah we can edit removeEdge in "tera-workflow", I guess you will want something like this:

```
const edges = wf.value.edges.filter(...)
edges.forEach(edge => {
  workflowService.removeEdge(wf.value, edge.id);  
});

```

mwdchang (approved):

PR #2278, opened 2023-11-20
task: project loading state

blanchco:
# Description

* Project loading state for better UI responsiveness
* This is mostly a look and feel thing I decided was better than how loading a project/switching between project is happening right now 

* In the before screen recording you can see that the previous project and its assets are temporarily visible even though the user can navigated to a new project

BEFORE:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/442411e8-0245-4215-a096-9e4372e260a9

AFTER:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/5957c0ab-dfe3-4ca3-9c68-4e77858dbc75


shawnyama (approved):
Nice

PR #2291, opened 2023-11-21
task: tera drilldown header

blanchco:
# Description

* Drilldown header component

Multiple tabs, full header design

https://github.com/DARPA-ASKEM/terarium/assets/33158416/7f4a7f93-cea7-40b2-b27f-6634d28019cf

Big close clickable area

https://github.com/DARPA-ASKEM/terarium/assets/33158416/ab985afc-faa1-4bd4-ba94-01e240c0c66b

Only one view (no tabs)
Screenshot 2023-11-22 at 12 06 11 PM

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
@@ -13,11 +13,9 @@
 			/>
 		</div>
 		<div class="tabs-row">
-			<div>
-				<TabView v-if="views.length > 1" :active-index="activeIndex" @tab-change="onTabChange">
-					<TabPanel v-for="(view, index) in views" :key="index" :header="view" />
-				</TabView>
-			</div>
+			<TabView v-if="views.length > 1" :active-index="activeIndex" @tab-change="onTabChange">
+				<TabPanel v-for="(view, index) in views" :key="index" :header="view" />
+			</TabView>
 			<a :href="documentationUrl" rel="noopener noreferrer">Documentation</a>
 		</div>
 	</header>
@@ -83,6 +81,7 @@ header .tabs-row:deep(.p-tabview .p-tabview-panels) {
 	align-items: center;
 	padding-right: 1.5rem;
 	color: var(--primary-color);
+	margin-left: auto;
 }
 
 :deep(.p-tabview-header:not(.p-highlight) .p-tabview-nav-link) {
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
@@ -53,19 +53,11 @@ header {
 	padding-right: 1rem;
 	height: 116px;
 }
-header .title-row {
-	display: flex;
-	justify-content: space-between;
-	align-items: center;
-	gap: 0.5rem;
-}
 
-header .tabs-row {
+header > * {
 	display: flex;
 	align-items: center;
 	justify-content: space-between;
-	/* padding-left: 1rem;
-    padding-right: 1rem; */
 }
 
 .title-container {
@@ -101,6 +93,10 @@ header .tabs-row:deep(.p-tabview .p-tabview-panels) {
 
 :deep(.p-tabview-header:not(.p-highlight) .p-tabview-nav-link) {
 	background: var(--tab-backgroundcolor-unselected);
+}
+
+:deep(.p-tabview .p-tabview-nav li .p-tabview-nav-link:focus) {
+	background-color: var(--surface-section);
 }
 
 .close-mask {
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
@@ -1,10 +1,8 @@
 <template>
 	<header>
 		<div class="title-row">
-			<div class="title-container">
-				<h4><slot /></h4>
-				<i v-if="props.tooltip" v-tooltip="tooltip" class="pi pi-info-circle" />
-			</div>
+			<h4><slot /> <i v-if="props.tooltip" v-tooltip="tooltip" class="pi pi-info-circle" /></h4>
+
 			<Button
 				class="close-mask"
 				icon="pi pi-times"
@@ -20,7 +18,7 @@
 					<TabPanel v-for="(view, index) in views" :key="index" :header="view" />
 				</TabView>
 			</div>
-			<a target="_blank">Documentation</a>
+			<a :href="documentationUrl" rel="noopener noreferrer">Documentation</a>
 		</div>
 	</header>
 </template>
@@ -34,6 +32,7 @@ const props = defineProps<{
 	activeIndex: number;
 	views: string[];
 	tooltip?: string;
+	documentationUrl?: string;
 }>();
 const emit = defineEmits(['close', 'tab-change']);
 
@@ -60,14 +59,9 @@ header > * {
 	justify-content: space-between;
 }
 
-.title-container {
-	display: flex;
-	align-items: center;
-	gap: 1rem;
-}
-
-.title-container > i {
+.title-row > h4 > i {
 	color: var(--text-color-secondary);
+	margin-left: 1rem;
 }
 
 .p-button.p-button-icon-only.p-button-rounded {
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
@@ -1,0 +1,76 @@
+<template>
+	<header>
+		<div class="actions">
+			<div>
+				<h5><slot /> <i v-tooltip="'Test tooltip.'" class="pi pi-info-circle" /></h5>
+			</div>
+			<Button icon="pi pi-times" text rounded aria-label="Close" @click="emit('close')" />
+		</div>
+		<div class="tabs">
+			<TabView
+				v-if="props.views.length > 1"
+				:active-index="props.activeIndex"
+				@tab-change="onTabChange"
+			>
+				<TabPanel v-for="(view, index) in props.views" :key="index" :header="view" />
+			</TabView>
+
+			<a target="_blank">Documentation</a>
+		</div>
+	</header>
+</template>
+
+<script setup lang="ts">
+import Button from 'primevue/button';
+import TabView, { TabViewChangeEvent } from 'primevue/tabview';
+import TabPanel from 'primevue/tabpanel';
+
+const props = defineProps<{
+	activeIndex: number;
+	views: string[];
+}>();
+const emit = defineEmits(['close', 'tab-change']);
+
+const onTabChange = (event: TabViewChangeEvent) => {
+	emit('tab-change', event);
+};
+</script>
+
+<style scoped>
+header {
+	display: flex;
+	flex-direction: column;
+	gap: 16px;
+	background-color: var(--surface-highlight);
+	padding-top: 20px;
+	padding-left: 16px;
+	padding-right: 16px;
+}
+header .actions {
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	gap: 0.5rem;
+}
+
+header .tabs {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	/* padding-left: 1rem;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
...
+	padding: 1rem;
+}
+.close-mask:before {
+	content: '';
+	position: absolute;
+	top: -100px;
+	right: -100px;
+	bottom: -12px;
+	left: -12px;
+}
```
blanchco, line 121: make close clickable area larger

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -1,27 +1,46 @@
...
+
+const props = defineProps<{
+	title: string;
+	tooltip?: string;
+}>();
 
 const emit = defineEmits(['on-close-clicked']);
+const slots = useSlots();
+
+const views = computed(() => Object.keys(slots));
```
blanchco, line 34: I'm going to give this one a think over when I start with the main container component.  The jist of this is you can pass in as many slots as you want into the drilldown and each will get their own tab view.

```
<tera-drilldown>
    <template #Wizard>
        <!--- dev adds wizard components here -->
    </template>
    <template #Notebook>
        <!--- dev adds notebook components here -->
    </template>
    <!-- ...dev can add as many templates as they want for each tab -->
</tera-drilldown>
```

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
...
+	align-items: center;
+	gap: 0.5rem;
+}
+
+header .tabs-row {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	/* padding-left: 1rem;
+    padding-right: 1rem; */
```
shawnyama, line 68: These share a lot you could do `header > *` for the shared rules

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
...
+	display: flex;
+	align-items: center;
+	padding-right: 1.5rem;
+	color: var(--primary-color);
+}
+
+:deep(.p-tabview-header:not(.p-highlight) .p-tabview-nav-link) {
+	background: var(--tab-backgroundcolor-unselected);
+}
+
```
shawnyama, line 105: Throw this rule in - the designers will love it
```
:deep(.p-tabview-nav li .p-tabview-nav-link:focus) {
	background-color: var(--surface-section);
}
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -238,10 +238,9 @@
 		<tera-drilldown
 			v-if="dialogIsOpened && currentActiveNode"
 			@on-close-clicked="dialogIsOpened = false"
+			:title="currentActiveNode.displayName"
+			:tooltip="'A brief description of the operator.'"
 		>
-			<template #header>
-				<h5>{{ currentActiveNode.displayName }}</h5>
-			</template>
```
YohannParis, line 244: I would have seen something more like
```html
<tera-drilldown v-if="dialogIsOpened && currentActiveNode" @on-close-clicked="dialogIsOpened = false">
  <tera-drilldown-header :tooltip="currentActiveNode.tooltip">
    { currentActiveNode.displayName }}
  </tera-drilldown-header>
  ...
</tera-drilldown>
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis, line 1: This file needs to be rewritten (in another PR).
Because the drilldowns are unique per operator, they should be defined in their folder.

**Current**
```html
<Teleport to="body">
  <tera-drilldown ... >
    <tera-calibrate-julia v-if="currentActiveNode.operationType === WorkflowOperationTypes.CALIBRATION_JULIA" ... />
    <tera-calibrate-ciemss v-if="currentActiveNode.operationType === WorkflowOperationTypes.CALIBRATION_CIEMSS" ... />
    ...
  </tera-drilldown>
</Teleport>
```

**New**
```html
<Teleport to="body">
  <tera-calibrate-julia-drilldown v-if="currentActiveNode.operationType === WorkflowOperationTypes.CALIBRATION_JULIA" ... />
  <tera-calibrate-ciemss-drilldown v-if="currentActiveNode.operationType === WorkflowOperationTypes.CALIBRATION_CIEMSS" ... />
  ...
</Teleport>
```

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -48,31 +64,12 @@ than the main application behind the modal when these render issues come, howeve
...
-	padding: 20px 16px;
-	height: 56px;
-}
-
 main {
 	margin: 0 0 0.5rem;
 	max-width: inherit;
 	/* contentHeight = fullscreen - modalMargin - headerHeight*/
-	height: calc(100vh - 1rem - 56px);
+	height: calc(100vh - 1rem - 116px);
```
YohannParis, line 71: that `headerHeight` should be hardcoded with a CSS variable in `tera-drilldown`

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -1,27 +1,43 @@
...
+				:active-index="selectedViewIndex"
+				:views="views"
+				:tooltip="tooltip"
+				@tab-change="handleTabChange"
+				@close="emit('on-close-clicked')"
+				>{{ props.title }}</tera-drilldown-header
+			>
+			<main>
+				<template v-for="(view, index) in views">
+					<slot v-if="selectedViewIndex === index" :name="view" />
```
YohannParis, line 14: I'm not sure we should switch between tabs like this, this is cause for re-render at every switch (see the spinner in the video).

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
...
+				class="close-mask"
+				icon="pi pi-times"
+				text
+				rounded
+				aria-label="Close"
+				@click="emit('close')"
+			/>
+		</div>
+		<div class="tabs-row">
+			<div>
```
YohannParis, line 18: why the extra `<div>`?

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
...
+				<TabView v-if="views.length > 1" :active-index="activeIndex" @tab-change="onTabChange">
+					<TabPanel v-for="(view, index) in views" :key="index" :header="view" />
+				</TabView>
+			</div>
+			<a target="_blank">Documentation</a>
```
YohannParis, line 23: Add an optional `documentationURL` prop.

```suggestion
			<a :href="documentationURL" rel="noopener noreferrer">Documentation</a>
```

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+	<header>
+		<div class="title-row">
+			<div class="title-container">
```
YohannParis, line 4: Feel free to use [CSS Grid](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Grids) instead of having `<div>` everywhere that do nothing but just convenient styling.

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
blanchco: Im slight changing this so that `display: none` is applied instead so that it does not render the tab each time:

```
<main>
    <div v-for="(view, index) in views" class="slot-container" :class="{'hide-slot': hideSlot(index)}">
	    <slot :name="view" />
    </div>
</main>
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
blanchco: agreed, was going to do that when I start migrating the operations over to the new drilldown

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
blanchco: My go to for layouts has usually been to use Flexbox hence the divs.   I can look into using the Grid styling more

packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis: I wrote an issue for it to avoid confusion.

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
YohannParis: Maybe

```vue
<section v-for="(view, index) in views" class="slot-container" :class="{'hide-slot': hideSlot(index)}">
  <slot :name="view" />
</section>
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
blanchco: I put the `tera-drilldown-header` directly in the `tera-drilldown` since I figured that all drilldowns would all have the same headers, do we forsee cases where dont need this header in the drilldown at all?

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
blanchco: for now I chagned this up a bit so that the title-container isn't even needed.
```
<div class="title-row">
  <h4><slot /> <i v-if="props.tooltip" v-tooltip="tooltip" class="pi pi-info-circle" /></h4>
  
  <Button
  class="close-mask"
  icon="pi pi-times"
  text
  rounded
  aria-label="Close"
  @click="emit('close')"
  />
  </div>
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis: I don't know but this would simplify the use of props, when we can just have it there.
For example we could remove the prop drilling about closing the drilldown:
```vue
<tera-drilldown v-if="dialogIsOpened && currentActiveNode">
  <tera-drilldown-header :tooltip="currentActiveNode.tooltip" @close="dialogIsOpened = false">
    {{ currentActiveNode.displayName }}
  </tera-drilldown-header>
  ...
</tera-drilldown>


```

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
blanchco: This was a way so that if there were no tabs, there would an empty element so that the Documentation link would always stick to the right:

```
header > * {
	display: flex;
	align-items: center;
}
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
blanchco: I get that, I guess something else to note is the tab functionality in the header.  The communication for opening/closing tabs is pretty tightly linked between the drilldown and drilldown-header based on the slots we use in the drilldown.  If we were to separate them out like this we would need to handle the tab behaviour on each drilldown we make rather than once inside the drilldown component itself.    

I figure we can keep it like this for now...Ill keep this in mind while I design the main container since I'm a little iffy on using the slots right now to determine the tabs and might end up rewriting the tab behaviour when I work on that PR

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
YohannParis: That why using CSS grid would avoid those kinds of hacks. Though you could do this just with flex-box.

blanchco: div has been removed, Ive kept flexbox. Did it with `margin-left: auto` on the <a> tag instead

YohannParis (approved):

PR #2292, opened 2023-11-21
start consolidating events

mwdchang:
### Summary
Mostly consolidate how events are passed - right now we have a mix of using event-emitters for bubbling up events, and using an event-bus for broadcasting across the Vue-app. This is a legacy architecture when some components were co-located and some where not.

Here we consolidate as much as we can, favouring using the vue-emit rather than the event-bus.
They are kind of the same thing with respect to the designs we have at this stage, but the idea is to pick a single method to make the logic and implementation easier to follow, less error-prone, and to indicate a more de-facto way of handling things rather than having mixed modalities. 

TODO: Once this has been achieved we can then further refactor and significantly simplify the vue-templates, and also make things more systematic to add new operators.

### Testing
Things should work as before except for Julia-stratify, which is slated to be decommissioned.

### DIFF
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -460,7 +460,7 @@ let workflowDirty: boolean = false;
 
 const isWorkflowLoading = ref(false);
 
-const currentActiveNode = ref<WorkflowNode<any> | null>();
+const currentActiveNode = ref<WorkflowNode<any> | null>(null);
 
 workflowEventBus.on('clearActiveNode', () => {
 	currentActiveNode.value = null;
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -294,6 +294,8 @@
 			<tera-dataset-transformer
 				v-if="currentActiveNode.operationType === WorkflowOperationTypes.DATASET_TRANSFORMER"
 				:node="currentActiveNode"
+				@append-output-port="(event) => appendOutputPort(currentActiveNode, event)"
+				@update-state="(event) => updateWorkflowNodeState(currentActiveNode, event)"
 			/>
 			<tera-model-transformer
 				v-if="currentActiveNode.operationType === WorkflowOperationTypes.MODEL_TRANSFORMER"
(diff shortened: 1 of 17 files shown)


packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis, line 1: Because all those events are identical:
```vue
<tera-OPERATOR
  v-if="currentActiveNode.operationType === WorkflowOperationTypes.OPERATOR_TYPE"
  :node="currentActiveNode"
  @append-output-port="(event) => appendOutputPort(currentActiveNode, event)"
  @update-state="(event) => updateWorkflowNodeState(currentActiveNode, event)"
/>
```

Why not updating `appendOutputPort` and `updateWorkflowNodeState` to fetch `currentActiveNode` themselves and only accept `event` as argument. 

**from**
```typescript
function appendOutputPort(node: WorkflowNode<any> | null, port: { type: string; label?: string; value: any }) {
  if (!node) return;
  ...
}
function updateWorkflowNodeState(node: WorkflowNode<any> | null, state: any) {
  if (!node) return;
  ...
}
```

**to**

```typescript
function appendOutputPort(port: { type: string; label?: string; value: any }) {
  const node = currentActiveNode.value;
  if (!node) return;
  ...
}
function updateOperatorState(state: any) {
  const node = currentActiveNode.value;
  if (!node) return;
  ...
}
```

Then we can update all those events to be:
```vue
<tera-OPERATOR
  v-if="currentActiveNode.operationType === WorkflowOperationTypes.OPERATOR_TYPE"
  :node="currentActiveNode"
  @append-output-port="appendOutputPort"
  @update-state="updateOperatorState"
/>
```

Or if those functions are used elsewhere, to just write function that does the same thing.

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
mwdchang: No particular strong reasons, but it lends a little more flexibility with this style of interfaces - in the sense that you can programmatically string together events that you'd have otherwise have to do physically. e.g. I am not limited to working with currentActiveNode.

Having said that, I am really hoping to have these repetitions condensed down in subsequent PRs, we can imagine something like this instead that would replace the dozen or so repeated drilldown-operation components we currently have.

```
<component
  v-if="currentActiveNode"
  name="currentActiveNode.name"
  :node="currentActiveNode"
  @append-output-port="(event) => appendOutputPort(currentActiveNode, event)"
  @update-state="(event) => updateWorkflowNodeState(currentActiveNode, event)"
/>
```

PR #2293, opened 2023-11-21
feat: operator actions

shawnyama:
# Description

- Created `tera-operator-actions.vue` which holds all operator buttons
- Open drilldown button is now the only button
- Passing functions of the different operators into `tera-operator-actions.vue` requires rethinking the structure of them


YohannParis (changes requested):
Let's talk about the usefulness of this component.

YohannParis:
This has been deemed unnecessary.

PR #2294, opened 2023-11-21
feat: operator title

shawnyama:
# Description

- Created `tera-operator-title` to display title of operator (typically the asset name)


packages/client/hmi-client/src/workflow/operator/tera-operator-title.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<h5 v-if="title">{{ title }}</h5>
```
YohannParis, line 2: I wonder if a slot would be better, as this is the only thing.

YohannParis (commented):
Why this is not part of `tera-header`???

shawnyama:
> Why this is not part of `tera-header`???

It's separate from the header. The header has its own title which is operation type. This title is for showing the asset type when applicable.

packages/client/hmi-client/src/workflow/operator/tera-operator-title.vue
shawnyama: What would the slot have other than an `h5`? I feel like this doesn't have to be a component.

YohannParis (approved):

PR #2296, opened 2023-11-22
refactor: tweaks to operator outputs

shawnyama:
# Description

Addressing @YohannParis comments from after the merge https://github.com/DARPA-ASKEM/terarium/pull/2276

Instead of using `isEmpty` I do a `display: none` for inputs/outputs when they are empty since they add there own padding/margin regardless


YohannParis (approved):
Thank you 🎉

PR #2299, opened 2023-11-22
Add notebook functionality to Mira Stratify

### LINKED ISSUE #2279: Beaker: Run code block through beaker kernel
There are a couple of ways we can send code into the beaker-kernel, 1. through a custom message type that we need to define for every type of operation, or 2. through a a more generic "evaluate code" message.  Let do a little bit of investigation before implementing a solution.

Option-2 would be more ideal ( through a a more generic "evaluate code" message) and it might be something we can just craft into the kernel-dispatcher and use it across the board. According to Matt we should be able to send something like this:

```
{
            msg_type="execute_request",
            content={
                "silent": False,
                "store_history": False,
                "user_expressions": {},
                "allow_stdin": True,
                "stop_on_error": False,
                "code": {{python_code}},
            },
            parent_header=parent_header,
            metadata={
                "trusted": True,
            },
}
```

And we would automatically get back a "preview" response. Let's dig a little deeper in terms of what we can do in terms of plumbing. In addition Matt has mentioned we may want to take a look at `tera-beaker-code-cell ` component to see how it handles the communications back and forth.

Jami159:
This PR adds the ability to stratify a model via the drilldown using the wizard or the notebook.
- if model is stratified via the wizard, then the executed code is reflected in the notebook
- if model is stratified via the notebook, then it is not reflected in the wizard -> there is a warning box stating this
- if models are stratified either via wizard or notebook, then a new output port is added to the stratify node
- if models are explicitly saved, then they are added to the project

Demo video: https://capture.dropbox.com/SBfR4JKnqO5xBEV0

Resolves #2279

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
+++ b/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
@@ -169,11 +169,11 @@ const stratifyRequest = () => {
 };
 
 const handleStratifyResponse = (data: any) => {
-	const code = data.content.executed_code;
-	codeText.value = code;
+	const executedCode = data.content.executed_code;
+	codeText.value = executedCode;
 
 	const state = _.cloneDeep(props.node.state);
-	state.strataCodeHistory.push({ code, timestamp: Date.now() });
+	state.strataCodeHistory.push({ code: executedCode, timestamp: Date.now() });
 	emit('update-state', state);
 };
 
@@ -262,7 +262,7 @@ const saveNewModel = async (modelName: string) => {
 	}, 800);
 };
 
-const initialize = async (editorInstance) => {
+const initialize = (editorInstance) => {
 	editor.value = editorInstance;
 };
 
--- a/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
+++ b/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
@@ -113,8 +113,8 @@ enum StratifyTabs {
 }
 
 interface SaveOptions {
-	addToProject: boolean;
-	appendOutputPort: boolean;
+	addToProject?: boolean;
+	appendOutputPort?: boolean;
 }
 
 const kernelManager = new KernelSessionManager();
--- a/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
+++ b/packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
@@ -21,6 +21,9 @@
 		<div class="left-side" v-if="activeTab === StratifyTabs.wizard">
 			<h4>Stratify Model <i class="pi pi-info-circle" /></h4>
 			<p>The model will be stratified with the following settings.</p>
+			<p v-if="node.state.hasCodeBeenRun" class="code-executed-warning">
+				Note: Code has been executed which may not be reflected here.
+			</p>
 			<tera-stratification-group-form
 				v-for="(cfg, index) in node.state.strataGroups"
 				:key="index"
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -105,6 +124,9 @@ const modelNodeOptions = ref<string[]>([]);
 const teraModelDiagramRef = ref();
 const newModelName = ref('');
 
+const editor = ref<VAceEditorInstance['_editor'] | null>(null);
```
mwdchang, line 127: Does `editor` need to be a ref? I am always a little cautious about sticking 3rd party stuff down into refs cause you never quite know how they would behave when they become reactive.

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -208,16 +234,89 @@ const inputChangeHandler = async () => {
 	}
 };
 
-const saveNewModel = async () => {
-	if (!model.value || !newModelName.value) return;
-	model.value.header.name = newModelName.value;
+const saveNewModel = async (modelName: string, options: Record<string, boolean>) => {
```
mwdchang, line 237: Since you know exactly what may be passed in, please avoid generic`Record<...>`, just create a local interface of it or something, this makes things a bit clearer.

```
interface SaveOptions {
  addToProject?: boolean;
  addToPortXYZ?: boolean;
}
```

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -208,16 +234,89 @@ const inputChangeHandler = async () => {
...
+		const configurationList = await getModelConfigurations(newModel.id);
+		configurationList.forEach((configuration) => {
+			emit('append-output-port', {
+				id: uuidv4(),
+				label: `${newModel.header.name} - ${configuration.name}`,
+				type: 'modelConfigId',
+				value: [configuration.id]
+			});
+		});
+	}, 800);
```
mwdchang, line 275: This section is not needed. The outputs should be `models` and not `model-configurations`

We still have some issues with the separation of concerns between model and model-configurations, but while that is been worked out let's remove the the entire setTimeout section for now.

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -208,16 +234,89 @@ const inputChangeHandler = async () => {
...
+		})
+		?.register('error', (data) => {
+			logger.error(`${data.content.ename}: ${data.content.evalue}`);
+		})
+		?.register('model_preview', (data) => {
+			handleModelPreview(data);
+
+			if (executedCode) {
+				const state = _.cloneDeep(props.node.state);
+				state.hasCodeBeenRun = true;
```
mwdchang, line 315: This is fine, but you need to do the reverse, when you stratify via the wizard, you need to set `hasCodeBeenRun` to false.

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
Jami159: done

mwdchang (approved):

PR #2300, opened 2023-11-23
feat: dynamic component routing for operation drilldown

mwdchang:
### Summary
Using `<component>` to instantiate the drilldown-components dynamically. This will:
- Keep the size of the template section nearly constant
- Allows for a more plug-and-play style: we can import components at runtime via `await import(path_to_component)` and add them into the registry
- Better enforcement of behaviours across components

Next up are the operation-nodes, they will need to go through some major refactoring, as some of them have the the core-logic embedded in the nodes and need to be shifted to the drilldowns. 

Also: 
- I also cleaned out the Julia stratify from the workflow.
- Cleaned out droppedAssetId hook to set the state, we can directly set the state on node-creation

### Testing
No functional changes, workflow should work as before.

### DIFF
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -120,11 +120,6 @@
 						@append-output-port="(event) => appendOutputPort(node, event)"
 						@update-state="(event) => updateWorkflowNodeState(node, event)"
 					/>
-					<!--
-					<tera-stratify-node-julia
-						v-else-if="node.operationType === WorkflowOperationTypes.STRATIFY_JULIA"
-					/>
-					-->
 					<tera-stratify-node-mira
 						v-else-if="node.operationType === WorkflowOperationTypes.STRATIFY_MIRA"
 					/>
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -446,6 +357,23 @@ import {
 const workflowEventBus = workflowService.workflowEventBus;
 const WORKFLOW_SAVE_INTERVAL = 8000;
 
+// FIXME: check if there is a component typing instead of any
```
shawnyama, line 360: I think `Component` is a type. Also I am not sure if using a map is ideal, I thought objects were faster when it comes to reading.

shawnyama: I believe you can just pass a string to the `is` prop in `<component>` and it'll know what you mean. You can do something like `:is="`Tera${Operation.name}`"`. Maybe that loses strictness though

mwdchang: Our component naming scheme isn't as uniform as you think, there are some slight variations that will fail.

Object access is faster, but barely noticeable until we get around to hundreds of thousands of accumulated accesses. Here is a mini-test, obj is a little over twice as fast in this somewhat trivial test. 200K access attempt, 7 milliseconds vs 3 milliseconds.

Result
```
obj 3
map 6.799999952316284
true
```

Test
```
const map = new Map();
const obj = {};
const limit = 80000
for (let i = 0; i < limit; i++) { map.set(i, i); obj[i] = i; }

let start = 0;
let end = 0;
let seq = [];
let objCount = 0;
let mapCount = 0;

for (let i = 0; i < 200000; i++) { seq.push(Math.round(Math.random() * limit)); }

start = performance.now();
for (let i = 0; i < seq.length; i++) {
  if (obj.hasOwnProperty(seq[i])) {
    objCount ++;
  }
}
end = performance.now();
console.log('obj', end - start)


start = performance.now();
for (let i = 0; i < seq.length; i++) {
  if (map.has(seq[i])) {
    mapCount ++;
  }
}
end = performance.now();
console.log('map', end - start)

console.log(objCount === mapCount);
```

YohannParis (approved):

PR #2301, opened 2023-11-23
feat: bitmasking first pass

shawnyama:
# Description

Using bitmasking to read the interaction state of the operator. For this first pass I assign`interactionStatus` to hover and drag states when appropriate. I turn on the `.focus` class when the node is being dragged.

### DIFF
--- a/packages/client/hmi-client/src/services/operator-bitmask.ts
+++ b/packages/client/hmi-client/src/services/operator-bitmask.ts
@@ -6,19 +6,16 @@ enum OperatorInteractionStatus {
 }
 
 function addState(status: number, state: OperatorInteractionStatus) {
-	// eslint-disable-next-line no-bitwise
 	status |= state;
 	return status;
 }
 
 function removeState(status: number, state: OperatorInteractionStatus) {
-	// eslint-disable-next-line no-bitwise
 	status &= ~state;
 	return status;
 }
 
 function isStateActive(status: number, state: OperatorInteractionStatus) {
-	// eslint-disable-next-line no-bitwise
 	return status & state;
 }
 
--- a/packages/client/hmi-client/src/services/operator-bitmask.ts
+++ b/packages/client/hmi-client/src/services/operator-bitmask.ts
@@ -1,73 +1,35 @@
-enum OperatorInteractionStatus {
+// These values represent different interaction states on an operator node
+// The operator header, border, etc. will be styled differently depending on which states are active
+enum Status {
 	Found = 0x0001,
 	Hover = 0x0010,
 	Focus = 0x0100,
 	Drag = 0x1000
 }
 
-function addState(status: number, state: OperatorInteractionStatus) {
+const addState = (status: number, state: Status) => {
 	status |= state;
 	return status;
-}
-
-function removeState(status: number, state: OperatorInteractionStatus) {
+};
+const removeState = (status: number, state: Status) => {
 	status &= ~state;
 	return status;
-}
-
-function isStateActive(status: number, state: OperatorInteractionStatus) {
-	return status & state;
-}
+};
+const isStateActive = (status: number, state: Status) => status & state;
 
 // Hover
-export function addHover(status: number) {
-	return addState(status, OperatorInteractionStatus.Hover);
-}
-
-export function removeHover(status: number) {
-	return removeState(status, OperatorInteractionStatus.Hover);
-}
-
-export function isHover(status: number) {
-	return isStateActive(status, OperatorInteractionStatus.Hover);
-}
-
+export const addHover = (status: number) => addState(status, Status.Hover);
+export const removeHover = (status: number) => removeState(status, Status.Hover);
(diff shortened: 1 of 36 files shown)


packages/client/hmi-client/src/workflow/tera-operator.vue
```diff
@@ -1,8 +1,16 @@
 <template>
-	<main :style="nodeStyle" ref="operator">
+	<main
+		:style="nodeStyle"
+		ref="operator"
+		@mouseenter="interactionStatus |= OperatorInteractionStatus.Hover"
+		@mouseleave="interactionStatus &= ~OperatorInteractionStatus.Hover"
```
YohannParis, line 6: I was thinking to create utility functions.

shawnyama: Do you mean functions for adding/removing bitmasks?

YohannParis: Yes and test: `isHover()`, or `addHover()`, `removeHover()`, etc.

packages/client/hmi-client/src/utils/bitmask.ts
```diff
@@ -0,0 +1,76 @@
...
+	return addState(status, OperatorInteractionStatus.Found);
+}
+
+export function removeFound(status: number) {
+	return removeState(status, OperatorInteractionStatus.Found);
+}
+
+export function isFound(status: number) {
+	return isStateActive(status, OperatorInteractionStatus.Found);
+}
```
shawnyama, line 49: I wonder if writing a class might clean things up a bit. I don't wanna over engineer this yet as this the first pass.

packages/client/hmi-client/src/utils/bitmask.ts
```diff
@@ -0,0 +1,76 @@
+enum OperatorInteractionStatus {
+	Found = 0x0001,
+	Hover = 0x0010,
+	Focus = 0x0100,
+	Drag = 0x1000
+}
+
+function addState(status: number, state: OperatorInteractionStatus) {
+	// eslint-disable-next-line no-bitwise
```
YohannParis, line 9: I think you can just edit our eslint config to remove this `no-bitwise` rule.

packages/client/hmi-client/src/utils/bitmask.ts
YohannParis: if you want this to be less verbose you can use `const` instead of function:
```typescript
export const isFound = (status:number) => isStateActive(status, OperatorInteractionStatus.Found);
```

packages/client/hmi-client/src/utils/bitmask.ts
```diff
@@ -0,0 +1,76 @@
...
+	return removeState(status, OperatorInteractionStatus.Focus);
+}
+
+export function isFocus(status: number) {
+	return isStateActive(status, OperatorInteractionStatus.Focus);
+}
+
+// Drag
+export function addDrag(status: number) {
+	console.log(addState(status, OperatorInteractionStatus.Drag));
```
YohannParis, line 66:
```suggestion
```

YohannParis (approved):
thank you, those functions really makes the code easier to read.

mwdchang:
The way you had written `utils/bitmask.ts` is a mixture of domain logic (e.g. found, not found) and generic logic (e.g. bitwise-or, bitwise-and). So I am tempted to say just throw it into services folder as you can't really just take it and dump into a different project.

packages/client/hmi-client/src/utils/bitmask.ts
```diff
@@ -0,0 +1,76 @@
+enum OperatorInteractionStatus {
+	Found = 0x0001,
+	Hover = 0x0010,
+	Focus = 0x0100,
+	Drag = 0x1000
+}
```
YohannParis, line 6: Because it is not exported, feel free to keep the name simple to make your code less verbose if you want.
```suggestion
enum Status {
	Found = 0x0001,
	Hover = 0x0010,
	Focus = 0x0100,
	Drag = 0x1000
}
```

shawnyama: Makes sense, I just put a comment to clarify that its for interactions

packages/client/hmi-client/src/services/operator-bitmask.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const addState = (status: number, state: Status) => {
+	status |= state;
+	return status;
+};
+const removeState = (status: number, state: Status) => {
+	status &= ~state;
+	return status;
+};
+const isStateActive = (status: number, state: Status) => status & state;
```
YohannParis, line 18: Not to be pedantic, but I think you reverse the meaning of `state` and `status`?
```suggestion
const addState = (state: number, status: Status) => {
	state |= status;
	return state;
};
const removeState = (state: number, status: Status) => {
	state &= ~status;
	return state;
};
const isStateActive = (state: number, status: Status) => state & status;
```

PR #2309, opened 2023-11-24
[TASK] workflow update to handle multiple runs

YohannParis:
# Description

* Workflow can handle multiple runs within a drilldown.
* This update extends the `WorkflowPort` to keep track of multiple runs
* utility functions to change which run is visible, and how to change the state of the operator.


mwdchang:
The util function may be better served under "services/workflow.ts", that is where the majority of workflow related functions live. This util.ts file is very much simulate/calibrate specific and should get rolled up into something else in my opinin.

YohannParis:
> The util function may be better served under "services/workflow.ts", that is where the majority of workflow related functions live. This util.ts file is very much simulate/calibrate specific and should get rolled up into something else in my opinin.

@mwdchang the function has been moved per your recommendation, which makes sense.

PR #2316, opened 2023-11-27
Task: funman output should be slightly more inline with designs

Tom-Szendrey:
To Test you should be able to
- [ ] to see different states in the drop down
- [ ] correctly see the trajectory renderer update when a new state is selected
- [ ] see a table containing the upper and lower bounds for each parameter received from funman

I will add the true box rendering to the table in the next PR which will be under [this](https://github.com/DARPA-ASKEM/terarium/issues/1924) 

image

image


packages/client/hmi-client/src/components/funman/tera-funman-output.vue
```diff
@@ -108,6 +123,46 @@ watch(
...
 	align-items: center;
 	gap: 0.8125rem;
-	align-self: stretch;
+}
+
+.p-chart {
+	width: 100%;
+	height: 200px;
+	margin-top: 0.5em;
+}
```
mwdchang, line 132: Not used

packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -164,37 +164,89 @@ export const renderFumanTrajectories = (
 	element: HTMLElement,
 	processedData: FunmanProcessedData,
 	boxId: string,
-	options: RenderOptions
+	options: RenderOptions,
+	state: string
 ) => {
```
mwdchang, line 169: From an API design perspective, you generally want to have things like "options" or "kwargs" as the last parameter. Order matters:
- Less error prone because options are some times equates with optional, so you should not have parameters afterward
- Similarly, mandatory parameters should be specified before optional parameters for consistency

Tom-Szendrey: Makes sense I will throw options as last param instead of state

mwdchang (approved):

PR #2318, opened 2023-11-27
task: update drilldown nodes

blanchco:
# Description

* small change to the drilldown node so that we do not need to define a tab name if there is only 1 top level component

Updated all nodes to use the drilldown component:
https://github.com/DARPA-ASKEM/terarium/issues/2324

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/code-asset/tera-code-asset-wrapper.vue
+++ b/packages/client/hmi-client/src/workflow/ops/code-asset/tera-code-asset-wrapper.vue
@@ -1,7 +1,7 @@
 <template>
 	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
 		<section>
-			<tera-code :asset-id="props.node?.state?.codeAssetId ?? ''" is-preview />
+			<tera-code :asset-id="node.state?.codeAssetId ?? ''" is-preview />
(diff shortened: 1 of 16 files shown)


blanchco:
@YohannParis @mwdchang I'm going to update like 3-4 drilldown nodes at a time in separate PRs and merge them all at once so that I don't block any dev work.  Starting with some easy ones here

packages/client/hmi-client/src/workflow/ops/code-asset/tera-code-asset-wrapper.vue
```diff
@@ -1,15 +1,22 @@
 <template>
-	<tera-code :asset-id="props.node?.state?.codeAssetId ?? ''" is-preview />
+	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
+		<section>
+			<tera-code :asset-id="props.node?.state?.codeAssetId ?? ''" is-preview />
```
mwdchang, line 4: Should always have "node" and "node.state", no need for `?` there.

mwdchang:
Didn't test, but code looks fine to me.

YohannParis (commented):
It would have been nice to rename those component with the word `drilldown` to understand their purpose, like we do with `node`.

PR #2319, opened 2023-11-27
feat: operator image carousel

shawnyama:
# Description

Using Galleria prime vue component. I put the carousel in the model from code node for testing.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/2303cadc-077b-4632-8476-9966b61c0ac6

### DIFF
--- a/packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
+++ b/packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
@@ -4,7 +4,6 @@
 			<img :src="slotProps.item" alt="Alt" />
 		</template>
 	</Galleria>
-
 	<div v-else class="no-image">
 		<i class="pi pi-image" />
 	</div>


packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
```diff
@@ -0,0 +1,49 @@
...
+<style scoped>
+.p-galleria:deep(.p-galleria-indicators .p-galleria-indicator button) {
+	width: 0.5rem;
+	height: 0.5rem;
+}
+
+.p-galleria:deep(.p-galleria-item-container),
+.no-image {
+	height: 240px;
+	max-height: 240px;
```
YohannParis, line 34: Check with @ngraham76 for the dimensions, because it should not be a square but more landscapy.

packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
```diff
@@ -0,0 +1,49 @@
...
+import { isEmpty } from 'lodash';
+import { PropType } from 'vue';
+import Galleria from 'primevue/galleria';
+
+defineProps({
+	images: {
+		type: Array as PropType<string[]>,
+		default: () => []
+	}
+});
```
mwdchang, line 22: Can you add some comments on what format types are supported here? Are we constrained to use image-hrefs or can we also pass in base64 image data (as strings)? 

e.g. We can display the images, but where are we planning on storing said images and if the images change due to modelers' interactions?

packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
```diff
@@ -0,0 +1,49 @@
...
+	width: 0.5rem;
+	height: 0.5rem;
+}
+
+.p-galleria:deep(.p-galleria-item-container),
+.no-image {
+	height: 240px;
+	max-height: 240px;
+	border: 1px solid #e7e9ee;
+	border-radius: 0.5rem;
```
YohannParis, line 36: could we use a CSS variables from the theme, or add a new value.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code-node.vue
```diff
@@ -1,5 +1,14 @@
 <template>
 	<main>model from code node</main>
+	<tera-operator-image-carousel :images="images" />
```
YohannParis, line 3: Could we use `<slot>` instead?
I would prefer to look like the following examples. Like `tera-operator-image-carousel` is to make a list of images looks better and add functionality to a list of images.
```vue
<tera-operator-image-carousel>
  <img src="@/assets/svg/terarium-icon.svg" alt="Terarium Icon" />
  <img src="@/assets/svg/uncharted-logo-official.svg" alt="Uncharted Logo" />
  <img src="@/assets/svg/terarium-logo.svg" alt="Terarium Logo" />
</tera-operator-image-carousel>
```
Or if the dev prefers JS
```vue
<tera-operator-image-carousel>
  <img v-for="image in images" :src="image.src" :alt="image.alt" />
</tera-operator-image-carousel>
```
If this is too complex or difficult, then don't worry.

packages/client/hmi-client/src/workflow/tera-operator.vue
```diff
@@ -229,7 +229,10 @@ main > ul:empty {
 	gap: 0.25rem;
 }
 
-:deep(li:hover) {
+/* 	TODO: Now that it's supported this would be a great place to use css nesting.
+	The formatter doesn't recognize it yet though, maybe there could be a way to tweak that? 
+*/
```
YohannParis, line 234: Do you mean Prettier?

shawnyama: I don't think it's prettier preventing it. It seems to be a fundamental rule even after I change the formatter:
image
This could still be addressed by prettier settings though

packages/client/hmi-client/src/workflow/operator/tera-operator-image-carousel.vue
ngraham76: @YohannParis @shawnyama I made a mistake in the specs. I wrote: "Preview height is not fixed but (ideally) should not exceed 240px". That should read "...should not exceed **120px**" Sorry for the confusion, I'll update the image in the feature.

packages/client/hmi-client/src/workflow/tera-operator.vue
YohannParis: You have to use an ampersand before your next selector when you use a combinator.
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting#combinators

So it will be:
```css
.toolbar > span {
  & > li {
    background-color: red;
  }
}
```

shawnyama: Still doesn't work. Also you don't need to use `&` necessarily https://developer.chrome.com/articles/css-nesting/#nesting-examples.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code-node.vue
shawnyama: Yeah this is getting into fighting with primevue territory since `Galleria` expects a value and it uses the `#item` template.

YohannParis: Too bad!

packages/client/hmi-client/src/workflow/tera-operator.vue
YohannParis: For normal selector, but when you use a combinator `>, +, etc` you need the ampersand.
Try to update prettier.

shawnyama: I'll do the prettier upgrade in a separate PR since I would want to do even more changes to the CSS in `tera-operator.vue`

PR #2326, opened 2023-11-28
task: update nodes to drilldown

blanchco:
# Description

* Wrapping some more nodes in drilldown:
  - calibrate-ciemss
  - calibrate-ensemble-ciemss
  - calibrate-julia
  - dataset-transformer
  - model-transformer 
  - simulate-ciemss
  - simulate-ensemble-ciemss 
  - simulate-julia


blanchco:
The diff seems a lot worse than it actually is...
For every component just surrounded it with, some don't need to <section> since they're already single top-level component

```
<tera-drilldown>
  <section>
  ...component
  </section>
</tera-drilldown>
```

and added an emit called 'close'

mwdchang (approved):

PR #2327, opened 2023-11-28
fix: funman layout

blanchco:
# Description

* Just surrounding funman with a single div so that the layout is okay with the new tera-drilldown component


Tom-Szendrey (approved):
Drill downs need to be all within 1 div good to know!

PR #2328, opened 2023-11-28
Update nodes 3

blanchco:
# Description

* Updating the final 3 nodes:

  - funman
  - model-from-code
  - stratify-mira

Resolves #(issue)


blanchco:
Final 3 nodes here:

again, just adding a 'close' event and wrapping with tera-drilldown and section where applicable

mwdchang (approved):

PR #2330, opened 2023-11-29
chore: cleanup drag in tera-operator

mwdchang:
### Summary
For some odd reason, `canDrag` prop is causing unnecessary refreshes within the operator components (embedded tera-simulate-charts are refreshing when moving-in  and moving-out of the canvas). 

While the refreshing problem is still under investigation, upon closer look we can remove and simplify the canDrag logic; and also removed unused props while I am in the area.

### Testing
No functional changess, can drag operator nodes as normal.


YohannParis (approved):
Nice and clean!

PR #2333, opened 2023-11-29
feat: placeholder graphic for operators

shawnyama:
# Description

- Added component that generates the appropriate placeholder graphic depending on the node
- For now I just added placeholder graphics for the model and dataset node

https://github.com/DARPA-ASKEM/terarium/assets/28237258/28322800-e19c-4347-af9f-ca06050fb1b9


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -67,7 +67,11 @@
 						@append-output-port="(event: any) => appendOutputPort(node, event)"
 						@append-input-port="(event: any) => appendInputPort(node, event)"
 						@update-state="(event: any) => updateWorkflowNodeState(node, event)"
-					/>
+					>
+						<template #placeholder-graphic>
+							<tera-operator-placeholder-graphic :operation-type="node.operationType" />
```
YohannParis, line 72: This should not be there, the developer has to add it to the `node` itself.
If a new node is created without one, that's a bug, we should not prepare for that.

YohannParis (changes requested):

YohannParis (approved):
Gorgeous 🎉

PR #2335, opened 2023-11-29
refactor: prettier imports and addressing fixmes

mwdchang:
### Summary
The goal here is to provide both a prettier import/registration section in the workflow, and also systematic way of resolving components without needing to fully qualify the node/drilldown names.

If we need to add a different representation of the operator, for example, if we want it as a separate page, we would do something like:

```
import { XYZOp as operation } from './stratify-mira-operation';
import node from './tera-xyz-node.vue';
import drilldown from './tera-xyz/vue';
import page from './tera-xyz-page.vue' // operator in new context

const name = operation.name;
export { name, operation, node, drilldown, page };
```

Also general cleanup
- Clean up SVG-defs of hackathon related markers that are no longer relevant
- Remove refresh-model since we have separated model and model-config
- Remove weird logic where the predicate is on ciemss but the logic is julia

The majority of the work is under "tera-workfiow.vue", other files are mostly re-arranging mod.ts imports.

### Testing
Just reorganizing and removing outdated logic, workflow should work in the same manner as before.


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -399,21 +282,6 @@ function appendOutputPort(
...
-		const state = node.state as CalibrationOperationStateJulia;
-		if (state.chartConfigs.length === 0) {
-			// This only ends up showing the output of the first run, perhaps we should consider showing
-			// the output of the last run, or all runs?
-			state.chartConfigs.push({
-				selectedRun: port.value[0],
-				selectedVariable: []
-			});
-		}
-	}
```
mwdchang, line 416: This should be replaced by update-state emitter, it also doesn't make a lot of sense here - the predicate is on ciemss but operates on julia-state ... so ... chop chop.

YohannParis (approved):
So tidy and organized!

Do you keep this event bus commented to not loose track of it?

shawnyama (approved):

mwdchang:
> Do you keep this event bus commented to not loose track of it?

Nope, that section should be removed, thanks for pointing it out.

YohannParis (approved):

PR #2338, opened 2023-11-30
update: stratify mira drilldown

blanchco:
# Description

* update startify-mira drilldown to now be sectioned with the tera-drilldown component
* create concept of a shared section in drilldown where section can be shared between tabs to avoid duplication

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
@@ -13,6 +13,10 @@
 				<template v-for="(tab, index) in tabs" :key="index">
 					<component :is="tab" v-show="selectedViewIndex === index" />
 				</template>
+
+				<section v-if="slots.shared">
+					<slot name="shared" />
+				</section>
 			</main>
 			<footer v-if="slots.footer">
 				<slot name="footer" />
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -13,6 +13,10 @@
 				<template v-for="(tab, index) in tabs" :key="index">
 					<component :is="tab" v-show="selectedViewIndex === index" />
 				</template>
+
+				<section v-if="slots.shared">
+					<slot name="shared" />
+				</section>
```
blanchco, line 19: dev can add a shared section so that sections that are shared between tabs can be written once

mwdchang: I find the naming of "shared" slightly weird because we also have the notions of sharing projects and assets, if in all our design sketches this right-pane is used for "preview" would it be better to call it as such?

Also @Jami159 and @Tom-Szendrey should be tagged to review here.

blanchco: preview works as well - I guess I was thinking more of a cases perhaps when a section needs to be shared that isn't a preview necessarily?  But I don't think there is a case where it the shared section wouldn't be shared so Ill rename as such

Tom-Szendrey (approved):

Jami159 (approved):
LGTM!

PR #2340, opened 2023-11-30
Task: funman - 4 

Tom-Szendrey:
**Done:**

- Adding multiple trajectories at once as we should not be restricting these to just one.
- Adding rendering boxes per table row

**Still todo:**

- Input text/sliders should have some consequence (not being used anywhere) 
- Look into `processFunman` as its false boxes are not being parsed correctly
- Look into table row when there are no boxes to display (last picture)
- Trajectory colouring depending on true vs false 

image

image


packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -179,14 +178,12 @@ export const renderFumanTrajectories = (
 	d3.select(element).selectAll('*').remove();
 	const svg = elemSelection.append('svg').attr('width', width).attr('height', height);
 
-	const points = trajs.filter((d: any) => d.boxId === boxId);
-
```
mwdchang, line 183: This doesn't look right, the plots, they do not look like mathematical functions

If we are to remove the boxId-filter, then we need to "group by" the boxIds.

What you have here is that you aggregated all the boxes together and it can give weird results.

Tom-Szendrey: Thanks for catching this!

packages/client/hmi-client/src/components/funman/tera-funman-boundary-chart.vue
```diff
@@ -0,0 +1,55 @@
...
+}>();
+
+const boundaryRef = ref();
+const width = 60;
+const height = 60;
+const renderOptions = { width, height };
+
+onMounted(async () => {
+	const funModel = await getQueries(props.funModelId);
+	const processedData = processFunman(funModel);
```
mwdchang, line 29: You are rendering this component multiple times with the same funman response, so instead of doing the same processing over and over I think we should just pass in the processed funman data structure instead of incurring repeated network and processing cost.

Tom-Szendrey: Makes sense. I wasnt sure here whether or not having a massive ref would be better or worse than multiple calls

I will change this

mwdchang (approved):
great, thanks!

PR #2342, opened 2023-12-01
feat: facet preview

shawnyama:
# Description

- Made a facet preview component that accepts an array of key value pairs. 
- Back when I tried to use the facet package I made some small edits related to it that I think are worth keeping.

*Image is just an example I am not keeping the facet preview in the model from code node*
image

### DIFF
--- a/packages/client/hmi-client/src/utils/filters-util.ts
+++ b/packages/client/hmi-client/src/utils/filters-util.ts
@@ -63,7 +63,9 @@ function isEqual(a: Filters, b: Filters) {
 }
 
 function findPositiveFacetClause(filters: Filters, field: ClauseField) {
-	return _.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
+	if (filters)
+		_.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
+	return null;
 }
 
 function removeClause(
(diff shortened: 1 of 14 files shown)


packages/client/hmi-client/src/utils/filters-util.ts
```diff
@@ -63,7 +63,9 @@ function isEqual(a: Filters, b: Filters) {
 }
 
 function findPositiveFacetClause(filters: Filters, field: ClauseField) {
-	return _.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
+	if (filters)
+		_.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
+	return null;
```
YohannParis, line 68: I think you are missing a return here

```js
function findPositiveFacetClause(filters: Filters, field: ClauseField) {
	if (filters)
		return _.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
	return null;
}
```

or 
```js
function findPositiveFacetClause(filters: Filters, field: ClauseField) {
	if (!filters) return null;
  return _.find(filters.clauses, (clause) => clause.field === field && clause.isNot === false);
}
```

YohannParis (approved):

YohannParis (approved):

PR #2353, opened 2023-12-01
feat: tera-carousel

shawnyama:
# Description

This version accepts any series of divs as a slot and puts them in a carousel. 

This component combines features of the carousel made within the asset cards and the image carousel that PrimeVue provides. Numbers or indicators can be used for navigation (prop option)

https://github.com/DARPA-ASKEM/terarium/assets/28237258/08ef71ad-a9a0-4445-beae-3a5351411e00

How text looks:
image


packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code-node.vue
YohannParis, line 1: That's the stuff!

YohannParis (approved):
Well done, better than I thought

PR #2360, opened 2023-12-04
Fix drilldown header height

blanchco:
# Description

* Fix the drilldown header height and move the documentation link based on wether there are tabs or not
* documentation link is optional

https://github.com/DARPA-ASKEM/terarium/assets/33158416/da8d7a82-b283-4719-b008-b8dcc217f680


packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
YohannParis, line 1: This really show that we should leverage CSS grids.

YohannParis (approved):

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
blanchco: perhaps - but i think I made it work with flex and flex grow.  We no longer need to define any heights here

PR #2364, opened 2023-12-04
julia calibrate refactor part 1

### LINKED ISSUE #2248: Julia calibration UI refinement - update node/drilldown design
Currently for Julia stratify, most of the work and interactions are on the operation-node, in our updated design most of the work and interactions will be on the operation-drilldown, the node itself becomes a relatively simple preview.

1. Check in with @Jami159 and @pascaleproulx on what the Julia calibrate node and drilldown should look like. Update the task here with screenshots and/or links to the Figma files

2. Do a first round of implementation of the above, basically the roles of the operation-node and operation-drilldown swaps.

3. Amend lingering issues from prototype
  - The projection chart jumps around during calibration
  - The final parameter values should be fetched from `result.json`, instead of the lastly received intermediate result ✅ #2280

Jami159:
This code strips out the code from `tera-calibrate-node-julia.vue` and puts it in `tera-calibrate-julia.vue`.
This isn't the final version of the workflow node or the drilldown.

Open questions/concerns that need to be fixed in the future/separate PR:
- What should the workflow node show after a calibrate was started from the drilldown and a user exited out of the drilldown?
- What should happen while a calibrate is running but a user has exited out of the drilldown?
  - to further clarify on this point, because the the drilldown component gets unmounted when it's closed, we stop querying the message queue, so in that time the simulation could finish running but we may not have all the intermediate results? (TODO: still need to verify the exact details of what happens in this scenario)

Resolves #2248


packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -226,6 +285,177 @@ onMounted(() => {
...
+		type: CalibrationOperationJulia.outputs[0].type,
+		label: `${port.label} - Output ${runList.value.length}`,
+		value: runIdList
+	});
+
+	// show the latest run in the dropdown
+	selectedRun.value = runList.value[runList.value.length - 1];
+};
+
+// Tom TODO: Make this generic... its copy paste from node.
```
Tom-Szendrey, line 451: Outdated comment: this is no longer copy paste code as you just removed the copy

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -246,33 +475,35 @@ function addMapping() {
...
 // Set up csv + dropdown names
-// Note: Same as calibrate-node
 watch(
 	() => datasetId.value,
 	async () => {
 		const { filename, csv } = await setupDatasetInput(datasetId.value);
 		currentDatasetFileName.value = filename;
 		csvAsset.value = csv;
-		datasetColumnNames.value = csv?.headers;
+		datasetColumnNames.value = csv?.headers?.map((col) => col.trim());
```
Tom-Szendrey, line 499: If we want to trim we will want to trim for ciemss too
Can change this in the setupDatasetInput

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -205,17 +250,31 @@ const selectedRunId = computed(
...
-const chartConfigurationChange = (index: number, config: ChartConfig) => {
-	const state = _.cloneDeep(props.node.state);
-	state.calibrateConfigs.chartConfigs[index] = config.selectedVariable;
+// refs to keep track of intermediate states and parameters
+const currentIntermediateVals = ref<{ [key: string]: any }>({ timesteps: [], solData: {} });
+const parameterResult = ref<{ [index: string]: any }>();
 
-	emit('update-state', state);
-};
+const showSpinner = ref(false); // TODO: showSpinner isn't actually showing a spinner -> rename this or maybe show a spinner
```
Tom-Szendrey, line 262: This todo looks out of date, show spinner is being used above?

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -246,33 +475,35 @@ function addMapping() {
...
-// Note: Same as calibrate-node
 watch(
 	() => modelConfigId.value,
 	async () => {
 		const { modelConfiguration, modelColumnNameOptions } = await setupModelInput(
 			modelConfigId.value
 		);
 		modelConfig.value = modelConfiguration;
-		modelColumnNames.value = modelColumnNameOptions;
+		modelColumnNames.value = modelColumnNameOptions?.map((state) => state.trim());
```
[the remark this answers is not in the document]
Jami159: thanks for the suggestion! done :)

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
Jami159: i just updated the `headers` property directly so that we don't have to go touch the ciemss code

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
Jami159: good catch

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -226,6 +285,176 @@ onMounted(() => {
 	}
 });
 
+onUnmounted(() => {
+	poller.stop();
```
Tom-Szendrey, line 289: As we have a poller defined in the drill down and the node we are going to get errors when we open and close the drill down now.
By closing the drill down we are unmounting it and therefore running into a failure case.

image

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
```diff
@@ -1,398 +1,60 @@
...
-import {
-	CalibrationOperationJulia,
-	CalibrationOperationStateJulia,
-	CalibrateMap,
-	CalibrateMethodOptions,
-	CalibrateExtraJulia
-} from './calibrate-operation';
+import { CalibrationOperationStateJulia } from './calibrate-operation';
+
+// TODO: FIGURE OUT WHAT TO SHOW IN THIS WORKFLOW NODE WHILE CALIBRATION IS RUNNING FROM THE DRILLDOWN
```
Tom-Szendrey, line 25: Should be safe to remove this. Every single node that can be run will have this same question answered this will not be a unique issue
If it is unique enough to track we should create a github issue rather than leaving this TODO as this todo will be lost to time

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
Jami159: just got rid of it

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-node-julia.vue
Jami159: alrighty removed

packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
Jami159: leaving this as it is for now as discussed on slack. will address in a future PR after figuring out how to handle polling if drilldown is unmounted.

Tom-Szendrey (approved):

PR #2372, opened 2023-12-05
update: model from code drilldown

blanchco:
# Description

* Just updating the model from code node here so that it conforms to the general drilldown layout.  Not final design..

Screenshot 2023-12-05 at 1 18 25 PM


shawnyama (approved):
LGTM

PR #2373, opened 2023-12-05
feat: code to model accepts code id

shawnyama:
# Description

image


shawnyama:
A few naming related questions:
- Are we sticking with "Model from code" or is it now named "Code to model"?
- Why are we calling the code "code asset" in some places in the UI and some code variables as well? Does calling it "code asset" help us understand that there are multiple code files within a code asset?

YohannParis (approved):

blanchco (approved):
LGTM!  I think we are calling it code asset since a code asset can refer to either one or multiple code files.

PR #2376, opened 2023-12-05
docs: update readme, add vscode project files.

kbirk:
This adds the recommended extensions to properly run the project in VSCode, as well as disabling the gradle extensions build server which causes lots of headaches with the debugger.

Fix issue in hmiServerDev.sh pulling the wrong containers depending on the deployment arg.


dgauldie (approved):
LGTM

chris-dickson (approved):

PR #2378, opened 2023-12-05
feat: coupling models

mwdchang:
### Summary
A version-0 of the model-coupling operator, you can plug in 0 or more decapodes models and run Julia code, as long as you have 

```
decapode = <something>
```

Effectively the model(s) you plug into the operator are turned into Julia variables that you can then manipulate in the notebook, but because variable names can collide and have constraints that do not exactly map to the model names (e.g. I can't use `halfar's ice` as variable name because of the space and tick), currently I am just using model_0, model_1, model_2 ... placeholders. This will be addressed in subsequent PR.

image

Todos and next steps
- Remove debugging information and replace preview with equations derived from decapode
- Add heuristic to generate usable names instead of model_0, model_1 ... etc
- Add in reset (once it is ready on beaker)
- Type checking so we don't mix decapodes and petrinets

### Testing
- add "Halfar Ice DC" decapode onto canvas
- add "Glens Law DC" decapode onto the canvas
- add the model coupling operator and connect
- drilldown into the operator
- run the following Julia code

```
ice_dynamics_composition_diagram = @relation () begin
  dynamics(Γ,n)
  stress(Γ,n)
end
ice_dynamics_cospan = oapply(ice_dynamics_composition_diagram,
  [
    Open(model_1, [:Γ,:n]),
    Open(model_0, [:Γ,:n])
  ]
)

decapode = apex(ice_dynamics_cospan)
```

After maybe 20-30 seconds the output will be shown in the preview, and you can then save the result into a new decapode.

### Reference:
- https://algebraicjulia.github.io/Decapodes.jl/stable/overview/#Merging-Multiple-Physics


packages/client/hmi-client/src/services/model.ts
```diff
@@ -44,7 +44,7 @@ export async function getBulkModels(modelIDs: string[]) {
  * @return Array<Model>|null - the list of all models, or null if none returned by API
  */
 export async function getAllModelDescriptions(): Promise<Model[] | null> {
-	const response = await API.get('/models/descriptions?page_size=200');
+	const response = await API.get('/models/descriptions?page_size=500');
```
mwdchang, line 47: I need to up this because we now have over 300 models

packages/client/hmi-client/src/workflow/ops/model-coupling/model-coupling-operation.ts
```diff
@@ -0,0 +1,22 @@
+import { Operation, WorkflowOperationTypes } from '@/types/workflow';
+
+export interface ModelCouplingState {
+	modelIds: string[];
```
Tom-Szendrey, line 4: What is the point of this? Are these different from the inputs?
I do not see this being used anywhere

Tom-Szendrey (approved):
Aside from not understanding the state LGTM
Its a WIP so maybe the state is just a half step ahead

packages/client/hmi-client/src/workflow/ops/model-coupling/model-coupling-operation.ts
mwdchang: Not being used - will clean up.

packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling-node.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+	<main>Model coupling</main>
```
YohannParis, line 2: You are missing the [tera-operator-placeholder](https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/components/operator/tera-operator-placeholder.vue).

packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling.vue
```diff
@@ -0,0 +1,244 @@
...
+				/>
+
+				<template #footer>
+					<Button style="margin-right: auto" label="Run" @click="runCodeModelCoupling" />
+				</template>
+			</tera-drilldown-section>
+		</div>
+		<template #preview>
+			<tera-drilldown-preview>
+				<div>
```
YohannParis, line 26: Why do you have a `<div>` here?

PR #2382, opened 2023-12-06
task: generate model from code asset

blanchco:
# Description

* generate model from code asset in model from code operator

https://github.com/DARPA-ASKEM/terarium/assets/33158416/5491c718-16ad-49c6-a8aa-3d942b2eeb84

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
@@ -199,8 +199,10 @@ onUnmounted(() => {
 });
 
 function buildJupyterContext() {
-	const contextName = 'decapodes_creation';
-	const languageName = 'julia-1.9';
+	const contextName =
+		selectedModelFramework.value === ModelFramework.Decapodes ? 'decapodes_creation' : null;
+	const languageName =
+		selectedProgrammingLanguage.value === ProgrammingLanguage.Julia ? 'julia-1.9' : null;
 
 	return {
 		context: contextName,
--- a/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
@@ -68,36 +68,29 @@
 		</div>
 		<template #preview>
 			<tera-drilldown-preview v-model:output="selectedModelFramework" :options="modelFrameworks">
-				<section v-if="modelValid === false || previewHTML === ''">
-					<div v-if="isProcessing">
-						<i class="pi pi-spin pi-spinner" :style="{ fontSize: '2rem' }"></i>
-					</div>
-				</section>
+				<tera-progress-spinner v-if="isProcessing" :font-size="2" />
 				<section>
-					<template v-if="selectedModelFramework === ModelFramework.Petrinet && model">
-						<tera-model-diagram :model="model" :is-editable="false"></tera-model-diagram>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -183,14 +199,8 @@ onUnmounted(() => {
...
-	const contextName =
-		selectedModelFramework.value === ModelFramework.Decapodes ? 'decapodes_creation' : null;
-	const languageName =
-		selectedProgrammingLanguage.value === ProgrammingLanguage.Julia ? 'julia-1.9' : null;
-	if (contextName === null || languageName === null) {
-		logger.warn("Can't work with current language. Do nothing.");
-		return null;
-	}
+	const contextName = 'decapodes_creation';
+	const languageName = 'julia-1.9';
```
blanchco, line 203: statically settings these for now - We will only run the kernel when the decapodes options is selected anyways

YohannParis: I agree with you, but why not leaving that nice test there?

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -235,22 +256,28 @@ function getModel() {
...
+				model_version: 'v1.0',
+				schema: 'modelreps.io/DecaExpr',
+				schema_name: 'DecaExpr'
+			}
+		};
+
+		modelValid.value = false;
+
+		kernelManager
+			.sendMessage('save_amr_request', messageContent)
```
YohannParis, line 278: Why is the AMR saved via the kernel?

blanchco: saving for petrinet will happen in another PR - This save message is just for the decapodes flow not the petrinet

YohannParis: I agree, still why is it happening within the Kernel?

blanchco: I couldnt tell you why its happening that way with the kernel, maybe something for @mwdchang to answer?  Are you saying you would want get the kernel response and then save it with a REST call?

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: Youre right - I've added the tests for contextName and languageName but removed the warning and the early null return.  We still want the context to be setup when the component is mounted and returning null wouldn't let that happen

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: in any case I feel that change would be better in a followup PR as well

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -144,28 +144,41 @@ enum ModelFramework {
 const isProcessing = ref(false);
 
 const editor = ref<VAceEditorInstance['_editor'] | null>(null);
-const modelName = ref('');
```
blanchco, line 147: removing the modelname for now - I will handle this in a later PR.  But the flow to name a model is a little confusing at the moment

YohannParis (approved):

PR #2386, opened 2023-12-07
Ciemss calibrate initial UI refactor

Jami159:
This code strips out the code from `tera-calibrate-node-ciemss.vue` and puts it in `tera-calibrate-ciemss.vue`.
This isn't the final version of the workflow node or the drilldown.

Open questions/concerns that need to be fixed in the future/separate PR:
- What should the workflow node show after a calibrate was started from the drilldown and a user exited out of the drilldown?
- What should happen while a calibrate is running but a user has exited out of the drilldown?
  - to further clarify on this point, because the the drilldown component gets unmounted when it's closed, we stop querying the message queue, so in that time the simulation could finish running but we may not have all the intermediate results? (TODO: still need to verify the exact details of what happens in this scenario)

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
@@ -1,85 +1,97 @@
 <template>
-	<!--Probably rename tera-asset to something even more abstract-->
 	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
-		<section>
-			<tera-asset stretch-content>
-				<template #edit-buttons>
-					<SelectButton
-						:model-value="view"
-						@change="if ($event.value) view = $event.value;"
-						:options="viewOptions"
-						option-value="value"
-					>
-						<template #option="{ option }">
-							<i :class="`${option.icon} p-button-icon-left`" />
-							<span class="p-button-label">{{ option.value }}</span>
-						</template>
-					</SelectButton>
-				</template>
-				<Accordion
-					v-if="view === CalibrationView.Input && modelConfig"
-					:multiple="true"
-					:active-index="[0, 1, 2, 3, 4]"
-				>
-					<AccordionTab :header="modelConfig.configuration.name">
-						<tera-model-diagram :model="modelConfig.configuration" :is-editable="false" />
-					</AccordionTab>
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
```diff
@@ -179,11 +177,144 @@ const modelConfigId = computed<string | undefined>(() => props.node.inputs[0]?.v
...
+
+onMounted(() => {
+	const runIds = querySimulationInProgress(props.node);
+	if (runIds.length > 0) {
+		getStatus(runIds[0]);
+	}
+});
+
+onUnmounted(() => {
+	poller.stop();
```
Tom-Szendrey, line 214: as we are defining a poller in the drill down and the node.
And also stopping them on unmount
We are getting errors when opening and closing the node.

image

packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -1,190 +1,40 @@
...
 import { logger } from '@/utils/logger';
-import { subscribe, unsubscribe } from '@/services/ClientEventService';
-import {
-	CalibrationOperationCiemss,
-	CalibrationOperationStateCiemss,
-	CalibrateMap
-} from './calibrate-operation';
+import { CalibrationOperationStateCiemss } from './calibrate-operation';
+
+// TODO: FIGURE OUT WHAT TO SHOW IN THIS WORKFLOW NODE WHILE CALIBRATION IS RUNNING FROM THE DRILLDOWN
```
Tom-Szendrey, line 25: Should be safe to remove this. Every single node that can be run will have this same question answered this will not be a unique issue
If it is unique enough to track we should create a github issue rather than leaving this TODO as this todo will be lost to time

packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
Jami159: leaving this as it is for now as discussed on slack. will address in a future PR after figuring out how to handle polling if drilldown is unmounted.

Tom-Szendrey (approved):

PR #2388, opened 2023-12-07
task: semantic table component

blanchco:
# Description

* Moving the semantic tables and its logic into its own component
* This can allow us to use the semantic tables in other places (i.e. the model-from-code operator)
* readonly option on semantic tables
* empty state handling in model from code operator

https://github.com/DARPA-ASKEM/terarium/assets/33158416/155ce364-af3d-49ca-927c-41863cba0ea1


packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -67,14 +67,32 @@
...
+				/>
+				<tera-operator-placeholder-graphic
+					v-if="!isProcessing && !selectedModel && !previewHTML"
+					:operation-type="node.operationType"
+					style="height: 100%"
+				/>
 				<section>
-					<template v-if="selectedModelFramework === ModelFramework.Petrinet && selectedModel">
+					<template
+						v-if="
```
YohannParis, line 84: You could move the `v-if="selectedModel && !isProcessing"` to the `<section>`
Maybe something like the following. What do you think?
```vue
<section v-if="selectedModel && (!isProcessing || !previewHTML)"?>
```

YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: I'll probably clean up this logic a bit once we know what we should be showing between the Petrinet & Decapodes views...I think the logic here wouldn't show the Decapodes view since that depends on the previewHTML.  I will definitely be keeping an eye on this in later PRs

PR #2394, opened 2023-12-08
task: fix funman traj manip

Tom-Szendrey:
Funman traj was pushing a lot of incorrect timesteps due to ugly string manip to get timesteps
Change how we are getting these trajs
image

image

### DIFF
--- a/packages/client/hmi-client/src/services/models/funman-service.ts
+++ b/packages/client/hmi-client/src/services/models/funman-service.ts
@@ -51,6 +51,7 @@ export const processFunman = (result: any) => {
 	// List of states and parameters
 	const states = result.model.petrinet.model.states.map((s) => s.id);
 	const params = result.model.petrinet.semantics.ode.parameters.map((p) => p.id);
+	const timesteps = result.request.structure_parameters.at(0).schedules.at(0).timepoints;
 
 	// "dataframes"
 	const points = [['id', 'label', 'box_id', ...params]];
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -105,25 +106,19 @@ export const processFunman = (result: any) => {
...
-
 				timesteps.forEach((t) => {
 					const traj: any = {
 						boxId,
 						pointId: point.id,
 						timestep: t
 					};
 					states.forEach((s) => {
-						traj[s] = filteredVals[`${s}_${t}`];
+						if (filteredVals[`${s}_${t}`] !== undefined) {
```
YohannParis, line 116: `!!` will check for not being `undefined` or `null`. Unless you really only want to check if it's not empty.
Otherwise use Lodash `!isEmtpy()`

```suggestion
						if (!!filteredVals[`${s}_${t}`]) {
```
```suggestion
						if (!isEmpty(filteredVals[`${s}_${t}`])) {
```

YohannParis (approved):

packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -105,25 +106,24 @@ export const processFunman = (result: any) => {
...
 					const traj: any = {
 						boxId,
 						pointId: point.id,
 						timestep: t
 					};
 					states.forEach((s) => {
-						traj[s] = filteredVals[`${s}_${t}`];
+						// Only push states that have a timestep key pair
+						if (Object.keys(filteredVals).includes(`${s}_${t}`)) {
+							traj[s] = filteredVals[`${s}_${t}`];
```
mwdchang, line 119: It looks like you want to use `hasOwnProperty` or `hasOwn` instead of this "keys(...).includes(...)" pattern.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

mwdchang (approved):

PR #2397, opened 2023-12-11
feat: handle multiple outputs in drilldown

blanchco:
# Description

* Initial proposal to handle multiple outputs within a drilldown
* This isnt the full dropdown design, but more of a v1 to handle output behaviour

https://github.com/DARPA-ASKEM/terarium/assets/33158416/7e530e6b-83b4-4234-a9ca-d2df8ffc2f17

**EDIT 
This is more or less the final of the outputs design, decided to merge in everything with this ticket


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -88,9 +89,10 @@ ResponseEntity<ExtractionResponse> postCodeToAMR(
 		@RequestParam("code_id") String codeId,
 		@RequestParam(name = "name", required = false) String name,
 		@RequestParam(name = "description", required = false) String description,
-		@RequestParam(name = "dynamics_only", required = false) Boolean dynamicsOnly
+		@RequestParam(name = "dynamics_only", required = false) Boolean dynamicsOnly,
+		@RequestParam(name = "llm_assisted", required = false) Boolean llmAssisted
```
blanchco, line 93: looks like an additional param was added for codeToAmr

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -318,24 +318,6 @@ export function selectOutput(
...
-			isOptional: false
-		} as WorkflowOutput<any>;
-		operator.outputs.push(current);
-	}
-
-	// Update the current state within the outputs
-	current.state = cloneDeep(operator.state);
-	current.operatorStatus = operator.status;
-	current.timestamp = new Date();
-
```
blanchco, line 338: this wasn't needed anymore as I am using the `'append-output-port'` to add a new output

yarn.lock
```diff
@@ -226,13 +226,21 @@ __metadata:
...
+"@babel/parser@npm:^7.23.0":
+  version: 7.23.6
+  resolution: "@babel/parser@npm:7.23.6"
+  bin:
+    parser: ./bin/babel-parser.js
+  checksum: 140801c43731a6c41fd193f5c02bc71fd647a0360ca616b23d2db8be4b9739b9f951a03fc7c2db4f9b9214f4b27c1074db0f18bc3fa653783082d5af7c8860d5
+  languageName: node
+  linkType: hard
+
 "@babel/parser@npm:^7.23.3":
```
blanchco, line 238: i dont know why we needed to add this, but this fixed the failing tests,  just ran 'yarn install' locally and this was updated

YohannParis:
Just to be clear, this is a version 1 of this task.
Do you plan to have multiple PR or do all the work in this one?

blanchco:
> Just to be clear, this is a version 1 of this task. Do you plan to have multiple PR or do all the work in this one?

Multiple PRs, I want to unblock @Jami159 as well so that he can work with his operator's outputs and make sure we are using the same behaviour.  The checkboxes and styling can come in followup PRs

Jami159:
@YohannParis you may have the same question for [my PR](https://github.com/DARPA-ASKEM/terarium/pull/2398), so just letting you know that the answer is the same as what @blanchco just mentioned :)

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -349,16 +349,57 @@ function removeCodeBlock(index: number) {
...
+);
+
+watch(
+	() => props.node.active,
+	async () => {
+		if (props.node.active) {
+			selectedOutputId.value = props.node.active;
+		}
+	}
+);
```
mwdchang, line 397: You have quite a few watchers here, are we confident that they are mutually exclusive and will not interfere with each other?

blanchco: I was a little sad at the watchers, but I couldn't really find a way to get around them here since we need to listen to the prop changes.  In any case I believe that the watchers shouldn't be interfering with one another.

blanchco:
Will need a re-review here...merging everything in this PR, here's a video:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/bb4c6cc1-d857-46cc-889a-f95812aae0ed

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -281,18 +284,28 @@ function appendInputPort(
...
 		value: isArray(port.value) ? port.value : [port.value],
 		isOptional: false,
-		status: WorkflowPortStatus.NOT_CONNECTED
-	});
+		status: WorkflowPortStatus.NOT_CONNECTED,
+		state: port.state,
+		timestamp: new Date()
+	};
+
+	if ('isSelected' in port) outputPort.isSelected = port.isSelected;
```
blanchco, line 304: I only want to add the isSelected key if its explicitly defined in the port, see `tera-operator-outputs` comment for more detail

packages/client/hmi-client/src/workflow/operator/tera-operator-outputs.vue
```diff
@@ -47,12 +47,19 @@ const emit = defineEmits([
...
 		default: () => []
 	}
 });
+
+const selectedOutputs = computed(() =>
+	props.outputs?.filter((output) => {
+		if (!('isSelected' in output)) return true;
+		return output.isSelected;
+	})
+);
```
blanchco, line 62: A way to expose output ports whilst making it backwards compatible with the way we are currently appending output ports from other nodes where the isSelected key isnt present

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
YohannParis: Why do you need to watch for those node information?

packages/client/hmi-client/src/services/workflow.ts
YohannParis: Where do you save an output that was not `Run`?

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -302,6 +315,22 @@ function updateWorkflowNodeState(node: WorkflowNode<any> | null, state: any) {
...
+function updateOutputPort(node: WorkflowNode<any> | null, workflowOutput: WorkflowOutput<any>) {
+	if (!node) return;
+	let outputPort = node.outputs.find((port) => port.id === workflowOutput.id);
+	if (!outputPort) return;
+	outputPort = {
+		...workflowOutput
+	};
+	workflowDirty = true;
+}
+
```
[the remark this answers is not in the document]
YohannParis: This would be better in the `services/workflow.ts`

YohannParis (commented):
This could be addressed in another PR.
But the `Run` button should be disabled if the state has not changed since last click on it.

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -302,6 +315,22 @@ function updateWorkflowNodeState(node: WorkflowNode<any> | null, state: any) {
 	workflowDirty = true;
 }
 
+function selectOutput(node: WorkflowNode<any> | null, selectedOutputId: string) {
+	if (!node) return;
+	workflowService.selectOutput(node, selectedOutputId);
+	workflowDirty = true;
+}
```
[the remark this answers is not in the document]
blanchco: I can add  that statement to the service.  I figure we would want to save the node state everytime we change the output which is would I made the workflow dirty.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: I was able to make the selectedOutputId a computed property instead of a watcher ✅ .  I believe I need to watch the state however.  Since the drilldown itself has the ref of `clonedState`, we need to listen whenever the output is changed which in turn changes the state and copies to `clonedState`.

packages/client/hmi-client/src/workflow/tera-workflow.vue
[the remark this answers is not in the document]
blanchco: nevermind - that statement is needed here since the node can be null

packages/client/hmi-client/src/workflow/tera-workflow.vue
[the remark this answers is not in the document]
blanchco: done

packages/client/hmi-client/src/services/workflow.ts
blanchco: What would be an example of this?  For something like save model we are leveraging the 'append-output-port' as well

blanchco:
@YohannParis added ability to name model with modal

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -116,10 +109,23 @@
...
+		</template>
+		<form @submit.prevent>
+			<label for="new-model">Enter a unique name for your model</label>
+			<InputText id="new-model" type="text" v-model="newModelName" placeholder="New model" />
+		</form>
+		<template #footer>
+			<Button @click="createModel">Create model</Button>
+			<Button class="p-button-secondary" @click="isNewModelModalVisible = false">Cancel</Button>
+		</template>
+	</tera-modal>
```
blanchco, line 124: we can't reuse the `tera-model-modal` here as it strictly creates an empty new model.  We have a unique case here since we already have the model create, we just need to rename it, so custom functions are need.

packages/client/hmi-client/src/services/model.ts
```diff
@@ -94,3 +95,22 @@ export function isModelEmpty(model: Model) {
...
+		logger.info('Model name cannot be empty - please enter a different name');
+		return false;
+	}
+	if (existingModelNames.includes(name.trim())) {
+		logger.info('Duplicate model name - please enter a different name');
+		return false;
+	}
+
+	return true;
+}
```
blanchco, line 116: made this a function to be used anywhere to validate a model name on a project

packages/client/hmi-client/src/services/workflow.ts
YohannParis: When I made changes to the state, but have not clicked `Run`, then select a different output to visualize its state. I would expect to be able to get back to that last un-`run` state. So it needs to be saved even if it wasn't ran, therefore does not have an output.

YohannParis: This will be solved in another PR.

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-modal.vue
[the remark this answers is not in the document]
blanchco: https://github.com/DARPA-ASKEM/terarium/assets/33158416/43d1f58a-2a94-4bf4-8f1d-6c84f72ff840

PR #2399, opened 2023-12-11
Upgrading tooling

YohannParis:
# Description

* Updating Yarn to v4.0.2, which removing the workspace plugin, now part of main codebase.
* Updating Vite to v5.0.7
* Updating Node LTS to 20.10.0
* Updated Prettier, and run the formatting on all the existing files
* Update Dockerfile to remove GitHub linting comments
* Other tooling


mwdchang:
There are quite many formatting changes here, are those automatic or did you do them manually? 

Ideally I think it would be good if we don't massively change coding-styles en mass at the current time as it makes it harder to track history/changes.

YohannParis:
> There are quite many formatting changes here, are those automatic or did you do them manually?
> 
> Ideally I think it would be good if we don't massively change coding-styles en mass at the current time as it makes it harder to track history/changes.

Those are done by Prettier upgraded version. I solved all the issues resolving from the upgrading of the tooling. As doing in multiple round would be annoying.

containers/permissions/Dockerfile
YohannParis, line 1: @bigglesandginger could take a quick look at the changes I made to remove GitHub linting issues.

mwdchang:
Seems generally okay to me - prettier has caused some headaches in the past because it fights with the linter, so we may want to keep an eye out and be extra careful in the next little while. In the past I have seen it eitehr:
- makes thing on into infinite loops (prettier changes X to Y, linter changes Y back to X)
- produce non-optimal code for some reason

mwdchang (approved):

PR #2405, opened 2023-12-12
feat: add asset from new popup menu

shawnyama:
# Description

At the moment I am showing a logger message so the user knows that the asset has been added to the project. 

There isn't an elegant way of showing the project names within the asset card so I'll leave it for another time. I am iffy about fetching all the assets from every project and then comparing them to the search results. The `getAll` for projects doesn't include `assets`. In order to fetch assets each project needs to be called individually. I think that this comparison should be solved in the backend when the search results are fetched.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/9733f7c3-f90a-4a05-99d2-bd0084e7b413


packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -99,9 +101,62 @@ const props = defineProps({
...
+						// FIXME: handle cases where assets is already added to the project
+						const datasetId = selectedAsset.value.id;
+						// then, link and store in the project assets
+						const assetType = AssetType.Datasets;
+						if (datasetId) {
+							response = await useProjects().addAsset(assetType, datasetId, project.id);
+							assetName = selectedAsset.value.name;
+						}
+					}
+					console.log(selectedAsset.value);
```
YohannParis, line 141:
```suggestion
```

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -99,9 +101,62 @@ const props = defineProps({
 	}
 });
 
-const previewedAsset = ref<ResultType | null>(null);
+const selectedAsset = ref();
 
-const emit = defineEmits(['toggle-data-item-selected']);
+const projectOptions = computed(() => [
+	{
+		label: 'Add this to...',
```
YohannParis, line 108:
```suggestion
		label: 'Select Project to add asset',
```

It will be nicer if `asset` was the type of the asset, but that's not vital.

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -99,9 +101,62 @@ const props = defineProps({
...
+						// then, link and store in the project assets
+						const assetType = AssetType.Datasets;
+						if (datasetId) {
+							response = await useProjects().addAsset(assetType, datasetId, project.id);
+							assetName = selectedAsset.value.name;
+						}
+					}
+					console.log(selectedAsset.value);
+
+					if (response) logger.info(`Added ${assetName} to ${project.name}`);
```
YohannParis, line 143: Add an error handling with a `logger.warn` explaining that the asset was not added to the project.

packages/client/hmi-client/src/page/data-explorer/components/tera-search-item.vue
```diff
@@ -6,50 +6,34 @@
 		:highlight="searchTerm"
 		@click="emit('toggle-asset-preview')"
 	>
-		<Button
-			:icon="`pi ${statusIcon}`"
-			class="p-button-icon-only p-button-text p-button-rounded"
-			@click.stop="emit('toggle-selected-asset')"
-		/>
+		<Button @click.stop="toggle" icon="pi pi-plus" text rounded />
```
YohannParis, line 9: a spinner after clicking until the user gets an answer would be neat.

packages/client/hmi-client/src/services/project.ts
```diff
@@ -154,16 +154,16 @@ async function getPublicationAssets(projectId: string): Promise<ExternalPublicat
  * @assetId string - represents the id of the asset to be added. This will be the internal id of some asset stored in one of the data service collections
  * @return any|null - some result if success, or null if none returned by API
  */
-async function addAsset(projectId: string, assetsType: string, assetId: string) {
+async function addAsset(projectId: string, assetType: string, assetId: string) {
```
YohannParis, line 157: This function could be a good place to add the logger informing the user that the asset has been added to the project or not. Especially that the `assetType` is a string and therefore easy to display.

YohannParis (approved):

packages/client/hmi-client/src/services/project.ts
shawnyama: I didn't do that because I'd have to pass the `assetName` and `projectName` which are different

PR #2411, opened 2023-12-13
Fix stratify mira reset

Jami159:
I'm calling `reset_request` followed by `stratify_request`/`execute_request`. These are asynchronous. Meaning that it's possible `stratify_request`/`execute_request` finishes before `reset_request` which could lead to issues. This PR makes sure `stratify_request`/`execute_request` are only called once `reset_request` finishes.


Tom-Szendrey (approved):
ty for tagging me, i wasnt aware

PR #2413, opened 2023-12-13
feat: prepare model-template canvas, sidebar and cards

shawnyama:
# Description

- Initialized infinite canvas and template card for model templating
- Testing can be done here: http://localhost:8080/model-template-test 
- Not sure what method we want to use for adding the "Latin Modern Math" font so I set it as serif for now
- Ports change color when hovered over and ellipsis button appears when card is hovered over (Both of these rules work if the card is in the canvas)
- Next PR I'll handle dragging cards into the canvas

https://github.com/DARPA-ASKEM/terarium/assets/28237258/a05218bd-20ce-4d98-971a-ad0041744eb5


YohannParis:
Can you add all the design elements in this PR, like hover/click design, the drag buttons, the three dots buttons on the bottom right. The list of blocks in the right side, etc.

packages/client/hmi-client/src/components/model-template/tera-model-template-card.vue
```diff
@@ -0,0 +1,116 @@
...
+	& > .pi {
+		font-size: 0.75rem;
+		color: var(--text-color-subdued);
+	}
+}
+</style>
+
+<style>
+/* When a card is placed in the data layer of the infinite canvas */
+.data-layer .card-container {
```
mwdchang, line 106: Why do we need to spell out the explicit coupling here between infinite-canvas and the template-card? Are we expecting to use the card somewhere else where you would want to have a different style and/or behaviour?

mwdchang:
I think if you are going to delay building the foundation layers, we should at least start thinking about making `model-template-test.vue` data driven, e.g. have something that represents a collection of model-templates in the next iteration as oppose to pure templating html.

packages/client/hmi-client/src/components/model-template/tera-model-template-card.vue
shawnyama: Since the cards are also shown in the sidebar, I thought those ones shouldn't look like they have selectable ports

mwdchang: See thanks for the explanation, though we might need to revisit this when things become more complicated.

mwdchang (approved):

YohannParis:
Did you fixed the two issue I brought?

PR #2414, opened 2023-12-13
changed project id to UUID

dgauldie:
# Description

Resolves #(issue)


kbirk (approved):

YohannParis:
that's going to make the URLs massive... We can live with that, but not great aesthetically

PR #2422, opened 2023-12-14
task visualize code blocks

blanchco:
# Description

* Showing all input code blocks
* Saving only added code blocks on state
* cannot delete a input code block
* visualize code snippets from code blocks

https://github.com/DARPA-ASKEM/terarium/assets/33158416/7b1d4cc8-4270-4525-9eec-083c0cfbe1b1

Note:  Running with multiple dynamics will need to be done in another PR, this is more just displaying multiple dynamics both inputted ones and ones that we added


packages/client/hmi-client/src/components/widgets/tera-expandable-panel.vue
blanchco, line 1: I've seen this panel in more than one operator, i figure it would be good to make it it's own component:
Screenshot 2023-12-15 at 4 08 39 PM

The default slot in the panel can be defined by the user when they use the component

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -334,7 +334,5 @@ export function selectOutput(
 export function updateOutputPort(node: WorkflowNode<any>, updatedOutputPort: WorkflowOutput<any>) {
 	let outputPort = node.outputs.find((port) => port.id === updatedOutputPort.id);
 	if (!outputPort) return;
-	outputPort = {
-		...updatedOutputPort
-	};
+	outputPort = Object.assign(outputPort, updatedOutputPort);
```
blanchco, line 337: This wasn't being done properly before 😬

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -377,39 +363,76 @@ async function handleDecapodesPreview(data: any) {
...
+				codeContent: codeSnippet ?? '',
+				name: 'Code Block',
+				includeInProcess: true,
+				type: CodeBlockType.INPUT
+			};
+		});
+
+		return inputCodeBlocksForFile;
+	});
+
```
blanchco, line 414: Codeblocks are a pain to work with I want to convert something like this:
```
{
  "files": {
    "chime-1.1.5/.dockerignore": {
      "dynamics": {
        "name": "test",
        "description": "",
        "block": [
          "L1-L9"
        ]
      }
    },
    "chime-1.1.5/.bldr.jl": {
      "dynamics": {
        "name": "test",
        "description": "",
        "block": [
          "L28-L33",
          "L1-L9"
        ]
      }
    }
  }
}
```

 into a more workable format, the format being the CodeBlock interface

YohannParis: Feel free to update the back end enpoints to send data better formatted for your needs.
That is why it is there.

YohannParis (approved):
Do not forget to add the number of code blocs in the output ports on the Code Asset.

blanchco:
> Do not forget to add the number of code blocs in the output ports on the Code Asset.

Will do as a part of this task https://github.com/DARPA-ASKEM/terarium/issues/2431

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: will do, might do i in another PR though as I finalize everything

PR #2424, opened 2023-12-14
Update simulation operators

Jami159:
This PR refactors both the julia and ciemss simulation operators to:
1. move functionality from workflow node to drilldown
2. update drilldown to match new wizard/preview layout
3. replace existing output selection implementation with new output selection implementation

TODO: Jami - add demo

Resolves: #2425


YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -293,17 +293,16 @@ function appendOutputPort(
 	if (!node) return;
 
 	const uuid = uuidv4();
-	const timestamp = new Date();
 
 	const outputPort: WorkflowOutput<any> = {
 		id: uuid,
 		type: port.type,
-		label: `${port.label} ${timestamp.toLocaleTimeString()}`,
+		label: port.label,
```
mwdchang, line 300: What is the label logic suppose to be?

If we want to have add a timestamp we would want it to be centralized - the way you have it is that each operator would add their own timestamp which seems to be quite tediously if at the end of the day they are all suppose to add timestamps to their labels.

But in additional, using `toLocaleTimeString` is not very good here, because the label gets locked into a string, so if a modeller were to look at the workflow in a different timezone things can get screwy.

@YohannParis what is expected here in terms of the port design?

Jami159: Yeah I wasn't entirely sure if it made sense to always show the timestamp alongside the output port label (I saw it in one design I think, but not in general). So I removed that from here. Instead if we do want the timestamp added for a specific operator or something, it can be added directly for that operator.

But I don't have any strong opinions about this. I'm cool with implementing whatever designs require.

PR #2426, opened 2023-12-15
Feat: Auto mapping for calibration V0

Tom-Szendrey:
- I have moved the calibration map definition to the service file as it is shared between calibration types.

- I changed modelOptions and datasetColumnOptions to be objects instead of string[]. Theres no need to restrict their information as we do not want to refetch this information elsewhere

- adding a new service function to do a whole ton of string comparisons for the auto mapping button

image

To test I used the following:
Model: 739f6a1b-4e56-40cf-95d5-194348d3f0b3
Dataset: 528787a0-4607-4d8f-a5e3-2829edf44a94
workflow: /5/workflows/811fca52-31ad-4fee-bd2c-5db972af5c14

I updated a dataset to include a grounding for one column that is just marked "test" and added the same key to the model


packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -108,3 +115,28 @@ export const renderLossGraph = (
 	}
 	yAxisGroup.attr('transform', `translate(${marginLeft}, 0)`).call(yAxis);
 };
+
+export const autoCalibrationMapping = async (modelOptions: State[], datasetOptions: any[]) => {
+	const result = [] as CalibrateMap[];
+	modelOptions.forEach((modelOption) => {
```
Tom-Szendrey, line 121: I know everytime i write a for loop people ask me to use forEach instead but I cannot use `break` in this. which means when I find a correct result it keeps chugging away 
Will yall complain if i write a for loop for this instead of a forEach?

mwdchang: There's nothing wrong with a `for (....)` construct:
- If it makes sense to use it, use it.
- If it makes the code easier to comprehend, use it.

mwdchang (approved):

PR #2429, opened 2023-12-15
feat: draggable template card

shawnyama:
# Description
Test here: http://localhost:8080/model-template-test

- Wrote `tera-canvas-item.vue` to wrap around `tera-model-template-card` and `tera-operator` as they both share similar drag/drop behaviour.
- resize and interactionStatus (bitmasking) are still exclusively done in `tera-operator`. I don't think `tera-model-template-card` need those - we can move them into `tera-canvas-item` later if needed.
- Side note: got rid of `h7` rule as that's not even a valid tag

https://github.com/DARPA-ASKEM/terarium/assets/28237258/e293fcf2-b9f7-48d5-af7b-c6b7cc3ba331


shawnyama:
I don't know why specifically the document node expands as I move it left. It's as if the `left` css attribute is effecting things (from tera-canvas-item). My main changes around this area are moving the `minWidth` style into `tera-canvas-item` and removing the `main {width: 15rem}` from `tera-operator`.


https://github.com/DARPA-ASKEM/terarium/assets/28237258/cdcbc1b3-c165-43ce-87a5-4815fd6d2db0

packages/client/hmi-client/src/components/model-template/tera-model-template.vue
```diff
@@ -1,16 +1,25 @@
...
+					<header>Model framework</header>
 					<h5>{{ model.header.schema_name }}<i class="pi pi-info-circle"></i></h5>
 				</section>
 				<section>
-					<h7>Model templates</h7>
+					<header>Model templates</header>
 					<ul>
 						<li v-for="(_, index) in 5" :key="index">
-							<tera-model-template-card />
+							<tera-model-template-card :card="newCard" draggable="true" />
```
mwdchang, line 22: Note: `#foreground` is meant as an overlay for the infinite canvas, you could put things like messages/status etc etc. I am not entirely sure if it is the best place to put a side panel. Having said that I don't really have any objections either.

shawnyama: in the workflows the top toolbar is placed in `#foreground` so I saw this as a toolbar on the side.

mwdchang:
> I don't know why specifically the document node expands as I move it left. It's as if the `left` css attribute is effecting things (from tera-canvas-item). My main changes around this area are moving the `minWidth` style into `tera-canvas-item` and removing the `main {width: 15rem}` from `tera-operator`. Note that this happens with a document with a long title though I still can't narrow down where its coming from.
> 
>  Screen.Recording.2024-01-02.at.1.53.28.PM.mov

@shawnyama is this still an issue? I can't reproduce.

YohannParis (approved):

PR #2432, opened 2023-12-18
Design paper-cuts

YohannParis:
# Description

* Work with @jamiewaese-uncharted to fix small paper-cuts
* Move the search bar from the header to the data-explorer

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/components/tera-filter-bar.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/components/tera-filter-bar.vue
@@ -1,5 +1,5 @@
 <template>
-	<div class="flex align-items-center pt-2 pb-2 pl-8 pr-8">
+	<div class="flex align-items-center">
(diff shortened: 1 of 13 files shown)


packages/client/hmi-client/src/page/data-explorer/components/tera-filter-bar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="flex align-items-center pt-2 pb-2 pl-8 pr-8">
+	<div class="flex align-items-center">
```
blanchco, line 2: We don't like these class names right? 😈

blanchco (approved):
Looks good 💃, checked out the branch as well and is working fine

packages/client/hmi-client/src/page/data-explorer/components/tera-filter-bar.vue
YohannParis: indeed

PR #2433, opened 2023-12-18
Fix `updateXxx` methods to check for existence and return 404s

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectService.java
```diff
@@ -1,37 +1,50 @@
...
-	public void delete(final UUID id){
-		projectRepository.deleteById(id);
+	public Optional<Project> updateProject(final Project project) {
+		if (!projectRepository.existsById(project.getId())) {
+			return Optional.empty();
+		}
+		return Optional.of(projectRepository.save(project));
+	}
+
+	public void deleteProject(Project project) {
```
dvince2, line 46: This is going to be some conflicts :D

dvince2 (approved):

PR #2434, opened 2023-12-18
Inital proof of concept

bigglesandginger:
Stands up the `hmi-server` listening on two ports (by default `3000` (secure) and `3001` (insecure))

Adds `InternalController` that serves requests to URL that start with `/internal/`

Adds functionality to allow only calls on `3001` to route to `/internal/`

Calls to `/internal/` from `3000` will fail (with 403)
Calls to not `/internal/` from `3001` will fail (with 403)

To test: `curl -v localhost:3001/internal/projects/<project-id> | jq`

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/internal/InternalController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/internal/InternalController.java
@@ -1,0 +1,47 @@
+package software.uncharted.terarium.hmiserver.controller.internal;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import io.swagger.v3.oas.annotations.tags.Tags;
+import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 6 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/internal/InternalController.java
```diff
@@ -0,0 +1,47 @@
...
+	final ProjectService projectService;
+
+	@Operation(summary = "Gets a project by ID")
+	@ApiResponses(value = {
+		@ApiResponse(responseCode = "200", description = "Project found.", content = {
+			@Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = Project.class)) }),
+		@ApiResponse(responseCode = "500", description = "Error finding project", content = @Content),
+		@ApiResponse(responseCode = "404", description = "Project not found", content = @Content) })
+	@GetMapping("/projects/{id}")
+	public ResponseEntity<Project> getProject(
```
dvince2, line 39: For the final version, will we need to duplicate all of the endpoints we want our internal services to hit like this?

bigglesandginger: Correct, those that are required

bigglesandginger: I am not certain we need all that Swagger stuff... maybe it is vital for the internal calls <shrug>

PR #2435, opened 2023-12-18
Mock rebacservice. Add project controller tests.

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
```diff
@@ -1,50 +1,59 @@
...
-
 @Service
 @Slf4j
-@RequiredArgsConstructor
 public class ReBACService {
 
 	Keycloak keycloak;
 
-	private final Config config;
+	@Autowired
```
dvince2, line 53: Was there an issue using `@RequiredArgsConstructor` over `@Autowired`?  I believe that the accepted "best practice" is to use the former over the later

kbirk: There was but because the Mock inherits that it was complaining. Let me take another swing

dvince2: Fair!  I'll ✅ and then you can merge it either like this or if you find a way with the requiredArgs.

dvince2 (approved):

PR #2438, opened 2023-12-18
fix: model output port not appearing on drag-n-drop interaction

mwdchang:
### Summary
This fix the situations where the output-port is not created properly when you drag-and-drop a model from the resource pane onto the canvas.

This does not address existing models that do not have outputs, they will need to be recreated.

### Testing
Drag any model onto the canvas, should see one output port representing the propagation of said model.


mwdchang:
I think I must have mangled up the commit somewhere - the fix was pushed into main and merged already.

```
https://github.com/DARPA-ASKEM/terarium/commit/296802dead7f65930dfa2546edf32da4ad6265cd
```

And then this got merged with main and became a blank PR ...

mwdchang:
Already fixed.

PR #2439, opened 2023-12-18
2242 soft delete

dvince2:
Now setting the deleted timestamp in services, and the generic get all calls only return active assets


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ModelConfigurationService.java
```diff
@@ -31,6 +31,7 @@ public List<ModelConfiguration> getModelConfigurations(Integer page, Integer pag
 		final SearchRequest req = new SearchRequest.Builder()
 				.index(elasticConfig.getModelConfigurationIndex())
 				.size(pageSize)
+				.query(q -> q.bool(b -> b.mustNot(mn-> mn.exists(e->e.field("deleted_on")))))
 				.sort(new SortOptions.Builder()
 						.field(new FieldSort.Builder().field("timestamp").order(SortOrder.Asc).build()).build())
 				.build();
```
mwdchang, line 37: Out of curiosity, how would we different between a query-context versus a filter-context going through the ElasticSearch's Java API?

kbirk (approved):

PR #2443, opened 2023-12-18
fix: relinking edges

mwdchang:
### Summary
Edges can become detached from their respective port through either node-size changing, addition/deletion of ports, etc etc. This PR propose a logic to reattach the edges back to their designated ports.

To avoid unnecessary flickering, a `threshold` is introduced:
- if distance from the port to the edge's end point < threshold, do nothing
- else reassign endpoint's X/Y coordinate

### Testing
Probably the best tests are under "Simulate Integration test Project" under public projects folder. It has a lot of workflows that had been generated programmatically with quirky edges. When you open these workflows now the edges should look normal.


packages/client/hmi-client/src/components/operator/tera-operator-inputs.vue
```diff
@@ -2,6 +2,7 @@
 	<ul>
 		<li
 			v-for="(input, index) in inputs"
+			:id="input.id"
 			:key="index"
```
YohannParis, line 6: Why not use the `input.id` here and get read of the `index`?

packages/client/hmi-client/src/components/operator/tera-operator.vue
```diff
@@ -131,20 +125,42 @@ function openInNewWindow() {
...
+function resizeHandler() {
+	emit('resize', props.node);
+}
+
+onMounted(() => {
+	if (!operator.value) return;
+
+	operator.value.addEventListener('mousedown', startDrag);
+	resizeObserver = new ResizeObserver(resizeHandler);
+	resizeObserver.observe(operator.value);
```
YohannParis, line 150: Nice !

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -583,6 +584,72 @@ function onPortMouseleave() {
...
+const dist2 = (a: Position, b: Position) => (a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y);
+const threshold2 = 5.0 * 5.0;
+
+/**
+ * Relink edges that have become detatched
+ *
+ * [output-port](edge source => edge target)[input-port]
+ *
+ * FIXME: not efficient, need cache/map for larger workflows
+ * */
```
YohannParis, line 601:
```suggestion
 **/
```

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
mwdchang: I think this is prettier being annoying and decide to format comments.

packages/client/hmi-client/src/components/operator/tera-operator-inputs.vue
mwdchang: Sure, will do.

PR #2455, opened 2023-12-25
Model config operator drilldown - first pass

### LINKED ISSUE #2483: Model config drilldown implementation - first pass

Jami159:
This PR updates the model config drilldown with the simple functionality of updating the name, description, initial expressions, and parameter values of a model to create a new model configuration.

Things not implemented in this PR but should be added later:
- the ability to edit parameter distributions
- update UI styling to match the designs more closely
- figure out what to do about existing configurations
  - they show up in the dropdown but don't have state because state is only created for new configurations
  - they aren't selectable/unselectable right now to show/hide them from the workflow node

Demo:

https://github.com/DARPA-ASKEM/terarium/assets/10509899/cbbce55a-0bf6-48f1-983d-3115d30d1e65

Resolves #2483


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -1,18 +1,284 @@
...
+				@update:output="onUpdateOutput"
+				@update:selection="onUpdateSelection"
+				is-selectable
+			>
+				<div v-if="configCache[selectedConfigId]">
+					<TeraModelSemanticTables
+						:model="configCache[selectedConfigId].configuration"
+						:is-editable="false"
+					/>
+				</div>
```
shawnyama, line 51: No need for div also component should be kebab case.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -1,18 +1,284 @@
 <template>
 	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
-		<section>TODO</section>
+		<section :tabName="ConfigTabs.Wizard">
+			<tera-drilldown-section>
+				<div>
+					<Steps :model="formSteps" :readonly="false" @update:active-step="activeIndex = $event" />
+				</div>
```
shawnyama, line 7: No need for div

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -0,0 +1,217 @@
+<template>
+	<div class="value-editor">
+		<div>
```
shawnyama, line 3: For divs that hold larger sections/containers I'd suggest to use `section`. eg. the parent div and the children just after that.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -0,0 +1,217 @@
...
+			<div>
+				<h4>Initials</h4>
+				<Accordion v-if="stratifiedModelType && !_.isEmpty(initials)" :multiple="true">
+					<AccordionTab v-for="[init, vals] in initials" :key="init">
+						<template #header>
+							{{ init }}
+						</template>
+						<div v-for="val in vals" :key="val">
+							<Button
```
shawnyama, line 12: Nitpick: using a ul > li > Button here would be more semantic

shawnyama (approved):

packages/client/hmi-client/src/workflow/ops/model-config/model-config-operation.ts
```diff
@@ -1,9 +1,14 @@
...
+import { Initial, ModelParameter } from '@/types/Types';
 
 export const name = 'ModelConfigOperation';
 
 export interface ModelConfigOperationState {
 	modelId: string | null;
+	configName: string;
+	configDescription: string;
+	configInitials?: Initial[];
+	configParams?: ModelParameter[];
```
shawnyama, line 11: Remove the `config` prefix form these names, they're already wrapped around `ModelConfigOperationState`. Reflect this in the ref names in `tera-model-config.vue`

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -1,18 +1,281 @@
...
+
+const updateConfigParam = (param) => {
+	const state = _.cloneDeep(props.node.state);
+	// find param with the same id and update it
+	state.configParams = state.configParams?.map((p) => {
+		if (p.id === param.id) {
+			return param;
+		}
+		return p;
+	});
```
mwdchang, line 167: This looks a bit suspicious, are you updating from yourself - what is the intention here?

YohannParis:
I wonder if we should "disable" the preview section, as it can be confusing why editing the right does not change the left panel. What do you think @ngraham76?

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -1,18 +1,281 @@
...
+				isSelected: false,
+				state: {
+					modelId: modelConfig.modelId,
+					configName: modelConfig.name,
+					configDescription: modelConfig.description,
+					configInitials: modelConfig.configuration.semantics?.ode.initials,
+					configParams: modelConfig.configuration.semantics?.ode.parameters
+				}
+			});
+		}, 1000);
```
mwdchang, line 215: This setTimeout mechanism is not really necessary, since you already have the values you want to propagate forward. 

Also it would be better named `createModelConfiguration` to better reflect what you are actually doing. Note there is another savings part; in the semantics-table that doesn't seem to work that we should sync up on. 

Nitpick: I might suggest inverting the logic, instead of
```
if (xyz) {
   ...do stuff...
}
```
This makes the code a little bit less steep in terms of indentation
```
if (!xyz) return
...do stuff...
```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -0,0 +1,217 @@
...
+watch(
+	[() => props.initials, () => props.parameters],
+	() => {
+		if (fieldType.value && selectedField.value) {
+			getClickedField(fieldType.value, selectedField.value);
+		}
+	},
+	{ immediate: true }
+);
+</script>
```
mwdchang, line 184: Can we do a bit or reorganization for the `<script>` section to make things more idiomatic? 

We typically have the pattern of:
- imports
- props/emits declarations
- refs/computed declarations
- functions
- watchers
- lifecycle-hooks

You have things a bit more intermingled.

mwdchang: Sorry, posted on wrong file

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -1,18 +1,281 @@
...
+};
+
+watch(
+	() => selectedConfigId.value,
+	() => {
+		lazyLoadModelConfig(selectedConfigId.value);
+	},
+	{ immediate: true }
+);
 </script>
```
mwdchang, line 273: Can we do a bit or reorganization for the <script> section to make things more idiomatic?

We typically have the pattern of:
- imports
- props/emits declarations
- refs/computed declarations
- functions
- watchers
- lifecycle-hooks
- 
You have things a bit more intermingled.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
mwdchang: Okay I see, the naming of things is a bit confusing, too many params. 

I think it would be clearer if we can do this:
```
const idx = state.configParams.findIndex(d => d.id === param.id);
if (idx) {
  state.configParams[idx] = param
}

```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -0,0 +1,221 @@
...
+	getUnstratifiedParameters,
+	getUnstratifiedInitials
+} from '@/model-representation/petrinet/mira-petri';
+import { getStratificationType } from '@/model-representation/petrinet/petrinet-service';
+
+const props = defineProps<{
+	model: Model;
+	initials?: Initial[];
+	parameters?: ModelParameter[];
+}>();
```
mwdchang, line 132: Would it be easier to make **initials** and **parameters** required? As in, even if a model has no parameters it would be an empty array instead of null.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -0,0 +1,221 @@
...
+							class="asset-button"
+							plain
+							text
+							size="large"
+							@click="getClickedField('parameter', vals[0])"
+						/>
+					</li>
+				</ul>
+			</div>
+		</section>
```
mwdchang, line 72: Okay for now, but this `<section>` is a bit wonky to use if when you have a long initials/parameters list, you have to constantly scroll back and forth. Let's flag this as a FIXME.

Jami159: fixme added

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
Jami159: done

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: reorganized as suggested

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: updated w ur suggestion

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: moved the `emit(append-output-port...` out of the `setTimeout` but keeping the `setTimeout` here for caching for now.

mwdchang:
> I wonder if we should "disable" the preview section, as it can be confusing why editing the right does not change the left panel. What do you think @ngraham76?

@YohannParis @ngraham76 
We flagged a short listing of FIXMEs and next-steps, some of these items revolve around design ambiguities such as this. Let's take the conversation  there instead as we may have better understanding when viewed in the context of other bugs/annoyances. Please see this task item: https://github.com/DARPA-ASKEM/terarium/issues/2485

mwdchang (approved):
Great first round, thanks!

PR #2464, opened 2024-01-02
chore: general clean up, show available names in comments

mwdchang:
### Summary
Instead of using model_0, model_1, ... model_n as variables, derive more meaningful names from the model header and output information as code comment into the notebook tab.

Also did a few general cleanup in the space.

image

### Testing
No functional change, below code should still work - you may need to change the names to match.

```
ice_dynamics_composition_diagram = @relation () begin
  dynamics(Γ,n)
  stress(Γ,n)
end
ice_dynamics_cospan = oapply(ice_dynamics_composition_diagram,
  [
    Open(halfar, [:Γ,:n]),
    Open(glen, [:Γ,:n])
  ]
)

decapode = apex(ice_dynamics_cospan)

```

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling.vue
@@ -6,7 +6,6 @@
 		<div :tabName="ModelCouplingTabgs.Notebook">
 			<tera-drilldown-section>
 				<h4>Code Editor - Julia</h4>
-				{{ modelMap }}
 				<v-ace-editor
 					v-model:value="codeText"
 					@init="initializeEditor"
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling.vue
```diff
@@ -223,13 +215,34 @@ const runCodeModelCoupling = () => {
...
 });
+
+watch(
+	() => props.node.inputs,
+	async () => {
+		const inputs = props.node.inputs;
+
+		// reset
+		modelMap.value = {};
+		codeText.value = '# The following <variable>: <model> mappings are available\n';
```
Tom-Szendrey, line 229: Whenever this watch is hit (when we open drill down) we will rewrite this codeText value
I do not see this code being saved in state right now but we will likely want to keep it later?

Approving as this is not an issue while we do not have this saved in state

Tom-Szendrey (approved):

packages/client/hmi-client/src/workflow/ops/model-coupling/tera-model-coupling.vue
mwdchang: Code fragment is not saved right now. I think we might be able to just update the "info" section - but need to think about this a bit.

PR #2467, opened 2024-01-02
Add basic auth service request functionality

kbirk:
Removed the trusted port approach in favour of https://github.com/unchartedsoftware/pantera-platform-core/pull/408


packages/server/src/main/java/software/uncharted/terarium/hmiserver/security/SecurityConfig.java
```diff
@@ -11,74 +10,70 @@
...
-//	public SecurityFilterChain userSecurityFilterChain(HttpSecurity http) throws Exception {
-//		http
-//			.securityMatcher("/api/**")
-//			.authorizeHttpRequests((authorize) -> authorize
-//				.anyRequest().hasRole(Roles.USER)
-//			);
-//		http.oauth2ResourceServer(configurer -> configurer.jwt(jwtConfigurer -> jwtConfigurer.jwtAuthenticationConverter(authenticationConverter)));
-//		return http.build();
-//	}
+	// @Bean
```
bigglesandginger, line 52: If you want you can totally feel free to delete these Order(2) functions - they are something that should have been deleted a while ago but I was feeling unsure.

bigglesandginger:
A brief review (as I will be unavailable for the next two days) looks good.  I approve but have not tested to see if the secure route vs unsecure is working (one should be able to use the app normally, not able to cURL an app api endpoint without creds, and equally not be able to use 3001 with the client but be able to cURL without creds... .

packages/server/src/main/java/software/uncharted/terarium/hmiserver/TerariumApplication.java
```diff
@@ -13,50 +13,42 @@
...
-import java.util.Set;
+import software.uncharted.terarium.hmiserver.configuration.Config;
 
 @SpringBootApplication
 @EnableFeignClients
 @EnableScheduling
 @PropertySource("classpath:application.properties")
 public class TerariumApplication {
+
+	@Autowired
```
dvince2, line 24: Not a deal breaker but in general we're now trying to avoid `@Autowired` in favour of a `@RequiredArgsConstructor`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/configuration/Config.java
```diff
@@ -24,6 +42,40 @@ public class Config {
 	 */
 	String baseUrl;
 
+	/**
+	 * The private internal API port of the application, will bypass keycloak
+	 * authentication (if within the selected controllers).
+	 */
+	Integer trustedPort = 3001;
```
dvince2, line 49: Before merging to main we should have this defined in `application.properties`

dvince2 (approved):

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/security/SecurityConfig.java
kbirk: Done!

PR #2475, opened 2024-01-04
Random Paper cuts

YohannParis:
Work to remove some random paper-cuts that @jamiewaese-uncharted found in the application:

- Change the _Workflow_ context menu to use `ContextMenu`.

### DIFF
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -353,19 +353,19 @@ const categories: Record<string, { label: string; icon?: string; separator?: boo
 	code: {
 		label: 'Code operators'
 	},
-	document: {
-		label: 'Document operators'
-	},
+	// document: {
+	// 	label: 'Document operators'
+	// },
 	dataset: {
 		label: 'Dataset operators'
 	},
 	simulate: {
 		separator: true,
 		label: 'Simulate'
-	},
-	llm: {
-		label: "Ask 'em LLM tool"
 	}
+	// llm: {
+	// 	label: "Ask 'em LLM tool"
+	// }
 };
 const operationContextMenuList = [
 	// Model
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -346,31 +346,25 @@ const removeNode = (event) => {
 const largeNode = { width: 420, height: 220 };
 
 // Menu categories and list items are in order of appearance for separators to work
-const categories: Record<string, { label: string; icon: string; separator?: boolean }> = {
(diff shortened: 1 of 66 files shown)


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -356,116 +357,127 @@ const removeNode = (event) => {
...
+const addOperatorToWorkflow: Function = (operator: any) => () => {
+	workflowService.addNode(wf.value, operator.operation, newNodePosition);
+	workflowDirty = true;
+};
+const addLargeOperatorToWorkflow: Function = (operator: any) => () => {
+	workflowService.addNode(wf.value, operator.operation, newNodePosition, {
+		size: { width: 420, height: 220 }
+	});
+	workflowDirty = true;
+};
```
mwdchang, line 369: Would prefer to just have a single method that passes in optional size, rather than normal, large, and what other sizes that comes up.

```
addOperatorToWorkflow(operator, size?) 
```

packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -356,116 +357,127 @@ const removeNode = (event) => {
...
+const contextMenuItems: MenuItem[] = [
+	// Model
+	{
 		label: 'Model operators',
-		icon: 'pi pi-share-alt'
+		items: [
+			{
+				label: ModelOp.operation.displayName,
+				command: addOperatorToWorkflow(ModelOp)
+			},
```
mwdchang, line 380: This was previously generated, so we didn't have to repeat label/command over and over to keep the 3rd-party interface at a minimum. What are the benefits of changing this?

packages/client/hmi-client/src/workflow/tera-workflow.vue
mwdchang: Also technically speaking `operator` is typed, the typing is available in services/workflow.ts though it may not be exported.

```
interface OperatorImport {
  name: string;
  operation: Operation;
  node: Component;
  drilldown: Component;
}

```

mwdchang (changes requested):

packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis: Both are done

packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis: The use of function to avoid a repetition when we have to list all the `name` none the less mase the code unclear. Now we also have to group those operator. So listing them is non-negotiable.
This is not DRY, but it's clear, and simple to understand what is happening.

mwdchang (approved):

PR #2480, opened 2024-01-05
chore: funman hardening part 2

mwdchang:
### Summary
Bring Funman closer to existing operator paradigm and the design specification.

- Adding in the output dropdown/switcher
- UI tweak to make selecting "interested variables" more usable using a multi-select
- Fix state saving and restoration bugs, add requestParameters into saved state spec
- Harden some of the state typing; remove extraneous checks
- Consolidate usage between model-configuration and model - we don't really need to fetch model
- Misc cleanups

There is a bit of a change in terms of how states are handles, rather than proxying state directly, state values are copied into separate refs, and are initialized in the setModelOptions function. This is to facilitate switching between different output-ports, as well it makes it a little easier to write changes back into the actual state-object on the node.

### Testing
Should be able to run several funman valudations with different configuration values, these should be available in the output dropdown. When switching between different output previews the state should reflect that of the selected output.

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
+++ b/packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
@@ -50,7 +50,19 @@
 						</div>
 						<Slider v-model="tolerance" :min="0" :max="1" :step="0.01" />
 						<div class="section-row">
+							<label>Select parameters of interest<i class="pi pi-info-circle" /></label>
+							<MultiSelect
+								ref="columnSelect"
+								:modelValue="variablesOfInterest"
+								:options="requestParameters.map((d: any) => d.name)"
+								:show-toggle-all="false"
+								@update:modelValue="onToggleVariableOfInterest"
+								:maxSelectedLabels="1"
+								placeholder="Select columns"
+							/>
+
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/workflow/ops/funman/tera-funman.vue
```diff
@@ -280,106 +328,138 @@ const addConstraintForm = () => {
...
+			.filter((d: any) => d.label === 'all')
+			.map((d: any) => d.name);
+	} else {
+		toast.error('', 'Provided model has no parameters');
 	}
 };
 
-const setRequestParameters = async (modelParameters) => {
+const setRequestParameters = (modelParameters: ModelParameter[]) => {
+	if (props.node.state.requestParameters) {
```
Tom-Szendrey, line 399: Need to check if `requestParameters` is not empty.
node is initiated with state.requestParameters = [] which will satisfy this if

mwdchang: Yeah, I just noticed this as well - nice catch.

Tom-Szendrey (approved):

PR #2481, opened 2024-01-05
Remove IProject interface

### LINKED ISSUE #2476: [TASK]: Use new ProjectAssets and remove duplicate Project and Workflow model impl. on front end
**Describe the task**
Currently, in `Project.ts` we have a second implementation of the Project model.  This duplicate project contains a field `assets` which contains all of the project's actual assets (documents, models, workflows, etc etc) in addition to the standard Project fields.  This is a holdover from using TDS, and is not idea because we don't need to have all of these assets held in memory.  In the new Project model as defined in `Types.ts` there is a `projectAssets` field which contains the metadata about an asset but not the asset itself.  This is what should be used, then just when needed, we can ask the back end for the actual asset to display to the user.
When the correct calls have been made we will also run into an issue with Workflows, as workflows have been entirely defined in the front end with the back end models being dummy objects.  I've updated our back end to have the correct typing but the front end will need to be changed to use the workflow objects from Types.ts

**Tasks to do**
- [ ] Remove the IProject model in favour of the Project model
- [ ] Get rid of the call to `/projects/${projectId}/assets` in `services/project.ts`
- [ ] Replace lightweight front end display of assets to use `Project.projectAssets`
- [ ] Replace manipulation of actual assets with a call to the correct back end controller to get the actual Dataset/Model/Workflow
- [ ] Replace workflows and their related objects with the new types.ts models.

YohannParis:
# Description

* Remove unused `ProjectPages` enum values.
* Update `useResourcesStore` to remove unused project methods.
* Update `ProjectMenu` to use new `Project` interface.
* Add sorting difference between `createdOn` and `updatedOn` on Homepage
* Update `Project` service to returns `userName` from `User` service as well
* Remove `getAssets()` from the Project service
* Remove `types/Project.ts`
* Update Project Composables
* Added a `assetName` to `ProjectAsset` Java model

Resolves #2476


packages/client/hmi-client/src/types/Project.ts
```diff
@@ -1,37 +1,8 @@
...
-	AssetType,
-	Code,
-	Dataset,
-	DocumentAsset,
-	ExternalPublication,
-	Model,
-	Project
-} from '@/types/Types';
-import { Workflow } from '@/types/workflow';
+import { AssetType } from '@/types/Types';
```
dvince2, line 1: This entire class i think can go away once this is done. That `isProjectAssetTypes` is only used in one place and i think its only for the deprecated assettype.. And that enum of one value seems unused?

YohannParis: That is the goal

PR #2482, opened 2024-01-05
Load resources fix

dvince2:
This is not a completed fix yet, but, will be blocking soon on the front end revamp with some of the back end changes so I want to do a WIP commit to our branch.  The code here is a little jank because we're at a crossroads on how assets are loaded, but, the front end is not quite there yet to load assets on demand and as such we need to still have the "getAssets" working correctly to get assets to the front end.

Assets now correctly appear in the UI with their names, but, can't be opened quite yet. However the skeleton there is working correctly for the next refactor.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/OperatorStatus.java
```diff
@@ -0,0 +1,35 @@
+package software.uncharted.terarium.hmiserver.models.dataservice.workflow;
+
+public enum OperatorStatus {
+	/*
```
kbirk, line 4: Should we just remove this commented stuff?

kbirk (approved):

PR #2484, opened 2024-01-08
rewrite document asset

blanchco:
# Description
rewrite the document asset view

Screenshot 2024-01-09 at 11 38 26 AM

Resolves #(issue)


packages/client/hmi-client/src/components/documents/tera-extractions.vue
blanchco, line 1: I moved the extractions into its own component so that it can be used by both the drilldown and the normal document asset view

packages/client/hmi-client/src/components/documents/tera-extractions.vue
```diff
@@ -0,0 +1,123 @@
...
+					<h5>{{ equation.name }}</h5>
+				</template>
+				<Image id="img" :src="equation.asset.metadata?.url" :alt="''" preview />
+			</tera-asset-block>
+		</AccordionTab>
+		<AccordionTab v-if="!isEmpty(clonedState.figures)">
+			<template #header>
+				<header>
+					Figure Images
+					<span style="font-size: var(--font-caption)">({{ countFigures }} selected)</span>
```
shawnyama, line 26: This should be in the css

packages/client/hmi-client/src/components/documents/tera-extractions.vue
```diff
@@ -0,0 +1,123 @@
...
+				</header>
+			</template>
+			<tera-asset-block
+				v-for="(figure, i) in clonedState.figures"
+				:key="i"
+				:is-included="figure.includeInProcess"
```
shawnyama, line 32: Is there still a need to show the 'Include in process' toggle if we are not viewing this in a workflow drilldown?

shawnyama (approved):

packages/client/hmi-client/src/components/documents/tera-extractions.vue
blanchco: The designers still wanted this toggle to show up, just so when we drag the asset in the drilldown it will already show which extractions they want to use.  We dont actually have the ability to save this state at the moment so its not entirely functional when looking at the asset page

packages/client/hmi-client/src/components/documents/tera-extractions.vue
blanchco: done

packages/client/hmi-client/src/components/documents/tera-extractions.vue
```diff
@@ -0,0 +1,127 @@
...
+import Accordion from 'primevue/accordion';
+import AccordionTab from 'primevue/accordiontab';
+import { computed, ref } from 'vue';
+import { cloneDeep, isEmpty } from 'lodash';
+import { AssetBlock } from '@/types/workflow';
+import Image from 'primevue/image';
+import TeraAssetBlock from '../widgets/tera-asset-block.vue';
+
+const props = defineProps<{
+	state: DocumentOperationState;
```
YohannParis, line 82: I do not think this should be aware of a state. I think it should receive the extractions directly and let the drilldown deal with state management. This could be confusing in the future.

blanchco: fair, Ill look into changing this up

PR #2493, opened 2024-01-09
fix: allow decimal input

mwdchang:
### Summary
Model config editor did not allow fractional numbers

### Test
Being able to enter fraction in model-config-editor


mwdchang:
See https://github.com/DARPA-ASKEM/terarium/pull/2494

PR #2496, opened 2024-01-09
feat: model template edges and template name edit

shawnyama:
# Description

- link nodes with straight edges
- rename model cards

File name changes:
- `tera-model-template.vue` -> `tera-model-template-editor.vue`
- `tera-model-template-card.vue` -> `tera-model-template.vue` (this change looks messy)

https://github.com/DARPA-ASKEM/terarium/assets/28237258/d57eb1d9-a538-46a1-be95-d879d53ce0df

Resolves #(issue)


mwdchang:
Maybe consider instead of the renaming of
```
tera-model-template.vue -> tera-model-templator.vue
```

do

```
tera-model-template.vue -> tera-model-template-editor
or
tera-model-template.vue -> tera-model-template-composer
```

to better reflect their roles?

packages/client/hmi-client/src/components/model-template/tera-model-template.vue
```diff
@@ -1,150 +1,163 @@
...
-	color: var(--text-color-subdued);
-	cursor: help;
-}
+<style>
+/* When a card is placed in the data layer of the infinite canvas 
+(eg. ports shouldn't look selectable if card is in the sidebar) */
+.data-layer .card-container {
+	& .card:hover > .p-button {
+		display: block;
+	}
```
mwdchang, line 156: Not a fan of this kind of coupling across components, as it is prone to creating bugs. 
We can leave this in as-is for this PR, but I think we should think of ways to have the constraint in a more spelled-out manner, maybe something like:

```
<tera-model-template :readonly="true" /> 
```

or something to this effect.

packages/client/hmi-client/src/components/model-template/tera-model-template.vue
```diff
@@ -1,150 +1,163 @@
...
+				</header>
+				<section>Diagram/Equations</section>
+			</main>
+			<Button icon="pi pi-ellipsis-v" rounded text />
+		</section>
+		<ul>
+			<li
+				v-for="(variable, index) in fakeVariables"
+				class="port"
 				:key="index"
```
mwdchang, line 20: Avoid `:key=index` if you can, it can have unexpected reactivity issues because object in the list can change but the key is the same.

mwdchang (approved):

PR #2497, opened 2024-01-09
rewrite dataset

blanchco:
# Description
- v1 rewrite of dataset
- tera-columnar-panel: a component which allows us to evenly divide a view into sections (now used in drilldown as well)
- tera-asset-card: a component which is surrounded with a grey background and border, user can define whatever they want to go in there
- adding tabs to tera-asset

https://github.com/DARPA-ASKEM/terarium/assets/33158416/c64004d0-2c45-4333-bac9-c7d366ba2965

Resolves #(issue)


packages/client/hmi-client/src/components/widgets/tera-asset-card.vue
shawnyama, line 1: Since the list rules are shared in `tera-model-description` and `tera-dataset-description` you can apply them here if you use `:deep`

packages/client/hmi-client/src/workflow/ops/dataset/tera-dataset-drilldown.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
+		<section tabName="Description">
+			<tera-drilldown-section :is-loading="fetchingDataset">
+				<div>
+					<tera-dataset-description :dataset="dataset" :raw-content="rawContent" />
+				</div>
```
shawnyama, line 7: remove uneeded div

packages/client/hmi-client/src/workflow/ops/dataset/tera-dataset-drilldown.vue
```diff
@@ -0,0 +1,70 @@
...
+			</tera-drilldown-section>
+		</section>
+		<section tabName="Data">
+			<tera-drilldown-section :is-loading="fetchingDataset">
+				<div class="datatable-container">
+					<tera-dataset-datatable :rows="100" :raw-content="rawContent" />
+				</div>
```
shawnyama, line 14: remove div and move this container class to the component

packages/client/hmi-client/src/components/widgets/tera-columnar-panel.vue
```diff
@@ -0,0 +1,18 @@
...
+
+<script setup lang="ts"></script>
+
+<style scoped>
+main {
+	display: grid;
+	grid-auto-flow: column;
+	grid-template-columns: repeat(auto-fit, minmax(0, 1fr));
+	overflow: hidden;
+	padding: 1rem;
```
shawnyama, line 15: use `var(--gap)` ;)

shawnyama, line 15: Maybe move this padding to `tera-drilldown` since it'll only be used there. That way you don't have to do a `padding: 0` in tera-model-description for example

shawnyama (approved):

packages/client/hmi-client/src/components/widgets/tera-columnar-panel.vue
blanchco: good idea, added!

packages/client/hmi-client/src/workflow/ops/dataset/tera-dataset-drilldown.vue
blanchco: done

packages/client/hmi-client/src/workflow/ops/dataset/tera-dataset-drilldown.vue
blanchco: i needed to move this div to the description, component.  It was hiding content without it

packages/client/hmi-client/src/components/widgets/tera-asset-card.vue
blanchco: I've renamed the component (apparently we had another component named tera-asset-card as well oops).  I think the idea I had for this component was so that it can be used outside of just the model/dataset description cards.  Its meant to just be a really basic dumb component that a dev can use whenever they want a component with that grey background/border.  So whenever a dev uses this they can define whatever styling they want and its not necessarily going to be those shared styles we have for the info cards in model/dataset.  Let me know what you think

packages/client/hmi-client/src/workflow/ops/dataset/tera-dataset-drilldown.vue
shawnyama: hmm it needed a proper parent element anyways I guess

packages/client/hmi-client/src/components/widgets/tera-asset-card.vue
shawnyama: If this is going to be used outside of assets then it should be named something like `tera-grey-card` in `/components/widgets`. I guess we can reconsider making an info card component if there are going to be more than what we have so far.

blanchco: Ya that makes sense, might add the asset card later on, but for now have renamed to `tera-grey-card` 😄

PR #2510, opened 2024-01-11
chore: remove unused files

mwdchang:
### Summary
- Clean up unused files
- Remove unused stratify modules and components
- Merge the two petrinet-services together to make things less confusing

There are likely objects that can be cleaned up as well on the server side, will do this after the server migration has completed.


YohannParis (approved):
Looks good to me!

shawnyama (approved):

PR #2515, opened 2024-01-12
Model config matrices WIP

Jami159:
Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
@@ -4,12 +4,22 @@
 		<section>
 			<div>
 				<h4>Initials</h4>
-				<Accordion v-if="stratifiedModelType && !_.isEmpty(initials)" :multiple="true">
-					<AccordionTab v-for="[init, vals] in initials" :key="init">
-						<template #header>
-							{{ init }}
-						</template>
-						<ul>
+				<!-- TODO: create a separate component for this custom "Accordion" -->
+				<div v-if="stratifiedModelType && !_.isEmpty(initials)">
+					<div v-for="([init, vals], idx) in initials" :key="init">
+						<div>
+							<Button
+								:icon="openInitials.includes(idx) ? 'pi pi-chevron-down' : 'pi pi-chevron-right'"
+								class="p-button-sm p-button-text"
+								@click="toggleAccordion(FieldTypes.StratifiedInitial, idx)"
+							/>
+							<Button
+								:label="init"
+								class="p-button-text"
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
```diff
@@ -116,18 +204,25 @@
...
+import { StratifiedMatrix } from '@/types/Model';
+
+enum FieldTypes {
+	Parameter = 'parameter',
+	Initial = 'initial',
+	StratifiedParameter = 'stratified-parameter',
+	StratifiedInitial = 'stratified-initial'
+}
 
 const props = defineProps<{
```
mwdchang, line 227: Can you add a explanation comment here? Because a model already has initials and parameters, why do we need to pass them in separately?

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -132,27 +132,35 @@ const updateState = (updatedField) => {
...
-	if (idx !== -1) {
-		state.parameters[idx] = param;
+	if (updateAll) {
+		state.parameters = params;
+	} else {
+		// find param with the same id and update it
+		const idx = state.parameters.findIndex((p) => p.id === params[0].id);
+		if (idx !== -1) {
+			state.parameters[idx] = params[0];
+		}
```
mwdchang, line 146: The update functions have confusing signatures - I had to dive deep into the implementation to understand how the array and boolean flag interact with each other. 

We can leave it alone for now, but let's have a chat when you get around to clean things up.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-editor.vue
Jami159: these initials and parameters are the ones that are stored in state: https://github.com/DARPA-ASKEM/terarium/blob/bd5323b5a071de2f614fb7c4db602253168bce48/packages/client/hmi-client/src/workflow/ops/model-config/model-config-operation.ts#L10-L11

I thought it would be better to just store the initials/parameters in state instead of the whole model. This also makes it a bit easier (I think) to modify and track the fields in the wizard as well as switch between different outputs.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Jami159: I did initially think about having separate functions for updating (e.g. updateSingleParam and updateAllParams or something), but thought this approach miigghttt be okay LOL. But happy to make changes if necessary.

mwdchang (approved):

PR #2529, opened 2024-01-15
fix: sidepanel scroll persists

blanchco:
# Description

* A fix to allow the scroll position in the asset nav sidebar to persist, we now directly go to the overview page when clicking on a nav item.  Removing the key attribute will no longer make the component refresh so it keeps the scroll position

BEFORE:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/bb50851d-e15d-46a4-8759-f381fc196d75

AFTER:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/4f70789e-4335-4a9e-9d9f-376d69fca665

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/App.vue
+++ b/packages/client/hmi-client/src/App.vue
@@ -9,7 +9,7 @@
 	</header>
 	<main>
 		<router-view v-slot="{ Component }">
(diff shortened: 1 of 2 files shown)


shawnyama (approved):
cool 😎

YohannParis (approved):
Cool indeed 🎉

PR #2537, opened 2024-01-16
[TASK] Add type to import Types

YohannParis:
# Description

* Some issues with types not being imported properly on the front-end.
* This is a full project replace of `import (\{.*\} from '@\/types\/Types';)` to `import type $1`.
* Fix the issue or `enum` within `types/Types.ts` are imported without the `type` keyword.


shawnyama (approved):
I wonder if it would be worth generating a seperate `Enums` file for the enums

docs/Typescript_Model_Generation.md
```diff
@@ -63,7 +63,7 @@ tl;dr `yarn dev` and the rest is handled.
 
 The generated types are added to `packages/client/hmi-client/src/types/Types.ts` and can be imported as follows:
 ```typescript
-import { EventType } from '@/types/Types';
+import type { EventType } from '@/types/Types';
```
shawnyama, line 66: `EventType` is an enum not a type

YohannParis:
> I wonder if it would be worth generating a seperate `Enums` file for the enums

Those are generated automatically from the Java backend, so I do not think it's worth our time.
I'm sure this is something will get use to quite quickly.

PR #2542, opened 2024-01-16
feat: template cards connect through junctions

shawnyama:
# Description

- Added junctions to target card ports
- For now a junction appears 500px to the right once a new edge is created (and its port isn't referring to any junctions already)
- Junctions are draggable
- Usability should be tested by designers

Things to improve - maybe @mwdchang has some ideas about these:
- AStar calculations are laggy
- Need to make the edges more curvy like the design

https://github.com/DARPA-ASKEM/terarium/assets/28237258/7954a008-546b-4c98-b5a4-ef415f2000a8


shawnyama:
^ started fixing a little edge case will get back to it later

mwdchang:
@shawnyama I didn't realize you added in A-Star algorithm already, it is problematic if you try to evaluate it every rendering cycle for all edges, I will try to take a closer look, but my general strategy would be
- increase the algorithm's internal grid size (basically you'd reach your goal faster)
- since each edge is calculated independently of other edges, only update the elements that are actually moving, 

Or conversely if we stick strictly to the notion of bipartite graphs (where there is a vertical line that the elements in each partition cannot cross) we can come up with our own algorithms that will be much cheaper.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -140,26 +161,42 @@ let currentPortPosition: Position = { x: 0, y: 0 };
...
+		}
+	}
+	return false;
+}
+
 function interpolatePointsForCurve(a: Position, b: Position): Position[] {
-	const controlXOffset = 0;
-	return [a, { x: a.x + controlXOffset, y: a.y }, { x: b.x - controlXOffset, y: b.y }, b];
+	return getAStarPath(a, b, collisionFn);
 }
```
mwdchang, line 193: Most of the performance will be gained by not calculating the AStar-path in rendering.

If we stick with AStar,
- You want to calculate out `edge.points` on node-dragging events, and then 
- Here you'd just return the calculated `edges.points` ... e.g. you can remove this function


pseudo code:
```
const updatePosition = ({ x, y }, node: any) => {
  .... stuff ...
  const edges_going_in = get_edges_that_are_input(node)
  const edges_going_out = get_edges_that_are_output(node)
  
  edges_going_in.forEach(e => {
    e.points = getAStarPath(_.first(e.points), getPortPosition(node, e))
  })
  edges_going_out.forEach(e => {
    e.points = getAStarPath(_.last( getPortPosition(node, e), _.last(e.points))
  })
}
```

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -140,26 +161,42 @@ let currentPortPosition: Position = { x: 0, y: 0 };
...
 	() => newEdge.value && newEdge.value.points && newEdge.value.points.length === 2
 );
 
+function collisionFn(p: Position) {
+	const buffer = 50;
+	for (let i = 0; i < modelTemplates.value.length; i++) {
+		const checkingNode = modelTemplates.value[i].metadata.templateCard;
+		// FIXME: Thi is  a hack to get around hierarhical geometries, will need to
+		// relax this guard.
+		// if (node.nodes && node.nodes.length > 0) continue;
```
mwdchang, line 181: Let's remove the copy-pasta comments - we don't have nested hierarchies, at least not in a way that the edges will need to cross nesting levels.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -240,6 +332,42 @@ function onDrop(event) {
 	newModelTemplate.value = null;
 }
 
+const updatePosition = ({ x, y }, node: any) => {
+	if (!isMouseOverCanvas) return;
+
+	const isJunction = node.edges !== undefined;
```
mwdchang, line 335: Should add a comment or fixme to explain, cause it is a bit misleading on first reading - cause it isn't clear you got two different types of nodes.

mwdchang (approved):

PR #2546, opened 2024-01-17
feat(hmi-server): add endpoint to see if name already exists

### LINKED ISSUE #2465: [FEAT]: Add endpoint to see if name already exists
From @mwdchang:
Something came to mind while cleaning up a handful of old and out-of-scope tickets. We have at occasions reported a desire to not-have duplicate names on workflows/models etc etc. Can we craft an API end point to check for this? Maybe something like
`GET /asset-name/:type/:name?[projectId=:id]`
That returns true if name is available, false otherwise.
If projectId is provided it checks for duplicate only within the scope of project's assets, otherwise it checks the global scope.

dvince2:
Adding a new endpoint to the server to check if an asset already exists with that name. This can optionally be done within the scope of a specific project or across all entries of that asset type.  In addition, blowing away old unused `toString()` functionality inside of AssetType as it was a nightmare to work with and unused.

**NOTE**
This is being merged into the new dataservice branch as it makes use of new database calls, and will be a part of `main` once that branch is merged in.

Resolves #2465

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/AssetController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/AssetController.java
@@ -1,0 +1,104 @@
+package software.uncharted.terarium.hmiserver.controller.dataservice;
+
+import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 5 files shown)


YohannParis (approved):
LGTM

PR #2548, opened 2024-01-18
Funman drilldown boundary chart

mwdchang:
### Summary
Continue to add to Funman features

- Add in drilldown for the boundary chart, and interactions to select a specific box
- Add in a (temporary) solution with Funman exits and zero-fills the remainder timestamps, we need to remove the zeros
- Add in parameters to constraint groups, and conversely filter out initial state-values
- Small stylistic tweaks

image

### DIFF
--- a/packages/client/hmi-client/src/components/funman/tera-funman-output.vue
+++ b/packages/client/hmi-client/src/components/funman/tera-funman-output.vue
@@ -11,7 +11,31 @@
 	</p>
 
 	<!-- TODO: add boxes modal per row https://github.com/DARPA-ASKEM/terarium/issues/1924 -->
-	<div class="variables-table">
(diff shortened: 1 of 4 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/funman/tera-funman-output.vue
shawnyama, line 1: May wanna wrap this component with a parent tag, Vue may give warnings if you don't.

shawnyama (approved):

packages/client/hmi-client/src/components/funman/tera-funman-output.vue
mwdchang: Vue3 does supporting multiple roots within template sections.

PR #2552, opened 2024-01-18
tds migration frontend fixes

shawnyama:
# Description

- models appear in project view and preview panel
- dropdown options and selection works in asset nodes

image

image


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/ModelDescription.java
```diff
@@ -29,10 +39,13 @@ static public ModelDescription fromModel(Model model) {
 					.setModelVersion(model.getHeader().getModelVersion())
 					.setExtractedFrom(model.getHeader().getExtractedFrom()));
 		}
-
 		return desc;
 	}
 
+	@Id
```
dvince2, line 45: In its current form, you can get rid of the `@Id` and `@GeneratedValue` annotations I think (tagging @kbirk here to be sure)  - this ID doesnt get added to the postgres database so these aren't autogenerated, and from the look of the code these are all being set off of their parent's ID in the block above?

dvince2 (approved):
One minor note on annotations but this looks good

PR #2556, opened 2024-01-19
redesign model config drilldown

blanchco:
# Description

https://github.com/DARPA-ASKEM/terarium/assets/33158416/2bd8a72b-24b6-44a0-bf06-e9f698ce9a16

- model configuration redesign
- ability to edit values with a table view
- tera-output-dropdown component created so that the output dropdown can be used outside the preview

Resolves #(issue)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -0,0 +1,206 @@
...
+	{ label: 'Distribution', value: ParamType.DISTRIBUTION },
+	{ label: 'Time Varying', value: ParamType.TIME_SERIES }
+];
+const props = defineProps<{
+	modelConfiguration: ModelConfiguration;
+	data: any[];
+	stratifiedModelType: StratifiedModel | null;
+	tableType: StratifiedMatrix;
+	hideHeader?: boolean;
+}>();
```
mwdchang, line 137: it feels like we should be able to have some notion of typing for `data` at this stage? Since you created the tableFormattedXYZ right beforehand?

mwdchang (changes requested):

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -0,0 +1,206 @@
...
+					<label>Max</label>
+					<InputNumber
+						class="p-inputtext-sm"
+						inputId="numericInput"
+						mode="decimal"
+						:min-fraction-digits="1"
+						:max-fraction-digits="10"
+						v-model.lazy="slotProps.data.value.distribution.parameters.maximum"
+						@update:model-value="emit('update-value', [slotProps.data.value])"
+					/>
```
mwdchang, line 63: What resolution are you testing on, the min/max inputs are a bit messed up for me

image

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -0,0 +1,206 @@
...
+					:model-value="slotProps.data.type"
+					:options="typeOptions"
+					optionLabel="label"
+					optionValue="value"
+					placeholder="Select a parameter type"
+					@update:model-value="(val) => changeType(slotProps.data.value, val)"
+				/>
+			</template>
+		</Column>
+		<Column field="value" header="Initial value" style="width: 15%">
```
mwdchang, line 33: I would consider renaming this to just `Value` - "initial values" means they can change during the process of a simulation, but these are just constants or sampling distributions.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -0,0 +1,206 @@
...
+				<span v-else-if="slotProps.data.type === ParamType.EXPRESSION">Expression</span>
+				<Dropdown
+					v-else
+					:model-value="slotProps.data.type"
+					:options="typeOptions"
+					optionLabel="label"
+					optionValue="value"
+					placeholder="Select a parameter type"
+					@update:model-value="(val) => changeType(slotProps.data.value, val)"
+				/>
```
mwdchang, line 30: When I switch from constant => distribution, it seemed to wipe out my existing distribution min/max values.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
blanchco: Screenshot 2024-01-23 at 11 20 18 AM

I've added a
Screenshot 2024-01-23 at 11 21 35 AM
 class to the container so that the elements in the distribution will always show vertically:

packages/client/hmi-client/src/components/widgets/tera-output-dropdown.vue
```diff
@@ -0,0 +1,85 @@
...
+import { WorkflowOutput, WorkflowPortStatus } from '@/types/workflow';
+import Dropdown from 'primevue/dropdown';
+import Checkbox from 'primevue/checkbox';
+
+defineProps<{
+	options?: WorkflowOutput<any>[] | { label: string; items: WorkflowOutput<any>[] }[];
+	output?: WorkflowOutput<any>['id'];
+	isSelectable?: boolean;
+	isLoading?: boolean;
+}>();
```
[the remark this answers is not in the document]
shawnyama: Passing props like this should help https://vuejs.org/guide/components/props.html#prop-validation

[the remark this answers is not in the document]
blanchco: makes sense, Ive changed the options and output props so that they are no longer optional in this component:
```
defineProps<{
	options: WorkflowOutput<any>[] | { label: string; items: WorkflowOutput<any>[] }[];
	output: WorkflowOutput<any>['id'];
	isSelectable?: boolean;
	isLoading?: boolean;
}>();
```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -261,13 +395,12 @@ onMounted(async () => {
...
 		if (m) {
 			model.value = m;
-
-			// TODO: set this only if configs don't exist already
-			// set the state with the model initials and params
-			const state = _.cloneDeep(props.node.state);
-			state.initials = m.semantics?.ode.initials;
-			state.parameters = m.semantics?.ode.parameters;
-			emit('update-state', state);
+			if (outputs.value.length === 0) {
```
shawnyama, line 398: use `isEmpty` from lodash

packages/client/hmi-client/src/components/widgets/tera-output-dropdown.vue
```diff
@@ -0,0 +1,85 @@
...
+	font-size: var(--font-body-small);
+}
+
+:deep(.p-checkbox .p-checkbox-box) {
+	border-radius: var(--border-radius-medium);
+}
+
+.dropdown-option {
+	display: flex;
+	gap: 0.5rem;
```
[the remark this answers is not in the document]
blanchco: sorry just copy pasted this over from the preview section where it was previously

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
blanchco: The way we distinguish between a constant vs a distribution as far as I know is if the parameter has a distribution key on it.  When I switch between constant vs distribution it adds/deletes the key for distribution which is why it resets.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -18,41 +26,39 @@
 					/>
 				</div>
 				<div v-else-if="activeIndex === 1">
```
shawnyama, line 28: No need for this div

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
mwdchang: Okay - let's flag this as a FIXME as we may need to revisit this behaviour later.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
blanchco: I've got away with removing both tableType and stratifiedModelType.  tableType was a little tricky, I needed to add an additional emitter to the component instead called 'update-config' so that the parent can can the updates.  stratifiedModelType was extracted from the model-configuration as we discussed

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
blanchco: ive changed these in favour of v-show to keep the ui state when switching between pages.  So now the div is needed now!

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -0,0 +1,206 @@
...
+		v-model:expanded-rows="expandedRows"
+		dataKey="id"
+		:row-class="rowClass"
+		class="p-datatable-sm"
+		:class="{ 'hide-header': hideHeader }"
+	>
+		<Column expander style="width: 5rem" />
+		<Column field="id" header="ID" style="width: 10%"></Column>
+		<Column field="name" header="Name" style="width: 15%"></Column>
+		<Column field="type" header="Value type" style="width: 10%">
```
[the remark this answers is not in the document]
blanchco: I've personally found it easier for datatables to have their column widths stored inline, do we need 6 different css classes to define the column width for every column? But Im happy to change this if this is just a me thing 😬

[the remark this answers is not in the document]
shawnyama: Maybe you could use `nth-child` if you don't want to create classes. But yeah depending on the situation maybe its better to do inline

mwdchang (approved):

PR #2562, opened 2024-01-22
feat: staleness check and duplicate branch

mwdchang:
### Summary
Add provisional logic to 
- check if current operator.state is out of sync with operator.outputs[active].state
- copy logic to duplicate branches in the workflow

### Testing
yarn workspace hmi-client run test


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -358,3 +358,117 @@ export function updateOutputPort(node: WorkflowNode<any>, updatedOutputPort: Wor
...
+) => {
+	const hasKey = Object.prototype.hasOwnProperty;
+
+	const keys = Object.keys(outputState);
+	for (let i = 0; i < keys.length; i++) {
+		const key = keys[i];
+		if (!hasKey.call(operatorState, key)) return false;
+		if (!_.isEqual(operatorState[key], outputState[key])) return false;
+	}
+	return true;
```
YohannParis, line 376: Why not a simpler test on the array?

```suggestion
	return keys.includes(key => 
    hasKey.call(operatorState, key) && _.isEqual(operatorState[key], outputState[key])
  );
```

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -358,3 +358,117 @@ export function updateOutputPort(node: WorkflowNode<any>, updatedOutputPort: Wor
 	if (!outputPort) return;
 	outputPort = Object.assign(outputPort, updatedOutputPort);
 }
+
+// Check if the current-state matches that of the output-state.
+// Note operatorState subsumes the keys of the outputState
+export const isOperatorStateInSync = (
```
YohannParis, line 364: If this function doesn't use any out of scope information, why not making it a function?
```suggestion
export function isOperatorStateInSync(
```

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -358,3 +358,117 @@ export function updateOutputPort(node: WorkflowNode<any>, updatedOutputPort: Wor
...
+		n.y += offset;
+	});
+	copyEdges.forEach((e) => {
+		if (!e.points || e.points.length < 2) return;
+		e.points[0].y += offset;
+		e.points[e.points.length - 1].y += offset;
+	});
+
+	// 6. Finally put everything back into the workflow
+	copyNodes.forEach((node) => wf.nodes.push(node));
```
shawnyama, line 472: I think it'd be cleaner to do:
`wf.nodes.push(...copyNodes)`

shawnyama (approved):

packages/client/hmi-client/src/services/workflow.ts
mwdchang: No particular reason other than most of stuff in the file are of the form `export const x = (...) => { ... }` to keep things consistent

packages/client/hmi-client/src/services/workflow.ts
mwdchang: `includes` does not take in a predicate function, and all keys need to meet the conditions, so to do this in one shot I think you will need to do something like "negative of lodash _.some(...)" which I think comes at the expense of comprehension for 1-2 lines fewer code.

YohannParis: I see, otherwise `every()` could do that. But fair enough, just curious.

packages/client/hmi-client/src/services/workflow.ts
YohannParis: I see, am used to use `const` when I need access to reactive values from out of scope of a function.
But it does not matter.

packages/client/hmi-client/src/services/workflow.ts
mwdchang: done

PR #2563, opened 2024-01-22
Jamie: tweak design of sticky header on model page

jamiewaese-uncharted:
Currently when you scroll down the asset pages, the sticky header at the top of the asseet changes to a white background and a smaller font. It's a tiny bit disorienting. With this tweak they stay the same.

**BEFORE**
sticky header - before

**AFTER**
sticky header - after


packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -174,25 +174,24 @@ header {
...
+	top: 0px;
 	z-index: 100;
 	isolation: isolate;
-	background-color: rgba(255, 255, 255, 0.85);
+	background-color: var(--surface-ground-transparent);
 	backdrop-filter: blur(6px);
 	padding: var(--gap-small) var(--gap);
 	border-bottom: 1px solid var(--surface-border-light);
-	box-shadow: 0px 4px 8px -7px #b8b8b8;
+	/* box-shadow: 0px 4px 8px -7px #b8b8b8; */
```
YohannParis, line 194:
```suggestion

```

YohannParis (approved):

PR #2577, opened 2024-01-23
feat: move model editing procedures into its own service

shawnyama:
# Description

- moved model editing events to their own service
- cards can be deleted
- moved `Position` into `common.ts`
- noticed that some Workflow types aren't imported as types so I adjusted that (`import ...` => `import type ...`)

Aside:
- tried out debouncing  `updatePosition` but it keeps my card in a dragged state

https://github.com/DARPA-ASKEM/terarium/assets/28237258/4fdb10e2-37dc-460a-b8ee-e66d47e25000


packages/client/hmi-client/src/components/model-template/tera-model-template.vue
```diff
@@ -18,7 +18,8 @@
 				</header>
 				<tera-model-diagram :model="model" :is-editable="isEditable" is-preview />
 			</main>
-			<Button v-if="isEditable" icon="pi pi-ellipsis-v" rounded text />
+			<Button v-if="isEditable" @click="toggle" icon="pi pi-ellipsis-v" rounded text />
+			<Menu ref="menu" :model="cardOptions" :popup="true" />
```
YohannParis, line 22: [In the PrimeVue documentation](https://primevue.org/menu/#popup) you can see that they use the `aria-` attributes to do the job of your `toggle` function. That would be cleaner.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -180,8 +140,8 @@ const isCreatingNewEdge = computed(
 
 function collisionFn(p: Position) {
 	const buffer = 50;
-	for (let i = 0; i < modelTemplateCards.value.length; i++) {
-		const checkingNode = modelTemplateCards.value[i];
+	for (let i = 0; i < cards.value.length; i++) {
+		const checkingNode = cards.value[i];
```
YohannParis, line 144:
```suggestion
function collisionFn(p: Position): boolean {
    const buffer = 50;
    
    return cards.value.some(({ x, y, width, height }) => {
        const withinXRange = p.x >= x - buffer && p.x <= x + width + buffer;
        const withinYRange = p.y >= y - buffer && p.y <= y + height + buffer;
        return withinXRange && withinYRange;
    });
}
```

YohannParis (approved):
LGTM 🎉

packages/client/hmi-client/src/components/model-template/tera-model-template.vue
shawnyama: No the toggle function is actually still there. You gotta toggle the full code 😛

PR #2581, opened 2024-01-24
Cache user fetching

bigglesandginger:
# Description

Adds caches to user table in a hope to reduce request load to PostGres as this is processed for every request to the server and will almost never change

To see caching hits/missed in the logs add the following to application.properties:

`logging.level.org.springframework.cache=TRACE`

Resolves #(issue)

if the incorrect ENVIRONMENT is supplied to `hmiServerDev.sh` then the script errors as `break` is called inappropriately, this corrects that.  Also adds a lean local variant for when one's docker disk is too small for normal local

### DIFF
--- a/hmiServerDev.sh
+++ b/hmiServerDev.sh
@@ -227,7 +227,8 @@ case ${COMMAND} in
 
       start
         ENVIRONMENT
-          remote | local | full (default: remote)  Indicate which environment to develop against
+          remote | local | full | ll (default: remote)  Indicate which environment to develop against
+              (ll: local_lean to run local with the absolute minimal support to run hmiServer for development)
 
         run (default: null) Indicate whether to run the server after starting the containers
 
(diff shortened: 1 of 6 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/configuration/CacheConfiguration.java
```diff
@@ -19,6 +25,14 @@
 @Slf4j
 public class CacheConfiguration implements CachingConfigurer {
 
+	@Bean
+	public CacheManager cacheManager() {
+		SimpleCacheManager cacheManager = new SimpleCacheManager();
+		Cache booksCache = new ConcurrentMapCache("users");
+		cacheManager.setCaches(Arrays.asList(booksCache));
```
chris-dickson, line 32: Rename the cache from the example :P

chris-dickson (approved):
Looks good with the one minor change

hmiServerDev.sh
YohannParis, line 1: Could you update the bottom helper with the new `ll` option?

PR #2585, opened 2024-01-24
time-varying params

blanchco:
# Description

Initial implementation of time varying params in model config:

https://github.com/DARPA-ASKEM/terarium/assets/33158416/0b4acd27-19b8-4889-b836-05f55c50fd6f

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
@@ -208,7 +208,7 @@ const validateTimeSeries = (values: string) => {
 
 		const [time, value] = pair.split(':').map((part) => part.trim());
 
-		if (!/^\d+$/.test(time) || !/^\d+$/.test(value)) {
+		if (!/^\d+$/.test(time) || !/^\d+(\.\d+)?$/.test(value)) {
 			isValid = false;
 		}
 	});
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
@@ -189,34 +189,15 @@ const updateTimeseries = (id: string, value: string) => {
 };
 
 const validateTimeSeries = (values: string) => {
-	let isValid = true;
+	const message = 'Incorrect Format (e.g., 0:500, 10:550, 25:700 etc)';
 	if (typeof values !== 'string') {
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -165,26 +181,80 @@ const openMatrixModal = (datum: ModelConfigTableData) => {
...
+			isValid = false;
+		}
+	});
+
+	errorMessage.value = '';
+	if (!isValid) {
+		errorMessage.value = 'Incorrect Format (e.g., 0:500, 10:550, 25:700 etc)';
+	}
+	return isValid;
+};
```
YohannParis, line 221: You could simplify a lot this function now that you defined it.

```suggestion
const validateTimeSeries = (values: string): boolean => {
    const message = 'Incorrect Format (e.g., 0:500, 10:550, 25:700 etc)';
    
    if (typeof values !== 'string') {
        errorMessage.value = message;
        return false;
    }

    const isPairValid = (pair: string): boolean => /^\d+:\d+$/.test(pair.trim());
    const isValid = values.split(',').every(isPairValid);

    errorMessage.value = isValid ? '' : message;
    return isValid;
};
```

YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -165,26 +181,80 @@ const openMatrixModal = (datum: ModelConfigTableData) => {
...
+			isValid = false;
+			return;
+		}
+
+		const [time, value] = pair.split(':').map((part) => part.trim());
+
+		if (!/^\d+$/.test(time) || !/^\d+(\.\d+)?$/.test(value)) {
+			isValid = false;
+		}
+	});
```
mwdchang, line 214: FYI: This forEach is a bit inefficient, you will keep on parsing even if isValid is false

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
blanchco: changed

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
blanchco: changed in favour for the function yohann provided

PR #2593, opened 2024-01-25
codeblocks to amr

blanchco:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
@@ -10,7 +10,7 @@
 						:options="programmingLanguages"
 						@change="setKernelContext"
 					/>
-					<Button label="Add code block" icon="pi pi-plus" text @click="addCodeBlock" disabled />
+					<Button label="Add code block" icon="pi pi-plus" text @click="addCodeBlock" />
 				</header>
 				<tera-operator-placeholder
 					v-if="allCodeBlocks.length === 0"
@@ -25,7 +25,7 @@
 						:key="i"
 						:is-editable="asset.type !== CodeBlockType.INPUT"
 						:is-deletable="asset.type !== CodeBlockType.INPUT"
-						@delete="removeCodeBlock(i)"
+						@delete="removeCodeBlock(i - inputCodeBlocks.length)"
 						:is-included="allCodeBlocks[i].includeInProcess"
 						@update:is-included="
 							allCodeBlocks[i].includeInProcess = !allCodeBlocks[i].includeInProcess
@@ -139,6 +139,7 @@ import { getCodeAsset } from '@/services/code';
 import { codeToAMR } from '@/services/knowledge';
 import { CodeBlock, CodeBlockType, getCodeBlocks } from '@/utils/code-asset';
 import TeraAssetBlock from '@/components/widgets/tera-asset-block.vue';
+import TeraModelModal from '@/page/project/components/tera-model-modal.vue';
 import { ModelFromCodeState } from './model-from-code-operation';
 
 const props = defineProps<{
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/TDSCodeController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/TDSCodeController.java
@@ -24,9 +24,11 @@ import software.uncharted.terarium.hmiserver.models.code.GithubRepo;
 import software.uncharted.terarium.hmiserver.models.dataservice.PresignedURL;
 import software.uncharted.terarium.hmiserver.models.dataservice.code.Code;
 import software.uncharted.terarium.hmiserver.models.dataservice.code.CodeFile;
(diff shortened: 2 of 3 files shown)


packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -276,12 +277,29 @@ async function handleCode() {
...
-		const modelId = await codeToAMR(
-			props.node.inputs[0].value?.[0],
-			'temp model',
-			'temp model description',
-			true
-		);
+		const codeContent = allCodeBlocks.value
+			.filter((block) => block.includeInProcess)
+			.reduce((acc, block) => `${acc}${block.asset.codeContent}\n`, '');
+		const file = new File([codeContent], 'tempFile');
```
blanchco, line 283: put the code blocks all in 1 file and upload it to a "temporary" (public) code-asset which will be used to call the KM code-to-amr endpoint

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/TDSCodeController.java
```diff
@@ -404,6 +400,19 @@ private ResponseEntity<Integer> uploadCodeHelper(UUID codeId, String fileName, H
...
+
+			Map<String, CodeFile> fileMap = code.get().getFiles();
+
+			if(fileMap == null){
+				fileMap = new HashMap<>();
+			}
+			fileMap.put(fileName, codeFile);
+			code.get().setFiles(fileMap);
+			codeService.updateCode(code.get());
+
```
blanchco, line 415: adding file information back which was removed in the migration

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -106,6 +122,38 @@ ResponseEntity<ExtractionResponse> postCodeToAMR(
...
+
+		if(response.getStatusLine().getStatusCode() != 200) {
+			throw new Exception("Error uploading file");
+		}
+		// 3. create model from code asset
+		return ResponseEntity.ok(knowledgeMiddlewareProxy.postCodeToAMR(createdCode.getId().toString(), "temp model", "temp model description", false, false).getBody());
+		} 
+		catch (Exception e) {
+			log.error("unable to upload file", e);
+			return ResponseEntity.internalServerError().build();
```
dvince2, line 153: Instead of `return`ing an error, we've moved towards throwing a `ResponseStatusException` as it lets us include an error for the front end:

```
					org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR,
					"YOUR PLAINTEXT ERROR MESSAGE HERE");
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -106,6 +122,38 @@ ResponseEntity<ExtractionResponse> postCodeToAMR(
 						.getBody());
 	}
 
+		// Create a model from code blocks
+	@PostMapping(value = "/code-blocks-to-model", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+	@Secured(Roles.USER)
+	public ResponseEntity<ExtractionResponse> codeBlocksToModel(@RequestPart Code code, @RequestPart("file") MultipartFile input) throws IOException {
```
dvince2, line 128: You're also able to add swagger docs via annotations here now. Something like...
```
@Operation(summary = "Create a model from code blocks")
	@ApiResponses(value = {
			@ApiResponse(responseCode = "200", description = "BLAH BLAH", content = @Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = ExtractionResponse.class))),
			@ApiResponse(responseCode = "500", description = "BLAH BLAH", content = @Content)
	})
```

dvince2 (approved):

YohannParis (commented):
I do not understand why you are sending form data to the knowledge controller?
Why you are not creating a code asset via the data-service first, then send that like the classic code-to-amr.

blanchco:
> I do not understand why you are sending form data to the knowledge controller? Why you are not creating a code asset via the data-service first, then send that like the classic code-to-amr.

@YohannParis I figure that it would be better to do everything in one endpoint as opposed to 3 calls on the front end.  

We call new endpoint code-blocks-to-amr which:
creates a new code asset -> uploads the code blocks as a file to the new code asset -> calls code to amr on that code asset and returns the job id all in the backend.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
blanchco: good to know, added

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
blanchco: added 👍

PR #2594, opened 2024-01-26
Reduce database requests

bigglesandginger:
# Description

Caches Roles and Authorities - details calculated on every call

Does not save the creation of a User POJO from JWT data - the code then compares the POJO to the retrieved POJO in the database and updates as required

Changes the last_updated_ms to only store it to the minute level

Adds (commented out) logging for Hibernate and Caching - for debugging inspection

Resolves #(issue)


dvince2 (approved):

YohannParis (approved):
LGTM

PR #2603, opened 2024-01-29
2602 add common parent for terarium assets

### LINKED ISSUE #2602: Add common parent for terarium assets
Add a common `TerariumAsset` abstract class for all our assets to inherit from. This should have common properties like id, timestamps and temporary

dvince2:
# Description

All assets now derive from the same parent class. Also adding the flags for `temporary` and `publicAsset` but these features aren't in yet.

Resolves #2602

### DIFF
--- a/packages/client/hmi-client/src/services/knowledge.ts
+++ b/packages/client/hmi-client/src/services/knowledge.ts
@@ -78,9 +78,12 @@ export const equationsToAMR = async (
  * Given a model, enrich its metadata
  * Returns a runId used to poll for result
  */
-export const profileModel = async (modelId: string, documentId: string | null = null) => {
+export const profileModel = async (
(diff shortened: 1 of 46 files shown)


packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -78,9 +78,12 @@ export const equationsToAMR = async (
  * Given a model, enrich its metadata
  * Returns a runId used to poll for result
  */
-export const profileModel = async (modelId: string, documentId: string | null = null) => {
+export const profileModel = async (
+	modelId: string | undefined,
```
YohannParis:
```suggestion
	modelId: TerariumAsset["id"],
```

YohannParis (approved):

PR #2607, opened 2024-01-29
2544 - Async taskrunner with GoLLM `model_card` implementation

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/Model.java
```diff
@@ -32,6 +32,8 @@ public class Model extends TerariumAsset implements SupportAdditionalProperties,
 
 	private Map<String, Object> model;
 
+	private String modelCard;
```
kbirk, line 35: Not sure where the model card is supposed to be written. This is probably not correct.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -0,0 +1,134 @@
...
+		});
+		return handler;
+	}
+
+	@PostMapping("/model_card")
+	@Secured(Roles.USER)
+	public ResponseEntity<TaskResponse> createModelCardTask(
+			@RequestParam(name = "document-id", required = true) UUID documentId,
+			@RequestParam(name = "model-id", required = true) UUID modelId)
+			throws JsonProcessingException, IOException {
```
dvince2, line 89: Rather than throwing these for the client to deal with do we want to catch them and then throw a `ResponseStatusException` with a `INTERNAL_SERVER_ERROR` or whatever seems appropriate?

packages/taskrunner/README.md
```diff
@@ -0,0 +1,11 @@
+# Terarium Task Runner
```
dvince2, line 1: I wasn't sure where else to put this comment, so here will do :)

We'll need to update the `hmiServerDev.sh` to start this up locally

dvince2 (approved):
Just a few questions/formatting ideas. LGTM

packages/taskrunner/README.md
kbirk: I've added the container to the `docker-compose-common.yml` and that should really be enough to kick it off locally.

Getting a local dev container that hotswaps both the taskrunner code and the GoLLM code will be a `nice to have` that I'll get in after this PR.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
kbirk: Fixed!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -0,0 +1,134 @@
...
+			} catch (IOException e) {
+				log.error("Failed to write model card to database", e);
+			}
+		});
+		return handler;
+	}
+
+	@PostMapping("/model_card")
+	@Secured(Roles.USER)
+	public ResponseEntity<TaskResponse> createModelCardTask(
```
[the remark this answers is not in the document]
kbirk: Fixed!

packages/taskrunner/README.md
dvince2: 👍

PR #2612, opened 2024-01-30
chore: clean vite-local

mwdchang:
### Summary
The `TGPTController` is responsible for returning the connection information for beaker - e.g. the websocket address, jupyter-token ... etc. Currently instead of returning these information, the server returns the client address with dummy-routes, and then the client routes again based on the running configuration. This creates a weird circular double-proxy situation where the client needs to know the server and the server needs to know the client.

- client asks server how to connect to beaker
- server returns dummy client routes
- client connects through these routes
- the routes are intercepted by the client and rerouted to actual beaker host

This can be simplified by just getting the client to connect directly, so
- client asks server how to connect to beaker
- client connects through these routes


mwdchang:
Will revisit later

PR #2618, opened 2024-01-30
profile model when creating a model from code

blanchco:
# Description

* profile a model when there is a document attached to the model from code operator and the user runs

**to note KM needs to be updated for the profiling to work

https://github.com/DARPA-ASKEM/terarium/assets/33158416/688e58de-9c0a-4bf3-a80d-a4728f0dd891

sample of the model card in the preview:

Screenshot 2024-01-30 at 11 21 26 AM


shawnyama (approved):

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -346,6 +354,12 @@ async function handleCode() {
 	}
 }
 
+async function profile(modelId: string) {
+	const profileModelJobId = await profileModel(modelId, documentId.value);
+	await fetchExtraction(profileModelJobId);
+	const model = await getModel(clonedState.value.modelId);
+	selectedModel.value = model;
+}
```
YohannParis, line 362: This should be part of the model service.

PR #2619, opened 2024-01-30
feat: create subset data operator

shawnyama:
# Description

- Created a subset operator that accepts an esgf dataset
- Latitude and longitude coordinates can be entered as integers
- Then click Run to generate the subset
- Can save the subset as a new dataset
- Preview images still have to be worked on in the backend 
- Spatial and time skipping isn't passed to the endpoint yet

### Issues
- ~After the first run I get an error - it seems like a new subset can't be generated after that (it worked sometimes before though)~ resolved with poller

### Testing notes
- Search "historical" and switch the dropdown source to ESGF in the explorer and add one of those datasets to a project
- Make sure you use `yarn dev` since staging connects to the climate data service 

https://github.com/DARPA-ASKEM/terarium/assets/28237258/2de83641-526a-4c71-868e-14b529923fe5


packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
```diff
@@ -0,0 +1,396 @@
...
+	if (dataset.value?.esgfId && dataset.value?.id) {
+		updateState();
+		isSubsetLoading.value = true;
+
+		const subsetId = await getClimateSubsetId(
+			dataset.value.esgfId,
+			dataset.value.id,
+			`${longitudeStart.value},${longitudeEnd.value},${latitudeStart.value},${latitudeEnd.value}`,
+			`${fromDate.value.toISOString()},${toDate.value.toISOString()}`
+			// isSpatialSkipping.value ? spatialSkipping.value ?? undefined : undefined // Not sure if its this or timeSkipping
```
mwdchang, line 288: Do we have an answer now?

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -50,6 +50,25 @@ async function getClimateDataset(datasetId: string): Promise<Dataset | null> {
 	return response?.data ?? null;
 }
 
+async function getClimateSubsetId(
+	esgfId: string,
+	parentDatasetId: string,
+	envelope: string,
+	timestamps?: string,
+	thinFactor?: string
+): Promise<string | null> {
```
mwdchang, line 59: This functional interface design is unclear, if you have consecutive optionals they become difficult to reason with. e.g. if I want to specify `thinFactor` but not `timestamps` then timestamps is no longer optional.

If you have a bunch of things that may or may not be specified, place them into an options object:
```
function abc( a: string, b: string, c: number, options: { .... })
```

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -184,10 +184,13 @@ const fetchDataset = async () => {
 				} else {
 					// We are assuming here there is only a single csv file. This may change in the future as the API allows for it.
 					image.value = undefined;
-					rawContent.value = await downloadRawFile(
-						props.assetId,
-						datasetTemp?.fileNames?.[0] ?? ''
-					);
+					// Tmep solution to avoid downloading raw file for netcdf files
```
YohannParis, line 187:
```suggestion
					// TODO = Temporary solution to avoid downloading raw NetCDF files, which can be massive
					// A better solution would be to check the size of an asset before downloading it, and/or 
					// downloading a small subset of it for presentation purposes.
```

packages/client/hmi-client/src/components/drilldown/tera-output-dropdown.vue
```diff
@@ -43,6 +43,7 @@ const emit = defineEmits(['update:selection']);
 
 <style scoped>
 .p-dropdown {
+	max-width: 20rem;
```
YohannParis, line 46: We might also truncate the names of output with an ellipsis if they are too big.
@ngraham76 we should look on improving those output dropdown to be more complex, with multiline information containing Name, status, createdOn/updatedOn time, if it's a saved asset or not.

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -21,7 +21,7 @@ async function getAll(): Promise<Dataset[] | null> {
 	return response?.data ?? null;
 }
 
-async function getAllClimate(query: string): Promise<Dataset[] | null> {
+async function searchClimateDatasets(query: string): Promise<Dataset[] | null> {
```
YohannParis, line 24: I prefer using the word `question` instead of `query`, it clarify that this is an English string written by a human instead of a software created query. If that makes sense?
```suggestion
async function searchClimateDatasets(question: string): Promise<Dataset[] | null> {
```

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -50,6 +50,25 @@ async function getClimateDataset(datasetId: string): Promise<Dataset | null> {
...
+): Promise<string | null> {
+	const url = `/climatedata/queries/subset-esgf/${esgfId}?parent-dataset-id=${parentDatasetId}&envelope=${envelope}`;
+	if (timestamps) url.concat(`&timestamps=${timestamps}`);
+	if (thinFactor) url.concat(`&thin-factor=${thinFactor}`);
+
+	const response = await API.get(url).catch((error) => {
+		logger.error(
+			`Error: climate data service was not able to retrieve the subset of the dataset ${esgfId} ${error}`
+		);
+	});
```
YohannParis, line 71: Errors are handled by `api/api.ts` and displays whatever the controllers sends back.
Here you can only handle `2XX` HTTP message. So instead test that is not a 200.
```suggestion
	const response = await API.get(url);
	if (response.status !== 200 || isEmtpy(response.data)) {
		logger.error(`Climate-data service was not able to retrieve the subset of the dataset ${esgfId}`, error);
		return null
	}
```

packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
```diff
@@ -0,0 +1,398 @@
+<template>
+	<tera-drilldown :title="node.displayName" @on-close-clicked="emit('close')">
+		<div :tabName="SubsetDataTabs.Wizard">
+			<tera-drilldown-section>
+				<!-- <InputText placeholder="What do you want to do?" /> -->
```
YohannParis, line 5:
```suggestion
```

packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
```diff
@@ -0,0 +1,398 @@
...
+				<div class="number-entry">
+					<label class="float-label">Keep every nth time slice</label>
+					<InputNumber v-model="timeSkipping" :disabled="!isTimeSkipping" />
+				</div>
+				<template #footer>
+					<!--FIXME: A lot of these drilldowns have this margin auto for the left footer's button.
+						We should make it so the left footer is aligned left and the right
+						footer is aligned right-->
+					<Button
+						:style="{ marginRight: 'auto' }"
```
YohannParis, line 54:
```suggestion
						class="mr-auto"
```

packages/client/hmi-client/src/workflow/ops/subset-data/subset-data-operation.ts
```diff
@@ -0,0 +1,42 @@
...
+	isSpatialSkipping: boolean;
+	spatialSkipping: number | null;
+	isTimeSkipping: boolean;
+	timeSkipping: number | null;
+	notebookSessionId?: string;
+}
+
+export const SubsetDataOperation: Operation = {
+	name: WorkflowOperationTypes.SUBSET_DATA,
+	displayName: 'Subset data',
```
YohannParis, line 20:
```suggestion
	displayName: 'Subset dataset',
```

YohannParis (approved):
Nice job 🎉

packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
```diff
@@ -0,0 +1,396 @@
...
+		isSubsetLoading.value = true;
+
+		const subsetId = await getClimateSubsetId(
+			dataset.value.esgfId,
+			dataset.value.id,
+			`${longitudeStart.value},${longitudeEnd.value},${latitudeStart.value},${latitudeEnd.value}`,
+			`${fromDate.value.toISOString()},${toDate.value.toISOString()}`
+			// isSpatialSkipping.value ? spatialSkipping.value ?? undefined : undefined // Not sure if its this or timeSkipping
+		);
+		isSubsetLoading.value = false;
```
bigglesandginger, line 291: How does this deal with the "it takes 30 minutes to generate a new subset" thing?  I was not seeing an looping (and arguably we shouldn't have any (Yohann, Kevin, and I have a neat messaging thing almost perfect)) but what happens when the UI get yet another `202` response from the server, and needs to check back in later?

packages/client/hmi-client/src/services/dataset.ts
dvince2: I vote on the original code there - `response.status` can _only_ be a 2XX in this case because if it was a 4XX or 5XX error `api.ts` would catch the error, so this new setup wouldn't do much.

With the catch on the error like how it is it matches how we do it everywhere else, and deals with the off chance that api.ts runs into an issue.

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
shawnyama: It doesn't download it - it straight up doesn't work since it expects csvs or something

packages/client/hmi-client/src/services/dataset.ts
YohannParis: The new code catch non 200 status, like a `204 No Content`, or `202 Accepted`, or that the response.data is null.

packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
shawnyama: After asking Satchel it seems like the UI doesn't match with how the climate data service handles this.

If you look at where it says `thin-factor` a bit below here https://github.com/DARPA-ASKEM/climate-data?tab=readme-ov-file#preview

It only accepts one `n` value and then you can exclude certain fields to be effected using the `thin-fields` param (this param isn't in our call yet).

Maybe for this pass we can ignore this?

CC @dvince2 @bigglesandginger

packages/client/hmi-client/src/services/dataset.ts
shawnyama: A human searching something isn't always going to be phrased as a question though

YohannParis: Not a question in the english grammar sense, but a query in english

packages/client/hmi-client/src/workflow/ops/subset-data/tera-subset-data.vue
YohannParis: We will work on that later, as long as the subset result is saved as a dataset for now.

shawnyama: This is actually important - the first Run attempt won't work for other users. It works for me since I was using the same numbers.

PR #2626, opened 2024-01-30
Fix image and types

kbirk:


packages/client/hmi-client/src/types/Types.ts
```diff
@@ -1148,6 +1157,15 @@ export enum EvaluationScenarioStatus {
     Stopped = "STOPPED",
 }
 
+export enum TaskStatus {
```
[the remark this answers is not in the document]
dgauldie: well.... similar fields.... but worth combining

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -710,6 +711,14 @@ export interface TimeSpan {
...
 }
 
+export interface TaskResponse {
+    id: string;
+    script: string;
+    status: TaskStatus;
+    output: any;
+    additionalProperties: any;
+}
+
```
chris-dickson, line 721: If these are client-only types they need to be put into their own files.  This is generated and the build will fail if it doesn't have a matching java type.

chris-dickson (changes requested):

packages/client/hmi-client/src/types/Types.ts
kbirk: Those were generated

packages/client/hmi-client/src/types/Types.ts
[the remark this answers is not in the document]
kbirk: Made an issue for https://github.com/DARPA-ASKEM/terarium/issues/2627 . Doing it properly so that tasks are guarded against invalid states to prevent race conditions will take time

dgauldie (approved):

chris-dickson (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/Model.java
```diff
@@ -32,6 +31,7 @@ public class Model extends TerariumAsset implements SupportAdditionalProperties,
 
 	private Map<String, Object> model;
 
+	@TSOptional
 	private String modelCard;
```
YohannParis, line 35: the `modelCard` is part of the `metadata` property, we do not need a new property.

PR #2629, opened 2024-01-30
fix: ensure generated type linting occurs on server changes

chris-dickson:
Previously, the generated type linting was occuring with in the client linting step.  Since this only happens when a file is changed in the `packages/client/**` path, it was possible to modify a server file, have the branch build pass only to fail in main.  

This PR ensures that branch builds will fail that modify a `@TSModel` annotated entity in Java and do not update the Types.ts file.

### DIFF
--- a/.github/workflows/lint-types.yaml
+++ b/.github/workflows/lint-types.yaml
@@ -8,8 +8,8 @@ on:
     branches: ['main']
   pull_request:
     paths:
-        - 'packages/client/**'
-        - 'packages/server/**'
+      - 'packages/client/**'
+      - 'packages/server/**'
     branches: ['main']
 jobs:
   lintTypes:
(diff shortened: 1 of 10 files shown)


.github/workflows/lint-types.yaml
```diff
@@ -0,0 +1,40 @@
+name: Lint Generated Types
+on:
+  workflow_call:
+  push:
+    paths:
+      - 'packages/client/**'
```
Nithos, line 6: do you still need/want this to be triggered when only client changes are made? I think it may be sufficient to just trigger on server changes no?

chris-dickson: It's to capture the case of modifying the types file manually but it's probably better to just specify the exact path of that file instead of all client files.

Nithos (approved):

PR #2639, opened 2024-01-31
feat: call amr to template and add flat view

shawnyama:
# Description

https://github.com/DARPA-ASKEM/terarium/assets/28237258/ff0be308-d6af-494f-8d5d-9cf8cf3c1fce

- call `amr_to_template` from beaker
- flat view and flattened model state are available
- rearranged sidebar layout

### Testing
Put a valid model id into `model-template-test.vue` and go to `http://localhost:8080/model-template-test`

### DIFF
--- a/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
+++ b/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
@@ -388,7 +388,7 @@ watch(
 				x: 100,
 				y: 100
 			};
-			console.log(flattenedModel);
+
 			modelTemplatingService.addCard(flattenedTemplates.value, flattenedModel);
 
 			// Initialize beaker kernel
--- a/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
+++ b/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
@@ -144,12 +144,8 @@ let canvasTransform = { x: 0, y: 0, k: 1 };
 let isMouseOverPort = false;
 let junctionIdForNewEdge: string | null = null;
 
-const decomposedTemplates = ref<ModelTemplates>(
-	modelTemplatingService.initializeModelTemplates(props.model)
-);
-const flattenedTemplates = ref<ModelTemplates>(
-	modelTemplatingService.initializeModelTemplates(props.model)
-);
+const decomposedTemplates = ref<ModelTemplates>(modelTemplatingService.initializeModelTemplates());
+const flattenedTemplates = ref<ModelTemplates>(modelTemplatingService.initializeModelTemplates());
 const modelFormatOptions = ref([EditorFormat.Decomposed, EditorFormat.Flattened]);
 const currentModelFormat = ref(EditorFormat.Decomposed);
 
--- a/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
+++ b/packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
@@ -114,7 +114,7 @@ import * as d3 from 'd3';
 import type { Position } from '@/types/common';
 import type { Model } from '@/types/Types';
 import type {
-	ModelTemplateEditor,
+	ModelTemplates,
 	ModelTemplateCard,
 	ModelTemplateJunction
 } from '@/types/model-templating';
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/model-template/tera-model-template.vue
```diff
@@ -81,7 +81,7 @@ const cardWidth = computed(() => cardRef.value?.clientWidth ?? 0);
 const card = computed<ModelTemplateCard>(
 	() =>
 		props.model.metadata.templateCard ?? {
-			id: -1,
+			id: '',
```
mwdchang, line 84: What is this `model.metadata.templateCard` business? Do we need it? If not it might be nice to clean up early before things get really gnarly with the junctions logic.

mwdchang: I see ... let's have a quick chat when you have sometime. I think the name-spaces are getting a bit confusing with a lot of model and model-templates floating about. Let's see if we can make things a little easier to read.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -108,29 +119,49 @@ import type {
...
 let isMouseOverPort = false;
 let junctionIdForNewEdge: string | null = null;
 
 const modelTemplateEditor = ref<ModelTemplateEditor>(
-	modelTemplatingService.emptyModelTemplateEditor()
+	modelTemplatingService.initializeModelTemplateEditor(props.model)
+);
+const flatModelTemplateEditor = ref<ModelTemplateEditor>(
+	modelTemplatingService.initializeModelTemplateEditor(props.model)
 );
```
mwdchang, line 152: This initialization section is confusing. You have a watcher below that watches for `props.model` changes and react to it, yet your two editors seem to assume that props.model does not change and do no re-initialize. 

Also your editor construction accepts nulls, I think this is a bit of a paralysis by so many different options and could become error prone down the line. I think we should tighten things up so the behaviours can be inferred by other people in a more straightforward manner.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -108,29 +119,49 @@ import type {
...
 
-defineProps<{
+const props = defineProps<{
 	model?: Model;
 }>();
 
+enum EditorView {
+	Template = 'Template',
+	Flat = 'Flat'
+}
```
mwdchang, line 137: It isn't so much that we have template-view vs flat-view; they are both templates, just one is flat and one is not. 

Maybe it is better to call them `Decomposed` and `Flattened` instead? Any thoughts?

Also you are calling the enum EditorView, but you are using them where we are setting "formats"....so perhaps we should rename the enum to EditorFormat to match, or vice versa.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -108,29 +119,49 @@ import type {
...
 
 const modelTemplateEditor = ref<ModelTemplateEditor>(
-	modelTemplatingService.emptyModelTemplateEditor()
+	modelTemplatingService.initializeModelTemplateEditor(props.model)
+);
+const flatModelTemplateEditor = ref<ModelTemplateEditor>(
+	modelTemplatingService.initializeModelTemplateEditor(props.model)
 );
+const modelFormatOptions = ref([EditorView.Template, EditorView.Flat]);
+const modelFormat = ref(EditorView.Template);
```
mwdchang, line 154: I would suggest modelFormat => currentModelFormat. to be a bit more consistent.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -325,25 +356,94 @@ function mouseUpdate(event: MouseEvent) {
...
+			// Initialize beaker kernel
+			try {
+				const context = {
+					context: 'mira_model',
+					language: 'python3',
+					context_info: {
+						id: props.model.id
+					}
+				};
+				await kernelManager.init('beaker_kernel', 'Beaker Kernel', context);
```
mwdchang, line 407: If you anticipate the possibility of re-initializing the kernelManager, you should calling a "shutdown" before

packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -20,15 +21,18 @@ export const modelTemplateOptions = [
...
-	};
+	} as ModelTemplateCard;
 	return modelTemplate;
 });
 
-export function emptyModelTemplateEditor(name = 'name', description = '') {
+export function initializeModelTemplateEditor(model: Model | null = null) {
+	const name = model?.header?.name ?? 'name';
+	const description = model?.header?.description ?? 'description';
+
```
mwdchang, line 35: Calling something an `Editor` suggest that it performs some actions that you can invoke, this is just a data structure. For example we didn't call "Workflow" a "WorkflowEditor" because we choose to let the WorkflowService perform the actions rather than the Workflow object itself.

Are you planning to add functionalities directly into this object?

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
shawnyama: Yeah, those enum names make sense. The designs just had them as `Template` and `Flat`. Will apply these.

packages/client/hmi-client/src/services/model-templating.ts
shawnyama: Not planning to add functionality I was just having trouble naming it. I changed it to `ModelTemplates` now

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
shawnyama: Thanks, I added the watcher at the end so I forgot to update these refs

mwdchang (approved):

PR #2652, opened 2024-02-02
feat: reflect model template actions in mira

shawnyama:
# Description
- Add templates, add junctions/edges are reflected in beaker
- Code returned from beaker must be run in the notebook in order to apply changes to the model
- Every template that's added from the sidebar is given a number appended to all of its model variables
- Adding junctions from `flattenedToDecomposed` kinda works just need to use port position instead of the card

https://github.com/DARPA-ASKEM/terarium/assets/28237258/d5edd640-3b07-4462-9412-edaabd3d2aeb


packages/client/hmi-client/vite.config.local.ts
```diff
@@ -14,13 +14,13 @@ config.server.proxy = {
...
 	'^/beaker_ws': {
-		target: `ws://${localhost}:3050`,
+		target: `ws://${localhost}:8888`,
 		rewrite: (path_str) => path_str.replace(/^\/beaker_ws/, ''),
 		changeOrigin: true,
 		ws: true
 	},
 	'^/beaker': {
-		target: `http://${localhost}:3050`,
+		target: `http://${localhost}:8888`,
```
mwdchang, line 23: Just a reminder do not commit your local/personal configuration settings.

Tom-Szendrey:
Cant see model until i drag on the graph?

mwdchang:
@shawnyama I don't think being able to zoom and drag the template-nodes in the cards are desirable behaviours.

I think we turned them off with pointer-events:none in the workflow-nodes, alternatively I think you can use useZoom:false when the renderer is initialized.

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -352,33 +379,35 @@ function mouseUpdate(event: MouseEvent) {
...
-	});
+	try {
+		const context = {
+			context: 'mira_model_edit',
+			language: 'python3',
+			context_info: {
+				id: isModelLoaded ? props.model?.id ?? '' : ''
+			}
+		};
+		await kernelManager.init('beaker_kernel', 'Beaker Kernel', context);
```
mwdchang, line 393: Something looks a bit off here, it seems like you are opening up far more web-socket connections to beaker than it is necessary? How many do you actually need?

image

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
```diff
@@ -352,33 +379,35 @@ function mouseUpdate(event: MouseEvent) {
...
+	}
 }
 
 watch(
 	() => props.model,
-	async () => {
+	() => {
 		if (props.model) {
 			// Create flattened view of model
 			const flattenedModel: any = cloneDeep(props.model);
```
mwdchang, line 413: Sorry not sure where is the best place to add this observation:

To what extend does is the flattened-view editing functional? I can drag templates onto the flattened view, but the templates disappear, however, if I switch to the decomposed view the templates I had created would be there.

packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -79,16 +77,223 @@ export function junctionCleanUp(modelTemplates: ModelTemplates) {
...
+	modelTemplate: any
+) {
+	const { name } = modelTemplate.header;
+
+	// If a decomposed card is added, add it to the kernel
+	if (Object.values(DecomposedModelTemplateTypes).includes(name)) {
+		const addTemplateArguments: AddTemplateArguments = {};
+
+		// Append a number to the model variables to avoid conflicts
+		modelTemplate = appendNumberToModelVariables(modelTemplate, modelTemplates.models.length);
```
mwdchang, line 140: I don't think this will always work if you are basing it on `modelTemplates.models.length`, because through the additional and removal of template actions the length-value will get repeated. 

Having said that, we can put a giant FIXME on this and address it subsequently. As it also partially plays into whether we want to run immediately or bach stuff up in the notebook.

packages/client/hmi-client/src/components/llm/tera-notebook-jupyter-input.vue
```diff
@@ -28,7 +28,7 @@ const props = defineProps<{
 	contextInfo: any;
 }>();
 
-const emit = defineEmits(['llm-output']);
+const emit = defineEmits(['output-code']);
```
mwdchang, line 31: Any particular reason for the name change? I was thinking this may not necessarily be outputting code based on the props passed in?

packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -79,16 +77,223 @@ export function junctionCleanUp(modelTemplates: ModelTemplates) {
...
+			outputCode(d);
+
+			model.header.name = name;
+			model.metadata.templateCard.name = name;
+			if (model.model.transitions[0] && model.semantics.ode.rates[0]) {
+				model.model.transitions[0].id = name;
+				model.model.transitions[0].properties.name = name;
+				model.semantics.ode.rates[0].target = name;
+			}
+		});
```
mwdchang, line 241: The way you are using and passing around the kernel manager, now that there are most instance of its usage, is a bit odd. It feels like we should almost invert the pattern because the kernel manager is the wrapper. 

e.g. the pattern I am seeing is:
```
function blah (a, b, c, kernelManager) {
  kernelManagerr.sendMessage(...).on('some_response', () => {
    // do stuff with a, b, c
  });
```
If may be easier to separate responsibilities, and easier to test, if we do

```
function handleResult(a, b, c) { .... }

kernelManager
  .sendMessage(...)
  .on('some_response', () => {
    handleResult( ... )
  });
```

Food for thought, let me know what you think and we should have a chat about it before/if you actually start to change things.

packages/client/hmi-client/src/components/llm/tera-notebook-jupyter-input.vue
Tom-Szendrey: The intention of this is to have LLM output. Currently this is only registered to `code_cell` responses but thats mostly because i still need to work on this component. 
It should be llm-output IMO

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
shawnyama: I somehow wasn't thinking about editing the flattened view, just made a fix so cards are added in there properly. When cards are linked in the flattened view it's supposed to merge the cards once they are connected.

packages/client/hmi-client/src/components/llm/tera-notebook-jupyter-input.vue
shawnyama: I was making the passed emit names consistent in `tera-model-edit`. I guess they can be different though

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
shawnyama: Now that I put this in tera-model-edit, I'll use the one created there instead of creating a new one. Not exactly sure why there would be more than 2 since I am shutting down the kernel if I want to rerun it. Will look into this

shawnyama: Only one session is used within `tera-model-edit` now

packages/client/hmi-client/src/components/model-template/tera-model-template-editor.vue
shawnyama: ^ adding to flattened template should be its own new issue

packages/client/hmi-client/src/services/model-templating.ts
mwdchang: Will handle in subsequent PRs

mwdchang (approved):

Issue #2638, opened 2024-01-31
[TASK]: Support Decapodes AMR


YohannParis:
**Describe the task**

We need to support _Decapodes_ [AMR](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/decapodes/json/amr_schema.json) in the data-service.

**Task Items**
- [ ] Create new model `DecapodesContext.java` 
```Java
// Just be a wrapper around a JSON blob.
private ModelHeader header;
private Object context;
@JsonProperty("parent_model")
private UUID modelId;
```
- [ ] Create new model `DecapodesConfiguration.java`
```Java
// Just be a wrapper around a JSON blob.
private ModelHeader header;
private Object configuration;
@JsonProperty("parent_context")
private UUID decapodesContextId;
```
- [ ] `/api/models/decapodes` endpoint to generate a _Decapodes_ AMR **Model**
[example of a JSON output](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/decapodes/examples/grigoriev/model.json)
```JS
{
   "header": { 
      "id": UUID,
      // ...all properties of 'header'
    },
   "model": // ...all properties of 'model'
}
```

- [ ] `/api/model-contexts/decapodes/{id}` endpoint to generate a _Decapodes_ AMR **Context**
[example of a JSON output](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/decapodes/examples/grigoriev/context.json)
```JS
{
   "header": { 
      "id": UUID,
      "parent_model": UUID,
      // ...all properties of 'header'
    },
   "context": // ...all properties of 'context'
}
```

- [ ] `/api/model-contexts/decapodes/{id}` endpoint to generate a _Decapodes_ AMR **Configuration**
[example of a JSON output](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/decapodes/examples/grigoriev/configuration.json)
```JS
{
   "header": { 
      "id": UUID,
      "parent_context": UUID,
      // ...all properties of 'header'
    },
   "configuration": // ...all properties of 'configuration'
}
```

- [ ] `/api/model-combined/decapodes/{DecapodesConfigurationId}` endpoint to generate a _Decapodes_ AMR **Combined** 
[example of a JSON output](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/decapodes/examples/grigoriev/combined.json)
This one needs to combine a Model, DecapodesContext, and DecapodesConfiguration in one result.
[...]

PR #2653, opened 2024-02-04
chore: decapodes data structures

mwdchang:
### Summary
Add objects for decapode configuration and decapodes context and decapodes configuration.

Some fields are captured as Object, since the fields captures disjunctive typing. e.g. `{ value: 123, format: 'x' }` or `{ uri: 'abc' }` or `8` ... they can be typed out more carefully if needed, but most of these objects are just pass-throughs.

### Testing
CRUD operations of configuration/context through swagger-ui.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ConfigurationDataset.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ConfigurationDataset.java
@@ -15,6 +15,7 @@ import java.io.Serializable;
 @Accessors(chain = true)
 @TSModel
 public class ConfigurationDataset implements Serializable {
+	private String _type;
 	private String type;
 	private String name;
 	private String description;
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ConfigurationDataset.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ConfigurationDataset.java
@@ -1,0 +1,23 @@
+package software.uncharted.terarium.hmiserver.models.dataservice.multiphysics;
+
+import java.util.List;
+import java.lang.Number;
+
+import jakarta.persistence.*;
+import com.fasterxml.jackson.annotation.JsonAlias;
+import lombok.Data;
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import lombok.experimental.Accessors;
+import java.io.Serializable;
+
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class ConfigurationDataset implements Serializable {
+	private String type;
+	private String name;
+	private String description;
+
+	private ConfigurationDatasetFile file;
+}
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
@@ -1,12 +1,10 @@
 package software.uncharted.terarium.hmiserver.models.dataservice.multiphysics;
 
-import jakarta.persistence.*;
 import com.fasterxml.jackson.annotation.JsonAlias;
 import lombok.Data;
 import lombok.experimental.Accessors;
 import software.uncharted.terarium.hmiserver.annotations.TSModel;
 import java.io.Serializable;
-import java.util.Map;
 
 @Data
 @Accessors(chain = true)
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
@@ -11,11 +11,11 @@ import java.util.Map;
 @Data
 @Accessors(chain = true)
 @TSModel
-public class ConfigurationHeader implements Serializable {
+public class ContextHeader implements Serializable {
 	private String id;
 	private String description;
 	private String name;
 
-	@JsonAlias("parent_context")
+	@JsonAlias("parent_model")
 	private String parentContext;
 }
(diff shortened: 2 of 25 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
```diff
@@ -0,0 +1,19 @@
...
+@Data
+@Accessors(chain = true)
+@TSModel
+public class ContextHeader implements Serializable {
+	private String id;
+	private String description;
+	private String name;
+
+	@JsonAlias("parent_model")
+	private String parentContext;
```
YohannParis, line 18: This should be `parentModel`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesContext.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.experimental.Accessors;
+
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import software.uncharted.terarium.hmiserver.models.dataservice.TerariumAsset;
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class DecapodesContext extends TerariumAsset implements Serializable {
+	private ContextHeader header;
```
YohannParis, line 15: Because you are using `TerariumAsset` you do not need a `ContextHeader`.
`TerariumAsset` provides an `id, name, description`, only the `parentModel` is missing, that's why I'm using the `ModelHeader` in ticket #2638
```java
private ModelHeader header;
private Object context;
@JsonProperty("parent_model")
private UUID parentModel;
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesConfiguration.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.experimental.Accessors;
+
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import software.uncharted.terarium.hmiserver.models.dataservice.TerariumAsset;
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class DecapodesConfiguration extends TerariumAsset implements Serializable {
+	private ConfigurationHeader header;
```
YohannParis, line 14: Because you are using `TerariumAsset` you do not need a `ConfigurationHeader`.
`TerariumAsset` provides an `id, name, description`, only the `parentContext` is missing, that's why I'm using the `ModelHeader` in ticket #2638
```java
private ModelHeader header;
private Object configuration;
@JsonProperty("parent_context")
private UUID parentContext;
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesContext.java
```diff
@@ -0,0 +1,17 @@
...
+	private Context context;
```
YohannParis, line 16: Do we need to define those so perfectly?
I do not mind, but this can change at any point and will be quite difficult to support. Same thing for `configuration`.
Maybe we should add support for the `modelVersion` in the `ModelHeader`. But so close to the hackathon...

YohannParis (commented):
LGTM, but still not sure about not using `ModelHeader` here.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextPrimalDualRelation.java
```diff
@@ -0,0 +1,16 @@
...
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import java.io.Serializable;
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class ContextPrimalDualRelation implements Serializable {
+	private String primal;
+	private String dual;
+	private Object method;
```
dvince2, line 15: FWIW if these are typed as `JsonNode` it will still go to the front end as `any` but give us the option of extracting fields out of the Json in the back end without having to do strict typing.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ConfigurationDataset.java
```diff
@@ -0,0 +1,24 @@
...
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+import lombok.experimental.Accessors;
+import java.io.Serializable;
+
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class ConfigurationDataset implements Serializable {
+	private String _type;
```
dvince2, line 18: Out of curiousity, why does this use `_type` and `type`?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesContext.java
mwdchang: It needs to be `header.parent_model`, so as far as I am aware, without using a DSL and a generator like intertypes to do selective unions, we would either need to 
- extend ModelHeader and have extra fields we dont need, e.g. "extracted_from", or
- create new class with exactly the field that were prescribed

I went with the 2nd option.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesContext.java
mwdchang: Switching to Object/JsonNode would be trivial.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/ContextHeader.java
mwdchang: done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/multiphysics/DecapodesContext.java
mwdchang: I think if we want to have extra fields that may be okay, but if want to change the data structure that can be quite annoying as there may be schema-validity checkers downstream, so either way we need to adhere to the agreed upon schema.

YohannParis: Okay, I personally prefer to have the controller do that job, but the model will do.

PR #2659, opened 2024-02-05
ciemss simulate using inferred_parameters

mwdchang:
### Summary

Depends on: https://github.com/DARPA-ASKEM/terarium/pull/2669

Part 2 of of updating pyciemss calibrate. Lots of smaller things
- Add optional input-port for ciemss simulate, allow ciemss simulate to use inferred_parameters
- Removed ciemss extra operators, Sabina said they are largely not needed
- Fix client-event subscriptions for ciemss
- Fix MQ erroring out on having messages in the queue but no consumers attached
- Add loss graph, changed y-axis to use scientific notation
- Fix simulate-chart jitters via unnecessary rerenders

Remaining issues
- Loss values not persisted
- You need to refresh on the workflow view to get the preview to show up
- CIEMSS loss values are extremely high, not sure if there is some weird bug somewhere

CIEMSS calibrate drilldown:

image

CIEMSS simulate with inferred_parameters:

image


packages/client/hmi-client/src/workflow/tera-simulate-chart.vue
```diff
@@ -40,8 +46,14 @@ const props = defineProps<{
 	initialData?: CsvAsset;
 	mapping?: { [key: string]: string }[];
 	runType?: RunType;
+	size?: { width: number; height: number };
 }>();
 
+const chartSize = computed(() => {
+	if (props.size) return props.size;
+	return { width: 390, height: 190 };
+});
```
YohannParis, line 55: Why not a `reactive` which could make sure both values are set and have a default?

```suggestion
const chartSize = reactive({
    width: props.size?.width ?? 390,
    height: props.size?.height ?? 190
});
```

YohannParis (approved):
LGTM 🎉 

I would have prefer removing the use of `ClientEvent` but that should be a different PR.

packages/client/hmi-client/src/workflow/tera-simulate-chart.vue
mwdchang: I am a bit confused - `reactive` does not guarantee whether values should exist or not. But I want to keep sizing static where possible, for now, because that was partial to a lot of graph-jittering problems with chart.js.

YohannParis: Reactive was just a suggestion, no need to use it.
I wanted to say, why not testing that both value are set properly while at it?

```suggestion
const chartSize = computed(() => {
	return { 
		width: props?.size?.width ?? 390, 
		height: props?.size?.height ?? 190 
	};
});
```

this is a trivial question, not worth spending time about it. Sorry for the confusion.

PR #2662, opened 2024-02-06
gollm integration

blanchco:
# Description

* adding support to generate gollm model card
* data structures seem to be inconsistent when retrieiving data from data card so display of the card is based on keys and a little ugly for now
* dropdown to switch between ta1(skema) and ta4(gollm) card generation services
* added ability to add dropdown and other header actions in the drilldown header\

**NOTE KM does not work locally, so this isn't fully tested


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/services/DownloadService.java
```diff
@@ -55,7 +55,7 @@ public static Boolean IsPdf(final byte[] data) {
 
 		// for pdf version 1.3
 		// check if data ends with %%EOF
-		if (data[5] == 0x31 && data[6] == 0x2E && data[7] == 0x33) {
+		if (data[4] == 0x31 && data[5] == 0x2E && data[6] == 0x33) {
```
blanchco, line 58: apparently this is correct for checking for pdf, this was incorrect before and was flooding the document-assets with large text objects

dvince2: wow :o

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -7,8 +7,11 @@
 				:tooltip="tooltip"
 				@tab-change="handleTabChange"
 				@close="emit('on-close-clicked')"
-				>{{ props.title }}</tera-drilldown-header
-			>
+				>{{ props.title }}
+				<template #action-row>
+					<slot name="header-action-row"></slot>
```
YohannParis, line 12:
```suggestion
					<slot name="header-action-row" />
```

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,53 @@
+import API from '@/api/api';
+import useAuthStore from '@/stores/auth';
+import { TaskStatus } from '@/types/Types';
+import { logger } from '@/utils/logger';
+import { EventSourceMessage, fetchEventSource } from '@microsoft/fetch-event-source';
+
+export async function modelCard(documentId: string, modelId: string): Promise<any> {
```
YohannParis, line 7: Add some JSDoc please.
And handle all the HTTP responses available for the endpoint.
Also add a try, catch.

YohannParis, line 7:
```suggestion
export async function modelCard(documentId: string, modelId: string): Promise<TaskResponse> {
```

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,53 @@
...
+		params: {
+			'document-id': documentId,
+			'model-id': modelId
+		}
+	});
+
+	return response.data ?? null;
+}
+
+export async function handleTaskById(id: string, successHandler: (message: any) => void) {
```
YohannParis, line 18: Could you refactor most of this function in the the API.ts utils `/src/api/api.ts` to be reuse easily by other SSE listener.

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,53 @@
...
+			console.log(message);
+			const data = JSON.parse(message.data);
+			if (data?.status === TaskStatus.Failed) {
+				logger.error('Failed to generate model card.');
+				controller.abort();
+			}
+
+			if (data.status === TaskStatus.Success) {
+				successHandler(data);
+				console.log('Aborting connection on success');
```
YohannParis, line 35:
```suggestion
				logger.info('Aborting connection on success', { showToast: false });
```

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,53 @@
...
+				controller.abort();
+			}
+		},
+		async onopen(response: Response) {
+			console.log('open');
+			console.log(response);
+		},
+		onerror(error: Error) {
+			throw error;
```
YohannParis, line 44: Adding an logger.error here would be great.

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -54,40 +65,41 @@
...
 					/></span>
-					<Button
-						:disabled="isProcessing || allCodeBlocks.length === 0"
-						label="Run"
-						icon="pi pi-play"
-						severity="secondary"
-						outlined
-						@click="handleCode"
-					/>
+					<span style="margin-right: auto">
```
YohannParis, line 76: If you have to do it, use the primevue classes;
https://primeflex.org/margin#classes

```suggestion
					<span class="mr-auto">
```

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -54,40 +65,41 @@
...
+						<label>Service</label>
+						<Dropdown
+							size="small"
+							v-model="clonedState.modelService"
+							:options="modelServices"
+							@change="emit('update-state', clonedState)"
+						></Dropdown>
```
YohannParis, line 83:
```suggestion
						/>
```

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
```diff
@@ -469,6 +503,24 @@ function isSaveModelDisabled(): boolean {
 	return !selectedModel.value || !!activeProjectModelIds?.includes(selectedModel.value.id);
 }
 
+async function generateModelCard(docId, modelId) {
```
YohannParis, line 506: Should be in a service... with JSDoc

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,53 @@
...
+import { logger } from '@/utils/logger';
+import { EventSourceMessage, fetchEventSource } from '@microsoft/fetch-event-source';
+
+export async function modelCard(documentId: string, modelId: string): Promise<any> {
+	const response = await API.post('/gollm/model_card', null, {
+		params: {
+			'document-id': documentId,
+			'model-id': modelId
+		}
+	});
```
mwdchang, line 13: What is with the `null` parameter in your POST request?

packages/client/hmi-client/src/workflow/ops/model-from-code/tera-model-from-code.vue
blanchco: 😈

packages/client/hmi-client/src/services/goLLM.ts
blanchco: I thought that represented the body for the post request but its not needed, Ive removed it

packages/client/hmi-client/src/services/goLLM.ts
blanchco: I think we need to refactor how we handle errors in responses, they are being intercepted and we are not able to handle the  response errors in the API call, instead we just get a null response.  I written the logic here for now, but I think this needs to be improved in its own ticket, happy to discuss.

packages/client/hmi-client/src/services/goLLM.ts
blanchco: Oh nevermind, it is needed...ya the 2nd parameter represents the body for POST, but there is no body

packages/client/hmi-client/src/api/api.ts
```diff
@@ -181,4 +186,82 @@ export class Poller<T> {
...
+	const fetchEventSourceOptions: FetchEventSourceInit = {
+		...options,
+		headers: {
+			...options?.headers,
+			Authorization: `Bearer ${authStore.token}`
+		},
+		onmessage(message: EventSourceMessage) {
+			const data = message?.data;
+			const abort = () => controller.abort();
+			handlers.onMessage(data, abort);
```
YohannParis, line 237: Why use CamelCase for the handlers method, if the goal is to replicate the SSE ones?

YohannParis (approved):

packages/client/hmi-client/src/api/api.ts
blanchco: there wasn't really intention behind it Ive changed it to flatcase

packages/client/hmi-client/src/api/api.ts
```diff
@@ -181,4 +186,135 @@ export class Poller<T> {
...
+		const authStore = useAuthStore();
+		const fetchEventSourceOptions: FetchEventSourceInit = {
+			...this.options,
+			headers: {
+				...this.options?.headers,
+				Authorization: `Bearer ${authStore.token}`
+			},
+			onmessage: (message: EventSourceMessage) => {
+				const data = message?.data;
+				this.handlers.onmessage(data);
```
YohannParis, line 294: This might be better called `ondata()`

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Fetches model card data from the server.
+ * @param {string} documentId - The document ID.
+ * @param {string} modelId - The model ID.
+ * @returns {Promise<TaskResponse|null>} A promise resolving to the task response or null if unsuccessful.
+ */
+export async function modelCard(documentId: string, modelId: string): Promise<TaskResponse | null> {
+	try {
+		const response = await API.post('/gollm/model_card', null, {
```
YohannParis, line 14: We use `kebab-case` for URL, could you please update the back-end controller as well.
```suggestion
		const response = await API.post('/gollm/model-card', null, {
```

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,63 @@
...
+					break;
+				}
+				case 500:
+				default: {
+					logger.error('There was an issue dispatching the request');
+					break;
+				}
+			}
+			return null;
+		}
```
YohannParis, line 35: You are right those should be part of the API interceptors, feel free to remove it. we can table that discussion another day.

packages/client/hmi-client/src/api/api.ts
```diff
@@ -181,4 +186,135 @@ export class Poller<T> {
...
+	private status: SSEStatus;
+
+	/**
+	 * Create a SSEHandler.
+	 * @param {string} url - The URL to connect to.
+	 * @param {SSEHandlers} handlers - The handlers for the SSE events.
+	 * @param {FetchEventSourceInit} options - The options for the fetchEventSource function.
+	 */
+	constructor(url: string, handlers: SSEHandlers, options?: FetchEventSourceInit) {
+		this.url = url;
```
YohannParis, line 246: Could you update this to prepand `/api` to the URL like we do for the API?

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,63 @@
...
+	} catch (err) {
+		logger.error(err);
+		return null;
+	}
+}
+
+/**
+ * Handles SSE connection for a given task ID.
+ * @param {string} id - The task ID.
+ * @param {(message: any) => void} successHandler - The success handler function.
```
YohannParis, line 46:
```suggestion
 * @param {(data: any) => void} successHandler - The success handler function.
```

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export async function handleTaskById(id: string) {
+	const sseHandler = new SSEHandler(`/api/gollm/${id}`, {
+		onmessage(message) {
+			const data = JSON.parse(message);
```
YohannParis, line 51: Why not parse the `JSON.parse()` on the SSE handler directly.
That's why I would like to be called `ondata()` because it's confusing here of what I'm getting, the message from SSE or the message.data.

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -0,0 +1,63 @@
...
+			if (data?.status === TaskStatus.Failed) {
+				throw Error('Task failed');
```
YohannParis, line 53: You are throwing up an error, but I don't see it handle anyware. I'm worry this is an issue.
You could just log it and
```js
sseHandler.setStatus(SSEStatus.DONE);
sseHandler.closeConnection();
```

packages/client/hmi-client/src/services/model.ts
```diff
@@ -116,9 +118,36 @@ export function validateModelName(name: string): boolean {
...
+}
+
+/**
+ * Generates a model card based on the provided document ID, model ID, and model service type.
+ *
+ * @param {string} documentId - The ID of the document.
+ * @param {string} modelId - The ID of the model.
+ * @param {ModelServiceType} modelServiceType - The type of the model service.
+ *
+ * @returns {Promise<string | null>} The ID of the model if the operation is successful, null otherwise.
```
YohannParis, line 134: This function should not return anything.
We should just display that they were an error that happened via popup.

packages/client/hmi-client/src/services/goLLM.ts
blanchco: fair enough, that makes sense.  Changed accordingly

packages/client/hmi-client/src/services/goLLM.ts
blanchco: throwing an error here will be caught in the onerror function in the SSEHandler (tried and tested).  The error is then logged there.

packages/client/hmi-client/src/services/model.ts
blanchco: Changed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -82,13 +83,14 @@ private TaskResponseHandler getModelCardResponseHandler() {
...
-			log.info(response.toString());
+
+		handler.onRunning((TaskResponse resp) -> {
+			log.info(resp.toString());
 		});
 		return handler;
 	}
 
-	@PostMapping("/model_card")
+	@PostMapping("/model-card")
```
dvince2, line 93: thanks!

blanchco:
merging for now - will make additional improvements along with attach the gollm card to the document asset in the next PR

PR #2665, opened 2024-02-06
Fix: minor fix to model edit dropdown

Tom-Szendrey:
# Description
Missed this emit from previous PR

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
@@ -35,6 +35,7 @@
 				title="Model Preview"
 				v-model:output="selectedOutputId"
 				@update:output="onUpdateOutput"
+				@update:selection="onUpdateSelection"
 				:options="outputs"
 				is-selectable
 			>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
```diff
@@ -295,6 +302,13 @@ const onUpdateOutput = (id: string) => {
 	emit('select-output', id);
 };
 
+const onUpdateSelection = (id) => {
+	const outputPort = _.cloneDeep(props.node.outputs?.find((port) => port.id === id));
+	if (!outputPort) return;
+	outputPort.isSelected = !outputPort?.isSelected;
+	emit('update-output-port', outputPort);
+};
```
YohannParis, line 310: I was thinking about what the `_.cloneDeep()` does if `outputs?.find()` returns `undefined` and it does returns `undefined` as well.
But should it makes more sense to write it this way? Just for the thought, no need to change your code at all.

```suggestion
const onUpdateSelection = (id) => {
	const outputPort = props.node.outputs?.find((port) => port.id === id);
	if (!outputPort) return;
	const selectedOutputPort = _.cloneDeep(outputPort);
	selectedOutputPort.isSelected = !outputPort?.isSelected;
	emit('update-output-port', selectedOutputPort);
};
```

YohannParis (approved):

shawnyama (approved):

packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
Tom-Szendrey: This is a good thought.
I will keep it as it is for now though as this code is repeated in a few nodes so it will be more recognizable and it will be easier to find and refactor out should we want to

PR #2668, opened 2024-02-06
2667 bug unable to update document asset if the id is not in the payload

dgauldie:
# Description

Resolves #(issue)


dgauldie:
the actual fix is line 144 in document controller

dvince2 (approved):

PR #2669, opened 2024-02-06
Fix mq complexeties

### LINKED ISSUE #2666: Fix MQ complexeties

dvince2:
#Description

Removing the more complex per-user queues in hmi-server

Resolves #2666


kbirk:
Heads up, when you have multiple consumers on the same queue, they will round robin messages. So if we ever run multiple instances of the `hmi-server`, it may not route the messages to the correct instance holding the `SseEmitter`. You want to use a unique queue per `hmi-server` that is bound to the same exchange that receives the message, as each queue will receive the message from the bound exchange. 

I've set it up this way in the `TaskService`. It's not pretty, but it works:

```
@RabbitListener(bindings = @QueueBinding(value = @org.springframework.amqp.rabbit.annotation.Queue(autoDelete = "true", exclusive = "false", durable = "${terarium.taskrunner.durable-queues}"), exchange = @Exchange(value = "${terarium.taskrunner.response-exchange}", durable = "${terarium.taskrunner.durable-queues}", autoDelete = "false", type = ExchangeTypes.DIRECT), key = ""))
```

This creates an anonymous queue to the exchange that will receive the message. For some reason I could not get the `@RabbitListener` annotation to work with `exclusive = "true"`. Ideally we want the queues to be exclusive, but it fails to create them if I set that...

dvince2:
@kbirk from our discussion, because this issue will be quite large to fix with logic changes made in the various simulation services etc we're going to shelve it for now. I've added a TODO comment to the suspect areas and opened a new issue [here](https://github.com/DARPA-ASKEM/terarium/issues/2679)

dgauldie (approved):

PR #2670, opened 2024-02-06
Small fixes. Add cache bust to puill latest GoLLM in taskrunner build

kbirk:
The issue @blanchco  was encountering is because the dockerfile layer that pulls the GoLLM repo was getting cached, so it was running code that was out of date :)

Ideally we want to actually hardcode a GoLLM release, but for now, since we are iterating so quickly its nice to just use the latest by default.

### DIFF
--- a/packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/TaskServiceTest.java
+++ b/packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/TaskServiceTest.java
@@ -10,12 +10,14 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.security.test.context.support.WithUserDetails;
 
+import lombok.extern.slf4j.Slf4j;
 import software.uncharted.terarium.hmiserver.TerariumApplicationTests;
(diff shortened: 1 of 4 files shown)


dvince2 (approved):

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/TaskServiceTest.java
```diff
@@ -106,7 +108,7 @@ public void testItCanSendGoLLMModelCardRequest() throws Exception {
 		req.setScript("gollm:model_card");
 		req.setInput(content.getBytes());
 
-		List<TaskResponse> responses = taskService.runTaskBlocking(req);
+		List<TaskResponse> responses = taskService.runTaskBlocking(req, 300);
```
dgauldie, line 111: what this magic number for?

dgauldie (approved):

j2whiting (approved):

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/TaskServiceTest.java
kbirk: seconds, just bumping the test to wait 5 minutes

PR #2674, opened 2024-02-06
add file ext

blanchco:
# Description

* adding file extensions to temp code files

### DIFF
--- a/packages/client/hmi-client/src/utils/data-util.ts
+++ b/packages/client/hmi-client/src/utils/data-util.ts
@@ -2,7 +2,7 @@ import { ResourceType, ResultType } from '@/types/common';
 import { Filters } from '@/types/Filter';
 import { isEmpty } from 'lodash';
 import { FACET_FIELDS as MODEL_FACET_FIELDS } from '@/types/Model';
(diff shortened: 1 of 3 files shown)


dvince2 (approved):

packages/client/hmi-client/src/utils/data-util.ts
```diff
@@ -2,7 +2,7 @@ import { ResourceType, ResultType } from '@/types/common';
 import { Filters } from '@/types/Filter';
 import { isEmpty } from 'lodash';
 import { FACET_FIELDS as MODEL_FACET_FIELDS } from '@/types/Model';
-import type { Model, Document, Dataset, DocumentAsset } from '@/types/Types';
+import { Model, Document, Dataset, DocumentAsset, ProgrammingLanguage } from '@/types/Types';
```
YohannParis, line 5: You should not remove those `type` in front of import.
If what you are importing is an `enum`, then you should add a new line:
```ts
import type { Model, Document, Dataset, DocumentAsset } from '@/types/Types';
import { ProgrammingLanguage } from '@/types/Types';
```

blanchco: oops, I didn't even realize that got removed...

PR #2676, opened 2024-02-07
Add the server url for swagger

YohannParis:
Swagger on _staging_ https://server.staging.terarium.ai/swagger-ui/index.html does not let you authorize with keycloak.
Maybe setting the URL.


dgauldie (changes requested):
we don't need this anymore

PR #2684, opened 2024-02-07
Replaced empty state and loading svgs with Lottie file animations

jamiewaese-uncharted:
Subtle animations... so nice....

The it's using a new-ish format called a Lottie file. Very compact.

If you don't like these animations they're easy to tweak. 

Lottie animation empty seed

Lottie animation loading and empty state


mwdchang (changes requested):
@jamiewaese-uncharted we wouldn't normally have a `package-lock.json` file in the project, since the packages are under the governance of the yarn package manager (via yarn.lock).

Can you please remove package-lock.json file and test if everything still works? If not let us know what were the installation instructions for vue3-lottie and we can setup the dependencies on your behalf.

jamiewaese-uncharted:
Hope this works... 

image

mwdchang (approved):

PR #2689, opened 2024-02-09
chore: house cleaning feb 2024

mwdchang:
### Summary
Clean up petrinet-renderer and model-diagram components: There are a good chunk of editing-related features that are no longer in-use, this will reduce complexity and unnecessary rendering cost.

General clean up, mostly around stratification with typed-systems


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelSemantics.java
```diff
@@ -20,5 +19,5 @@ public class ModelSemantics implements SupportAdditionalProperties {
 	private List<Object> span;
 
 	@TSOptional
-	private TypingSemantics typing;
+	private Object typing;
```
dvince2, line 22: I think we want to keep these `Object` as `JsonNode` just for the fexibility of reading it easy on the back in just in case (it will still translate to `any` for the client)

dvince2 (approved):

YohannParis (approved):

PR #2690, opened 2024-02-09
Add netcdf asset

bigglesandginger:
# Description

Resolves #(issue)


dvince2:
@bigglesandginger You'll need to run `./gradlew generateTypeScript` from the top level directory to generate the types for the frontend.

packages/server/src/main/resources/application-local.properties
```diff
@@ -41,7 +41,7 @@ spicedb.target=localhost:50051
 ########################################################################################################################
 aws-access-key-id=admin
 aws-secret-access-key=admin123
-aws-url=http://minio:9000
+aws-url=http://localhost:9000
```
dvince2, line 44: You'll need to add `minio` to your `/etc/hosts/` file for this. As-is `localhost` may have issues working with docker, etc

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/NetCDFService.java
```diff
@@ -0,0 +1,133 @@
...
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class NetCDFService {
```
dvince2, line 31: We've moved to centralizing all of the service layer code between the controllers and ES, so this class should now just extend `S3BackedAssetService`. As an example, see `DatasetService`.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NetCDFController.java
```diff
@@ -0,0 +1,232 @@
...
+				error);
+		}
+	}
+
+	@PutMapping("/{id}")
+	@Secured(Roles.USER)
+	@Operation(summary = "Updates an netCDF")
+	@ApiResponses(value = {
+		@ApiResponse(responseCode = "200", description = "NetCDF updated.", content = @Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = NetCDF.class))),
+		@ApiResponse(responseCode = "204", description = "NetCDF not found", content = @Content),
```
dvince2, line 117: This case is returning a `ResponseEntity.notFound().build()` which is `404`

dvince2 (changes requested):

bigglesandginger:
> @bigglesandginger You'll need to run `./gradlew generateTypeScript` from the top level directory to generate the types for the frontend.

I kept getting the following error:
```
* What went wrong:
Execution failed for task ':compileJava'.
> org.gradle.api.GradleException: Unable to enhance class : software.uncharted.terarium.hmiserver.filters.ServiceRequestFilter$HeaderMapRequestWrapper
```

All because I had forgotten that Java updates with `brew upgrade` and Java is the new Python.  Why can this code base not use Java 21 but must be 17 <grumble>

packages/server/src/main/resources/application-local.properties
bigglesandginger: local is used within docker?  we should have an application-docker.properties, but ok

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NetCDFController.java
bigglesandginger: I copied DocumentController (or something similar) so that is the origin of that error - I was confused but assumed it was on purpose

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/NetCDFService.java
bigglesandginger: Always a joy when a major refactor sweeps through

PR #2693, opened 2024-02-09
model card on document upload

blanchco:
# Description

* GOLLM model card generated on model upload.
* Im not entirely happy with how to implement this but we will need to move the pdfextractions calls to the backend and have everything run there for the ideal UX.  For now gollm needs to wait for pdfextractions to finished in order to get the text to runs the card.

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/page/project/components/tera-upload-resources-modal.vue
+++ b/packages/client/hmi-client/src/page/project/components/tera-upload-resources-modal.vue
@@ -72,6 +72,7 @@ import InputText from 'primevue/inputtext';
 import { useToastService } from '@/services/toast';
 import TeraImportGithubFile from '@/components/widgets/tera-import-github-file.vue';
 import { extractPDF } from '@/services/knowledge';
+import { modelCard } from '@/services/goLLM';
 
 defineProps<{
 	visible: boolean;
@@ -153,6 +154,11 @@ async function upload() {
 				const newAsset = useProjects().addAsset(assetType, id);
 				if (name && name.toLowerCase().endsWith('.pdf')) {
 					extractPDF(id);
+				} else if (
+					(name && name.toLowerCase().endsWith('.txt')) ||
+					(name && name.toLowerCase().endsWith('.md'))
+				) {
+					modelCard(id);
 				}
 				return newAsset;
 			})
--- a/packages/client/hmi-client/src/services/document-assets.ts
+++ b/packages/client/hmi-client/src/services/document-assets.ts
@@ -2,10 +2,12 @@
  * Documents Asset
  */
 
-import API from '@/api/api';
+import API, { PollerState } from '@/api/api';
 import type { AddDocumentAssetFromXDDResponse, Document, DocumentAsset } from '@/types/Types';
 import { logger } from '@/utils/logger';
 import { Ref } from 'vue';
+import { fetchExtraction } from './knowledge';
+import { modelCard } from './goLLM';
 
 /**
  * Get all documents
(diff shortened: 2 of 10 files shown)


packages/client/hmi-client/src/services/model.ts
```diff
@@ -141,6 +141,6 @@ export async function generateModelCard(
 	}
 
 	if (modelServiceType === ModelServiceType.TA4) {
-		await modelCard(documentId, modelId);
+		await modelCard(documentId);
```
mwdchang, line 144: Sometimes you await for modelCard and sometimes you do not await for modelCard - if there is something special that people need to be aware of can you please add some comments to the function declaration?

blanchco: I can add a comment for clarity.  The way its being used right now is to just fire and forget mostly except for that special case where I fetch the model again in the `tera-model-from-code` and `tera-model-from-document-drilldown` components.

blanchco:
```
async function generateCard(docId, modelId) {
	if (!docId || !modelId) return;

	if (clonedState.value.modelService === ModelServiceType.TA1 && card.value) {
		return;
	}

	if (clonedState.value.modelService === ModelServiceType.TA4 && goLLMCard.value) {
		return;
	}

	await generateModelCard(docId, modelId, clonedState.value.modelService);
	fetchModel();
}
```

packages/client/hmi-client/src/page/project/components/tera-upload-resources-modal.vue
```diff
@@ -153,6 +154,11 @@ async function upload() {
 				const newAsset = useProjects().addAsset(assetType, id);
 				if (name && name.toLowerCase().endsWith('.pdf')) {
 					extractPDF(id);
+				} else if (
+					(name && name.toLowerCase().endsWith('.txt')) ||
+					(name && name.toLowerCase().endsWith('.md'))
+				) {
+					modelCard(id);
```
YohannParis, line 161: Why this is done here?

packages/client/hmi-client/src/services/document-assets.ts
```diff
@@ -209,16 +211,25 @@ async function createDocumentFromXDD(
...
 		logger.error('Error upload file from doi');
 		return null;
 	}
 
+	if (response.data.extractionJobId) {
+		const result = await fetchExtraction(response.data.extractionJobId);
+		if (result.state === PollerState.Done) {
+			modelCard(response.data.documentAssetId);
+		}
+	}
```
YohannParis, line 232: This should be done in the controller of the `/document-asset/create-document-from-xdd`.
But I guess we have no way to know that it was done don't we?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -73,12 +69,16 @@ private TaskResponseHandler getModelCardResponseHandler() {
...
+				log.info("Writing model card to database for document {}", props.getDocumentId());
+				final DocumentAsset document = documentAssetService.getAsset(props.getDocumentId())
 						.orElseThrow();
 				final ModelCardResponse card = objectMapper.readValue(resp.getOutput(), ModelCardResponse.class);
-				model.getMetadata().setGollmCard(card.response);
-				modelService.updateAsset(model);
+				if (document.getMetadata() == null){
+					document.setMetadata(new java.util.HashMap<>());
+				}
+				document.getMetadata().put("goLLM_card", card.response);
```
YohannParis, line 79: We agreed that `snake_case` are ~the devil~ not to be used anyware in the HMI.
For JSON properties we use `camelCase`, so in this case:
```suggestion
				document.getMetadata().put("gollmCard", card.response);
```

blanchco: changed all instances to `gollmCard`

packages/client/hmi-client/src/services/document-assets.ts
blanchco: Everything should probably be done in the backend, Im not a fan of waiting for extractions/model card generation in the front end.  But I think that would require a pretty big refactor

packages/client/hmi-client/src/page/project/components/tera-upload-resources-modal.vue
blanchco: For txt and md we dont need to wait for extractions since the text field is already available

blanchco: I guess for these files in particular, this could be done in the backend

YohannParis: That would be great indeed.

YohannParis (approved):

PR #2694, opened 2024-02-09
feat: model edit polish

shawnyama:
# Description

- Inverted beaker calls
- Determine number to append to the end of a model template card's variables
- Auto run works properly
- Rename template sanity check
- Added `templateCard` property to `ModelMetadata`
- Removed old template card component (I think it came back from some merge)

https://github.com/DARPA-ASKEM/terarium/assets/28237258/2978bfd9-90b7-4105-b023-bc06c5b217ac

- ~`tera-model-diagram` zoom/pan doesn't work. It seems like it has to do with it not being seen in the window once it's mounted. If I put `tera-model-diagram` in the Wizard view it works as it should.~


packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -196,95 +260,143 @@ export function addCard(
...
-			}
 		});
 }
 
-export function removeCard(
-	modelTemplates: ModelTemplates,
+/**
+ * Update decomposed template name
+ */
+export function updateDecomposedTemplateNameInView(model: any, newName: string) {
```
mwdchang, line 312: You have a fair amount of `model: any` which I think can and should probably be typed:
- model can correspond to `types/Types#Model`, while
- model.model can, for the time, correspond to `types/Types#PetrinetModel`

packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -78,14 +82,66 @@ export function junctionCleanUp(modelTemplates: ModelTemplates) {
...
+				// Transition properties
+				if (properties?.name) valuesToCheck.add(properties.name);
+				if (input) input.forEach((value: string) => valuesToCheck.add(value));
+				if (output) output.forEach((value: string) => valuesToCheck.add(value));
+
+				// Check if the expression contains any variables with a number after it
+				if (expression) {
+					const matches = expression.match(/([a-zA-Z])(\d+)/g);
+					if (matches) matches.forEach((match: string) => valuesToCheck.add(match));
+				}
```
mwdchang, line 123: For what it is worth, I don't think you need to check for expressions, since they are derived from parameters and initials and you included them already from above.

packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -294,107 +406,101 @@ export function addEdge(
...
-				y: yPos
-			} as ModelTemplateCard;
-			yPos += 200;
-
-			addCard(decomposedTemplates, kernelManager, outputCode, modelTemplate);
-			allInitals.push(...modelTemplate.semantics.ode.initials);
-		});
+	// Insert template card data into decomposed models
+	let yPos = 100;
+	const allInitals: any[] = [];
```
mwdchang, line 441: Should be `types/Types#Initial` I think

packages/client/hmi-client/src/services/model-templating.ts
shawnyama: I was using any since `ModelMetadata` doesn't hold a `templateCard` property, I'll see about adding it in.

shawnyama: Added the types, which made me add quite a bit of checks for properties

mwdchang (approved):

PR #2700, opened 2024-02-12
2695 update ES index mappings

dgauldie:
# Description

Resolves #(issue)


chris-dickson (commented):
Couple questions/comments: 

1. Any reason why you're manually adding `id` as a keyword field instead of using native ES id's?  Id queries are quite fast in ES and this adds overhead.
2. You're likely going to want to make indices default to 1 shard.  I don't think these are going to have a ton of data in them to warrant 16.  
3. You can reduce a ton of duplication by using [Component Templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-component-template.html).

kbirk (approved):

PR #2702, opened 2024-02-12
2699 integrate climate data within hmi server

bigglesandginger:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/AssetType.java
```diff
@@ -12,5 +12,6 @@ public enum AssetType {
 	WORKFLOW,
 	ARTIFACT,
 	CODE,
-	DOCUMENT
+	DOCUMENT,
+	NETCDF
```
YohannParis: Why do we have this here?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
```diff
@@ -555,4 +564,39 @@ private static CsvColumnStats getStats(final List<String> aCol) {
...
+
+		try {
+			final Optional<PresignedURL> url = datasetService.getDownloadUrl(id, filename);
+			try (NetcdfFile ncFile = NetcdfFiles.open(url.get().getUrl())) {
+				ImmutableList<Attribute> globalAttributes = ncFile.getGlobalAttributes();
+				for (Attribute attribute : globalAttributes) {
+					String name = attribute.getName();
+					Array values = attribute.getValues();
+					//				log.info("[{},{}]", name, values);
+				}
```
YohannParis: This is the only way to know if the Dataset contains a NetCDF or not?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/AssetType.java
bigglesandginger: You should be seeing that change - that was a merge error but I am not showing that as current in the PR

YohannParis (approved):

PR #2707, opened 2024-02-13
Have searchable assets write to existing indices

kbirk:
The searchable document and model datasets now ingest into the `model` and `document` indices with the correct mappings.

Document source text is added to S3 as well.


packages/es-ingest/src/main/resources/application.properties
```diff
@@ -9,7 +9,35 @@ logging.pattern.console=%d{yyyy-MM-dd HH:mm:ss} [%level] %msg [%c:%L]%n
...
+terarium.amazon.credential.access-key=${aws-access-key-id}
+terarium.amazon.credential.secret-key=${aws-secret-access-key}
+terarium.amazon.s3.region=us-east-1
+terarium.amazon.s3.url=${aws-url}
+
+########################################################################################################################
+# Ingest configuration
+########################################################################################################################
+terarium.ingestParams[0].name="Model Dataset"
+terarium.ingestParams[0].inputDir=/home/kbirk/Downloads/model
```
YohannParis, line 33: You should add those to the readme to know that those needs to be changed when running the application

kbirk: It's all in the README.

dgauldie, line 33: should these default to something other than /home/kbirk?? maybe something that will fail like "/your/model/directory"

kbirk: Fixed

packages/es-ingest/src/main/resources/application-local.properties
```diff
@@ -5,20 +5,10 @@ terarium.elasticsearch.url=http://localhost:9200
...
-terarium.esingest.ingestParams[1].topics=climate, epidemiology
-terarium.esingest.ingestParams[1].outputIndexRoot=searchable_document
-terarium.esingest.ingestParams[1].ingestClass=software.uncharted.terarium.esingest.ingests.document.DocumentIngest
-terarium.esingest.ingestParams[1].clearBeforeIngest=true
-terarium.esingest.ingestParams[1].ignoreMissingDocumentsDuringUpdates=false
+terarium.file-storage-s3-bucket-name=askem-local-storage
+terarium.amazon.credential.access-key=admin
+terarium.amazon.credential.secret-key=admin123
+terarium.amazon.s3.region=us-east-1
+terarium.amazon.s3.url=http://localhost:9000
```
bigglesandginger, line 14: I believe this is supposed to be `minio` not `localhost` for running locally (within Docker-world).  I have taken to adding an `application-intellij.properties` so that I can override `minio` to `localhost` and not have a /etc/hosts entry for that :) - Currently that property file is in a PR of mine - 2702

packages/server/src/main/resources/application-local.properties
```diff
@@ -41,7 +41,7 @@ spicedb.target=localhost:50051
 ########################################################################################################################
 aws-access-key-id=admin
 aws-secret-access-key=admin123
-aws-url=http://minio:9000
+aws-url=http://localhost:9000
```
bigglesandginger, line 44: see above comment

bigglesandginger (changes requested):
Derek informed me as to why `minio` vs `localhost`

packages/es-ingest/src/main/resources/application-local.properties
kbirk: Ooo okay I was wondering. Mind renaming it to `application-ide.properties` so us vscode plebs don't feel left out? :)

kbirk: Added it to this PR actually.

packages/server/src/main/resources/application-local.properties
kbirk: Done!

bigglesandginger (approved):

PR #2717, opened 2024-02-14
feat: add template to flat view

shawnyama:
# Description

- Model template editing can be done in the flattened view
- As soon as a decomposed card is linked to the flattened card the same edit is applied on the decomposed side
- Any decomposed cards that are unlinked to the flattened model are ignored

## How it works
- Key system behaviour (how it always worked): The flattened model updates whenever an edit is done in the decomposed view
- When I add cards/links on the flattened side it is pure UI
- Once a decomposed card is connected to the flattened card all changes translate to an equivalent edit on the decomposed side (see `reflectFlattenedEditInDecomposedView`)
- These decomposed edits then trigger the key system behaviour ^ which "reflattens" or "merges" what's on the flattened view

https://github.com/DARPA-ASKEM/terarium/assets/28237258/04be82e6-57de-45b9-8169-749924b0ed4d


mwdchang (approved):
Per discussion, we should improve the naming in subsequent PRs. Between the usages of template, model, card, and combinations of the 3 terms it is a bit difficult to understand what is going on.

PR #2724, opened 2024-02-14
fix: ensure workflow save on unload

mwdchang:
### Summary
This tries to ensure we have a chance to save before the page is unloaded, e.g. by hard-refreshing for because of the keycloak revnewal

### Test
This is a bit hard to test because of non-blocking, probably the easiest way to convince this works is to add logging to WorkflowController#updateWorkflow and check it is firing on page refresh.


YohannParis (approved):

blanchco (approved):
💃

PR #2728, opened 2024-02-15
Update ProjectAsset when updating Asset name

YohannParis:
# Description

* Issue to refresh projectAsset when an Asset has been updated.
* One more time with the AssetType lowercased


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -127,6 +140,41 @@ public Optional<T> updateAsset(final T asset) throws IOException, IllegalArgumen
...
+			return ((Model) asset).getHeader().getName();
+		} else if (asset instanceof Workflow) {
+			return ((Workflow) asset).getName();
+		} else if (asset instanceof Code) {
+			return ((Code) asset).getName();
+		} else if (asset instanceof DocumentAsset) {
+			return ((DocumentAsset) asset).getName();
+		}
+		return null;
+	}
```
mwdchang, line 179: I don't think you are really using the generics `<T>` if that is the intent here, as far as I remember you can do instanceof without needing the <T> part.

YohannParis: Removed and changed to `Object`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SimulationController.java
```diff
@@ -214,7 +214,7 @@ public ResponseEntity<ProjectAsset> createFromSimulationResult(
 			if (project.isPresent()) {
 				final Optional<ProjectAsset> asset = projectAssetService.createProjectAsset(project.get(),
 						AssetType.DATASET,
-						dataset.getId());
+						dataset);
```
dgauldie, line 217: should this be called datasetId then?

dvince2 (approved):
just conflicts to resolve. Types.ts may need to be regenerated after merging main too

PR #2736, opened 2024-02-16
operator active output behaviour

mwdchang:
### Summary
This modifies the behaviour such that a successful run of an operator will make that operator the active output. Moreover, because we agreed to have a singular exposed output per operator, the emission `append-output-port` is renamed to `append-output` to better reflect what is actually happening.

```
I am going to take a crack at harmonizing the behaviour of the preview-dropdown we have in the operators, just chitchatting with Pascale
- if we run something, the output should be selected automatically, currently this seems to be a mix of auto and manual selection after "run" is clicked
- successive runs will replace the existing output. 
- when selected output changes it will propagate the carried value downstream and mark downstream operators as invalid (to be done later but before eval)
```


YohannParis (approved):
LGTM

PR #2750, opened 2024-02-19
stratify refinement

mwdchang:
### Summary
Per discussion, change the stratify operator such that a "stratify run" will create a new output, instead of creating output on "model save". Updated design to match the output dropdown.

TODO: I added new fields `directed` and `structure` - this are not set currently, will need to check with Nelson to see what are they dependencies and default values.

image

### Testing
Create two different stratifications, should be able to toggle between the two.


packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -276,29 +290,17 @@ const inputChangeHandler = async () => {
...
-		emit('append-output', {
-			id: uuidv4(),
-			label: modelName,
-			type: 'modelId',
-			value: [modelData.id]
-		});
-		emit('close');
-	}
+	await projectResource.addAsset(AssetType.Model, modelData.id, projectId);
+	useToastService().success('', `Saved to project ${projectResource.activeProject.value?.name}`);
```
YohannParis, line 303: Please use `logger.info()` or add a new `LogLevels` for success.

YohannParis (changes requested):

packages/client/hmi-client/src/workflow/ops/stratify-mira/tera-stratify-mira.vue
mwdchang: Why? There is no need to log this event.

YohannParis: Once [my PR](https://github.com/DARPA-ASKEM/terarium/pull/2759) is approved you can do the following:
`logger.success(`Saved to project ${projectResource.activeProject.value?.name}`, { silent: true });`
This will display it green and do not send the log to the back-end.

We want to log everything to the back-end to be able to do a "detailled report" if necessary in the future.

mwdchang: Cannot use `logger.success` because it maps to `console[level]` and `console["success"]` does not exist.

Issue #2753, opened 2024-02-20
[BUG]: Model configuration not showing initials/parameters correctly


mwdchang:
Model configuration operator is incorrectly displaying the parameters/initials, namely it is treating everything as if it is stratified. The `getUnstratifiedInitials(...)` can be used safely only if the model has been stratified as it has heuristics to truncate strings otherwise.

```
190 const parameters = computed<Map<string, string[]>>(() => {
191   if (!model.value) return new Map();
192   return getUnstratifiedParameters(model.value);
193 });
194
195 const initials = computed<Map<string, string[]>>(() => {
196   if (!model.value) return new Map();
197   return getUnstratifiedInitials(model.value);
198 });
```

![Image](https://github.com/DARPA-ASKEM/terarium/assets/1220927/9358d671-bce6-4a2f-8853-ebce77dace5a)

PR #2754, opened 2024-02-20
fix initials/parameters display due to incorrect stratification usage

mwdchang:
### Summary
Fix model configuration display issues because it is treating the model as if it is stratified when it is not. This ends up hiding parameters that have underscores in their names that arose naturally, as oppose to underscores that arose from stratifications through MIRA.

image

Fixes: https://github.com/DARPA-ASKEM/terarium/issues/2753


YohannParis (approved):
LGTM

Tom-Szendrey (approved):

PR #2755, opened 2024-02-20
funman changes

mwdchang:
### Summary
- Add a compartmental sanity check (mass conservation + non-negativity), this is turned on by default
- Lower the default tolerance value, 0.89 was too high and we almost always need to lower it
- Add calculation for normalization constant = `sum(eval(initial state expressions))`

### Testing
Maybe a bit difficult as funman had not been stable since last sync against their develop-branch. Probably want to use simple models like SIR.


Tom-Szendrey (approved):

YohannParis (approved):
LGTM

PR #2758, opened 2024-02-20
Ts/2500 implement notebook beaker integrate within model config

Tom-Szendrey:
# Description
This PR is for 2 things - Sorry!
1) This adds the beaker context and components to the notebook side of things.
To do this we have had to add a temporary config to this drilldown so it has access to a configuration when there are none yet created. This can then be passed to beaker for reading and writing.

2) This updates how we are using local variables in this component to be more aligned with how we are doing that in every other component.
We are going 
From state updates local, local sometimes updates state, wait for debounce to do things
To local updates state

# Testing:
Please note I am testing with this branch of beaker for preview messages: https://github.com/DARPA-ASKEM/beaker-kernel/pull/85

https://github.com/DARPA-ASKEM/terarium/assets/17088680/e956d00c-0dcc-4202-b701-b2d13bf0bd6b

Should be able to:
Ask agent for snippets of code (not many only the 2 tools in the agent)
Write code you want and update the model_config object as well
Save changes
See results


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -411,55 +518,104 @@ const onUpdateSelection = (id) => {
...
+		knobs.value.initials = state.initials;
+		knobs.value.parameters = state.parameters;
+		knobs.value.timeseries = state.timeseries;
+	}
+
+	// Create a new session and context based on model
+	try {
+		const jupyterContext = buildJupyterContext();
+		if (jupyterContext) {
+			await kernelManager.init('beaker_kernel', 'Beaker Kernel', jupyterContext);
```
Tom-Szendrey, line 577: I may want to shutdown before hand. This is a potential issue in all of our nodes that use this atm I believe. When switching outputs we are not shutting this down as we are not hitting unmount. Are we just creating new sessions on top of eachother?

`kernelManager.shutdown();`

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -411,55 +528,104 @@ const onUpdateSelection = (id) => {
...
+		knobs.value.initials =
+			model.value?.semantics?.ode.initials !== undefined
+				? model.value?.semantics?.ode.initials
+				: [];
+		knobs.value.parameters =
+			model.value?.semantics?.ode.parameters !== undefined
+				? model.value?.semantics?.ode.parameters
+				: [];
+		knobs.value.timeseries =
+			model.value?.metadata?.timeseries !== undefined ? model.value?.metadata?.timeseries : [];
```
blanchco, line 573: timeseries should default to in empty object I believe `{}`

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -143,16 +180,134 @@ const emit = defineEmits([
...
+	if (!code) return;
+
+	const messageContent = {
+		silent: false,
+		store_history: false,
+		user_expressions: {},
+		allow_stdin: true,
+		stop_on_error: false,
+		code
+	};
```
mwdchang, line 245: Nitpick: Do we need all these attributes?

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -143,16 +180,134 @@ const emit = defineEmits([
...
+
+	knobs.value.initials =
+		model.value?.semantics?.ode.initials !== undefined ? model.value?.semantics?.ode.initials : [];
+	knobs.value.parameters =
+		model.value?.semantics?.ode.parameters !== undefined
+			? model.value?.semantics?.ode.parameters
+			: [];
+	knobs.value.timeseries =
+		model.value?.metadata?.timeseries !== undefined ? model.value?.metadata?.timeseries : [];
+};
```
mwdchang, line 303: Can we show Demeter some love and not have the same chain repeated over and over?

```
const ode = model.value?.semantics?.ode;
if (ode) return;

knobs.value.initials  = ...
knobs.value.parameters = ...

```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -411,55 +528,104 @@ const onUpdateSelection = (id) => {
...
-		configInitials.value = props.node.state.initials;
-		configParams.value = props.node.state.parameters;
-		configTimeSeries.value = props.node.state.timeseries;
+	() => props.node.inputs[0],
+	async () => {
+		await initialize();
 	},
-	{ immediate: true, deep: true }
+	{ immediate: true }
+);
```
mwdchang, line 600: This may be better placed in `onMounted` hook, it is effectively watching the same conditions as we don't really expect inputs to change after the component is created. onMounted gives you a bit more guarantee that the scaffolds are there to use (e.g. the DOM tree).

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -411,55 +528,104 @@ const onUpdateSelection = (id) => {
...
+	state.initials = knobs.value.initials;
+	state.parameters = knobs.value.parameters;
+	state.timeseries = knobs.value.timeseries;
+	state.tempConfigId = data.id;
+	emit('update-state', state);
+};
+
+// Fill the form with the config data
+const initialize = async () => {
+	const state = _.cloneDeep(props.node.state);
```
mwdchang, line 553: I feel like we should clean up the initialize section a bit, first you don't seem to be updating state here, so no need to clone?

But further more, it is a bit confusing here, because if I were to unroll `createTempModelConfig`, you have
- state updating knob.values, and
- knob.values updating state
While this may end up being the same thing mathematically (I don't actually know) it would be easier to reason with if we have a more uni-directional flow.

It is also weird that in `createTempModelConfig` you have state updated via knobs, as opposed to from the model-configuration you just created.

Let me know if you'd like to have a chat re: organizing.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -143,16 +180,134 @@ const emit = defineEmits([
...
+	// for now only save the last code executed, may want to save all code executed in the future
+	const codeHistoryLength = props.node.state.modelEditCodeHistory.length;
+	const timestamp = Date.now();
+	if (codeHistoryLength > 0) {
+		state.modelEditCodeHistory[0] = { code, timestamp };
+	} else {
+		state.modelEditCodeHistory.push({ code, timestamp });
+	}
+	emit('update-state', state);
+};
```
mwdchang, line 284: Q: is this pattern used everywhere we have this agent/notebook pairing?

mwdchang (changes requested):

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Tom-Szendrey: No, Fair point

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Tom-Szendrey: I still need an individual check for each parameter and initials as theyre both potentially undefined within `ode` but yes I can use `const ode` to shorten the length of nonsense here

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Tom-Szendrey: fair point

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
Tom-Szendrey: yes I believe so. This can be factored when we have some time

mwdchang (approved):

PR #2759, opened 2024-02-20
Improving Toast, add success and do not hide errors

YohannParis:
# Description

* Replace DEBUG with SUCCESS
* Stop ERROR logs to automatically hide


shawnyama (approved):

packages/client/hmi-client/src/utils/logger.ts
```diff
@@ -84,18 +58,9 @@ class Logger {
 		try {
 			throw new Error();
 		} catch (error: any) {
-			const stack = error.stack.split('\n');
-			const callerLine = stack[3];
-			return callerLine;
+			return error.stack.split('\n')[3];
```
mwdchang, line 61: Did anyone ever check if this magic-number-3 is valid or not?

mwdchang:
```
```
try { throw new Error("hello world"); } catch (error) { console.log(error.stack.split("\n")[3]); }

>  undefined
```

PR #2760, opened 2024-02-20
Jamie operator placeholder image bonanza

jamiewaese-uncharted:
**BEFORE**:
unconnected
image

connected
image

**AFTER**:
unconnected
image

connected
image


packages/client/hmi-client/src/components/operator/tera-operator-placeholder.vue
```diff
@@ -8,32 +8,52 @@
...
+import simulateProbabilistic from '@assets/svg/operator-images/simulate-probabilistic.svg';
+import coupleModels from '@assets/svg/operator-images/couple-models.svg';
+import stratifyModel from '@assets/svg/operator-images/stratify-model.svg';
+import createModelFromCode from '@assets/svg/operator-images/create-model-from-code.svg';
+import transformDataset from '@assets/svg/operator-images/transform-dataset.svg';
+import createModelFromEquation from '@assets/svg/operator-images/create-model-from-equation.svg';
+import transformModel from '@assets/svg/operator-images/transform-model.svg';
+import dataset from '@assets/svg/operator-images/dataset.svg';
+// import validateModelConfiguration from '@assets/svg/operator-images/validate-model-configuration.svg'
+import document from '@assets/svg/operator-images/document.svg';
```
mwdchang, line 31: Instead of calling items `deterministic` or `probabilistic` it is better to name them per their performer group(s). Internally everywhere we have been naming these xyz-julia and xyz-ciemss respectively. The notions of a deterministic/probabilistic split is incorrect.


Also strictly speaking, we are optimizing a model-configuration, the optimization process does not change the model itself per se.


Also some of your images look identical, I don't know if it would be better to have a single physical file, but import the same file as different semantic instances. Maybe @YohannParis has some preference here.

jamiewaese-uncharted: Re some images looking identical... yes, but that is temporary. The main work here was setting up the page so I can easily swap SVG's when we have a better idea of what we want to put in each node. They will all be different eventually.

jamiewaese-uncharted:
@mwdchang Can we approve this now?

mwdchang (approved):

PR #2767, opened 2024-02-21
Clearer stratify

mwdchang:
### Summary
This enables fuller MIRA stratify capabilities, and try to use an easier to understand language.
- directed: this is hidden and assume to be true
- structure: this is proxied by useStructure (true | false), where `true = None = all combinations` and `false = [] = nothing`

Cartesian =~ Allow existing transitions to involve multiple strata
Structure =~ Create new transitions between strata

image

### Testing
Structure toggle works, you need to have beaker instance running.

### DIFF
--- a/packages/client/hmi-client/src/components/stratification/tera-stratification-group-form.vue
+++ b/packages/client/hmi-client/src/components/stratification/tera-stratification-group-form.vue
@@ -2,11 +2,13 @@
 	<div class="strata-group" :style="`border-left: 9px solid ${props.config.borderColour}`">
 		<div class="input-row">
 			<div class="sub-header">
-				<label>Directed</label>
-				<InputSwitch @change="emit('update-self', updatedConfig)" v-model="directed" />
+				<label class="multi-line-label">Create new transitions between stratas</label>
+				<InputSwitch @change="emit('update-self', updatedConfig)" v-model="useStructure" />
 			</div>
 			<div class="sub-header">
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/stratification/tera-stratification-group-form.vue
```diff
@@ -74,7 +77,8 @@ const updatedConfig = computed<StratifyGroup>(() => ({
 	groupLabels: labels.value,
 	cartesianProduct: cartesianProduct.value,
 	directed: directed.value,
-	structure: structure.value
+	structure: structure.value,
```
YohannParis, line 80: Why not removing `structure`? It is not used anywhere in this component, and you are using `useStructure` for the logic.

YohannParis (approved):

packages/client/hmi-client/src/components/stratification/tera-stratification-group-form.vue
mwdchang: So ..... my intuition is that we will need "structure" in the next iteration, in our discussion we were a bit lazy and went with the black/white route, either you have all structural combinations or no structural combinations. 

But this is kind of kicking the can down the road: model-configuration will become more difficult (resolving transitions that shouldn't be there), and simulations will be more expensive (solving relationships that result in 0s). So having a "structure" in place now makes it a little easier for the next round and the stratify-state remains compatible.

PR #2771, opened 2024-02-21
Tweak to allow loading simulation results into the dataset beaker context

mattprintz:


packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
```diff
@@ -182,13 +182,24 @@ const setKernelContext = (kernel: IKernelConnection, context_info) => {
...
 };
 
+// FIXME: this is a bit fragile
+const toAssetType = (t: string) => {
+	if (t.endsWith('Id')) {
+		return t.substring(0, t.length - 2);
+	}
+	throw new Error(`Cannot convert type ${t}`);
+};
+
```
YohannParis, line 192: Here a stronger method to put your mind at ease.

```suggestion
function toAssetType(type: string): AssetType {
    const lowercaseType = type.toLowerCase();

    for (const assetType of Object.values(AssetType)) {
        if (lowercaseType.includes(assetType.toLowerCase())) {
            return assetType;
        }
    }

    throw new Error(`Cannot convert ${type} to a known AssetType.`);
}
```

When tested with the following it works as you expect it:
```js
console.log(toAssetType('modelID'));
console.log(toAssetType('lolDataset'));
console.log(toAssetType('I love using beaker to analyse my simulations results'));
console.log(toAssetType('image'));
```

Screenshot 2024-02-23 at 08 16 43

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -136,7 +136,10 @@ export const addEdge = (
 
 	// Check if type is compatible
 	if (sourceOutputPort.value === null) return;
-	if (sourceOutputPort.type !== targetInputPort.type) return;
+
+	const allowedTypes = targetInputPort.type.split('|');
```
YohannParis, line 140: Should we think about making those `inputPort.type` an array instead?

YohannParis (approved):
LGTM 🎉

packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
mwdchang: Thanks for the offer, but no it does not. 
This is meant to convert the string to match askem-beaker's terminology, and and the terms used in Terarium. I will add a bit more explanation comment on why the function is a bit fragile.

packages/client/hmi-client/src/services/workflow.ts
mwdchang: I have not thought about this option. We can have a chat about it along with other potential improvement like using concrete classes instead of types.

PR #2772, opened 2024-02-21
feat: update search for document assets

shawnyama:
# Description

https://github.com/DARPA-ASKEM/terarium/assets/28237258/1f5af425-8ddd-4f1b-af1e-4e7eb7f92e06

- Document assets are fetched instead of the xdd version and they can be added into a project
- ~`fileNames` aren't included in the `getDocumentAsset` endpoint so I adjusted the conditionals to check for a `text` property. So for now TXT docs will appear but not PDFs. I let @kbirk know about this issue.~ Changed the JsonAlias to `filenames` in `DocumentAsset.java` so now I always get `fileNames`
- Made the search bar a textarea for more of a chat like experience (matches design) and to remove auto complete suggestions
- Facet bar is collapsed by default


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
```diff
@@ -35,7 +35,7 @@ public class DocumentAsset extends TerariumAsset {
 	private String userId;
 
 	@TSOptional
-	@JsonAlias("file_names")
```
YohannParis, line 38: Sorry do not do that, leave it as is and we will fix it properly in es with @kbirk

YohannParis (approved):

PR #2775, opened 2024-02-21
fix: ensemble ciemss

mwdchang:
### Summary
- Fix initial mapping/ensembleConfig not configured properly
- Rewrite and simplify mapping table, this had messed up rendering and invalid html

image

### Testing
Note ciemss currently will fail if sample > 1


Tom-Szendrey (approved):
Thanks for these changes!

Issue #33, opened 2022-09-29
Integrate Phantom UI into TERArium


YohannParis:
This is a complex task and will be broken down into multiple issues:

- [ ] Add React component to TERArium: looking into [Vite.js](https://github.com/n3ps/vite-vue-react-demo) and [front-end architecture](https://www.patterns.dev/posts/islands-architecture/)

Import relevant [Phantom UI components](https://github.com/DARPA-ASKEM/data-annotation/tree/main/phantom):
- [ ] Import `Table`, link to `data-annotations` API, write tests, and fit TERArium style
- [ ] Import `Annotation Form`, link to `data-annotations` API, write tests, and fit TERArium style

YohannParis:
This is not relevant anymore as we have decided to create or own component to fit _TERArium_ best.

PR #2794, opened 2024-02-23
client-event log spamming

bigglesandginger:
# Description

The issue is potentially caused by the Bearer Token becoming stale (see issue: "Avoid stale headers by supplying them as a function #33" https://github.com/Azure/fetch-event-source/issues/33)

As the `fetch-event-source` code is 4 years stale this is an attempt to replace with `extended-eventsource` (see issue "Resurrect repo #74" https://github.com/Azure/fetch-event-source/issues/74)

Initial testing appears to "work" - waited 15 minutes

Resolves #(issue)

Web-browser Network console being filled with `client-event` notifications ad nauseum.


bigglesandginger:
There are issues, by design, that you cannot have two or more tabs open on the app (against the same server).  The server will force disconnect a user if they connect with a different EventSource (ie the other tab) and then vicious circle of repeating force disconnecting.

bigglesandginger:
The Bearer Token is definitely refreshing correctly - in the server logs (which I augmented locally) I can see the client establish a new connection.  I will test the old code on the weekend to see if this occurred.

Screenshot 2024-02-23 at 4 30 33 PM

dgauldie (approved):

dgauldie (dismissed):
this PR fails type checking. You will need to fix onerror and onopen

YohannParis (approved):

PR #2796, opened 2024-02-23
cleanup preview and selection

mwdchang:
### Summary
This cleans up the update-port and select-port semantic in the workflow operators. There are numerous instances where the signals are crisscrossed and hard to figure out whether the intention is to select output or update the output-port's attributes.

Also remove the multi-select from the preview drilldown, only a single output will be exposed by any given operator.

### DIFF
--- a/packages/client/hmi-client/src/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/workflow/tera-workflow.vue
@@ -104,7 +104,6 @@
 						<component
 							:is="registry.getNode(node.operationType)"
 							:node="node"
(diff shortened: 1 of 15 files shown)


packages/client/hmi-client/src/workflow/tera-workflow.vue
YohannParis, line 1: Do we want to keep those console calls?

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
mwdchang: nope, will remove

PR #2799, opened 2024-02-26
Ts/2790 mapping 1 to 1

Tom-Szendrey:
# Description

1. Auto mapping now returns 1 to 1 results
2. Most logic is done in a generic function, so we can utilize this with `model + dataset`, or `model + model`  (or dataset + dataset)

# Testing
Manually adding groundings to a dataset and model pair, running this and checking the results match what is expected
Console.logs and hitting the API were used in double checking http://34.230.33.149:8771/docs#/entities/entity_similarity_api_entity_similarity_post


packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,182 @@ export const renderLossGraph = (
...
+	sourceEntities.forEach((ele) => {
+		if (ele.groundings) {
+			ele.groundings.forEach((grounding) => allSourceGroundings.push(grounding));
+		}
+	});
+	targetEntities.forEach((ele) => {
+		if (ele.groundings) {
+			ele.groundings.forEach((grounding) => allTargetGroundings.push(grounding));
+		}
+	});
```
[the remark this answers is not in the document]
Tom-Szendrey: Much cleaner, thank you

packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,172 @@ export const renderLossGraph = (
...
+	// Filter out anything with a similarity too small
+	const filteredSimilarity = allSimilarity.filter((ele) => ele.similarity >= acceptableDistance);
+	const validSources: any[] = [];
+	const validTargets: any[] = [];
+
+	// Join results back with source and target
+	filteredSimilarity.forEach((similarity) => {
+		// Find all Sources assosiated with this similarity
+		sourceEntities.forEach((source) => {
+			if (source.groundings && source.groundings.includes(similarity.source)) {
```
shawnyama, line 161: Can be shortened to `source.groundings?.includes(similarity.source)`

packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,172 @@ export const renderLossGraph = (
...
+			if (target.groundings?.includes(similarity.target)) {
+				validTargets.push({
+					targetId: target.id,
+					sourceKey: similarity.source,
+					targetKey: similarity.target,
+					distance: similarity.similarity
+				});
+			}
+		});
+	});
```
shawnyama, line 181: Instead of doing a check on every iteration it may be better to filter first.
```suggestion
filteredSimilarity.forEach((similarity) => {
		const matchingSources = sourceEntities.filter((source) => source.groundings?.includes(similarity.source));
		const matchingTargets = targetEntities.filter((target) => target.groundings?.includes(similarity.target));

		matchingSources.forEach((source) => {
			validSources.push({
				sourceId: source.id,
				sourceKey: similarity.source,
				targetKey: similarity.target,
				distance: similarity.similarity
			});
		});

		matchingTargets.forEach((target) => {
			validTargets.push({
				targetId: target.id,
				sourceKey: similarity.source,
				targetKey: similarity.target,
				distance: similarity.similarity
			});
		});
	});
```

packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,172 @@ export const renderLossGraph = (
...
+
+	// for each distinct source, find its highest matching target:
+	const distinctSource = [...new Set(validSources.map((ele) => ele.sourceId))];
+	distinctSource.forEach((distinctSourceId) => {
+		let currentDistance = -Infinity;
+		let currentTargetId = '';
+		validSources.forEach((source) => {
+			if (distinctSourceId === source.sourceId) {
```
shawnyama, line 189: You can do a filter here too

packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,172 @@ export const renderLossGraph = (
...
+			targetEntities.push({
+				id: col.name,
+				groundings: Object.entries(col.metadata?.groundings?.identifiers).map((ele) =>
+					ele.join(':')
+				)
+			});
+		} else {
+			targetEntities.push({ id: col.name });
 		}
 	});
```
shawnyama, line 246: copilot coming in with a clutch clean up
```suggestion
	// Fill sourceEntities with modelOptions
	modelOptions.forEach((state) => {
		const groundings = state.grounding?.identifiers
			? Object.entries(state.grounding.identifiers).map((ele) => ele.join(':'))
			: undefined;
		sourceEntities.push({ id: state.id, groundings });
	});

	// Fill targetEntities with datasetOptions
	datasetOptions.forEach((col) => {
		const groundings = col.metadata?.groundings?.identifiers
			? Object.entries(col.metadata.groundings.identifiers).map((ele) => ele.join(':'))
			: undefined;
		targetEntities.push({ id: col.name, groundings });
	});
```

YohannParis (approved):
I wonder if this should be in its own service as we might use it elsewhere about mapping?

packages/client/hmi-client/src/services/calibrate-workflow.ts
```diff
@@ -117,76 +127,172 @@ export const renderLossGraph = (
...
+	modelTwoOptions.forEach((state) => {
+		if (state.grounding?.identifiers) {
+			sourceEntities.push({
+				id: state.id,
+				groundings: Object.entries(state.grounding?.identifiers).map((ele) => ele.join(':'))
+			});
+		} else {
+			sourceEntities.push({ id: state.id });
+		}
+	});
```
shawnyama, line 289: Same pattern as above. Not sure if you want to make a helper function for this since it's used quite a bit
```suggestion
	// Fill sourceEntities with modelOneOptions
	modelOneOptions.forEach((state) => {
		const groundings = state.grounding?.identifiers
			? Object.entries(state.grounding.identifiers).map((ele) => ele.join(':'))
			: undefined;
		sourceEntities.push({ id: state.id, groundings });
	});

	// Fill targetEntities with datasetOptions
	modelTwoOptions.forEach((state) => {
		const groundings = state.grounding?.identifiers
			? Object.entries(state.grounding.identifiers).map((ele) => ele.join(':'))
			: undefined;
		targetEntities.push({ id: state.id, groundings });
	});
```

Tom-Szendrey:
> I wonder if this should be in its own service as we might use it elsewhere about mapping?

This is a good point. I will throw this into `src/services/concept.ts` where the `entity_similarity` mira call is made if you think that is better?

shawnyama (approved):

packages/client/hmi-client/src/services/calibrate-workflow.ts
Tom-Szendrey: I find this pattern harder to see explicitly what is happening but you do make a good point that the way I have written it its harder to see that `sourceEntities` is being pushed to either way
I will take your approach here

packages/client/hmi-client/src/services/calibrate-workflow.ts
Tom-Szendrey: Spoke offline

PR #2801, opened 2024-02-26
Facet types not matching

bigglesandginger:
# Description

Stops the 400 errors in the client network console

Fundamentally there is a disconnect between the UI's `Types.ts AssetType` values and `TaggableTypes.java` but I haven't looked to see where these are used completely.  

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ConceptController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ConceptController.java
@@ -1,5 +1,6 @@
 package software.uncharted.terarium.hmiserver.controller.dataservice;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


dvince2 (changes requested):
So @YohannParis had a similar issue with casing on `AssetType` that could be fixed with the lowercase name being marked as the `@JsonValue` for the front end. See [this](https://github.com/DARPA-ASKEM/terarium/pull/2728/files) PR under `AssetType.java`.  That will be preferable to having a mismatch of cases on that `type` field I think.

Also if you're modifying the models you'll need to regenerate the front end Types.ts with `./gradlew generateTypeScript`

YohannParis:
I think those taggable Asset could be removed from the app, I don't know what are we using them for?

bigglesandginger:
> I think those taggable Asset could be removed from the app, I don't know what are we using them for?

I moved the concept of this to your big PR - as for keeping removing Taggable Assets - I think that needs to be a discussion with Derek/Gauldie/Kevin as I have no idea what they are save they are used somehow through the magic of annotations and tracing that is beyond IntelliJ's capability.

PR #2802, opened 2024-02-26
mira model template preview for regnet/stockflow framework types

mwdchang:
### Summary
Using beaker to generate model preview. For:
- petrinet: use our own renderer
- stock and flow: use MMT preview
- regnet: use MMT preview

For example: Stock and flow SIR model
image

### Testing
Requires beaker, you can get an example stock-n-flow from: https://raw.githubusercontent.com/DARPA-ASKEM/simulation-integration/main/data/models/SEIR_stockflow.json

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model/tera-model-node.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model/tera-model-node.vue
@@ -8,7 +8,7 @@
 				@change="if ($event.value) view = $event.value;"
 				:options="viewOptions"
 			/>
-			<div class="container">
+			<div class="container" v-if="model && model.header.schema_name === 'petrinet'">
 				<tera-model-diagram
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model/tera-model-node.vue
```diff
@@ -88,11 +93,37 @@ async function onModelChange(chosenProjectModel: ProjectAsset) {
...
+		console.error(err);
+	}
+};
+
 onMounted(async () => {
 	const state = props.node.state;
 	if (state.modelId) {
 		model.value = await getModel(state.modelId);
 
+		await generateTemplatePreview();
```
YohannParis, line 125: Do we need the `await` here, I do not think we need to wait for it to keep doing logic.

YohannParis (approved):

PR #2805, opened 2024-02-26
2803 feat force typests generation before committing

dvince2:
# Description

When a user goes to check in code now we're running a quick and dirty check against Types.ts.  If there originally was no Types.ts we generate the file and fail the check so that the user can ensure that there are no breaking changes on the front end. If Types.ts was generated we take a hash of the file and run the generation again to see if there are any changes. If there are again we fail the check.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/Model.java
```diff
@@ -40,11 +40,12 @@ public class Model extends TerariumAssetThatSupportsAdditionalProperties {
 	@TSOptional
 	private ModelMetadata metadata;
 
+
 	public Model() {
 		super();
 	}
 	// Copy constructor
-    public Model(Model other) {
+    public Model(final Model other) {
```
dvince2, line 48: Model.java was how I was testing this script. Its not technically related to this PR but there is now a tiny little refactor thats happened on commit :)

.husky/typesVerification.sh
```diff
@@ -0,0 +1,28 @@
+echo "Running Types.ts generation test"
+
+
+TYPES_FILE="packages/client/hmi-client/src/types/Types.ts"
+
+MODIFIED_FILES=`git status -s | grep "Types.ts" | wc -l`
+if (( MODIFIED_FILES == 0 )); then
```
Tom-Szendrey, line 7: Sorry for my lack of knowledge here.
Does this mean I will need to generate for every PR? Even if it has no java changes?

Tom-Szendrey: I cant read

Tom-Szendrey (approved):

dvince2:
As @chris-dickson pointed out the performance on this check may not be great and will slow down the commit process. We may want to sit on this one or discuss post-hackathon

PR #2807, opened 2024-02-26
Jamie stratify papercuts pg 14 of pascales notes

jamiewaese-uncharted:
Cosmetic tweaks to the Stratify operator


YohannParis (approved):

mwdchang (changes requested):
Please check your local git repo/setup are up-to-date.

The changes you have here are dated Feb 26th and is seemingly based on the state of the repo on Feb 20th. Yet we have made changes on Feb 21sth that are not reflected here so something had gone awry: the telltale signs here is that based on discussions with Nelson  and Matt we have already removed `Directed` and `Cartesian product` as the terms are hard to understand - so it should not be possible for you to move the section to the bottom, after we have already removed it ahead of you.

Please check you are in sync'ed up with the main branch, otherwise PRs might overwrite other people's changes.

PR #2810, opened 2024-02-26
Jamie stratify semantic table papercuts pg 14 b of pascales notes

jamiewaese-uncharted:
cleaned up semantic table Concept columns, removed Extractions columns, formatted numbers, etc.


YohannParis (approved):

mwdchang:
@jamiewaese-uncharted I can't tell where we are at here, because some of the changes you made here overlaps with your other changes which we have already merged. Please have a look and either resolve or close the PR.

PR #2811, opened 2024-02-26
Fix model from equations

blanchco:
# Description

* model from equations fix to persist all equations
* add document asset to the one documentId port, it looks like we got rid of the other port so I just stick this logic for propagation of selected asset on the one port.
* gollm card generation feedback


packages/client/hmi-client/src/workflow/ops/model-from-document/tera-model-from-document-drilldown.vue
```diff
@@ -219,14 +220,14 @@ const goLLMCard = computed<any>(() => document.value?.metadata?.gollmCard);
 
 const isNewModelModalVisible = ref(false);
 const savingAsset = ref(false);
-
+const generatingCard = ref(false);
```
shawnyama, line 223: `isGeneratingCard` is superior naming

packages/client/hmi-client/src/workflow/ops/document/tera-document-operation-drilldown.vue
```diff
@@ -169,14 +169,4 @@ watch(
...
-	font-size: var(--font-caption);
-	font-weight: var(--font-weight-semibold);
-	max-height: 2em;
-	color: var(--gray-700);
-	display: -webkit-box;
-	-webkit-line-clamp: 3;
-	-webkit-box-orient: vertical;
-}
-</style>
+<style scoped></style>
```
YohannParis, line 172:
```suggestion
```

YohannParis (approved):

PR #2829, opened 2024-02-27
fix: model-edit operator save/restore state

mwdchang:
### Summary
model-edit operator is not utilizing output state, so when selecting from the output preview dropdown nothing happens. 
This PR fixes this issue.

TODO: We still need to work out whether we want to create new output on every run? or when the model is saved?

### Testing
Start from a simple model, e.g. SIR
- Add a new template A => B
- Save model to create output
- Remove I => R template
- Save model to create output

Should be able to exit, come back in,  and when switching between outputs see the changes restored, this might take a few seconds ad we need to rerun the code that was saved.

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
@@ -19,7 +19,7 @@
 						outlined
 						severity="secondary"
 						size="small"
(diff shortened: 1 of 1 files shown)


YohannParis (approved):
LGTM

@pascaleproulx mentioned to me that every run should create a new output... but the latest run needs to be active one by default.

This is ideal only if we update the downstream of the node linked output.

PR #2830, opened 2024-02-27
Update local km to use live services

YohannParis:
# Description

* Use `latest` tag
* Point to TA1 live services


dgauldie (approved):
lgtm

PR #2833, opened 2024-02-27
regenet semantic tables

blanchco:
# Description

* Update the semantic tables component to handle both petrinet and regnet tables and display appropriately
* Up next displaying a regnet model!

https://github.com/DARPA-ASKEM/terarium/assets/33158416/8c81609d-f281-4d72-9434-36bce1b376e8


YohannParis (approved):
Well done 🎉

PR #2838, opened 2024-02-27
Sim previews

mwdchang:
### Summary
Node previews on TA3 nodes.

Note this is in sort of an in-between state, ideally the nodes and the dirlldowns will function independently, but currently you need to wait for the simulation to finish within the drilldown in order for the node preview to work.

- Nodes/drilldowns need to handle in-progress
- Need to simplify in-progress list to just a single id

image


packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
```diff
@@ -14,19 +24,44 @@
...
+const selectedRunId = ref<any>(null);
+
+watch(
+	() => props.node.active,
+	async () => {
+		const active = props.node.active;
+		if (!active) return;
+
+		selectedRunId.value = props.node.outputs.find((o) => o.id === active)?.value?.[0];
+		if (!selectedRunId.value) return;
```
Tom-Szendrey, line 56: Should we set hasData.value back to false here?
Waiting to run this locally on pulling but it looks like we may run into a scenario with output switching where we are showing old results? 

I may be mistaken I have not yet run this

mwdchang: Probably, but can be cleaned up later.

Tom-Szendrey (approved):

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
Tom-Szendrey: Very fair

PR #2839, opened 2024-02-27
Update tera-asset.vue

jamiewaese-uncharted:
BEFORE
image

AFTER
image


shawnyama (approved):
Consider going into the PrimeVue themes and change the rules there next time. That way the problem will be solved for all components.

In this case you'd go to `assets/css/theme/extensions/_tabview.scss`

PR #2853, opened 2024-02-28
Fix and test other conversion tasks

kbirk:


packages/mira/tasks/sbml_to_petrinet.py
```diff
@@ -4,13 +4,29 @@
...
+def create_temp_file(name, contents):
+    with open(name, 'w') as f:
+        f.write(contents)
+
+
+def delete_temp_file(name):
+    try:
+        os.remove(name)
+    except:
+        pass
```
dvince2, line 27: Should this be logged?

dvince2 (approved):

packages/mira/tasks/sbml_to_petrinet.py
kbirk: Nah, its just to clean up any created file, if it never made it far enough to create, this just catches that "file does not exist"  error.

PR #2858, opened 2024-02-28
model config regnet tables

blanchco:
# Description

* Model config regnet editable parameters


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config-table.vue
```diff
@@ -181,6 +181,10 @@ const matrixModalContext = ref({
 	matrixId: ''
 });
 
+const modelType = computed(
+	() => props.modelConfiguration?.configuration?.header?.schema_name ?? AMRSchemaNames.PETRINET
+);
```
YohannParis, line 186: I think we need to create a util `isPetrinet` because we need to test the values `toLowerCase()`

blanchco: ah, Ill add a function called like 'getModelType' and we can pass in the model and it returns the schema_name with toLowercase on it

blanchco: added everywhere we get the model type, defaults to petrinet

YohannParis (approved):

PR #2859, opened 2024-02-28
Fix equations to model operator and edit config 

YohannParis:
# Description

* Display preview on model edit and equations to model
* Fix bug that did not display model edit


blanchco (approved):
✅

PR #2862, opened 2024-02-28
fix[hmi-server]: Dataset extract columns, doing it on GET if not present

### LINKED ISSUE #2784: Dataset extract columns, doing it on GET if not present

dvince2:
# Description

This is actually the fix for incorrect column information after a beaker-driven transform too. There is a corresponding change on beaker that blows away column data (https://github.com/DARPA-ASKEM/beaker-kernel/pull/95) since columns and/or data will have changed in modified datasets. Then on a GET call we repopulate the column data because its empty.  This will also cover any situation where a CSV/dataset was some how uploaded and the column data was empty too. In addition, this centralizes some code around getting CSV files and updating headers.

Resolves #2784

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
@@ -1,18 +1,20 @@
 package software.uncharted.terarium.hmiserver.controller.dataservice;
 
-import java.io.IOException;
-import java.io.StringReader;
-import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.UUID;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
+import co.elastic.clients.elasticsearch._types.FieldValue;
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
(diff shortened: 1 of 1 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
```diff
@@ -202,6 +187,45 @@ public ResponseEntity<Dataset> getDataset(@PathVariable("id") final UUID id) {
...
+		}
+
+		if(dataset.get().getColumns() != null && !dataset.get().getColumns().isEmpty()){
+			//columns are set. No need to extract
+			return dataset;
+		}
+		if(dataset.get().getFileNames() == null || dataset.get().getFileNames().isEmpty()){
+			//no file names to extract columns from
+			return dataset;
+		}
```
YohannParis, line 209: Let's change that to make sure that we update the missing columns fields.
Because the CSV might be updated and therefore need to re-run this for the new or remove columns.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
dvince2: These are quick failure cases - in this case here there are no files attached to the dataset (so no CSV to read headers from)

In the case above it we only continue when the column fields are empty (beaker blows away the column metadata when it does a transform now). We don't want to do this column check on every get because it is EXPENSIVE - we need to fetch the CSV file to do it

PR #2871, opened 2024-02-29
Jamie fix model edit canvas so items dont go outside canvas

jamiewaese-uncharted:
BEFORE
edit model canvas before

AFTER
edit model canvas after


packages/client/hmi-client/src/components/widgets/tera-infinite-canvas.vue
```diff
@@ -228,5 +228,6 @@ svg:active {
 
 .foreground-layer {
 	position: relative;
+	pointer-events: none;
 }
```
mwdchang, line 232: Infinite-canvas is used in other places (e.g. workflow), as well the top two layers (foreground, data) are meant to be interactive. 

Adding a `pointer-event: none` here will break other areas of the application.

mwdchang (changes requested):
`pointer-events: none` in infinite-canvas is causing weird effects elsewhere, please let us know why this is desired and we can investigate a work-around.

mwdchang (approved):

PR #2874, opened 2024-02-29
Compare Models

bigglesandginger:
# Description

Resolves #(issue)


dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -67,196 +63,12 @@ public class GoLLMController {
...
-				}
-
-			} catch (Exception e) {
-				log.error("Failed to configure model", e);
-				throw new RuntimeException(e);
-			}
-			log.info("Model configured successfully");
-		});
-		return handler;
+		taskService.addResponseHandler(new ModelCardResponseHandler(objectMapper, documentAssetService));
```
kbirk, line 68: We should autowire these as private members of the GoLLMController so we don't need to manually pass the deps. Just add @Component to the class definitions and the @RequiredArgsContructor stuff

bigglesandginger: I was thinking there might be some "nice" refactoring beyond what I have done, but as I cannot test anything currently I was thinking this might be good enough for now?

PR #2882, opened 2024-02-29
Optimize operator

Tom-Szendrey:
# Description
Creating the optimize box
Adding backend code for pyciemss optimize

# Still TODO
Many cleanup issues have been created
#2903
#2904
#2905 
#2906 
#2907

# Notes
This box's output is the same as the model configuration thrown in. There is no useful information as far as I am aware coming from this optimize call to append or update our model configuration.
This means this is essentially purely for visual inspection at the moment.

Also to not is that currently the visual inspection is essentially the same as the ciemss simulation output as I need to include the policy.json (which is essentially just `[1.0]`) into the output in some meaningful way

# Screenshots 

General page in screenshot
image

image

Issue 3 in screenshot
image

image


packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss.vue
```diff
@@ -0,0 +1,729 @@
...
+							<label for="target-variable">Target-variable(s)</label>
+							<MultiSelect
+								class="p-inputtext-sm"
+								:options="modelStateOptions.map((ele) => ele.id)"
+								v-model="knobs.targetVariables"
+								placeholder="Select"
+							/>
+						</div>
+						<div class="label-and-input">
+							<label for="statistic">Statistic</label>
```
mwdchang, line 106: Why do you have so many `<label for="..."` declarations? 
Are they suppose to be pointing to something?

Tom-Szendrey: Not for anything, relic of Jami's initial work on this I will remove these

dvince2 (approved):

mwdchang:
I think a bit or reorganization would be nice, so the asynchronous calls/requests are not daisy chained to each other - this makes things harder to test.  

For example, take this snippet here, it is difficult to test "A" independently without other side-effects

```
function Test() {
   A()
}
function A() {
  id = doStuff()
  B(id)
}
```

As opposed to this style

```
function Test() {
   id = A()
   B(id)
}
function A() {
  id = doStuff()
  return id
}
```

Tom-Szendrey:
> I think a bit or reorganization would be nice, so the asynchronous calls/requests are not daisy chained to each other - this makes things harder to test.
> 
> For example, take this snippet here, it is difficult to test "A" independently without other side-effects
> 
> ```
> function Test() {
>    A()
> }
> function A() {
>   id = doStuff()
>   B(id)
> }
> function B(id) {
>   blah()
> }
> ```
> 
> As opposed to this style
> 
> ```
> function Test() {
>    id = A()
>    B(id)
> }
> function A() {
>   id = doStuff()
>   return id
> }
> function B(id) {
>   blah()
> }
> ```

Thank you very good shout out. My mistake
I have changed the setOptimizeResults logic 
If you notice anywhere else I have this please let me know. I will double check

mwdchang (approved):

PR #2887, opened 2024-02-29
Fix task response handler lookup

kbirk:
Fixes how the handlers are executed.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ModelCardResponseHandler.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ModelCardResponseHandler.java
@@ -1,66 +1,68 @@
 package software.uncharted.terarium.hmiserver.service.tasks;
 
+import java.util.UUID;
+
+import org.springframework.stereotype.Component;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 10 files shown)


bigglesandginger (approved):
The whitespace changes are so irritating.  First Gauldie then you :). We need to either have some proper whitespace formatting or have orders that first come first server.  The fact that IntelliJ is inconsistent, is really frustrating, but then you have tabs and not whitespaces :)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ModelCardResponseHandler.java
```diff
@@ -1,66 +1,68 @@
...
-            final Response card = objectMapper.readValue(((TaskResponse)resp).getOutput(), Response.class);
-            if (document.getMetadata() == null){
-                document.setMetadata(new java.util.HashMap<>());
-            }
-            document.getMetadata().put("gollmCard", card.response);
+	@Override
+	public void onSuccess(TaskResponse resp) {
+		try {
+			final String serializedString = objectMapper
+					.writeValueAsString(((TaskResponse) resp).getAdditionalProperties());
```
bigglesandginger, line 51: I suppose we can remove the casting of resp - but if this is a show stopper let's get this through - then clean up later

PR #2893, opened 2024-02-29
2889 task vensim support

dgauldie:
# Description

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
@@ -1,6 +1,8 @@
 package software.uncharted.terarium.hmiserver.controller.mira;
 
 import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
 
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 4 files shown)


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -76,36 +78,59 @@ public ResponseEntity<Model> convertAndCreateModel(
 			@RequestBody final ModelConversionRequest conversionRequest) {
 
 		try {
-			TaskRequest req = new TaskRequest();
-			req.setInput(conversionRequest.getModelContent().getBytes());
 
-			if (endsWith(conversionRequest.getModelName(), List.of(".mdl"))) {
+			final Optional<Artifact> artifact = artifactService.getAsset(conversionRequest.artifactId);
```
dvince2, line 82: I just want to confirm that this is intentional to use `Artifact` here?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
YohannParis: Yes, we want to use them for saving user uploaded files that are not part of our java model, like AMR.

kbirk:
I made some changes that will need to be merged into this for the new sync / async mode param for GoLLMController and MiraController: https://github.com/DARPA-ASKEM/terarium/pull/2928

YohannParis (approved):

PR #2899, opened 2024-03-01
2898 contributors and authors not showing on project cards

bigglesandginger:
# Description

Shows Contributors (users/groups with edit permissions)
Screenshot 2024-03-01 at 1 42 19 PM

Shows list of Authors (users/groups with edit permissions)

https://github.com/DARPA-ASKEM/terarium/assets/5182085/0912dd76-31e7-4e6f-8655-6fe82c6aca06

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -130,6 +135,47 @@ public ResponseEntity<List<Project>> getProjects(
...
+		try {
+			List<RebacPermissionRelationship> permissionRelationships = rebacProject.getPermissionRelationships();
+			for (RebacPermissionRelationship permissionRelationship : permissionRelationships) {
+				Schema.Relationship relationship = permissionRelationship.getRelationship();
+				if (relationship.equals(Schema.Relationship.CREATOR)
+						|| relationship.equals(Schema.Relationship.ADMIN)
+						|| relationship.equals(Schema.Relationship.WRITER)) {
+					if (permissionRelationship.getSubjectType().equals(Schema.Type.USER)) {
+						PermissionUser user = reBACService.getUser(permissionRelationship.getSubjectId());
+						String name = user.getFirstName() + " " + user.getLastName();
```
YohannParis, line 160: Should we add a method in `PermissionUser` to return the fullname instead?

packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
```diff
@@ -161,25 +162,64 @@ function selectNewPermissionForUser(permission: string, userId: string) {
...
+	const updatedProject = cloneDeep(props.project);
+	if (updatedProject.authors) {
+		const index = updatedProject.authors.indexOf(name);
+		if (index !== undefined && index > -1) {
+			updatedProject.authors.splice(index, 1);
+		}
+		await useProjects().update(updatedProject);
+	}
+}
+
```
dvince2, line 210: I don't think that these remove/add user functions are technically doing anything.  The `Authors` field is set as read only/transient/optional and is only updated/set when we `GET` a project. Changing the authors list and updating won't actually change the object in the database.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -130,6 +135,47 @@ public ResponseEntity<List<Project>> getProjects(
 		return ResponseEntity.ok(projects);
 	}
 
+	private class Contributor {
```
YohannParis, line 138: What is the purpose of this class?

dvince2 (approved):
This all looks good, I'm just not sure if the update is needed/working as expected!

packages/client/hmi-client/src/services/project.ts
```diff
@@ -178,33 +178,41 @@ async function getPermissions(projectId: Project['id']): Promise<PermissionRelat
 	}
 }
 
-async function setPermissions(projectId: Project['id'], userId: string, relationship: string) {
+async function setPermissions(
+	projectId: Project['id'],
+	userId: string,
+	relationship: string
```
YohannParis, line 184: Should this use `PermissionProject["PermissionProject"]` type?
```suggestion
	relationship: PermissionProject["PermissionProject"]
```

packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
```diff
@@ -69,6 +69,7 @@ import Button from 'primevue/button';
 import { getUsers } from '@/services/user';
 import type { PermissionRelationships, PermissionUser, Project } from '@/types/Types';
 import { useProjects } from '@/composables/project';
+import { cloneDeep } from 'lodash';
```
YohannParis, line 72: We want to move away from `cloneDeep` and use the Web API [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) now that this is supported by all browser.

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
bigglesandginger: It is to update the front-end - which this currently forces the UI to refetch from the server, though maybe in a future there would be an "intelligent" update that does not do complete refetch.  Otherwise, the UI elements that are not the Share Dialog do not know that anything changed, and short of the user clicking the refresh page button, will not reflect the changes that the user just performed in said dialog.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
bigglesandginger: Great question.  There is seems to be a function required here and in the front end for this feature.  Ideally this is a task of its own to capture this need and make it system-wide.

packages/client/hmi-client/src/services/project.ts
bigglesandginger: Don't know - I didn't write this, but IntelliJ insisted this get reformatted for no good reason, all because I changed the return type.

PR #2900, opened 2024-03-01
refactor: model templating variable names

shawnyama:
# Description

### Main changes
- `ModelTemplates` -> `ModelTemplateCanvas` since this holds all the items in UI such as junctions, edges, model template
- along with the above change variables like `decomposedTemplates`, `flattenedTemplates`, `modelTemplates` now end with `Canvas` eg. `decomposedCanvas`
- variables named `modelTemplate` held a `Model` type, now I am just calling them `model` - it seems like a cleaner choice overall

Some smaller touch ups are easy to see in my last two commits.


mwdchang (approved):
Didn't test, but looks better, thanks!

PR #2911, opened 2024-03-04
Fix Model card issues

YohannParis:


packages/client/hmi-client/src/components/model/petrinet/tera-model-description.vue
```diff
@@ -193,7 +193,7 @@ const emit = defineEmits(['update-model', 'fetch-model', 'update-configuration',
 
 const teraModelDiagramRef = ref();
 
-const card = computed(() => {
+const card: Card | any | null = computed(() => {
```
mwdchang, line 196: This mixing of concrete types and any doesn't make sense to me. We either type it properly
- Card | null | soemthing, or we give up and just do
- `any` without additional typing information

mwdchang (approved):

PR #2918, opened 2024-03-05
feat: update input ports on different output selection

shawnyama:
# Description

Before the input of a downstream node wouldn't update if a different output from the upstream node was chosen. Now it updates properly. Also if the downstream node is connected to anything or has any outputs those are all invalidated

https://github.com/DARPA-ASKEM/terarium/assets/28237258/666f85cf-0dc4-4f11-af4d-4eba2312a977


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
mwdchang, line 1: Is this file just a refactoring change? I don't understand why it is here? is there a commit you didn't push?

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -345,25 +345,48 @@ export class WorkflowRegistry {
...
+		wf.edges.forEach((e) => {
+			if (e.source === targetNode.id) removeEdge(wf, e.id);
+		});
+		targetNode.outputs = [];
+		// Update the input port of the target node
+		const targetPort = targetNode.inputs.find((port) => port.id === edge.targetPortId);
+		if (!targetPort) return;
+		targetPort.label = selected.label;
+		targetPort.value = selected.value;
+	});
```
mwdchang, line 389: I think we will need some further discussion about this:
- I don't think clearing out the output is solution here, it might be better to just keep them around and mark the operator to an "invalid" status instead
- Your example works if it is `A => B => C` but it will not make sense if the chain is 4 or longer `A => B => C => D`, basically if you are messing with the graph itself the algorithm will need to be recursive.

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
shawnyama: I was playing around with this component to understand things, and I ended up changing the lodash import.

packages/client/hmi-client/src/services/workflow.ts
YohannParis: I agree that clearing the outputs of downstream operator is not the way to go.
Also we should use the `OperatorStatus.INVALID`

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -331,6 +331,20 @@ export class WorkflowRegistry {
...
+	nodesCache: Map<string, WorkflowNode<any>>
+) {
+	const outgoingEdges = wf.edges.filter((edge) => edge.source === sourceNode.id);
+	outgoingEdges.forEach((edge) => {
+		const targetNode = nodesCache.get(edge.target);
+		if (!targetNode) return;
+		targetNode.status = OperatorStatus.INVALID;
+		cascadeInvalidateDownstream(wf, targetNode, nodesCache);
+	});
+}
```
mwdchang, line 346: `Filter` is actually worse than `find` because you must traverse the entire array 😄 

My idea of is that you build `id => node[]` map/relationship ahead of time so you avoid doing any lookups or filters, can we do that?

shawnyama:
found a weird bug

https://github.com/DARPA-ASKEM/terarium/assets/28237258/9bf67be5-4d87-4dec-a9c3-5bf989429863

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -345,25 +356,63 @@ export class WorkflowRegistry {
...
+
+		// Collect node cache
+		if (nodeCache.has(edge.source)) {
+			// Add the node connected to this edge's source
+			nodeCache
+				.get(edge.source)
+				?.push(wf.nodes.find((node) => node.id === edge.target) as WorkflowNode<any>);
+			// Make a new relationship for the nodes attached to this edge's target node
+			if (!nodeCache.has(edge.target)) nodeCache.set(edge.target, []);
+		}
```
mwdchang, line 412: I think this is buggy, let's say we have this graph:
```
A => B => C => D
```

And my edge list is as follows (you can't rely on order because of deletes, duplicate and whatnot)
- C => D
- B => C
- A => B

If we invalidate on `A`,  your cache starts as `A => []` and you are not going to get any thing for B, C, D because the first two iterations will be skipped

YohannParis (approved):

PR #2928, opened 2024-03-05
Task caching and fixes

kbirk:
This PR accomplishes two things:

- Task caching / sharing: using a SHA256 of the task, determine if it has already been sent, if it has and is still pending or has finished successfully, use the existing task.
- Client sent events are sent _after_ request handlers have resolved. Now when the browser gets a `SUCCESS` it knows that any and all server logic has completed for that task.

This code uses redis for caching and locking and is written to support caching across multiple instances of the hmi-server if we ever need to scale it that way.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
@@ -56,8 +56,8 @@ public class MiraController {
 		public Model response;
 	};
 
-	private boolean endsWith(String filename, List<String> suffixes) {
-		for (String suffix : suffixes) {
+	private boolean endsWith(final String filename, final List<String> suffixes) {
+		for (final String suffix : suffixes) {
 			if (filename.endsWith(suffix)) {
 				return true;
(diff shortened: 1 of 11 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -1,113 +1,140 @@
...
+	void init() {
+		taskService.addResponseHandler(stellaToStockflowResponseHandler);
+		taskService.addResponseHandler(mdlToStockflowResponseHandler);
+		taskService.addResponseHandler(sbmlToPetrinetResponseHandler);
+	}
+
+	@PostMapping("/convert-and-create-model")
 	@Secured(Roles.USER)
-	@Operation(summary = "Dispatch a mira conversion task")
+	@Operation(summary = "Dispatch a Mira conversion task")
```
YohannParis, line 87: If you want to be pedantic... It's an acronym
```suggestion
	@Operation(summary = "Dispatch a MIRA conversion task")
```

YohannParis (approved):

PR #2932, opened 2024-03-05
reorg sciml simulate

mwdchang:
### Summary
This covers the Simulate operators, Calibrate and Ensemble will be dealt with in different PRs.

There are two themes to the PR here
- Reorganization and simplification of the operator logic, we no longer need to support multiple concurrent simulations within the same operator, and this means we can simplify state management and remove logic that never quite worked completely. 
- Enable a preview on the workflow within the operator node, the nodes will poll and update in a similar manner as the drilldown.

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
@@ -169,7 +169,6 @@ const updateState = () => {
 
 // Main entry point
 const run = async () => {
-	console.log('making request');
 	const simulationId = await makeForecastRequest();
 
 	const state = _.cloneDeep(props.node.state);
@@ -244,14 +243,12 @@ watch(
 
 watch(
 	() => props.node.active,
-	async (n, o) => {
-		if (!props.node.active || n === o) return;
+	async (newValue, oldValue) => {
+		if (!props.node.active || newValue === oldValue) return;
 		selectedOutputId.value = props.node.active;
 
 		// Update Wizard form fields with current selected output state timespan
 		timespan.value = props.node.state.currentTimespan;
-
-		console.log('hihihi', selectedRunId.value);
 
 		// Resume or fetch result
 		const simulationId = selectedRunId.value;
--- a/packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
@@ -117,34 +117,27 @@
 
 <script setup lang="ts">
 import _ from 'lodash';
-import { computed, onMounted, onUnmounted, ref, watch } from 'vue';
+import { computed, ref, shallowRef, watch } from 'vue';
 import Button from 'primevue/button';
 import InputNumber from 'primevue/inputnumber';
 import type { CsvAsset, SimulationRequest, TimeSpan } from '@/types/Types';
 import { ChartConfig, RunResults } from '@/types/SimulateConfig';
 
 import { getModelConfigurationById } from '@/services/model-configurations';
-import { Poller, PollerState } from '@/api/api';
 
-import {
-	getRunResult,
-	makeForecastJob,
-	querySimulationInProgress,
-	pollAction
-} from '@/services/models/simulation-service';
+import { getRunResult, makeForecastJob } from '@/services/models/simulation-service';
 import { createCsvAssetFromRunResults, saveDataset } from '@/services/dataset';
 import { csvParse } from 'd3';
-import { WorkflowNode } from '@/types/workflow';
 import InputText from 'primevue/inputtext';
 import TeraSimulateChart from '@/workflow/tera-simulate-chart.vue';
 import TeraDatasetDatatable from '@/components/dataset/tera-dataset-datatable.vue';
-import { useProjects } from '@/composables/project';
 import SelectButton from 'primevue/selectbutton';
 import TeraDrilldown from '@/components/drilldown/tera-drilldown.vue';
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
```diff
@@ -203,15 +202,21 @@ const updateState = () => {
...
-const runSimulate = async () => {
-	const modelConfigurationList = props.node.inputs[0].value;
-	if (!modelConfigurationList?.length) return;
+// Main entry point
+const run = async () => {
+	showSpinner.value = true;
+	const simulationId = await makeForecastRequest();
+	await pollResult(simulationId);
+	await processResult(simulationId);
+	await lazyLoadSimulationData(simulationId);
```
YohannParis, line 211: Does those steps needs to be in order? If yes, could we display multiple state of loading to let the user know what is happening?
Like: "running simulation", "processing simulation results" and "loading simulation results "

YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
mwdchang: They need to be in order, but for sciml simulate, it happens too fast that progress messages will look like flickering bugs.

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
```diff
@@ -24,31 +24,96 @@
...
+	state.chartConfigs.push([]);
+	emit('update-state', state);
+};
+
+watch(
+	() => props.node.state.inProgressSimulationId,
+	async (id) => {
+		if (!id || id === '') return;
+
+		const r = await pollResult(id);
```
YohannParis, line 100: What does `r` stand for?

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
```diff
@@ -169,23 +161,27 @@ const selectedRunId = computed(
...
 	state.currentTimespan = timespan.value;
 	emit('update-state', state);
 };
 
-const runSimulate = async () => {
-	const modelConfigurationList = props.node.inputs[0].value;
-	if (!modelConfigurationList?.length) return;
+// Main entry point
+const run = async () => {
+	console.log('making request');
```
YohannParis, line 172:
```suggestion
```

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
```diff
@@ -272,49 +225,37 @@ const onSelection = (id: string) => {
...
-			selectedOutputId.value = props.node.active;
-		}
+	async (n, o) => {
+		if (!props.node.active || n === o) return;
+		selectedOutputId.value = props.node.active;
 
 		// Update Wizard form fields with current selected output state timespan
 		timespan.value = props.node.state.currentTimespan;
+
+		console.log('hihihi', selectedRunId.value);
```
YohannParis, line 254:
```suggestion
```

YohannParis (approved):
This will work if the workflow is still running on the background when the drilldown is open, but if in the future we open those drilldowns in their own tab/window, we loose the Poller updates.

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
```diff
@@ -272,49 +225,37 @@ const onSelection = (id: string) => {
...
-			lazyLoadSimulationData(selectedRunId.value);
-		}
-	},
-	{ immediate: true }
+	() => props.node.state.inProgressSimulationId,
+	(id) => {
+		if (id === '') showSpinner.value = false;
+		else showSpinner.value = true;
+	}
 );
```
shawnyama, line 243: I think this could be done in a `computed` variable

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
```diff
@@ -272,49 +225,37 @@ const onSelection = (id: string) => {
...
 watch(
 	() => props.node.active,
-	() => {
-		// Update selected output
-		if (props.node.active) {
-			selectedOutputId.value = props.node.active;
-		}
+	async (n, o) => {
+		if (!props.node.active || n === o) return;
+		selectedOutputId.value = props.node.active;
```
shawnyama, line 249: Whats `n` and `o`?

shawnyama (approved):

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
mwdchang: new/old, will make them more explicit

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-node-julia.vue
mwdchang: made var more explicit

packages/client/hmi-client/src/workflow/ops/simulate-julia/tera-simulate-julia.vue
mwdchang: A watcher construct is better here, as it clearly signals the intention that the value being watched can and will change, whereas a computed just says it is dependent on other variables.

Also this construct allows you to do more things if needed.

PR #2934, opened 2024-03-06
Fix: Save dataset from simulation

Tom-Szendrey:
# Description
1) Move the save dataset from simulation into its own component
2) fix the java so this actually works


dvince2 (approved):

packages/client/hmi-client/src/components/dataset/tera-save-dataset-from-simulation.vue
```diff
@@ -0,0 +1,57 @@
...
+		:disabled="isSaveDisabled"
+		outlined
+		title="Saves the current version of the model as a new Terarium asset"
+		@click="showSaveInput = !showSaveInput"
+	>
+		<span class="pi pi-save p-button-icon p-button-icon-left"></span>
+		<span class="p-button-text">Save as new dataset</span>
+	</Button>
+	<span v-if="showSaveInput" style="padding-left: 1em; padding-right: 2em">
+		<InputText v-model="saveAsName" class="post-fix" placeholder="New dataset name" />
```
shawnyama, line 12: Put styles in class for the span. Also this `post-fix` class isn't in the css, maybe you forgot to move it?

packages/client/hmi-client/src/components/dataset/tera-save-dataset-from-simulation.vue
```diff
@@ -0,0 +1,57 @@
...
+import { useProjects } from '@/composables/project';
+import InputText from 'primevue/inputtext';
+import Button from 'primevue/button';
+
+const props = defineProps<{
+	simulationRunId: string;
+}>();
+
+const saveAsName = ref('');
+const showSaveInput = ref(<boolean>false);
```
shawnyama, line 39: should be written as ref<boolean>(false) but just do ref(false) it already assumes its boolean

packages/client/hmi-client/src/components/dataset/tera-save-dataset-from-simulation.vue
Tom-Szendrey: Thanks

packages/client/hmi-client/src/components/dataset/tera-save-dataset-from-simulation.vue
Tom-Szendrey: Dont know why it was like that, good shout

PR #2936, opened 2024-03-06
2778 logging and timing of terarium activities

### LINKED ISSUE #2778: Logging and timing of terarium activities
We have logging during the last eval.  This is in the client event code.
Verify that we still have logging, and then, is it acceptable - are there areas we should add it to?

Items to add/change

- [x] get rid of pause/resume and keep start/stop
- [x] log the time the user spends in each operator drilldown
- [x] Add precanned scenarios/tasks/descriptions
- [x] Add checkbox to denote multiple users

dvince2:
# Description

* Removing pause/resume operations from logging timing as it caused confusion
* New menu bar widget showing current testing status with stop button (Thanks @YohannParis!)
* Precanned scenario, task and descriptions
* Adding timing to drilldown operators

Resolves #2778


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -386,9 +389,23 @@ function updateOutputPort(node: WorkflowNode<any> | null, workflowOutput: Workfl
...
+	const timeSpent: number = Date.now() - startTime;
+	await EventService.create(
+		EventType.OperatorDrilldownTiming,
+		useProjects().activeProject.value?.id,
+		JSON.stringify({
+			node: currentActiveNode.value?.displayName,
+			timeSpent
+		})
+	);
+};
```
mwdchang, line 407: I am not sure what this is trying to measure, but note `dispalyName` is not unique as you can have several operators of the same type.

dvince2: We're just looking to see how long people are in each of the drilldowns... so I _think_ just using displayName here should be ok but we could log out more info if you think thats useful?

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
mwdchang: I don't know either, I suppose we can always change if needed. 

If in doubt we can serialize the entire `currentActiveNode.value`

PR #2941, opened 2024-03-07
Added box around suggested configurations section in tera-model-confi…

jamiewaese-uncharted:
BEFORE
Monosnap Terarium 2024-03-07 10-32-37

AFTER
Monosnap Terarium 2024-03-07 10-30-55


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -850,9 +855,65 @@ watch(
...
+		month: 'long',
+		day: '2-digit',
+		hour: '2-digit',
+		minute: '2-digit',
+		second: '2-digit',
+		timeZoneName: 'short'
+	});
+
+	return formatter.format(date);
+}
```
mwdchang, line 874: We already have several different date-formatters under `src/utils/date.ts`, can we just reuse one of the existing ones? If not, this function should be included into `src/util/dates.ts` rather than as a stand-alone in the component.

jamiewaese-uncharted: makes sense. I moved the function to `src/util/dates.ts` as suggested.

YohannParis (approved):

PR #2948, opened 2024-03-07
Fix the mira convert and create model

kbirk:


YohannParis (approved):

dgauldie (approved):
tested locally and approved

PR #2949, opened 2024-03-07
Styled dataset stats column with a boxplot for numbers and a list for categoricals

jamiewaese-uncharted:
**BEFORE**
image

**AFTER**
image


YohannParis (changes requested):
This should be its own component.

YohannParis (approved):

PR #2953, opened 2024-03-07
clone dataset-transform session on branching action

mwdchang:
### Summary
DataTransform has a `state.notebookSessionId` field that links to an ES document representing the notebook session, this means when we branch, subsequent modifications are written into the same doc from multiple sources creating weird inconsistencies. 

This PR breaks this by adding a "clone" method, when branching happens DataTransform sessions are cloned are reassigned new ids.

### Testing
- Add a dataset to the workflow
- Attach the dataset to a data-transform node
- Branch the data-transform node
- Do something, e.g. adding a column in one of the data-transform node
- The two data-transforms should be separate entities and showing different data

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
@@ -158,6 +158,7 @@ public class NotebookSessionController {
 	@Operation(summary = "Clone a session")
 	@ApiResponses(value = {
 			@ApiResponse(responseCode = "200", description = "NotebookSession cloned.", content = @Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = NotebookSession.class))),
+			@ApiResponse(responseCode = "500", description = "There was an issue cloning the session", content = @Content)
 	})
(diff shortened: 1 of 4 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -153,6 +153,27 @@ ResponseEntity<NotebookSession> updateNotebookSession(
 		}
 	}
 
+	@PostMapping("/{id}/clone")
+	@Secured(Roles.USER)
+	@Operation(summary = "Clone a session")
+	@ApiResponses(value = {
+			@ApiResponse(responseCode = "200", description = "NotebookSession cloned.", content = @Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = NotebookSession.class))),
```
dvince2, line 160: Needs the swagger doc `@ApiResponse` for the 500 error case

dvince2 (approved):

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
mwdchang: done

PR #2954, opened 2024-03-07
Added welcome video to the home page banner

jamiewaese-uncharted:
BEFORE
image

AFTER
image

Note, this is the video we made for the CHI submission. We can easily change videos when we have a new one.


mwdchang (changes requested):
Please do not commit larger files (say larger than 4MB) into the git repo. This makes things rather difficult to work with afterwards as it comes a part of the permanent history and makes things really slow with working with git.

We should put the video somewhere else that is more dedicated to housing multimedia....like maybe on a publicly available S3 bucket or in the documentation-image or something.

jamiewaese-uncharted:
I removed the video file and updated the thumbnail image. Should be good to go... ?
image

YohannParis (approved):

PR #2955, opened 2024-03-08
2947 feat improve error logging from the back end to the client

### LINKED ISSUE #2947: [FEAT]: Improve error logging from the back end to the client
Currently all error codes (http errors) from hmi-server are swallowed and the user is just given a generic error message like "unable to create new project"

The back end is sending up correct http error codes and error messages. We'd like to instead show these on the front end.  This should be done directly in `api.ts` so that the only error messages we need for the front end are legit front end only bugs not server issues

We'd thought to bubble the error codes up to the various vue components to handle but that will result in too much cruft

dvince2:
# Description

* `api.ts` no longer swallows error messages but rather will pop up a toast error message with the error code.
* Front end toasts that were just logging a null have been removed because the toast will catch that now. Error messages that were front end specific or as a part of a `catch` have been kept in.
* Some clean up on the dataservice side of hmi server
*  _NOTE_: This may surface some old bugs that had gone unseen because now 404 and 500 errors will pop to the front. We should deal with those problems as their own separately logged issues and not revert this change!!!

Resolves #2947

### DIFF
--- a/packages/client/hmi-client/src/api/api.ts
+++ b/packages/client/hmi-client/src/api/api.ts
@@ -41,7 +41,9 @@ API.interceptors.response.use(
 			if (error?.response?.data) {
 				const responseError: ServerError = error.response.data as ServerError;
 
-				// check to see if the 'message' property is set. If it is, set message to that value.  If not, check the 'trace' property and extract the error from that. It will be the substring between the first set of quotations marks
+				// check to see if the 'message' property is set. If it is, set message to that value.  
+				// If not, check the 'trace' property and extract the error from that. 
+				// It will be the substring between the first set of quotations marks.
 				if (responseError.message) {
 					message = responseError.message;
 				} else if (responseError.trace) {
--- a/packages/client/hmi-client/src/api/api.ts
+++ b/packages/client/hmi-client/src/api/api.ts
@@ -1,5 +1,5 @@
 import { logger } from '@/utils/logger';
-import axios, { AxiosHeaders } from 'axios';
+import axios, { AxiosError, AxiosHeaders } from 'axios';
 import { EventSource } from 'extended-eventsource';
 import { ServerError } from '@/types/ServerError';
 import useAuthStore from '../stores/auth';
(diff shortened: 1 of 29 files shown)


packages/client/hmi-client/src/api/api.ts
```diff
@@ -28,28 +28,37 @@ API.interceptors.request.use(
...
 			// redirect to login
 			const auth = useAuthStore();
 			auth.keycloak?.login({
 				redirectUri: window.location.href
 			});
+		} else {
+			let message = error.message;
+			let title = `${error.response?.statusText} (${error.response?.status})`;
+			if (error.response && error.response.data) {
+				const responseError: ServerError = error.response?.data as ServerError;
```
YohannParis, line 42:
```suggestion
			if (error?.response?.data) {
				const responseError: ServerError = error.response.data as ServerError;
```

packages/client/hmi-client/src/api/api.ts
```diff
@@ -28,28 +28,37 @@ API.interceptors.request.use(
...
+
+				// check to see if the 'message' property is set. If it is, set message to that value.  If not, check the 'trace' property and extract the error from that. It will be the substring between the first set of quotations marks
```
YohannParis, line 44:
```suggestion
				// check to see if the 'message' property is set. If it is, set message to that value.  
				// If not, check the 'trace' property and extract the error from that. 
				// It will be the substring between the first set of quotations marks.
```

packages/client/hmi-client/src/types/ServerError.ts
YohannParis, line 1: Are all those fields, except `message` be provided?

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -68,7 +68,6 @@ export const equationsToAMR = async (
 				return response.data.result;
 			}
 		}
-		logger.error(`Equations to AMR request failed`, { toastTitle: 'Error - Knowledge Middleware' });
```
YohannParis, line 71: Those will be caught by Axios? or do we need to Throw an error?

YohannParis (approved):

packages/client/hmi-client/src/types/ServerError.ts
dvince2: this looks to be the case yes!

packages/client/hmi-client/src/services/knowledge.ts
dvince2: In this case we'd never see status != 200 (because api.ts was eating the failure case), only response == null. And in that case, `api.ts` will now log the correct error to the toaster.

PR #2965, opened 2024-03-08
Equations to model update

YohannParis:
# Description

* Add the creation of a model when calling the equations to model contoller
* update the front-end to not communicate with the knowledge middleware.


YohannParis:
Replaced by 
#2967 
#2968 
#2969 
#2970 
#2971

PR #2966, opened 2024-03-09
Select which model card to display

YohannParis:
# Description

* Add more `--gap` CSS variables values.
* Clean the concept calls to not return `404` when nothing is returned from the DKG but `204`
* Remove Model search highlights
* Clean up how the model description are displayed
* Added a new component to display simple boolean switch
* Add a toggle to select which model_card to be displayed as description between TA1 and TA4

Screenshot 2024-03-09 at 14 54 33


packages/client/hmi-client/src/services/concept.ts
```diff
@@ -55,14 +55,10 @@ async function getFacets(type: AssetType, curies?: ClauseValue[]): Promise<Conce
...
-		const response = await API.get(`/mira/currie/${curies.toString()}`);
-		if (response?.status !== 200) return null;
-		return response?.data ?? null;
-	} catch (error) {
-		logger.error(error, { showToast: false });
-		return null;
-	}
+	const response = await API.get(`/mira/currie/${curies.toString()}`);
+	if (response?.status === 200 && response?.data) return response.data;
+	if (response?.status === 204) console.warn('No DKG entities found for curies:', curies);
```
dvince2, line 60: The changes made to `api.ts` for error messages will let a 204 slip by. If we want the user to know about this we'll need to pop up a toast too

dvince2 (approved):

packages/client/hmi-client/src/services/concept.ts
YohannParis: This is in purpose, kind of hiding DKG lack of biological ontologies

shawnyama (approved):
clean

PR #2968, opened 2024-03-10
Update model to equations call

YohannParis:
# Description

* Small clean up
* Remind to update the call to POST when the backend is ready, see PR https://github.com/DARPA-ASKEM/sciml-service/pull/167

### DIFF
--- a/packages/client/hmi-client/src/services/model.ts
+++ b/packages/client/hmi-client/src/services/model.ts
@@ -181,7 +181,7 @@ export async function generateModelCard(
 	}
 }
 
-// helper fucntion to get the model type, will always default to petrinet if the model is not found
+// helper function to get the model type, will always default to petrinet if the model is not found
 export function getModelType(model: Model | null | undefined): AMRSchemaNames {
 	const schemaName = model?.header?.schema_name?.toLowerCase();
 	if (schemaName === 'regnet') {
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/services/model.ts
```diff
@@ -194,17 +194,19 @@ export function getModelType(model: Model | null | undefined): AMRSchemaNames {
...
-		console.log(`getModelEquation: ${model.header.schema_name} not suported `);
+		console.warn(`getModelEquation: ${model.header.schema_name} not supported `);
 		return '';
 	}
 
-	const id = model.id;
-	const response = await API.get(`/transforms/model-to-latex/${id}`);
+	/* TODO - Replace the GET with the POST when the backend is ready,
+	 *        see PR https://github.com/DARPA-ASKEM/sciml-service/pull/167
+	 */
```
mwdchang, line 206: I am tempted to say just leave it as a GET request, sciml-service, for better or worse, is already tightly coupled with the hmi-server from top to bottom. Unless we have concrete plans to make sciml-service truly independent, my thinking is that it is easier to reason with if all the asset-fetch actions share the same behaviour/pattern.

YohannParis: It's fair, but I was thinking of asking @kbirk if we could add a Redis cache like he did for the task-runner with a hash of the POST as to not call sciml-service to often to just display equations.
Another solution would be to save the LaTeX responses to make sure it matches the `updatedOn` of the model... Still thinking about it.

mwdchang (approved):

PR #2970, opened 2024-03-10
Display nicer names for model framework

YohannParis:
# Description

* Create a util to transform PascalCase into Sentence Case
* Add more model framework compatibility, even if they are not supported yet.

Screenshot 2024-03-10 at 06 49 12


mwdchang (approved):
I am a bit confused with the renaming situation going on here and why it is needed; but otherwise this looks fine.

YohannParis:
> I am a bit confused with the renaming situation going on here and why it is needed; but otherwise this looks fine.

Because the Document to Model was renamed only in title to Equations to Model.
So I made sure the operator now matches that new title.

This PR should have been merge before the operator renaming, this PR was only about displaying nicer names for the dropdown. I should have done a better job when choosing my base.

PR #2974, opened 2024-03-11
Try to clean up Config table to add the +_ from jamie

YohannParis:
# Description

* Making a mess of everything, clearly not helping much


YohannParis:
This will be handle by a new UI/UX of the component.

PR #2977, opened 2024-03-11
Add link-amr, variable-extraction, cosmos-pdf-extraction

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/documentservice/ExtractionProxy.java
```diff
@@ -1,28 +1,57 @@
...
+import org.springframework.web.multipart.MultipartFile;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
 import software.uncharted.terarium.hmiserver.models.documentservice.autocomplete.AutoComplete;
 import software.uncharted.terarium.hmiserver.models.documentservice.responses.XDDExtractionsResponseOK;
 import software.uncharted.terarium.hmiserver.models.documentservice.responses.XDDResponse;
 
-@FeignClient(name = "extractionService", url = "${xdd-dev-service.url}")
+@FeignClient(name = "extractionService", url = "${xdd-prod-service.url}")
```
dvince2, line 20: Is this intentional/coordinated with the XDD team? I know that a number of features/docs have previously been unavailable in their prod instance.

dvince2: To add to this, I think that the xdd documents controller is still on dev not prod, which can cause inconsistencies between extractions and docs

kbirk: If i hit dev I get a 413 request too large. Only prod currently works. Not sure what we want to do here.

dvince2: We should at least keep the extractions and docs endpoints in sync. Reaching out to Ian at xDD may help?

kbirk: I'll revert back to dev so this PR doesn't break anything else. And we can debug why the dev endpoint for cosmos extractions is behaving wonky separately

PR #2980, opened 2024-03-11
feat: structural model comparison

shawnyama:
# Description

- displays the structural model comparison
- generated code from the dropdown works sometimes
- works for sure when pasting in code from the beaker documentation https://darpa-askem.github.io/beaker-kernel/contexts_mira.html#compare-the-three-models-and-visualize-and-display-them

Good code generation:

https://github.com/DARPA-ASKEM/terarium/assets/28237258/ca69a8f9-2468-4bf2-a975-cc726ec31200

Bad code generation:
Screenshot 2024-03-11 at 3 07 44 PM


packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
```diff
@@ -373,9 +374,11 @@ onUnmounted(() => {
 	flex-direction: column;
 }
 
-#notebook-section:deep(main) {
+.notebook-section:deep(main) {
 	gap: var(--gap-small);
 	position: relative;
+	/** TODO: Temporary solution, should be using the default overlay-container padding in tera-drilldown */
```
YohannParis, line 380: In my opinion, the overlay-container should not have any padding. We should let each drilldown decides how to be displayed.

YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue
shawnyama: I am leaning towards passing a prop like `removeDefaultPadding` in the case the padding needs to be specified.

shawnyama: Updated the comment to mention your opinion

PR #2981, opened 2024-03-11
Feat: Add optimize_result.json to the output panel

Tom-Szendrey:
# Description
I have thrown the result.dill file to result.json for us to work with on the frontend.
This saves the optimize run Id + gets the result file to utilize in the output panel.

# Blocked
Waiting on further design:
image

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
@@ -27,6 +27,7 @@ export interface OptimizeCiemssOperationState {
 	chartConfigs: string[][];
 	simulationsInProgress: string[];
 	simulationRunId: string;
+	optimzationRunId: string;
 	modelConfigName: string;
 	modelConfigDesc: string;
 }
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -27,6 +27,7 @@ export interface OptimizeCiemssOperationState {
 	chartConfigs: string[][];
 	simulationsInProgress: string[];
 	simulationRunId: string;
+	optimzationRunId: string;
```
mwdchang, line 30: Would recommend better naming; we are using `simulationId` rather loosely as anything-that-we-request-from-TA3, so if we are tracking multiple things it might be good to make things clear.

Tom-Szendrey: good point, i think i will have `forecastRunId` and `optimzationRunId` and avoid the ambiguous simulation altogether

blanchco (approved):

PR #2984, opened 2024-03-11
feat: add amr-to-mmt function to mira task runner

mwdchang:
### Summary
Add task to convert AMR (petrinet, regenet, stockflow) to MMTs, along with extra parameters/symbolics informations per resulting mira-template.

### Testing
Swagger POST /mira/amr-to-mmt


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/AMRToMMTResponseHandler.java
```diff
@@ -0,0 +1,46 @@
...
+	@Override
+	public TaskResponse onSuccess(final TaskResponse resp) {
+		try {
+			final Response modelResp = objectMapper.readValue(resp.getOutput(), Response.class);
+			resp.setOutput(modelResp.getResponse().getBytes());
+			// final Model model = modelService.createAsset(modelResp.getResponse());
+			// resp.setOutput(objectMapper.writeValueAsString(model).getBytes());
+		} catch (final Exception e) {
+			log.error("Failed to create model", e);
+			throw new RuntimeException(e);
```
dvince2, line 42: If we're just catching and releasing does it make sense to just rethrow `e` so that the original exception type isn't swallowed by `RuntimeException`?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -81,6 +82,35 @@ void init() {
...
+			// send the request
+			final TaskResponse resp = taskService.runTaskSync(req);
+			final JsonNode mmtInfo = objectMapper.readValue(resp.getOutput(), JsonNode.class);
+			return ResponseEntity.ok().body(mmtInfo);
+		} catch (final Exception e) {
+			final String error = "Unable to dispatch task request";
+			log.error("Unable to dispatch task request {}: {}", error, e.getMessage());
+			throw new ResponseStatusException(
+					org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR,
+					error);
```
dvince2, line 108: Because these exceptions are now shown on the front end, is there any risk around `e.getMessage()` being too verbose?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/AMRToMMTResponseHandler.java
mwdchang: sure, will do

mwdchang: Actually no I cannot, cause this is a java-override thing and doesn't throw Exceptions.
If we want to fix we need a bit of redesign CC @kbirk

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
mwdchang: Maybe, can revisit if people complains.

PR #2989, opened 2024-03-12
Add askem admins to all projects

bigglesandginger:
# Description

Allow ASKEM_Admins write permissions to all projects, both new and existing


YohannParis (approved):
Why is it easier to add a the ASKEM_ADMIN_GROUP permissions on all calls to rebac instead of giving the group admin on everything?

bigglesandginger:
> Why is it easier to add a the ASKEM_ADMIN_GROUP permissions on all calls to rebac instead of giving the group admin on everything?

Not really certain what you mean, but I think you are asking, why check ReBAC if the user is a member of the admin group?

The only place that there is knowledge that a given user (Yohann for instance) is in the admin group is within ReBAC, and ReBAC is designed to answer heirarchical permission calls.  Additionally, having this part of ReBAC means that as long as the code check ReBAC for permissions there are no edge cases to memorize.

PR #2993, opened 2024-03-12
decapodes operator

blanchco:
# Description

* decapodes operator, there really isnt much support for llms and decapodes right now, but this is a good base for the operator...It will generate julia code but it isnt runnable

https://github.com/DARPA-ASKEM/terarium/assets/33158416/2f042f58-47cc-4fc4-9f45-9b5282bce969

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
+++ b/packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
@@ -72,12 +72,14 @@ const emit = defineEmits(['close', 'update-state']);
 const kernelManager = new KernelSessionManager();
 let editor: VAceEditorInstance['_editor'] | null;
 
-const modelId = computed(() => props.node?.inputs?.[0]?.value?.[0]);
+const modelId = computed(() => props.node.inputs[0]?.value?.[0]);
 const notebookResponse = ref();
 const sampleAgentQuestions = ['Convert the model to equations please'];
-const contextLanguage = ref<string>('julia-1.10');
+const contextLanguage = 'julia-1.10';
 const isRunningCode = ref<boolean>(false);
 const selectedOutputId = ref<string>(props.node.active ?? '');
+const defaultCodeText =
+	'# This environment allows you to work with a decapode model that was passed in.';
 const outputs = computed(() => {
 	if (!isEmpty(props.node.outputs)) {
 		return [
@@ -93,9 +95,7 @@ const outputs = computed(() => {
 const initializeEditor = (editorInstance: any) => {
 	editor = editorInstance;
 };
-const codeText = ref(
-	'# This environment contains the variable "model_config" to be read and updated'
-);
+const codeText = ref();
 
 const buildJupyterContext = () => ({
 	context: 'decapodes',
@@ -165,7 +165,7 @@ const inputChangeHandler = async () => {
 
 	if (getModelType(model) !== AMRSchemaNames.DECAPODES) return;
 
-	codeText.value = props.node.state.codeHistory?.[0]?.code ?? '';
+	codeText.value = props.node.state.codeHistory?.[0]?.code ?? defaultCodeText;
 
 	// Create a new session and context based on model
 	try {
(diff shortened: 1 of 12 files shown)


packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
```diff
@@ -0,0 +1,209 @@
...
+
+const props = defineProps<{
+	node: WorkflowNode<DecapodesOperationState>;
+}>();
+
+const emit = defineEmits(['close', 'update-state']);
+const kernelManager = new KernelSessionManager();
+let editor: VAceEditorInstance['_editor'] | null;
+
+const modelId = computed(() => props.node?.inputs?.[0]?.value?.[0]);
```
mwdchang, line 75: `props.node` is not optional according to your declaration

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
```diff
@@ -0,0 +1,209 @@
...
+const notebookResponse = ref();
+const sampleAgentQuestions = ['Convert the model to equations please'];
+const contextLanguage = ref<string>('julia-1.10');
```
mwdchang, line 78: I am guessing `ref` is not needed since it will be static.

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return [];
+});
+
+const initializeEditor = (editorInstance: any) => {
+	editor = editorInstance;
+};
+const codeText = ref(
+	'# This environment contains the variable "model_config" to be read and updated'
+);
```
mwdchang, line 98: Copy-pasta error

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
```diff
@@ -0,0 +1,209 @@
...
+};
+
+const inputChangeHandler = async () => {
+	if (!modelId.value) return;
+
+	const model = await getModel(modelId.value);
+
+	if (getModelType(model) !== AMRSchemaNames.DECAPODES) return;
+
+	codeText.value = props.node.state.codeHistory?.[0]?.code ?? '';
```
mwdchang, line 168: Are we starting with blank code-text or your default code-text from above?

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-node.vue
```diff
@@ -0,0 +1,21 @@
...
+import { WorkflowNode } from '@/types/workflow';
+import Button from 'primevue/button';
+import { DecapodesOperationState } from './decapodes-operation';
+
+defineProps<{
+	node: WorkflowNode<DecapodesOperationState>;
+}>();
+
+const emit = defineEmits(['update-state', 'append-output', 'open-drilldown']);
+</script>
```
mwdchang, line 17: Seem like very little thing is actual in-use: e.g. emit, props, maybe some clean up?

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
blanchco: changed to static string for `contextlanguage`

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
blanchco: it now defaults to the `defaultText`

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
blanchco: changed!

packages/client/hmi-client/src/workflow/ops/decapodes/tera-decapodes-drilldown.vue
blanchco: true, I've removed the optional for node.  I sort of like to computed here so that I can use it everywhere I need it in the notebook (which is only 2 places in this case 😄 ) rather than write out something like `modelId = props.node....` everywhere i need it but happy to change that if there an issue with it being computed

mwdchang (approved):

PR #2994, opened 2024-03-12
fix: once a new output is generated update the selected output to it

shawnyama:
# Description

Before when a new output was generated it appeared to be the chosen one once the output list was updated. In reality it was just a side effect of the list changing. I now select the new output after a new one is added (within `appendOutput()`)

In the old version the new output looked like it was the selected one after a Run. However it was missing the connected label that should have been beside it. 

https://github.com/DARPA-ASKEM/terarium/assets/28237258/388c6f29-895c-4e24-ae54-1c94ff67fa70

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-output-dropdown.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-output-dropdown.vue
@@ -29,16 +29,29 @@
 </template>
 
 <script setup lang="ts">
+import { watch } from 'vue';
+import { isEmpty } from 'lodash';
 import { WorkflowOutput, WorkflowPortStatus } from '@/types/workflow';
 import Dropdown from 'primevue/dropdown';
 
-defineProps<{
+const props = defineProps<{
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/drilldown/tera-output-dropdown.vue
```diff
@@ -29,16 +29,29 @@
...
+// When a new output is added to the list choose it by default
+watch(
+	() => props.options,
+	(newOptions) => {
+		if (!isEmpty(newOptions)) {
+			const { items } = newOptions[0];
+			emit('update:selection', items[items.length - 1].id);
+		}
+	}
+);
```
mwdchang, line 54: Looks a little suspicious - let's have a sync on what are the symptoms and problems. 

As long as you invoke `append-output`, thing in theory should take care of itself, and while append-output can be triggered from update:selection, I don't think this is something you'd want to automatically invoke.

mwdchang (approved):

PR #2998, opened 2024-03-12
feat: save model compare state

shawnyama:
# Description

The state of structural model comparison is saved. Code is saved similar to other operators. Structural comparison image ids are saved into the state. Once the operator is opened the images are fetched by their id.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/dd3fd6ac-d9f3-4aa3-9557-654d238a1370


packages/client/hmi-client/src/workflow/ops/model-comparison/model-comparison-operation.ts
```diff
@@ -1,6 +1,15 @@
...
+export interface NotebookHistory {
+	code: string;
+	timestamp: number;
+}
+
+export interface ModelComparisonOperationState {
+	notebookHistory: NotebookHistory[];
+	hasCodeRun: boolean;
+	structuralComparisons: string[];
+}
```
mwdchang, line 12: Can you give us some ideas and/or examples of what `notebookHistory` and `structualComparisons` are? 

We want to be careful of not jamming too much information into a workflow because it is saved as a single document, so subsequent saving and retrieval can become quite prohibitive.

shawnyama: `structuralComparisons` hold the images that compare the models. 

As for `notebookHistory` I am following a similar pattern to `model-edit` and `model-config` notebooks where the code is saved within `codeHistory`. I was thinking to keep things consistent. There isn't an output planned for this component though so I am also thinking to just save a normal `code` state.

mwdchang: It may be worthwhile to take an hour to see if we can maybe save the image(s) to Redis instead of into the workflow...I am not sure what support we have in place so if it is too annoying to do I am okay with saving into workflow - but it will for sure slow things down a bit.

E.g. thinking something along this lines:
```
image = fetch_image_from_redis(imageId)
if image === null:
  image = genrate_image(code_history_or_whateever)
  imageId = cache_image(image)
```

Tom-Szendrey: Is that you [volunteering](https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/workflow/ops/model-edit/tera-model-edit.vue#L317)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ImageService.java
YohannParis, line 1: Is this saved in the file bucket as is?

packages/client/hmi-client/src/services/notebook.ts
```diff
@@ -0,0 +1,25 @@
...
+	// for now only save the last code executed, may want to save all code executed in the future
+	const notebookHistoryLength = node.state.notebookHistory.length;
+	const timestamp = Date.now();
+	if (notebookHistoryLength > 0) {
+		state.notebookHistory[0] = { code, timestamp };
+	} else {
+		state.notebookHistory.push({ code, timestamp });
+	}
+	return state;
+};
```
mwdchang, line 25: You may need to do a few sanity checks and not assume that `node.state` has all the fields that you are using.

packages/client/hmi-client/src/workflow/ops/model-comparison/tera-model-comparison.vue
```diff
@@ -222,9 +226,29 @@ function formatField(field: string) {
...
 
-// function submitGollmQuestion() {
-// 	console.log(gollmQuestion.value);
-// }
+function updateImagesState(newImageId?: string, clearImages = false) {
+	const state = cloneDeep(props.node.state);
+	if (newImageId) state.comparisonImageIds.push(newImageId);
+	if (clearImages) state.comparisonImageIds = [];
+	emit('update-state', state);
+}
```
mwdchang, line 234: Avoid using disjunctions and optionals in function signatures, it leads to confusion and bug, and is generally a bad design.

People should be able to read the function signature and reason with what it is doing, you cannot do that if the parameters are self-contradictory.

If you want to squeeze multiple features into a single function, the pattern I would recommend is to pass in an operator type and a data structure. So in your case it would be:

```
function updateImageState(operationType: string, imageIds: string[] | null = null) {
   ...
}
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ImageService.java
bigglesandginger: This writes a PNG to the S3 bucket (the bas64 is decoded to a "proper" PNG file)

YohannParis: fixed it myself

PR #3009, opened 2024-03-13
Replace bad defaults with actual filename

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -1,7 +1,26 @@
 package software.uncharted.terarium.hmiserver.controller.mira;
 
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
```
YohannParis, line 5: We need to sync our formatters... because VSCode put those Java util above the other imports, but IntelliJIDE does the opposite...
Same thing for tabs, etc.
@dgauldie

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/MdlToStockflowResponseHandler.java
```diff
@@ -34,7 +35,13 @@ public static class Response {
...
 		try {
 			final Response modelResp = objectMapper.readValue(resp.getOutput(), Response.class);
-			final Model model = modelService.createAsset(modelResp.getResponse());
+			Model model = modelResp.getResponse();
+			final ConversionAdditionalProperties props = resp
+					.getAdditionalProperties(ConversionAdditionalProperties.class);
+			// override the default stockflow name / description
+			model.getHeader().setName(props.getFileName());
+			model.getHeader().setDescription(props.getFileName());
+			model = modelService.createAsset(model);
```
YohannParis, line 44: We shoulsd also set the `TerariumAsset.name`
Also are you sure the `getFileName` is the best name, should we read the name from the json content?
```suggestion
			model.getHeader().setDescription(props.getFileName());
			model.setName(props.getFileName());
			model = modelService.createAsset(model);
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
dgauldie: yeah... not sure what kevin is using. We can sync up outside of this PR

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/MdlToStockflowResponseHandler.java
dgauldie: fixed. I'm pretty sure that the problem is the json content. It defaults to "SIR model"

YohannParis (approved):

PR #3010, opened 2024-03-13
Semantics table rework pt1

blanchco:
# Description

* rework the semantic tables so that they can be used individually everywhere (simply passing in a model configuration)
* added some columns to parameter and initial tables

Screenshot 2024-03-13 at 1 09 55 PM


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
```diff
@@ -44,9 +44,12 @@ public class ModelMetadata extends SupportAdditionalProperties implements Serial
...
 
-	/* Link user input string `source` to a parameter/variables of a model. */
-	@TSOptional 
-	private Map<String, Object> sources;
+
+	@TSOptional
+	private Map<String, Object> initials;
+
+	@TSOptional
+	private Map<String, Object> parameters;
```
blanchco, line 52: I could see the metadata becoming a little bloated, so Ive added initials and parameters here where we can place any metadata we want: i.e

```json
...metadata.initials: {
    s: {
         description: '',
         unit: '',
         source: '',
    }
}
```
TODO move timeseries over as well

mwdchang: Can you elaborate on what `metadata.initials` and `metadata.parameters` are used for? 

If you are talking about metadata for model variables they have their own schemes in their respective model-representation, and if kept out side of that it may cause problems downstream.

packages/client/hmi-client/src/workflow/ops/model-config/tera-parameter-table.vue
```diff
@@ -245,39 +284,153 @@ const typeOptions = [
...
+				} else {
+					param = props.modelConfiguration.configuration.semantics.ode.parameters.find(
+						(i) => i.id === v
+					);
+				}
+				const paramType = getParamType(param);
+				const timeseriesValue =
+					props.modelConfiguration.configuration.metadata?.timeseries[param!.id];
+				const parametersMetadata =
+					props.modelConfiguration.configuration.metadata?.parameters?.[param!.id];
```
mwdchang, line 348: It is easier to read if you don't repeat `props.modelConfiguration.configuration` so much, it is repeated 8 or 9 times here, can we just put that into a variable at the top of the function?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
blanchco: Forgive me if Im wrong, I thought the AMR schema was pretty static aside from `metadata` key where we could add anything we wanted, which is why we used the timeseries in metadata.  So for instance an initial has this structure:
```
export interface Initial {
    target: string;
    expression: string;
    expression_mathml: string;
}
```

But now I want an initial to have a `unit`, `description`, `source`.  I would happily add it to the Initial type, but as I mentioned before I recall we couldn't touch these things.

So now similar to the timeseries, I have a map on the metadata for initials which will have these additional metadata parameters, which we can then map in the front end.

We can discuss in the meeting further!

packages/client/hmi-client/src/workflow/ops/model-config/tera-parameter-table.vue
blanchco: done 😄

mwdchang (approved):

PR #3011, opened 2024-03-14
reorganize ciemss calibrate

mwdchang:
### Summary
- Sync ciemss-calibrate drilldown/node interactions and add in resuming logic for long running calibrations
- Tidy up code

Note unlike other operators, ciemss-calibrate is a two-stage process, so we need to track extra data, namely
- inProgressCalibrateId, which points to the dill file, and
- inProgressForecastId, which points to the simulation result

TODO in other PR(s)
- better handling of loss values and chart's variable selections, there seemed to be a few bugs 
- figure out what is going on with the state names coming back, we used to parse for `_state` but they don't seem to be there anymore, however @Tom-Szendrey is reporting parsing for `_state` in his optimization changes, so there may be some inconsistent naming issues we need to report.

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
@@ -166,10 +166,6 @@ import Column from 'primevue/column';
 import {
 	getRunResultCiemss,
 	makeCalibrateJobCiemss,
-	simulationPollAction,
-	querySimulationInProgress,
-	getCalibrateBlobURL,
-	makeForecastJobCiemss,
 	subscribeToUpdateMessages,
 	unsubscribeToUpdateMessages
 } from '@/services/models/simulation-service';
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
```diff
@@ -532,16 +392,23 @@ onMounted(async () => {
...
-	// currently in progress
-	if (runIds.length > 0) {
-		getCalibrateStatus(runIds[0]);
-	}
 });
 
-onUnmounted(() => {
-	poller.stop();
-});
+onUnmounted(() => {});
```
Tom-Szendrey, line 397: Can we remove?

Tom-Szendrey (approved):

packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
mwdchang: done

PR #3015, opened 2024-03-14
Add risk chart to optimize output panel

Tom-Szendrey:
# Description

Adding risk chart to the optimize output panel

Minor updates to API call as well as it has been ever changing
 -  Extra now includes solver method
 - Threshhold now properly used
 - risk now 100 - risk as requested

image


packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss.vue
```diff
@@ -570,6 +588,12 @@ const saveModelConfiguration = async () => {
 const setOutputValues = async () => {
 	const output = await getRunResultCiemss(knobs.value.forecastRunId);
 	simulationRunResults.value[knobs.value.forecastRunId] = output.runResults;
+	riskResults.value[knobs.value.forecastRunId] = await getRunResult(
+		knobs.value.forecastRunId,
+		'risk.json'
+	);
+	console.log(riskResults.value[knobs.value.forecastRunId]);
```
YohannParis, line 595:
```suggestion
```

packages/client/hmi-client/src/workflow/tera-optimize-chart.vue
```diff
@@ -0,0 +1,176 @@
...
+
+<style scoped>
+.multiselect-title {
+	font-size: smaller;
+	font-weight: 700;
+}
+
+.p-chart {
+	width: 100%;
+	/* height: 200px; */
```
YohannParis, line 164:
```suggestion
```

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-optimize-chart.vue
```diff
@@ -0,0 +1,175 @@
...
+	const minValue = data[0]; // data.getMin
+	const maxValue = data[data.length - 1]; // data.getMax
+	const stepSize = (maxValue - minValue) / (binCount - 1);
+	const bins: number[] = Array<number>(binCount).fill(0);
+	const binLabels: string[] = [];
+	for (let i = 0; i < binCount; i++) {
+		binLabels.push(
+			`${(minValue + stepSize * i).toFixed(4).toString()} - ${(minValue + stepSize * (i + 1))
+				.toFixed(4)
+				.toString()}`
```
mwdchang, line 98: `toFixed(...)` is already returning a string, you don't need the extra `toString(...)`

packages/client/hmi-client/src/workflow/tera-optimize-chart.vue
```diff
@@ -0,0 +1,175 @@
...
+			grid: {
+				color: '#fff',
+				borderColor: '#fff'
+			}
+		}
+	}
+});
+
+const getBinData = (data: number[]) => {
+	data.sort();
```
mwdchang, line 88: `sort(...)` is in place, e.g. it changes the source where `data` is pointing to, is this going to create a problem?

You might want to just use Math.min and Math.max which do not alter the underlying data.

packages/client/hmi-client/src/workflow/tera-optimize-chart.vue
```diff
@@ -0,0 +1,175 @@
...
+
+.p-chart {
+	width: 100%;
+	margin-top: 0.5em;
+}
+
+.p-multiselect {
+	width: 100%;
+	border-color: lightgrey;
+}
```
mwdchang, line 170: Please clean up css class assignments, you unused stuff declared here.
Likewise your templates have css-classes that point to nowhere. e.g. `class="simulate-chart"`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/OptimizeRequestCiemss.java
```diff
@@ -22,7 +22,7 @@ public class OptimizeRequestCiemss implements Serializable {
 	private String modelConfigId;
 	private TimeSpan timespan;
 	@TSOptional
-	private List<OptimizedIntervention> interventions;
+	private OptimizedIntervention interventions;
```
mwdchang, line 25: This part is a bit confusing since you are intermingling singular and plural variable names.

Can we make the language more consistent? 
If there is a todo like "we need to support a list of interventions some time in the future", please mark a FIXME some where.

Tom-Szendrey: Fair point. I will change OptimizedIntervention -> OptimizedIntervention**s** in a follow up PR
I will update pyciemss-service too so its up to date

PR #3017, opened 2024-03-14
Improve some error message

YohannParis:
# Description

* Improve the message from some services that just throw a lot of 500
* Revert error toast to disapears, but after 9 seconds.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -67,22 +67,6 @@ echo $CR_PAT | docker login ghcr.io -u USERNAME --password-stdin
 Login Succeeded
 ```
 
-## Error Message
-
-Effective error messaging is not about informing users of a problem; It’s about turning a negative experiences into
-opportunities for positive interactions and continued user engagement.
-
-* Be clear and concise. Use plain language and avoid passive voice to ensure users understand the message. Avoid
-  technical jargon that may confuse users.
-* Be specific and relevant by providing specific information about the error. Instead of a generic "Something went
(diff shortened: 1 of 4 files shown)


.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,22 @@ echo $CR_PAT | docker login ghcr.io -u USERNAME --password-stdin
...
+  technical jargon that may confuse users.
+* Be specific and relevant by providing specific information about the error. Instead of a generic "Something went
+  wrong" describe the nature of the error, like "Unable to save your changes because the server is temporarily
+  unavailable.”
+* Offer a solution or the next possible step, guide users on what to do next. If it's a simple fix they can perform,
+  include that in the message. If not, provide information on how to get further help.
+* Avoid negative words like "fail" or "error" that can increase user anxiety. Opt for neutral language that doesn't
+  assign blame and focus on the solution rather than the problem.
+  When in doubt please reach to the UX team for guidance.
+
```
mwdchang, line 85: Did you derive this from ChatGPT?  It sounds a bit weird when taken-in as a whole, feels like some of the points are contradicting each other.

I think we need to know our target users; if we expect them to be able to pick from an array of ODE solvers, understand monotonicity and setting intervention policies - they should (and probably appreciate) a detailed error message or two.

I would recommend we keep point-2 and point-3 and trash the rest, and remove the "Effective error messaging is not about informing users of a problem" part.

dvince2 (approved):

PR #3018, opened 2024-03-15
semantic table change pt2

blanchco:
# Description

*tables are now based off model

https://github.com/DARPA-ASKEM/terarium/assets/33158416/04e955eb-282d-4ebf-a75f-3320ed8af8e2

Follows this schema whether a field is editable or not:
# Model
Symbol
Name -> Editable
Description -> Editable
Concept -> Editable
Unit ->Editable
Value Type -> Editable
Value -> Editable
Source->Editable
# Model Configuration
Symbol
Name
Description
Concept
Unit -> Editable
Value Type -> Editable
Value-> Editable
Source-> Editable

Resolves #(issue)


mwdchang:
This has massive overlap with my changes and not exactly compatible, we should sync up.

blanchco:
> This has massive overlap with my changes and not exactly compatible, we should sync up.

Ya let's sync up Monday, just a general summary is that I'm switching to use model instead of model config as a prop since these need to be used in the model table.  Hope it isn't too bad!

YohannParis (approved):

PR #3020, opened 2024-03-15
Jamie make transform dataset notebook prettier

jamiewaese-uncharted:
Big face lift here. Untangled spaghetti and baked into a lasagna. My best suggestion is that you open this up and give it a shot. It seems to work well on my machine.

BEFORE
image

AFTER
image


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -380,7 +380,7 @@ public ResponseEntity<ProjectAsset> createAsset(
 					}
 				}
 			}
-			return ResponseEntity.status(HttpStatus.NOT_MODIFIED).build();
+			return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
```
dvince2, line 383: I don't think this is right. This will return a 401, which our front end will then capture and redirect you to the keycloak login page.

YohannParis (approved):

PR #3023, opened 2024-03-18
Feat: simulate ensemble appending ports

Tom-Szendrey:
# Description
Rather than 1 port that accepts multiple inputs, update the design for simulate ensemble to have appending ports

image

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
@@ -344,7 +344,10 @@ const runEnsemble = async () => {
 
 onMounted(async () => {
 	// FIXME: probably switch to multiport instead of multivalue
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
```diff
@@ -343,8 +343,10 @@ const runEnsemble = async () => {
 };
 
 onMounted(async () => {
-	// FIXME: probably switch to multiport instead of multivalue
-	const modelConfigurationIds = props.node.inputs[0]?.value;
+	const modelConfigurationIds: string[] = [];
+	props.node.inputs.forEach((ele) => {
```
Tom-Szendrey, line 347: Note not sure how to use `map` cleanly here as at least one ele (the last) will be `undefined` and I dont want to spend time looking into it

mwdchang: This is fine.

mwdchang (approved):

PR #3028, opened 2024-03-18
Jamie model page adjustments per pascale

jamiewaese-uncharted:
This was a Pascale request.

BEFORE
I didn't record a screenshot, but a lot of times when you scroll down the model page, if your mouse happens to be over the diagram when you scroll, the diagram zooms in or out instead of the page scrolling as expected.

AFTER
I added a variable called isLocked, and a button to toggle its status. When it's locked, pointer events are set to none. When it's unlocked, pointer events return and a green border is drawn around the section.

Unlock model diagram to adjust


shawnyama:
I think "Unlock to pan/zoom" would be a better name for the button

shawnyama (approved):

PR #3030, opened 2024-03-18
Remove last three KM endpoints

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/Dataset.java
```diff
@@ -73,7 +74,7 @@ public class Dataset extends TerariumAsset {
 	 * (Optional) Unformatted metadata about the dataset
 	 **/
 	@TSOptional
-	private Object metadata;
+	private Map<String, Object> metadata;
```
YohannParis, line 77: Why do we need to do that?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/Dataset.java
kbirk: The `profile_dataset` adds the card under the `"data_card"` field. So we at least need the metadata to be some "dictionary like" object.

YohannParis (approved):

bigglesandginger (approved):
Let's rock!

PR #3034, opened 2024-03-18
Added user content editor in the overview page

jryu01:
# Description

https://github.com/DARPA-ASKEM/terarium/assets/3279838/e531d78f-c945-4458-9d52-77672ff8382a

- Added text editor in overview page. 
- Automatically saves editor content change in every 3 seconds. 

Resolves #(issue)


package-lock.json
dvince2, line 1: I think this was unintentional

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
```diff
@@ -0,0 +1,117 @@
...
+
+const isContentSameAsLastSaved = computed(
+	() =>
+		lastSavedContent.value.length === editorContent.value.length &&
+		lastSavedContent.value === editorContent.value
+);
+const saveContent = async () => {
+	if (!props.project) return;
+	if (isContentSameAsLastSaved.value) return;
+	const res = await update({ ...props.project, overviewContent: editorContent.value });
```
YohannParis, line 58: Why not use `useProject` to update?

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
```diff
@@ -0,0 +1,117 @@
...
+					<li><strong>Upload stuff:</strong> Upload documents, models, code or datasets with the green button in the bottom left corner.</li>
+					<li><strong>Explore and add:</strong> Use the project selector in the top nav to switch to the Explorer where you can find documents, models and datasets that you can add to your project.</li>
+					<li><strong>Build a model:</strong> Create a model that fits just what you need.</li>
+					<li><strong>Create a workflow:</strong> Connect resources with operators so you can focus on the science and not the plumbing.</li>
+			</ul>
+	<br>
+			<p>Feel free to erase this text and make it your own.</p>
+	</div>
+`;
+const props = defineProps<{ project: Project | null }>();
```
mwdchang, line 33: I suspect that since we are on the project edit page, the project cannot be null.

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
jryu01: `useProject().update` fetches all projects again at the end and we don't need to do that in this page I think.

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
jryu01: I'm passing `activeProject` as a prop value from the parent e.g. `<tera-project-overview-editor :project="useProjects().activeProject.value" />` and it looks like `activeProject` is `Project | null`

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
YohannParis: You could update this method to not do that.

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
YohannParis: You do not need to pass it as a prop. You could simply call `useProjects().activeProject.value` within this component.

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
```diff
@@ -0,0 +1,117 @@
...
+			<ul>
+					<li><strong>Upload stuff:</strong> Upload documents, models, code or datasets with the green button in the bottom left corner.</li>
+					<li><strong>Explore and add:</strong> Use the project selector in the top nav to switch to the Explorer where you can find documents, models and datasets that you can add to your project.</li>
+					<li><strong>Build a model:</strong> Create a model that fits just what you need.</li>
+					<li><strong>Create a workflow:</strong> Connect resources with operators so you can focus on the science and not the plumbing.</li>
+			</ul>
+	<br>
+			<p>Feel free to erase this text and make it your own.</p>
+	</div>
+`;
```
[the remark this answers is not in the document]
jryu01: Done

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
jryu01: I just removed the props and used the composable.

packages/client/hmi-client/src/components/home/tera-project-overview-editor.vue
jryu01: I'm leaning toward to just use the `update` from the service here since it's simpler for this purpose and has no side effect to deal with.

YohannParis (approved):

PR #3035, opened 2024-03-18
multicharts for ciemss-simulate, reorg chart actions

mwdchang:
### Summary
- Allow node previews to show multiple charts instead of the first one.
- Reorg chart related actions (add, change config) to reduce duplicated logic across TA3 operators

NOTE: Existing **julia-calibrate** and **ciemss-simulate-ensemble** may not work properly, the two operators were using a modified version of chart-configuration that have now unified to the same data structure. 
We will need to run a re-indexer with something like this:

```
node.state.chartConfigs = node.state.chartConfigs.map((d: any) => d.selectedVariable);
```

TODO: We will probably want to add a remove/delete chart interaction.

image


Tom-Szendrey (approved):

packages/client/hmi-client/src/workflow/util.ts
```diff
@@ -1,6 +1,35 @@
-import { DataseriesConfig, RunType } from '@/types/SimulateConfig';
+import _ from 'lodash';
+import { DataseriesConfig, RunType, ChartConfig } from '@/types/SimulateConfig';
 import type { CsvAsset, TimeSpan } from '@/types/Types';
 
+export const chartActionsProxy = (state: any, updateStateCallback: Function) => {
```
YohannParis, line 5: A comment to explain the purpose of this function will go a long way.

YohannParis (approved):

YohannParis (approved):

PR #3037, opened 2024-03-18
Jamie prettify simulate with py ciemss

jamiewaese-uncharted:
Pascale asked me to fix the chart width issue.

BEFORE
image

AFTER
I also added the example string showing the time slices you selected, per the design. 

Monosnap Terarium 2024-03-18 16-50-17


packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -156,6 +172,23 @@ const emit = defineEmits(['append-output', 'update-state', 'select-output', 'clo
...
+		const value = start + i * step;
+		if (value % 1 === 0) {
+			values.push(Math.floor(value));
+		} else {
+			values.push(value.toFixed(2));
+		}
+	}
+	return values.join(', ');
+});
+
```
mwdchang, line 191: Number of samples is how many times you want to run the simulation, it has nothing to do with step-size. This exampleString is not necessary in CIEMSS, stepsize is always 1.

packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -198,11 +231,19 @@ const selectedRunId = computed(
 	() => props.node.outputs.find((o) => o.id === selectedOutputId.value)?.value?.[0]
 );
 
+const parentSize = ref({ width: 0, height: 270 }); // Set the initial height or any default height
+// Calculate the parent container's width
+onMounted(async () => {
+	const parentContainerWidth = document.querySelector('#output-panel').clientWidth - 48;
+	parentSize.value.width = parentContainerWidth;
+});
```
mwdchang, line 239: Do not identify elements by id, it doesn't always work. If you need to access HTML elements in Vue, use `ref`.

Also this function does not need to be `async`.

packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
jamiewaese-uncharted: Sorry, misunderstanding on my part based on some old designs. Removed.

packages/client/hmi-client/src/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
jamiewaese-uncharted: fixed. thanks.

mwdchang (approved):

PR #3048, opened 2024-03-19
Task: Change optimize output to be a simulation ID

Tom-Szendrey:
We are not currently saving interventions (see https://github.com/DARPA-ASKEM/terarium/pull/2990) so for now to be more usable we will output the simulation ID. 
image


YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss.vue
```diff
@@ -494,6 +494,14 @@ const runOptimize = async () => {
...
 	const simulationResponse = await makeForecastJobCiemss(simulationPayload);
 	getStatus(simulationResponse.id);
+	const state = _.cloneDeep(props.node.state);
+	emit('append-output', {
+		type: OptimizeCiemssOperation.outputs[0].type,
+		label: `Simulation output - ${props.node.outputs.length + 1}`,
+		value: simulationResponse.id,
+		isSelected: false,
+		state
+	});
```
mwdchang, line 504: I think `value` needs to be wrapped in an array`value: [simulationResponse.id]`

Tom-Szendrey: Youre right, i will change this in ciemss node too

PR #3053, opened 2024-03-19
apply suggested configs

blanchco:
# Description

* Apply suggested configs from table

https://github.com/DARPA-ASKEM/terarium/assets/33158416/cf1552a8-7183-4088-9ed3-93e7ecfdc9c9


YohannParis (approved):
Just add the `Symbol` column in the modal and good to go!

PR #3061, opened 2024-03-19
added edit mode to tables

jamiewaese-uncharted:
Added an "edit mode" for the Model variables and parameters table, per Pascale's request.

Frankly, it's a hack job. This whole thing is a bunch of interconnected spaghetti. Seems to work though... but it's ugly and the concepts are still editable when you're not in edit mode. And so are the Value Type dropdowns. TBH, this whole table should be re-written, now that we know what it's supposed to be.

edit mode for tables


YohannParis:
This will be handle with a new UI/UX of the whole component.

PR #3068, opened 2024-03-20
Update to more generic progress

kbirk:
Optional PR that modifies the extraction progress to be even more generic, and created an `estimated` progress value of t using a logarithmic easing function so that the first 50% takes n seconds, the next 25% takes n, next 12.5% takes n... etc.

Coupled with a css transition it behaves really nicely:

https://jsfiddle.net/hwpdb6yv/1/

 The main motivation for this is that it is less error prone since it doesn't rely on hardcoding exactly how many updates will be sent, and can be used in cases where that number isn't known ahead of time.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extractionservice/ExtractionStatusUpdate.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extractionservice/ExtractionStatusUpdate.java
@@ -1,24 +1,22 @@
 package software.uncharted.terarium.hmiserver.models.extractionservice;
 
+import java.util.UUID;
+
 import lombok.Data;
 import software.uncharted.terarium.hmiserver.annotations.TSModel;
-
-import java.util.UUID;
 
 @Data
 @TSModel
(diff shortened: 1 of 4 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extractionservice/ExtractionStatusUpdate.java
```diff
@@ -1,24 +1,22 @@
...
-    public ExtractionStatusUpdate(UUID documentId, Integer step, Integer totalSteps, String message, String error) {
-        this.documentId = documentId;
-        this.step = step;
-        this.totalSteps = totalSteps;
-        this.message = message;
-        this.error = error;
-    }
+	public ExtractionStatusUpdate(final UUID documentId, final Double t, final String message, final String error) {
+		this.documentId = documentId;
+		this.t = t;
```
YohannParis, line 18: Is this a percentage or a time value?
```suggestion
		this.percentage = percentage;
```

YohannParis (approved):

PR #3071, opened 2024-03-20
Jamie fix tooltip swatch colors in simulation chart

jamiewaese-uncharted:
**BEFORE**
Because the simulation chart uses semi-transparent lines, the tooltip color swatch is barely visible. It's nearly impossible to tell what line is what. 

Also, it's bad practice to encode color information with text alone. Our multiselect dropdown cleverly adjusts the text colors to double as a legend, but it's hard to read.
image

**AFTER**
I fixed both issues. I also reduced the opacity of the lines a bit, so you see them on the chart better.
image


packages/client/hmi-client/src/workflow/tera-simulate-chart.vue
```diff
@@ -286,6 +299,12 @@ onMounted(() => {
 </script>
 
 <style scoped>
+.custom-chip {
+	border: 1px solid var(--surface-border-light);
+	border-radius: var(--border-radius-bigger);
+	padding: var(--gap-xsmall) var(--gap);
+	color: var(--surface-0);
+}
```
YohannParis, line 307: 🤩

YohannParis (approved):

PR #3072, opened 2024-03-20
fix[hmi-client]: Handle Missing Queued Task

dvince2:
# Description

If a task is saved to the database and then not retreived in good time (IE hmi server gets shut down?) the climate data service will delete the status. However rather than returning a 404, they return a 200 with a body of 404 which blows everything up.

Resolves #(issue)


YohannParis:
@satchelbaldwin could we improve that on the climate-data service as well? I bet those are ESGF errors being bundle up.

YohannParis (approved):

PR #3078, opened 2024-03-20
Ts/3066 ensemble cleanups 3

Tom-Szendrey:
# Description
-Model Names rather than Ids are displayed to user.
-Rename weight enum to be more accurate
-disable run button if user provides no label

image

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
+++ b/packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
@@ -14,23 +14,23 @@
 							<label>
 								<input
 									type="radio"
-									v-model="ensembleCalibrationMode"
-									:value="EnsembleCalibrationMode.EQUALWEIGHTS"
+									v-model="ensembleWeightsMode"
+									:value="EnsembleWeightsMode.EQUALWEIGHTS"
 								/>
-								{{ EnsembleCalibrationMode.EQUALWEIGHTS }}
+								{{ EnsembleWeightsMode.EQUALWEIGHTS }}
 							</label>
 							<label>
 								<input
 									type="radio"
-									v-model="ensembleCalibrationMode"
-									:value="EnsembleCalibrationMode.CUSTOM"
+									v-model="ensembleWeightsMode"
+									:value="EnsembleWeightsMode.CUSTOM"
 								/>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
```diff
@@ -43,7 +43,7 @@
 									<th>Weight</th>
 								</thead>
 								<tbody class="p-datatable-tbody">
-									<tr v-for="(id, i) in ensembleConfigs.map((ele) => ele.id)" :key="i">
+									<tr v-for="(id, i) in listModelLabels" :key="i">
```
Tom-Szendrey, line 46: I am using index matching here. Im not sure if i should
add a frontend version of EnsembleModelConfigs that includes the name to avoid index matching

Im not a huge fan of index matching but also having a frontend extension of an existing object seems bleh too. thoughts?

packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
```diff
@@ -205,7 +207,7 @@ const MINBARLENGTH = 1;
 const showSpinner = ref(false);
 
 const listModelLabels = ref<string[]>([]);
-const ensembleCalibrationMode = ref<string>(EnsembleCalibrationMode.EQUALWEIGHTS);
+const ensembleWeightsMode = ref<string>(EnsembleWeightsMode.EQUALWEIGHTS);
```
shawnyama, line 210: Don't type this and it'll be autoset to `EnsembleWeightsMode` type

Tom-Szendrey: My mistake, thank you for catching this

packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss.vue
shawnyama: Maybe a computed variable could act as the frontend extension...not sure how this all works though. If it works for now maybe just save this for later

shawnyama (approved):

PR #3095, opened 2024-03-21
Update tera-parameter-table.vue

blanchco:
# Description

* add source for dataset/document when assigning params

Resolves #(issue)


YohannParis (approved):
This will only work for `PetriNet` right? Could you add it to the other table you made for the other frameworks. As well as initials.

blanchco:
> This will only work for `PetriNet` right? Could you add it to the other table you made for the other frameworks. As well as initials.

This will work for all model types since its only the metadata we are editing.  Ill try to add the suggested configs button the on the initials before the end of the day

YohannParis (approved):

PR #3099, opened 2024-03-21
Fix the model card request in extraction service

kbirk:
This moved the variable extraction code around to be called as well, but it currently times out and fails. I don't think it can handle the amount of text we send it?

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/AddDocumentAssetFromXDDRequest.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/AddDocumentAssetFromXDDRequest.java
@@ -1,11 +1,11 @@
 package software.uncharted.terarium.hmiserver.models.dataservice.document;
 
+import java.util.UUID;
+
 import lombok.Data;
 import lombok.experimental.Accessors;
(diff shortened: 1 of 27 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/AddDocumentAssetFromXDDRequest.java
```diff
@@ -15,4 +15,6 @@ public class AddDocumentAssetFromXDDRequest {
 	private Document document;
 
 	private UUID projectId;
+
+	private String domain = "epi"; // default value
```
dvince2, line 19: This feels like something we [should] have an enum for

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/AddDocumentAssetFromXDDRequest.java
YohannParis: Yes it should at the APP level. Let's table that for our talk on April 3

YohannParis (approved):

PR #3101, opened 2024-03-21
Optimize -> Proxy Chart

Tom-Szendrey:
# Description
Updating optimize to use the new proxy chart, thanks Dan!


mwdchang (approved):
thanks!

PR #3102, opened 2024-03-21
Jamie prettify ensemble simulate

jamiewaese-uncharted:
General prettification. Made the mapping section more better.

@mwdchang I did not apply your changes to the simulation output chart. This is still the version that doesn't fill the container.

**BEFORE**
image

**AFTER**
simulate ensemble - prettified


Tom-Szendrey (approved):
Thank you for this!

PR #3110, opened 2024-03-21
deal with new preview output

bigglesandginger:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/climateData/ClimateDataResultPreviews.java
YohannParis, line 1: Why not a map<year, image>?

YohannParis (approved):

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/climateData/ClimateDataResultPreviews.java
bigglesandginger: Because there is only one image in the preview - the subset gets a series (at least that is what I saw in the test request I did)

PR #3125, opened 2024-03-22
feat: show pyciemss errors

shawnyama:
# Description

I am using the operator state to save the error message. This is a temporary solution for eval purposes, we will use redis later instead.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/7b7b03c5-7e90-4898-9a82-14252a07ca87


YohannParis:
Why this is not visible right away?

shawnyama:
> Why this is not visible right away?

I am stuck on this - it should update but it doesn't

shawnyama:
This is fixed now.

YohannParis (approved):
I would have prefered to have those error message typed and used everywhere we have a preview panel. But that can be an new PR. Also check with @ngraham76 and @jamiewaese-uncharted because now we have an orange border when the preview is in an error state.
```typescript
interface previewError {
    name?: string;
    value?: string; 
    traceback?: string 
}
```

PR #3127, opened 2024-03-22
Better errors for KnowledgeController, log 4XX errors

kbirk:


dgauldie (approved):
tested, looks good as a first pass

PR #3131, opened 2024-03-22
fix[hmi-server]: Improving xDD errors

dvince2:
# Description

Improving the errors from xdd document and extraction search. If xDD gives us an exception we return that status code to the front end with an error message. In all other cases it's a server error, so we return 500


dgauldie (approved):
LGTM

PR #3138, opened 2024-03-24
add logging of 403 and 401 errors

dgauldie:
# Description

Resolves #(issue)


dgauldie:
I tested that 401 errors get logged, but I have not tested 403s. If anyone knows how to, please do so.

Also, if anyone wants to write tests for these new logging handlers, that would be great

bigglesandginger:
`KeycloakJwtAuthenticationConverter.extractResourceRoles()` checks to see if the user "isDirty" and if so updates the database (setting the user record and deleting all roles then adding all new roles)... will this fix the issue or make the issue revert to the old issue that PostGres will become hammered with database calls for every single request issued...  (what the caching fixed)?

bigglesandginger (approved):

PR #3143, opened 2024-03-25
Correct regrid context.

Tom-Szendrey:
May want to restyle the contextInfo on vue.
Also will need to clear out console.logs
image

image


bigglesandginger (approved):
Certainly more correct that what I wrote

PR #3160, opened 2024-03-25
fix[hmi-server]: Cosmos extraction fix

dvince2:
This actually works - tested with @iross and matt


iross:
Praise belongs to @mwestphall :

>...I did a bit of digging into the structure of the form data that that requests.post call generates, looks like it's along these lines:
> ```
> Headers:
> {'Content-Length': '350', 'Content-Type': 'multipart/form-data; boundary=SEPARATOR'}
> Body:
> --SEPARATOR
> Content-Disposition: form-data; name="compress_images"
> 
> False
> --SEPARATOR
> Content-Disposition: form-data; name="use_cache"
> 
> False
> --SEPARATOR
> Content-Disposition: form-data; name="pdf"; filename="sample.txt"
> 
> hello, world
> --SEPARATOR--
> ```
> So all 3 fields should go into the request body as different parts of a single multipart form

YohannParis (approved):

PR #3169, opened 2024-03-26
Download file for Dataset

YohannParis:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/dataset/tera-dataset.vue
+++ b/packages/client/hmi-client/src/components/dataset/tera-dataset.vue
@@ -55,8 +55,14 @@
 <script setup lang="ts">
 import { computed, onUpdated, PropType, Ref, ref, watch } from 'vue';
 import * as textUtil from '@/utils/text';
-import { cloneDeep, isString } from 'lodash';
-import { downloadRawFile, getClimateDataset, getDataset, updateDataset } from '@/services/dataset';
+import { cloneDeep, isEmpty, isString } from 'lodash';
+import {
+	downloadFile,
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -138,6 +138,14 @@ async function downloadRawFile(
 	return response?.data ?? null;
 }
 
+async function downloadFile(datasetId: string, filename: string): Promise<string | null> {
```
dvince2, line 141: I don't think this would work like that... `.../download-file` isn't returning a string, and then that string is treated like a url. That would be the `/download-url` endpoint we were discussing on the call no?

I believe that this endpoint is returning an `ArrayBuffer` like seen in `artifact.ts`?

dvince2 (changes requested):

dvince2 (approved):

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
YohannParis, line 1: The window open did not worked?
Otherwise can we just have the S3 link open in a new tab? Why do we have to pass the blob through?

dvince2: With the blob you don't get the file name - that can be done through a faked out `a` element click.

The initial PR had gone with the `download-file` endpoint which returns an arraybuffer/blob, and not `download-url`, so I rolled with that

YohannParis (commented):
LGTM

PR #3170, opened 2024-03-26
Protect getModel NPE error

bigglesandginger:
# Description

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
@@ -135,7 +135,11 @@ public class ModelController {
 							// Append the Document extractions to the Model extractions, just for the
 							// front-end.
 							// Those are NOT to be saved back to the data-service.
-							model.get().getMetadata().getAttributes().addAll(extractions);
+							if (extractions != null) {
+								model.get().getMetadata().getAttributes().addAll(extractions);
+							} else {
+								log.error("No attributes added to Model as DocumentAsset ({}) has no attributes.", documentId);
+							}
 						}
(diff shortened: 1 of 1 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -135,7 +135,11 @@ ResponseEntity<Model> getModel(@PathVariable("id") final UUID id) {
 							// Append the Document extractions to the Model extractions, just for the
 							// front-end.
 							// Those are NOT to be saved back to the data-service.
-							model.get().getMetadata().getAttributes().addAll(extractions);
+							if (extractions != null) {
+								model.get().getMetadata().getAttributes().addAll(extractions);
+							} else {
+								log.error("No attributes added to Model as DocumentAsset ({}) has no attributes.", documentId);
+							}
```
YohannParis, line 142: instead we should create an attributes and add those etractions to it.

YohannParis (changes requested):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
dgauldie: what extractions? they are null

bigglesandginger: The model's attributes are checked and created if missing (lines above).  The extractions are what is null - they come from the DocumentAsset that is read in from ES - why they are null is some other problem but not right here's fault.

YohannParis (approved):

dvince2 (approved):

PR #3177, opened 2024-03-26
fix[hmi-server]: Better error handling

dvince2:
Updating error handling

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
@@ -13,8 +13,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.http.HttpEntity;
 import org.apache.http.entity.ByteArrayEntity;
 import org.apache.http.entity.ContentType;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClients;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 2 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/CodeFile.java
```diff
@@ -15,29 +14,19 @@ public class CodeFile {
...
-        }
+        final String[] parts = fileName.split("\\.");
+        final String fileExtension = parts.length > 0 ? parts[parts.length - 1] : "";
+
+        final ProgrammingLanguage language = switch (fileExtension) {
+            case "py" -> ProgrammingLanguage.PYTHON;
+            case "jl" -> ProgrammingLanguage.Julia;
+            case "r" -> ProgrammingLanguage.R;
+            case "zip" -> ProgrammingLanguage.ZIP;
+            default -> null; // TODO: Do we need an "unknown" language?
```
YohannParis, line 27: I'll say by default we could define them as `ProgrammingLanguage.TXT`?

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -107,6 +105,8 @@ public ResponseEntity<UUID> equationsToModel(@RequestBody final JsonNode req) {
 		} catch (final FeignException e) {
 			final String error = "Skema Unified Service did not return any AMR based on the provided Equations";
 			log.error(error, e);
+			if(e.status()<100)
```
YohannParis, line 108: That is such a random number, do they return non HTTP error code sometimes?

YohannParis (approved):
LGTM

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
dvince2: SKEMA seems to love returning `-1` as an HTTP status code which has blown things up before... anything above 100 is a valid code (though anything 100-299 would be unexpected???)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/CodeFile.java
dvince2: I like that... but im not going to do it here because if we're hitting this path `null` is working for now and i dont want to blow things up :D

PR #3179, opened 2024-03-26
Fix Align AMR

YohannParis:
# Description

* Clean up some issue with the scenario form
* Update how we call the knowledgeController
* Move the extraction call into a runnable like PDF extraction
* Remove the optionals `annotateSkema` and `annotateMit` as Skema unified has them on by default.
* Remove the call to MIT as this is handle by the Skeme unified endpoint
* TODO - Move the align AMR into the extractionService
* TODO - Call align AMR after variable extraction if a model was provided
* TODO - Getting 500 when calling  SKEMA unified`link_amr`

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
@@ -145,6 +145,7 @@ const dialogActionCopy = computed(() => {
 	}
 	return `Use Document to ${result.toLowerCase()}`;
 });
+
 function openDialog() {
 	visible.value = true;
 }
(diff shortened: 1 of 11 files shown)


packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -183,62 +185,67 @@ const sendForEnrichment = async () => {
...
+	getRelatedDocuments();
 };
 
+function getRelatedDocuments() {
+	if (!props.assetType) return;
+
+	const provenanceType = mapAssetTypeToProvenanceType(props.assetType);
+	if (!provenanceType) return;
+
+	getRelatedArtifacts(props.assetId, provenanceType, [ProvenanceType.Document]).then((nodes) => {
```
dvince2, line 232: `getRelatedArtifacts` is still an async call - will need to `await` on this (and `getRelatedDocuments`).

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -360,145 +362,81 @@ public void run() {
...
+					 // add the attributes to the metadata
+					 if (document.getMetadata() == null) {
+						 document.setMetadata(new HashMap<>());
+					 }
+					 document.getMetadata().put("attributes", attributes);
+
+					 // update the document
+					 documentService.updateAsset(document).orElseThrow();
+
+				 } catch (final Exception e) {
```
dvince2, line 430: This will eat any exceptions thrown by the proxied services (skema unified, etc). Consider adding something like:

```
catch (final FeignException e) {
					final String error = "Transitive service failure";
							e.status()<100? HttpStatus.INTERNAL_SERVER_ERROR: HttpStatus.valueOf(e.status()),
							error + ": " + e.getMessage());
				}
```

dvince2 (approved):

PR #3181, opened 2024-03-27
sanity check warnings

mwdchang:
### Summary
Warnings when creating model-configurations

image

### Testing
model-configuration operator, use lower/upper bound of distribution of a parameter to be the same, try to run and should get a warning prompt dialog

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
@@ -670,7 +670,16 @@ const runSanityCheck = () => {
 		return errors;
 	}
 
-	modelToCheck.semantics?.ode?.parameters?.forEach((p) => {
+	let parameters: ModelParameter[] = [];
+	if ([AMRSchemaNames.PETRINET, AMRSchemaNames.STOCKFLOW].includes(modelType.value)) {
+		if (modelToCheck.semantics?.ode?.parameters) {
+			parameters = modelToCheck.semantics?.ode?.parameters;
+		}
+	} else if (modelToCheck.model.parameters) {
+		parameters = modelToCheck.model.parameters;
+	}
+
+	parameters.forEach((p) => {
 		const val = p.value || 0;
 		const max = p.distribution?.parameters.maximum;
 		const min = p.distribution?.parameters.minimum;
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -188,7 +188,7 @@
 							:disabled="isSaveDisabled"
 							label="Run"
 							icon="pi pi-play"
-							@click="createConfiguration"
+							@click="createConfiguration(false)"
```
YohannParis, line 191: Why adding the `false` when it is the default value?
```suggestion
							@click="createConfiguration"
```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -636,10 +662,49 @@ const updateConfigFromModel = (inputModel: Model) => {
...
+const createConfiguration = async (force: boolean = false) => {
 	if (!model.value) return;
 
 	const state = cloneDeep(props.node.state);
+
+	sanityCheckErrors.value = [];
+	if (force !== true) {
+		const errors = runSanityCheck();
+		if (errors.length > 0) {
+			console.log(errors);
```
YohannParis: Just to make it more digestible
```suggestion
      console.log("Model configuration sanity check errors: ");
			console.table(errors);
```

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -263,6 +263,30 @@
...
+				<section style="max-height: 22rem; overflow-y: scroll">
+					<div v-for="(errString, idx) of sanityCheckErrors" :key="idx">
+						{{ errString }}
+					</div>
+				</section>
+			</template>
+			<template #footer>
+				<Button label="Ok" class="p-button-primary" @click="sanityCheckErrors = []" />
+				<Button
+					label="Ignore"
```
YohannParis, line 282: I feel the label `Ignore` to be a bit confusing by itself. Maybe it's just me.
```suggestion
					label="Ignore warnings and use configuration"
```

YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
mwdchang: Because linter doesn't like it.

PR #3207, opened 2024-03-28
fix[hmi-server]: Download netcdf assets

### LINKED ISSUE #3206: [FEAT]: Allow saving of netcdf data

dvince2:
# Description
Adding support for downloading the original netcdf files from the source service. If there are download urls set this will take precidence over s3 stored assets, but we _should_ never have both set as subset datasets are new datasets without the dataset urls field set.

Resolves #3206

### DIFF
--- a/packages/client/hmi-client/src/types/Types.ts
+++ b/packages/client/hmi-client/src/types/Types.ts
@@ -160,6 +160,7 @@ export interface Dataset extends TerariumAsset {
     dataSourceDate?: Date;
     fileNames?: string[];
     datasetUrl?: string;
+    datasetUrls?: string[];
     columns?: DatasetColumn[];
     metadata?: any;
     source?: string;
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
@@ -17,7 +17,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVParser;
 import org.apache.commons.csv.CSVRecord;
-import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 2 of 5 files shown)


dvince2:
If this isn't needed for the eval scenario then lets leave this out today. There's no point in rocking the boat?

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -160,6 +160,7 @@ export interface Dataset extends TerariumAsset {
     dataSourceDate?: Date;
     fileNames?: string[];
     datasetUrl?: string;
+    datasetUrls?: string[];
```
YohannParis, line 163: Why do we need both?

dvince2: We don't, but I'm only about 80% sure we don't. `datasetUrl` is a holdover from old TDS. in `Dataset.java` I've added a comment with a link to a github issue to confirm this and remove after the eval - I dont want to commit that and have it blow up

bigglesandginger: I agree that it is most probably a holdover, but my knowledge is less with respect to this.  And I love the idea that during the eval, maybe not the time to delete it :)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
```diff
@@ -375,17 +376,54 @@ public ResponseEntity<StreamingResponseBody> getFile(
...
+		} catch (final IOException e) {
+			final String error = "Unable to get dataset";
 			log.error(error, e);
 			throw new ResponseStatusException(
 					org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR,
 					error);
 		}
+
+		if(dataset.get().getEsgfId() != null && !dataset.get().getEsgfId().isEmpty()
+			&& dataset.get().getDatasetUrls() != null && !dataset.get().getDatasetUrls().isEmpty()){
```
bigglesandginger, line 396: Liking - super solid

bigglesandginger (approved):

packages/client/hmi-client/src/types/Types.ts
dvince2: yes hahah :D

on the java model:
TODO: IS THIS NEEDED? IS THIS FROM OLD TDS? https://github.com/DARPA-ASKEM/terarium/issues/3194

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
dvince2: In general I prefer java to javascript... but I do love the brevity of just checking `if(!someVar)` and that does a null and empty check at the same time...

PR #3208, opened 2024-03-28
Move all to new paradigm. adding spinner

Tom-Szendrey:
# Description
TODO: Next PR add the chart itself

image

image


YohannParis (approved):

packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -18,16 +19,160 @@
...
+		if (response.state === PollerState.Done) {
+			// Start 2nd simulation to get sample simulation from dill
+			const simulationIntervetions = getSimulationInterventions();
+			const forecastResponse = await startForecast(simulationIntervetions);
+			const forecastId = forecastResponse.id;
+
+			const state = _.cloneDeep(props.node.state);
+			state.inProgressOptimizeId = '';
+			state.optimzationRunId = id;
+			state.inProgressForecastId = forecastId;
```
shawnyama, line 145: Is this the only place where `state.inProgressForecastId` gets mutated? If so you maybe don't need another watcher for that. You can just continue to do the simulation within here.

packages/client/hmi-client/src/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -33,7 +33,8 @@ export interface OptimizeCiemssOperationState {
 	threshold: number;
 	isMinimized: boolean;
 	chartConfigs: string[][];
-	simulationsInProgress: string[];
+	inProgressOptimizeId: string;
+	inProgressForecastId: string;
 	forecastRunId: string;
 	optimzationRunId: string;
```
shawnyama, line 39: typo, `optimzation` -> `optimization`.

packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
Tom-Szendrey: I could foresee there being issues if i put that in the watcher for `inProgressOptimizeId` as the initial check will fail when that is now back to empty. If I were to go into a different page and come back type thing

I could add one watcher for both but i dont think thats any cleaner

PR #3226, opened 2024-04-01
refactor: explorer source dropdown

shawnyama:
# Description

- Create enum for Document sources ('xDD' and 'Terarium')
- One variable for `sourceOptions` and `chosenSource`
- Simpler passing of `source` prop


packages/client/hmi-client/src/types/search.ts
YohannParis, line 1: it's customary for `enum` to be allcaps. make it's easier to see that this is not to be change etc.

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -22,25 +22,18 @@
 					<nav>
 						<SelectButton
 							:model-value="assetType"
-							@change="if ($event.value) assetType = $event.value;"
+							@change="if ($event.value) changeAssetType($event.value);"
```
YohannParis, line 25: Is the following possible?

```suggestion
							@change="changeAssetType"
```

YohannParis (approved):

YohannParis (approved):

PR #3238, opened 2024-04-02
fix[hmi-server]: Improved Funman Errors

dvince2:
Catching exceptions thrown by the proxy service for funman operations. Also small code cleanup in eval controller.


YohannParis (approved):

mwdchang:
@dvince2 funman-service has always been a bit quiet in terms of errors - what sort of reporting can we see now that we couldn't view before?

dvince2:
@mwdchang 

Without this, any exceptions would have thrown a generic exception from the controller without a message and with a 500 error code.  This will now properly bubble up the correct error code returned to us from the funman service and display a nicer error to the user.  I've just been going through all of our controllers which make use of feignclient proxies to add this for better visibility on the front end

PR #3247, opened 2024-04-03
only add interventions when provided.

Tom-Szendrey:
# Description
Only add interventions from `policy.json` when they exist.
Do not include `interventions: []` into the payload as that will impact how we are doing the check 

Post this week's eval I will need to look into how we are handling interventions much cleaner and we can talk about this:
https://github.com/DARPA-ASKEM/terarium/pull/2990

https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java#L121

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
+++ b/packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
@@ -143,7 +143,7 @@ const startForecast = async (simulationInterventions) => {
 		},
 		engine: 'ciemss'
 	};
-	if (_.isEmpty(simulationInterventions)) {
+	if (!_.isEmpty(simulationInterventions)) {
 		simulationPayload.interventions = simulationInterventions;
 	}
 	return makeForecastJobCiemss(simulationPayload);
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -137,13 +137,15 @@ const startForecast = async (simulationIntervetions) => {
...
 			end: props.node.state.endTime
 		},
-		interventions: simulationIntervetions,
 		extra: {
 			num_samples: props.node.state.numSamples,
 			method: props.node.state.solverMethod
 		},
 		engine: 'ciemss'
 	};
+	if (simulationIntervetions.length > 0) {
```
shawnyama, line 146: typo interveNtions. and use isEmpty

shawnyama (approved):

packages/client/hmi-client/src/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
Tom-Szendrey: no one said var names need to be english :O

PR #3253, opened 2024-04-03
Add matrix string field to config from dataset endpoint

kbirk:


dvince2 (approved):

YohannParis (approved):

j2whiting:
This will break GoLLM, no? 

We will need to update tool_utils, the prompt, task runner and a few other things to accept the new param.

j2whiting:
Following our Slack convo. I updated GoLLM to support the separate arg. 

We should revisit the idea of replacing AMR with a data structure that will include these mapping after eval.

j2whiting (approved):

PR #3255, opened 2024-04-03
3232 Allow charts to be removed in TA3 operators

### LINKED ISSUE #3232: [Task]: Allow charts to be removed in TA3 operators
We can add charts once a simulation has finished, but we do not currently allow the removal of charts, we should support this feature.

Chart actions are proxied, take a look at **src/workflow/util.ts#chartActionsProxy(...)** , add the ability to remove charts to these operators:
- simulate julia
- simulate ciemss
- calibrate julia
- calibrate ciemss
- ensemble simulate ciemss
- ensemble calibrate ciemss
- optimize ciemss (if applicable)

jryu01:
# Description

Added a button to `tera-simulate-chart.vue` which allows it to be removed from the list when clicked. 

Screenshot 2024-04-03 at 3 32 10 PM

Resolves #3232


mwdchang:
Looks great, thanks!

PR #3258, opened 2024-04-04
fix[hmi-server]: GoLLM exception handling

dvince2:
# Description

Like a responsible fisher, I'm going to catch and throw back all these exceptions

Resolves #noissue


YohannParis (approved):
Nice catch on the datasetText

PR #3259, opened 2024-04-04
forecast should include the calibration like opt

Tom-Szendrey:
Like the optimize call forecast should include the calibration inferred params if provided


mwdchang:
looks fine

mwdchang (approved):

PR #3264, opened 2024-04-04
feat: add string matching for automapping in calibration

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/services/concept.ts
+++ b/packages/client/hmi-client/src/services/concept.ts
@@ -231,7 +231,7 @@ const autoCalibrationMapping = async (modelOptions: State[], datasetOptions: Dat
 		const groundings = state.grounding?.identifiers
 			? Object.entries(state.grounding.identifiers).map((ele) => ele.join(':'))
 			: undefined;
-		sourceEntities.push({ id: state.id, groundings });
+		sourceEntities.push({ id: state.id.toLowerCase(), groundings });
 	});
 
 	// Fill targetEntities with datasetOptions
--- a/packages/client/hmi-client/src/services/concept.ts
+++ b/packages/client/hmi-client/src/services/concept.ts
@@ -144,6 +144,9 @@ const autoEntityMapping = async (
 	targetEntities: Entity[],
 	acceptableDist?: number
 ) => {
+	console.log(sourceEntities);
+	console.log(targetEntities);
+
 	const result = [] as EntityMap[];
 	const acceptableDistance = acceptableDist ?? 0.5;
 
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/concept.ts
```diff
@@ -154,64 +154,80 @@ const autoEntityMapping = async (
...
+			distinctSourceGroundings,
+			distinctTargetGroundings
+		);
+		if (!allSimilarity) return result;
+		// Filter out anything with a similarity too small
+		const filteredSimilarity = allSimilarity.filter((ele) => ele.similarity >= acceptableDistance);
+		const validSources: any[] = [];
+		const validTargets: any[] = [];
+
+		console.log(filteredSimilarity);
```
Tom-Szendrey, line 180: Remove console log

packages/client/hmi-client/src/services/concept.ts
```diff
@@ -154,64 +154,80 @@ const autoEntityMapping = async (
...
-	// Filter out anything with a similarity too small
-	const filteredSimilarity = allSimilarity.filter((ele) => ele.similarity >= acceptableDistance);
-	const validSources: any[] = [];
-	const validTargets: any[] = [];
 
-	// Join results back with source and target
-	filteredSimilarity.forEach((similarity) => {
-		// Find all Sources assosiated with this similarity
+	// Match id strings if there are no groundings in either source or target
+	if (isEmpty(distinctTargetGroundings) || isEmpty(distinctSourceGroundings)) {
```
Tom-Szendrey, line 159: Rather than 
If no grounding use exact mapping
If grounding use grounding

Do you think it would be better to:
Use exact mapping, then whatever is leftover check their groundings?

packages/client/hmi-client/src/services/concept.ts
```diff
@@ -210,9 +211,33 @@ const autoEntityMapping = async (
 				});
 			}
 		});
+
+		// Match by string if no target is found
+		if (isEmpty(currentTargetId)) {
+			targetEntities.some((target) => {
+				if (target.id.startsWith(distinctSourceId) || distinctSourceId.startsWith(target.id)) {
```
Tom-Szendrey, line 218: Why use `startsWith` for string comparison? 
What happens when you compare `infected` and `infected_old` both "start with" infected but do not match?

Is `stringOne.toLower() === stringTwo.toLower()` not better?

shawnyama: I considered that but take a look at this, the caps helps here:
image

Tom-Szendrey (approved):
Thanks

PR #3268, opened 2024-04-04
calibrate ensemble mapping

Tom-Szendrey:
# Description

- Updating mapping provided to ensemble calibrate
- Correcting `timespan` value in ensemble calibrate
- Adding optional input to `getTimespan` and using an interface for easier use of optional params.
- Updating default num particles to 1

# Testing
Running a regular calibrate to ensure the timespan changes are not breaking anything unexpectedly
Running ensemble calibrate with two different models with no indexes being titled the same

# Mapping Details
Pyciemss-service wants the dataset mapping as follows
{dataset timestamp col name}: "timestamp",
{X dataset column used}: {X dataset column used},
...

Example:
```
{
    "time": "timestamp",
    "S": "S",
    "I": "I"
}
```

### DIFF
--- a/packages/client/hmi-client/src/workflow/util.ts
+++ b/packages/client/hmi-client/src/workflow/util.ts
@@ -53,7 +53,7 @@ export interface GetTimespanParams {
 	timestampColName?: string;
 }
 
-export const getTimespan = (params: GetTimespanParams): TimeSpan => {
+export function getTimespan(params: GetTimespanParams): TimeSpan {
 	let start = 0;
 	let end = 90;
 	// If we have the min/max timestamp available from the csv asset use it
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/workflow/util.ts
```diff
@@ -47,26 +47,32 @@ export const chartActionsProxy = (node: WorkflowNode<any>, updateStateCallback:
...
 };
 
-export const getTimespan = (dataset?: CsvAsset, mapping?: CalibrateMap[]): TimeSpan => {
+export interface GetTimespanParams {
+	dataset?: CsvAsset;
+	mapping?: CalibrateMap[];
+	timestampColName?: string;
+}
+
+export const getTimespan = (params: GetTimespanParams): TimeSpan => {
```
YohannParis, line 56: It doesn't need out of scope awareness, a function can be just fine here.

```suggestion
export function getTimespan(params: GetTimespanParams): TimeSpan {
```

YohannParis (approved):

PR #3272, opened 2024-04-05
Return stdout / stderr on TaskResponse. Log stdout / stderr on failure.

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
```diff
@@ -323,6 +323,22 @@ private void onTaskResponseAllInstanceReceive(final Message message) {
 				rLock.unlock();
 			}
 
+			// if the task failed, lets log the stdout / stderr
+			if (resp.getStatus() == TaskStatus.FAILED) {
+				if (resp.getStdout() != null && resp.getStdout().length() > 0) {
+					log.error("Task {} failed, logging stdout", resp.getId());
+					System.out.print(resp.getStdout());
```
dvince2, line 330: Is there a reason for having both `log.error` and `System.out`?

dvince2: ohhh

kbirk: The log prefixes it with the time, filename, line etc and then the system out does an unformatted string. Just looks a bit nicer

dvince2 (approved):

YohannParis (approved):

PR #3274, opened 2024-04-05
fix: forget about using `Component` for now

shawnyama:
# Description

Resolves #(issue)


Tom-Szendrey (approved):
approving but also have no context here so be warned

PR #3276, opened 2024-04-08
Workflow refactor

kbirk:
- Moves workflows to postgres
- Rows are automatically synced elasticsearch if `public == true` and `temporary == false`
- Added deep cloning interfaces working with blackboxed types such as `WorkflowNode` / `WorkflowEdge`. This is also work for other similar types such as `Model`.
- Add import / export functionality built on top of cloning. For hierarchical assets across different tables we can override these to be cross-service.
- Added bulk syncing and index updating by incrementing versions and transferring alias
- Added a backwards syncing tool for migrating types from ES to postgres. (Ex. migrating existing models / documents). This populates PG, and re-syncs back to a new index, transferring the alias and leaving the original index intact.

This is all done in a backwards compatible way that will allow us to move each asset type / server over one by one.

I fixed a bunch of warnings... that added a bunch of more files to this PR....


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/Workflow.java
```diff
@@ -20,20 +31,51 @@
...
+			oldToNew.put(node.getId(), clonedNode);
+			clone.getNodes().add(clonedNode);
+		}
 
-	private Object edges;
+		clone.setEdges(new ArrayList<>());
+		for (final WorkflowEdge edge : edges) {
+			final WorkflowEdge clonedEdge = edge.clone();
+			clonedEdge.setSource(oldToNew.get(edge.getSource().getId()));
+			clonedEdge.setTarget(oldToNew.get(edge.getTarget().getId()));
```
mwdchang, line 75: The source/target assignments look suspicious, if we are cloning it I would expect the source/target nodes to be new and distinct instances? Or do we have a different meaning of cloning here?

mwdchang: The UI has a branching function that acts like a clone for a subgraph of the workflow-graph:

https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/services/workflow.ts#L445-L549

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
```diff
@@ -1,21 +1,32 @@
...
 import software.uncharted.terarium.hmiserver.configuration.Config;
 import software.uncharted.terarium.hmiserver.configuration.ElasticsearchConfiguration;
 import software.uncharted.terarium.hmiserver.models.dataservice.workflow.Workflow;
+import software.uncharted.terarium.hmiserver.repository.data.WorkflowRepository;
 import software.uncharted.terarium.hmiserver.service.elasticsearch.ElasticsearchService;
 
 @Service
-public class WorkflowService extends TerariumAssetService<Workflow>{
-
+public class WorkflowService extends TerariumAssetServiceWithES<Workflow, WorkflowRepository> {
```
mwdchang, line 12: What do we expect the workflow update pattern to be like? 

- Is it such that we'd send in a workflow object and the API will automatically figure out which nodes/edges it should update? or
- would we need to uniquely update each node/edge individually?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowEdge.java
```diff
@@ -1,27 +1,30 @@
...
 	private WorkflowNode source;
-	private String sourcePortId;
 	private WorkflowNode target;
-	private String targetPortId;
-	private WorkflowDirection direction;
 
-	public enum WorkflowDirection {
-		FROM_INPUT,
-		FROM_OUTPUT
+	public WorkflowEdge clone() {
```
bigglesandginger, line 22: Why not pass in the source and target, thus not rely on the caller remembering to set these after the fact?

bigglesandginger (changes requested):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/Workflow.java
kbirk: We are cloning the edge, and then setting the nodes to the newly cloned nodes, but using a lookup from prev id to cloned id with the `oldToNew` map which is `old_node_id -> new_cloned_node`

kbirk: If I am missing things that also need to be cloned, then I need to fix that

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowEdge.java
kbirk: Done!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
kbirk: Currently I left it at 1 row per Workflow. The only cubmersome part is that `WorkflowEdges` embed the entire node, rather than just referencing it by id. 

@mwdchang Are workflow nodes / edges ever shared between workflows? Can we get away by just having the edge reference the target / source by id?

kbirk: The `WorkflowNode<any>['id'];` syntax confused me, I thought those were the full types :) I'll fix my code

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -115,14 +118,14 @@ public List<T> getAssets(final Integer page, final Integer pageSize) throws IOEx
 	 * @param id The ID of the asset to delete
 	 * @throws IOException If there is an error deleting the asset
 	 */
-	public void deleteAsset(final UUID id) throws IOException {
+	public T deleteAsset(final UUID id) throws IOException {
 		final Optional<T> asset = getAsset(id);
 		if (asset.isEmpty()) {
-			return;
+			throw new NotFoundException("Asset not found: " + id.toString());
```
dvince2, line 124: I'm not 100% sure we want to be making this change here. I think that the `NotFoundException` is specifically a HTTP error 404 exception which is a little weird for this service to throw... especially if some other back end class talks to it, it would be unexpected (vs just returning the empty optional and letting the downstream deal with it).

At minimum we would probably want to add the `NotFoundException` to the method signature so anything downstream would know to expect that (because I wouldn't :D).  Same for `updateAsset` below.

dvince2: To add to this - by throwing a _different_ exception or returning empty optional our controller(s) can create a nice error message for the front end.  `NotFoundException` extends responseStatusException so the front end error would end up including that UUID which isn't overly user-helpful

packages/client/hmi-client/src/types/Types.ts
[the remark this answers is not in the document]
mwdchang: We had problems generating Typescript types for workflow because it uses features that are hard to construct in Jara - namely disjunctive typing and partials.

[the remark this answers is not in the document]
kbirk: Yeah, there are explicit types in `workflow.ts`, the java types aren't used anywhere.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/Workflow.java
```diff
@@ -19,21 +29,51 @@
...
+	@JdbcTypeCode(SqlTypes.JSON)
+	private List<WorkflowNode> nodes;
+
+	@Convert(converter = JsonConverter.class)
+	@JdbcTypeCode(SqlTypes.JSON)
+	private List<WorkflowEdge> edges;
+
+	public Workflow clone() {
+		final Workflow clone = new Workflow();
+		clone.setId(UUID.randomUUID());
```
dvince2, line 55: If this hasn't been saved in postgres, do we want to assign it a UUID at all? I thought that the UUID was auto generated at creation or insertion

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/JsonConverter.java
```diff
@@ -0,0 +1,33 @@
...
+			throw new RuntimeException("Error converting Object to JSON", e);
+		}
+	}
+
+	@Override
+	public Object convertToEntityAttribute(final String dbData) {
+		try {
+			return objectMapper.readValue(dbData, Object.class);
+		} catch (final IOException e) {
+			throw new RuntimeException("Error converting JSON to List<WorkflowNode>", e);
```
dvince2, line 30: `List<WorkflowNote>` feels too specific here

kbirk: Whoops

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
```diff
@@ -32,23 +36,27 @@ public abstract class TerariumAsset implements Serializable {
...
 	private Timestamp deletedOn;
 
 	@TSOptional
 	private Boolean temporary = false;
 
 	@TSOptional
 	private Boolean publicAsset = false;
+
+	public TerariumAsset clone() {
+		return this.clone();
```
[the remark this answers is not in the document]
kbirk: Great catch, that was supposed to be `super.clone()`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/Workflow.java
kbirk: Oooo shoot you are right, this is gonna be a bit tedious: Workflows have nested fields at refer to the parents ID. So we would need to write it out, get the created value and re-write it out after populating the nested fields. Lemme see if there's a more elegant way to get around this.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
kbirk: Reverted back to the way it was!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -133,13 +136,34 @@ public void deleteAsset(final UUID id) throws IOException {
 	 * @throws IOException If there is an error creating the asset
 	 */
 	public T createAsset(final T asset) throws IOException {
-		final UUID id = UUID.randomUUID();
-		asset.setId(id);
+		if (elasticService.documentExists(getAssetIndex(), asset.getId().toString())) {
+			throw new ResponseStatusException(
```
dvince2, line 140: Because this isn't a controller, do we want to throw a different kind of exception here to use it more generically? (Then let the controller package it nicely?)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -133,13 +136,34 @@ public void deleteAsset(final UUID id) throws IOException {
...
+	 * Create new assets and saves to ES
+	 *
+	 * @param asset The asset to create
+	 * @return The created asset
+	 * @throws IOException If there is an error creating the asset
+	 */
+	public List<T> createAssets(final List<T> assets) throws IOException {
+		for (final T asset : assets) {
+			if (elasticService.documentExists(getAssetIndex(), asset.getId().toString())) {
+				throw new ResponseStatusException(
```
dvince2, line 158: Because this isn't a controller, do we want to throw a different kind of exception here to use it more generically? (Then let the controller package it nicely?)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
kbirk: Done!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
kbirk: Done!

YohannParis (approved):
LGTM on the front-end.

PR #3277, opened 2024-04-08
Constraint form UI clean up

ngraham76:
I spent some more time trying to clean up the spacing and alignment of the time and bounds input fields in the constraints container a bit more. This is looking better, code is cleaner as well. Also adjusted the field class so that they are inputtext-md (not inputtext-sm) - they now match other fields on the page.

**BEFORE**
image

**AFTER**
image


YohannParis:
This looks good. But less plan a chat and I will show you the ~witchcraft~magic of CSS grids, which will help make those code cleaner and easier to manage.

YohannParis (approved):

PR #3285, opened 2024-04-09
Model config apply config also appends to node

Tom-Szendrey:
# Description

- Rename a function to something a bit more accurate
- Apply a model config will now also append it to the output so user does not need to hit run and create a new config 

# Testing:
Making a few distinct configs where changes can be easily seen.
Checking that the state is being properly updated when changing between outputs and is saved
Checking that output that is appended is updated correctly

https://github.com/DARPA-ASKEM/terarium/assets/17088680/a4b9c60d-c50f-4838-98d5-3cc781afcf78

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
@@ -70,7 +70,7 @@
 										<Button
 											class="use-button"
 											label="Apply configuration values"
-											@click="useSuggestedConfig(data)"
+											@click="applyConfigValues(data)"
 											text
 										/>
 									</template>
@@ -873,7 +873,8 @@ const initialize = async () => {
 	}
 };
 
-const useSuggestedConfig = (config: ModelConfiguration) => {
+const applyConfigValues = (config: ModelConfiguration) => {
+	const state = cloneDeep(props.node.state);
 	const amr = config.configuration;
 
 	knobs.value.name = config.name;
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -887,6 +888,34 @@ const useSuggestedConfig = (config: ModelConfiguration) => {
...
+	// Update output port:
+	const listOfConfigIds: string[] = props.node.outputs.map((output) => output.value?.[0]);
+	// Check if this output already exists
+	if (config.id && listOfConfigIds.includes(config.id)) {
+		// Select the existing output
+		const output = props.node.outputs.find((ele) => ele.value?.[0] === config.id);
+		emit('select-output', output?.id);
+	}
+	// If the output does not already exist
+	if (config.id && !listOfConfigIds.includes(config.id)) {
```
shawnyama, line 901: I'd start this function with
```
if (!config.id) {
 logger.error("Model configuration not found")
 return
}

...
```

Then this can be simplified to:
```
if (listOfConfigIds.includes(config.id)) {
		// Select the existing output
		const output = props.node.outputs.find((ele) => ele.value?.[0] === config.id);
		emit('select-output', output?.id);
}
// If the output does not already exist
else { ...
```

Tom-Szendrey: Fair point, thank you!

PR #3288, opened 2024-04-09
Fixed model diagram buttons - transparent BG bug.

ngraham76:
Fixed a bug where the Reset zoom and Lock to freeze buttons background was displaying as transparent in model diagram preview panes, causing legibility issues. The refinement to the code was made `inline`. Would it make more sense to create a new class of button for model diagrams in the CSS variables? Let me know, thanks!

**BEFORE**
image

https://github.com/DARPA-ASKEM/terarium/assets/141876103/7657f1f1-c58f-4b46-bdac-83e3084ed40c

**AFTER**
https://github.com/DARPA-ASKEM/terarium/assets/141876103/72b6395b-3765-4fc8-8cc2-7a79087a6aae
image


mwdchang:
@ngraham76 I can't tell what had changed here - the diff is only showing a slight code-format changes ... have you committed everything?

ngraham76:
> @ngraham76 I can't tell what had changed here - the diff is only showing a slight code-format changes ... have you committed everything?

@mwdchang  I've re-committed the changes - still not seeing the update? It's a small fix, the only difference should be that those buttons don't show a transparent BG on hover / click.

mwdchang (approved):

PR #3294, opened 2024-04-10
fix(model diagram): watcher shouldn't be called while template card moves

shawnyama:
# Description

The watcher in `tera-model-diagram` was being triggered due to it watching the entirety of `props.model`. This also meant watching for its `templateCard` property which holds its x and y coordinates on the model edit canvas. These values change while the card moves on the canvas causing unnecessary calls.

Now I am ignoring the `templateCard` property by making a shallow copy of the `props.model` without `templateCard` and watching for that.

I did a slight HTML clean that just wraps the conditional once around a couple of elements instead of checking for it twice


shawnyama:
This should suffice, we can leave the restructuring of where `templateCard` is for another PR

mwdchang (approved):

PR #3306, opened 2024-04-11
feat: transition matrices for model diagram

shawnyama:
# Description

- created `tera-tooltip` component 
- tooltip can appear within a specified position relative to the div it is wrapped around
- Stratified transition matrices appear when you click on them in the model diagram
- `tera-stratified-matrix-preview.vue` houses a preview of what the transition matrix looks like, this is what's put inside the `tera-tooltip` slot

## UX thoughts
- I have to click the unlock to adjust button for the transition hovering to work. That `pointer-events: none` rule for when the diagram is locked might be too much. I think just disabling the zoom may be good enough. 
- When I click on a node all the other ones turn a little transparent, I don't think there is a need for this since model editing isn't done in the diagram anymore.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/85f12c09-3827-42e2-8480-4b61662dad1d

### To do later

The design has these colourful borders based on which strata the cells belong to, thought I'd save that for later. 
image


YohannParis (approved):
Well done!

PR #3311, opened 2024-04-12
Small refactor for the clone methods

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
```diff
@@ -66,11 +66,20 @@ protected void onUpdate() {
...
 
+	// This is here just to satisfy the service interface.
 	public TerariumAsset clone() {
-		try {
-			return (TerariumAsset) super.clone();
-		} catch (final CloneNotSupportedException e) {
-			throw new RuntimeException("Error cloning TerariumAsset", e);
-		}
+		throw new RuntimeException(
+				"This should not be called. Use explicitly override this method on the derived class and call cloneSuperFields instead.");
```
YohannParis, line 72: Is that English?
Use explicitly override

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
kbirk: lol, hold on lemme fix.

PR #3312, opened 2024-04-12
Notification Persistence

kbirk:
This PR introduces two new tables to PG: `notification_groups` and `notification_events`. There is a one to many relationship between groups and events.

For any "task" a notification_group is created for the user under some id. Any updates / progress that generated is persisted as a notification event for the group. The `NotificationController` has a single `GET` endpoint that returns all notification groups, including all nested events, for a user for some window of time, defaulting to the last 2 hours.

Code as been added to work the existing `ClientEvent` and `TaskResponse` types to persist the notifications.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
```diff
@@ -0,0 +1,65 @@
...
+                                                                        implementation = NotificationGroup.class)))),
+                @ApiResponse(
+                        responseCode = "500",
+                        description = "There was an issue fetching the notifications",
+                        content = @Content)
+            })
+    public ResponseEntity<List<NotificationGroup>> getNotificationGroups(
+            @RequestParam(value = "since", required = false, defaultValue = "2") final long sinceInHours) {
+
+        final LocalDateTime localDateTime = LocalDateTime.now().minusHours(sinceInHours);
```
bigglesandginger, line 57: thinking `localDateTime` would be better named `sinceDateTime`

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -621,6 +621,23 @@ export interface EntitySimilarityResult {
     similarity: number;
 }
 
+export interface NotificationEvent {
```
bigglesandginger, line 624: These don't seem to be models required in the client... I appreciate this PR might not be the right place for this sort of clean up.  @YohannParis we should probably think about what models are actually shared with the UI and which should not be.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
kbirk: Done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
```diff
@@ -0,0 +1,65 @@
...
+import software.uncharted.terarium.hmiserver.models.notification.NotificationGroup;
+import software.uncharted.terarium.hmiserver.security.Roles;
+import software.uncharted.terarium.hmiserver.service.CurrentUserService;
+import software.uncharted.terarium.hmiserver.service.notification.NotificationService;
+
+@RequestMapping("/notification")
+@RestController
+@Slf4j
+@RequiredArgsConstructor
+public class NotificationController {
```
bigglesandginger, line 29: Should there be a `deleteNotificationGroup` or `acknowledgeNotificationGroup` function to clear groups that are complete and "acknowledged" by the user.

And maybe a cron job that deletes a group that was completed and is something like a month old (though probably on the NotificationService and not this Controller)

Thoughts for "completeness"

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private final CurrentUserService currentUserService;
+    private final NotificationService notificationService;
+
+    @GetMapping("/user")
+    @Secured(Roles.USER)
+    @Operation(summary = "Return all recent notification groups for a user")
```
YohannParis, line 36: I this should just be `/` and we should filter by user by default.
I do not see a reasons to see other users notifications.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
```diff
@@ -0,0 +1,65 @@
...
+                                                        schema =
+                                                                @io.swagger.v3.oas.annotations.media.Schema(
+                                                                        implementation = NotificationGroup.class)))),
+                @ApiResponse(
+                        responseCode = "500",
+                        description = "There was an issue fetching the notifications",
+                        content = @Content)
+            })
+    public ResponseEntity<List<NotificationGroup>> getNotificationGroups(
+            @RequestParam(value = "since", required = false, defaultValue = "2") final long sinceInHours) {
```
YohannParis, line 55: Why are we limiting ourselves to notification group less than 2 hours?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationGroup.java
```diff
@@ -0,0 +1,57 @@
...
+    @Schema(accessMode = Schema.AccessMode.READ_ONLY)
+    @Column(columnDefinition = "TIMESTAMP WITH TIME ZONE")
+    private Timestamp timestamp;
+
+    @OneToMany(mappedBy = "notificationGroup", fetch = FetchType.EAGER)
+    @OrderBy("timestamp DESC")
+    @Schema(accessMode = Schema.AccessMode.READ_ONLY)
+    @ToString.Exclude
+    @JsonManagedReference
+    private List<NotificationEvent> notificationEvents;
```
YohannParis, line 49: Is there a way to delete all the `notificationEvent` after one is a final or error?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
```diff
@@ -0,0 +1,65 @@
...
+                                                @ArraySchema(
+                                                        schema =
+                                                                @io.swagger.v3.oas.annotations.media.Schema(
+                                                                        implementation = NotificationGroup.class)))),
+                @ApiResponse(
+                        responseCode = "500",
+                        description = "There was an issue fetching the notifications",
+                        content = @Content)
+            })
+    public ResponseEntity<List<NotificationGroup>> getNotificationGroups(
```
jryu01, line 54: Can we have an option to limit the number of notification events returned for a notification group? I think most of the cases, we need the latest notification event per group and maybe we need to fetch more when needed or requested by user for the history.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
kbirk: Added this in.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
kbirk: No reason, just an arbitrary filtering value. I can bump it to 48 hours or something more reasonable.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
YohannParis: That's a great point, and kind of what I was asking above, maybe we need an option to do that.
Some Notification will need to keep every event, but most don't

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
YohannParis: Thanks, just pure curiosity on my part

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationGroup.java
kbirk: We track whether or not all events are acknowledged and only return those with unack'd events.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/notification/NotificationController.java
kbirk: So it turns out, hibernate does not let you filter join fields dynamically. And because hibernate is infuriating, I can't even filter them out in the java code without it persisting those changes without disabling all of join field functionality altogether.

YohannParis: So what would be a solution here?

bigglesandginger:
It's not too late to rip out all of the JPA and just write SQL

PR #3314, opened 2024-04-12
model function for getting and updating values

blanchco:
# Description

* just creating some functions to access and update model values

### DIFF
--- a/packages/client/hmi-client/src/model-representation/service.ts
+++ b/packages/client/hmi-client/src/model-representation/service.ts
@@ -153,7 +153,7 @@ export const getModelRenderer = (
  * @param {Model} model - The model object.
  * @returns {ModelParameter[]} - The model parameters.
  */
-export function getModelParameters(model: Model): ModelParameter[] {
+export function getParameters(model: Model): ModelParameter[] {
 	const modelType = getModelType(model);
 	switch (modelType) {
 		case AMRSchemaNames.REGNET:
@@ -171,7 +171,7 @@ export function getModelParameters(model: Model): ModelParameter[] {
  * @param {string} parameterId - The ID of the parameter.
  * @returns {ModelParameter | null} - The model parameter or null if not found.
  */
-export function getParameter(model: Model, parameterId: string): ModelParameter | null {
+export function getParameter(model: Model, parameterId: string): ModelParameter | undefined {
 	const modelType = getModelType(model);
 	switch (modelType) {
 		case AMRSchemaNames.REGNET:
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/model-representation/service.ts
```diff
@@ -145,3 +147,128 @@ export const getModelRenderer = (
...
+	metadataKey: string,
+	value: any
+) {
+	if (!model.metadata?.[semanticType]?.[semanticId]) {
+		model.metadata ??= {};
+		model.metadata[semanticType] ??= {};
+		model.metadata[semanticType][semanticId] ??= {};
+	}
+	model.metadata[semanticType][semanticId][metadataKey] = value;
+}
```
mwdchang, line 263: This feels both too abstract, and overly-fitting to the timeseries scenario. There are no clear indications that semanticType > semanticId > metadataKey.

Also this makes it not possible to do `metadata['notes'] = "i am a note"`cases of simple key-value assignment

packages/client/hmi-client/src/model-representation/service.ts
```diff
@@ -145,3 +147,128 @@ export const getModelRenderer = (
...
+	const modelType = getModelType(model);
+	switch (modelType) {
+		case AMRSchemaNames.REGNET:
+			return model.model?.parameters.find((p) => p.id === parameterId);
+		case AMRSchemaNames.PETRINET:
+		case AMRSchemaNames.STOCKFLOW:
+		default:
+			return model.semantics?.ode?.parameters?.find((p) => p.id === parameterId) ?? null;
+	}
+}
```
mwdchang, line 184: For `getParameter` and `getInitial` - you are distinguish between null and undefined, it would be good if we just stick to one or the other.

Also it would be nice if we have more consistent naming, e.g. you have:
- getModelParameters/getParameter
- getModelInitials/getInitial

packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -684,10 +685,10 @@ const updateConfigInitial = (inits: Initial[]) => {
 
 const updateConfigFromModel = (inputModel: Model) => {
 	if (isPetriNet.value || isStockFlow.value) {
-		knobs.value.initials = inputModel.semantics?.ode.initials ?? [];
-		knobs.value.parameters = inputModel.semantics?.ode.parameters ?? [];
+		knobs.value.initials = getModelInitials(inputModel) ?? [];
+		knobs.value.parameters = getModelParameters(inputModel) ?? [];
 	} else if (isRegNet.value) {
-		knobs.value.parameters = inputModel.model?.parameters ?? [];
+		knobs.value.parameters = getModelParameters(inputModel) ?? [];
```
mwdchang, line 691: It seems like your getModelparameters and getModelInitials will not return nulls anymore, so we can probably clean up the ?? mullish coalesces

blanchco: ah yes, ill change this.  

To note...I was saving this thought for Monday but i was thinking of just saving just a model configuration object to the state rather than (parameters, initials, sources, units, etc...) to make all this easier.  I didnt want to touch this file too much

packages/client/hmi-client/src/model-representation/service.ts
blanchco: https://unchartedsoftware.slack.com/archives/C05PETG2GTE/p1712952982343699?thread_ts=1712951482.874169&cid=C05PETG2GTE

YohannParis (approved):

PR #3317, opened 2024-04-15
Palantir java format

### LINKED ISSUE #3316: Setup a back-end formatter

YohannParis:
# Description

You can ignore anything in the `/src` folder.

* Added Spotless to manage java formatting
* Add plugin to Gradle
* Add a GitHub action `lint-server.yaml` to format and commit on PR
* Updated `lint-types.yaml` to also commit if a new types.ts is made

Resolves #3316


.github/workflows/publish.yaml
```diff
@@ -80,6 +84,7 @@ jobs:
       - lint
       - lint-client
       - test-client
+      - lint-server
```
dvince2, line 87: `lint-server` will need to run first, as it now has the potential to change `Types.ts` which could potentially break client side tests

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/code/CodeController.java
```diff
@@ -106,26 +106,26 @@ public ResponseEntity<StoredModel> transformCode(final String code) {
...
-            files = githubProxy
-                    .getGithubRepositoryContent(repoOwnerAndName, path)
-                    .getBody();
-        } catch (final FeignException e) {
-            final String error = "Error getting github repository content";
-            final int status = e.status() >= 400 ? e.status() : 500;
-            log.error(error, e);
-            throw new ResponseStatusException(org.springframework.http.HttpStatus.valueOf(status), error);
-        }
+public ResponseEntity<GithubRepo> getGithubRepositoryContent(
```
dvince2: Worryingly, this file was re-linted on your latest change and the linter wiped away the white space on it?

dvince2: GOTCHA! false alarm

dvince2 (approved):
Lets blow it up!

Aside - from our conversations offline, is it possible to wrap this commit up so we don't blanket our git history with this change?

.github/workflows/lint-server.yaml
```diff
@@ -0,0 +1,44 @@
...
+          distribution: "temurin"
+          java-version: 17
+          cache: gradle
+
+      - name: Run Spotless
+        run: |
+          ./gradlew spotlessApply
+
+      - name: Commit all changed files
+        uses: stefanzweifel/git-auto-commit-action@v5
```
Nithos, line 42: please switch to use a `sha` for the version

Nithos: and you sure you want to auto-commit these?

.github/workflows/publish.yaml
```diff
@@ -78,9 +82,10 @@ jobs:
     needs:
       - targets
       - lint
```
Nithos, line 84: what is basic `lint` step do now?

Nithos (commented):
Just looking at the workflow changes. Look good with minor request to use the sha for the version of new action.

.github/workflows/publish.yaml
YohannParis: it does dockerfiles and YAML config

.github/workflows/lint-server.yaml
YohannParis: Done. And yes I want to auto-commit those, it's always annoying to come back to re-run something that already run on the github action. Also it's easy to remove a commit from the branch if necessary.

PR #3318, opened 2024-04-15
Save sciml calibrate loss func

Tom-Szendrey:
# Description

- Save loss values in state
  **Note, this can be quite large (eg 2,000 elements of {iteration, value}). Is this a concern? I understand the state was not intended to just shove everything into**
- render graph onMount (watch immediate does not work for this referencing of HTML elements)
- while throwing this into the onMount I may as well throw other things into this initialize section to clean up a bit

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
+++ b/packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
@@ -144,7 +144,7 @@
 				</div>
 				<div class="form-section">
 					<h5>Loss function</h5>
-					<div v-if="inProgressSimulationId || selectedRunId" ref="lossPlot"></div>
+					<div ref="lossPlot"></div>
 				</div>
 				<div class="form-section">
 					<h5>Variables</h5>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/workflow/ops/calibrate-julia/tera-calibrate-julia.vue
```diff
@@ -440,11 +440,50 @@ async function getAutoMapping() {
...
+	initialize();
+});
+
 watch(
 	() => props.node.state.inProgressSimulationId,
 	(id) => {
 		if (id === '') {
 			unsubscribeToUpdateMessages([id], ClientEventType.SimulationSciml, messageHandler);
+			const state = _.cloneDeep(props.node.state);
+			state.intermediateLoss = lossValues;
```
Tom-Szendrey, line 485: Please note this can be quite large (eg 2,000 elements of {iteration, value}). 
Is this a concern? 
I understand the state was not intended to just shove everything into do we want to save this elsewhere?

mwdchang (approved):
Didn't test, but looks okay, thanks!

PR #3321, opened 2024-04-15
Treat Workflow transform as a local change

asylves1:
# Description

remove `workflowDirty` flag that triggering save on the zoom transformation change. 
Add the transformation change to the localstorage.

Resolves #(issue)


packages/client/hmi-client/src/workflow/tera-workflow.vue
```diff
@@ -670,7 +671,30 @@ function saveTransform(newTransform: { k: number; x: number; y: number }) {
...
+	const terariumWorkflowTransforms = localStorage.getItem('terariumWorkflowTransforms');
+	if (!terariumWorkflowTransforms) {
+		const transformation: WorkflowTransformations = { workflows: { [id]: canvasTransform } };
+		localStorage.setItem('terariumWorkflowTransforms', JSON.stringify(transformation));
+		return;
+	}
+	const workflowTransformations = JSON.parse(terariumWorkflowTransforms);
+	workflowTransformations.workflows[id] = canvasTransform;
+	localStorage.setItem('terariumWorkflowTransforms', JSON.stringify(workflowTransformations));
 }
```
YohannParis, line 698: This should be put in an utils file

YohannParis (approved):

packages/client/hmi-client/src/workflow/tera-workflow.vue
asylves1: Will do

PR #3323, opened 2024-04-15
feat[hmi-client]: The Documents / Models / Dataset switcher on the E…

### LINKED ISSUE #1257: [FEAT]: The Documents / Models / Dataset switcher on the Explorer page ought to show number of items in each
Currently you need to switch tabs to find out if there's any items on that page.
Monosnap Terarium - Workflow   etc 2023-06-21 10-51-22

This may require additional queries, but it will improve the UX a lot:
Monosnap Terarium - Workflow   etc 2023-06-21 10-54-58

dvince2:
# Description

This is a bit of a 'first pass' at this task because of the horribly disjointed state the explorer is in. Some of the searches (models and xdd documents) happen all at once so we can populate them.  Others (datasets) happen on demand. We also have two types of documents which makes this muddy (and in code there's only one type and then seperate checks??). As a result you won't get 100% accurate results in some of the cases until you click on the tabs which isn't great but its the best that can be done without doing the "big task" of addressing this data explorer...

Resolves #1257


packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -144,6 +144,9 @@ const rankedResults = ref(true); // disable sorted/ranked results to enable pagi
 // facets
 const facets = ref<{ [index: string]: XDDFacetsItemResponse }>({});
 const docCount = ref(0);
+const modelTotal = ref(0);
+const documentTotal = ref(0);
+const datasetTotal = ref(0);
```
mwdchang, line 149: These don't seem like they need to be `refs`

packages/client/hmi-client/src/page/data-explorer/DataExplorer.vue
```diff
@@ -155,9 +158,9 @@ const clientFilters = computed(() => queryStore.clientFilters);
 
 const assetType = ref(AssetType.Document);
 const assetOptions = ref([
-	{ label: 'Documents', value: AssetType.Document },
-	{ label: 'Models', value: AssetType.Model },
-	{ label: 'Datasets', value: AssetType.Dataset }
+	{ label: `Documents`, value: AssetType.Document },
+	{ label: `Models`, value: AssetType.Model },
+	{ label: `Datasets`, value: AssetType.Dataset }
```
mwdchang, line 163: Are these backticks? Just for consistency maybe single quotes `'` since these don't need to be template-literals.

mwdchang (approved):

PR #3324, opened 2024-04-15
simplify model config state

blanchco:
# Description

*Here I am simplifying the state of the model config operator.  Instead of holding each and every possible parameter, metadata value, initial etc... in state, this is all available in a `ModelConfiguration` so why not just hold that rather than separate them out?  Also much cleaner to read and work with than the previous state in my opinion.


mwdchang (approved):
Seems alright, thanks

PR #3329, opened 2024-04-16
Palantir java format with Tabs

YohannParis:
# Description

* Fix the code to use tabs instead of spaces for the code.
* Only format the `src/` folder.


packages/server/build.gradle
```diff
@@ -153,8 +153,10 @@ dependencyLocking {
 
 spotless {
     java {
+        target 'src/**/*.java'
```
dvince2, line 156: This looks to only be running in `/packages/server`, but we have other Java repositories (`es-ingest`, `db-migration`, `taskrunner`) that won't get these changes. I'm actually not sure if anything Spotless happens there - but we probably want it to!

YohannParis: Like I mentioned @bigglesandginger about it, I want to make sure we solve all the kinks and setup of that formater in one place for now before putting it everywhere.

PR #3335, opened 2024-04-16
fix infinite succes on extractions

blanchco:
# Description

* fixed issue where we would get infinite green success panels of extraction from document
* atob for base64 decoding rather than Textdecoder
* close connection on error

### DIFF
--- a/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
+++ b/packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
@@ -343,6 +343,7 @@ import { OperatorStatus } from '@/types/workflow';
 import { formatTimestamp } from '@/utils/date';
 import { logger } from '@/utils/logger';
 import { getInitials, getParameters } from '@/model-representation/service';
+import { b64DecodeUnicode } from '@/utils/binary';
 import { ModelConfigOperation, ModelConfigOperationState } from './model-config-operation';
 
 enum ConfigTabs {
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -508,7 +508,7 @@ const extractConfigurationsFromInputs = async () => {
 					}
 					if (data?.status === TaskStatus.Success) {
 						logger.success('Model configured from document(s)');
-						const outputJSON = JSON.parse(new TextDecoder().decode(data.output));
+						const outputJSON = JSON.parse(atob(data.output));
 						console.debug('Task success', outputJSON);
```
mwdchang, line 512: Is this causing the infinite successes? Why are we switching to `atob`? Do we need to handle non-ascii encodings?

blanchco: `data.output` is a base64 string.  I took a quick look at the [TextDecoder documentation](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder) and it look like it doesnt support base64, Im not sure why we were using it in the first place.  

Anyways, `TextDecoder` it was causing an uncaught error since it couldnt recognize base64 and the `onData` function would be called infinitely since the connection wasn't closing (hence infinite success toasts).  Ive added error handling in the `onData` function to close the connection when we get an error which was the root cause of the infinite toasts.

blanchco: @YohannParis done

YohannParis (approved):

PR #3337, opened 2024-04-16
Adding cancel to the pyciemss drilldowns

Tom-Szendrey:
# Description

image
image


packages/client/hmi-client/src/api/api.ts
```diff
@@ -80,6 +80,7 @@ export interface PollResponse<T> {
 	error: any;
 	progress?: any;
 	data: T | null;
+	cancelled?: boolean;
```
Tom-Szendrey, line 83: This is how I am enabling the simulation-service.ts `pollAction` to fwd the cancel state to this as previously it was only able to be done with the object itself using `stop()`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -201,6 +201,12 @@ public ResponseEntity<JobResponse> makeEnsembleCalibrateCiemssJob(
 				.getBody());
 	}
 
+	@GetMapping("ciemss/cancel/{id}")
+	@Secured(Roles.USER)
+	public ResponseEntity<JobResponse> cancelCiemssJob(@PathVariable("id") final UUID id) {
+		return ResponseEntity.ok(simulationCiemssServiceProxy.cancelJob(id).getBody());
```
dvince2, line 207: Is it worth it to try to catch a `FeignClientException` thrown by the proxy here? I can foresee some strange situation where the front end tries to cancel a job that is already done or has already been cancelled and I imagine the ciemss service would throw an exception in this case

Tom-Szendrey: Thank you for the tip, will do!

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
```diff
@@ -0,0 +1,35 @@
...
+
+const props = defineProps<{
+	simulationRunId?: string;
+	disabled: boolean;
+}>();
+
+const cancelSimulation = async () => {
+	if (!props.simulationRunId) return;
+	await cancelCiemssJob(props.simulationRunId);
+	logger.success(`Simulation ${props.simulationRunId} has been cancelled.`);
```
Tom-Szendrey, line 26: @YohannParis there is a logger here

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
```diff
@@ -0,0 +1,35 @@
...
+	<Button
+		:disabled="disabled"
+		outlined
+		severity="secondary"
+		title="Saves the current version of the model as a new Terarium asset"
+		@click="cancelSimulation"
+		label="Cancel"
+		icon="pi pi-times"
+	/>
+</template>
```
mwdchang, line 11: Your template has copy pasta errors:

```
		title="Saves the current version of the model as a new Terarium asset"
```

packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
```diff
@@ -234,6 +238,12 @@ const modelConfig = ref<ModelConfiguration>();
 
 const modelConfigId = computed<string | undefined>(() => props.node.inputs[0]?.value?.[0]);
 const datasetId = computed<string | undefined>(() => props.node.inputs[1]?.value?.[0]);
+const cancelRunId = computed(() => {
+	if (props.node.state.inProgressForecastId !== '') return props.node.state.inProgressForecastId;
+	if (props.node.state.inProgressCalibrationId !== '')
+		return props.node.state.inProgressCalibrationId;
+	return '';
+});
```
mwdchang, line 246: The the majority of your `cancelRunid` function can be simplified somewhat:
- You don't need to repeat `props.node.state` so many times times, it makes the code harder to read
- isn't this effectively just `return state. inProgressForecastId  || state inProgressCalibrationId` ?

packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
```diff
@@ -78,6 +78,8 @@ const getStatus = async (simulationId: string) => {
 	emit('update-state', state);
 
 	if (pollerResults.state === PollerState.Cancelled) {
+		state.inProgressSimulationId = '';
+		emit('update-state', state);
 		return pollerResults;
 	}
```
mwdchang, line 84: I think there is a better update pattern here, you are potentially calling `update-state` too much and that is generally not a a good thing.

Rather than

```
change_state_part_1
emit('update-state', ...)
if (blah) {
   change_state_part_2
   emit('update-state')
}
```

It would be better to have, if possible
```
change_state_part_1
if (blah) {
  change_state_part_2
}
emit('update-state')
```

mwdchang (changes requested):

mwdchang:
Please note that @blanchco is shuffling things around to better reorganize the workflow related files.

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
Tom-Szendrey: whoops - thank you!

packages/client/hmi-client/src/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
Tom-Szendrey: I will rewrite this to be cleaner

packages/client/hmi-client/src/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
Tom-Szendrey: This is a good point. I will double check these files for this

mwdchang (approved):

PR #3369, opened 2024-04-17
Split image building

YohannParis:
# Description

* Split all the images built with their own workflow
* ~Leverage `macos` to build arm version~ not possible until M2 available
* Remove the main `publish` file

All the images built, I deleted most of them, but kept the taskrunner until this is merged:
* https://github.com/orgs/DARPA-ASKEM/packages?tab=packages&q=task


.github/workflows/publish-client.yml
```diff
@@ -0,0 +1,165 @@
...
+
+      - name: Setup Node 20.10.0 LTS
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.10.0
+
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Build image using Makefile
```
Nithos, line 60: Why are you building here and then just below with the action? choose one method

.github/workflows/publish-client.yml
```diff
@@ -0,0 +1,165 @@
...
+# yamllint disable-line rule:truthy
+on:
+  push:
+    paths: [ 'packages/client/**' ]
+    branches: [ 'main' ]
+    tags: [ '*' ]
+
+jobs:
+  image-name:
+    name: Create image name
```
Nithos, line 16: you could use this for the tag generation, it gives you a few more perks but may be overkill here
[Metadata action](https://github.com/docker/metadata-action)

.github/workflows/publish-client.yml
```diff
@@ -0,0 +1,165 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Build image using Makefile
```
Nithos, line 94: same as before, seems you are building twice

.github/workflows/publish-client.yml
```diff
@@ -0,0 +1,165 @@
...
+      - arm64-emulation
+    steps:
+      - name: Login to registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create, inspect and publish manifest
```
Nithos, line 157: this looks correct to me

.github/workflows/publish-client.yml
YohannParis: It's just a poorly formatted name from the makefile as it doesn't build the image, but build the hmi-client (from source to distribution)

.github/workflows/publish-client.yml
YohannParis: After looking at it, yes I agree it is overkill.
Would be nice to have setup for Pantera though

dgauldie (approved):
ok

PR #3371, opened 2024-04-17
move workflow to components

blanchco:
# Description

* Move all the folders in src/workflow/ into src/components


mwdchang (approved):
didn't test, code looks fine

PR #3372, opened 2024-04-18
Fix notification for tasks to include the user id

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/notification/NotificationService.java
```diff
@@ -42,8 +42,15 @@ public Optional<NotificationGroup> getNotificationGroup(final UUID id) {
 	}
 
 	public NotificationGroup createNotificationGroup(final NotificationGroup notificationGroup) {
-		final User user = currentUserService.get();
-		notificationGroup.setUserId(user != null ? user.getId() : "anonymous");
+		if (notificationGroup.getUserId() == null
+				|| notificationGroup.getUserId().isEmpty()) {
+			try {
+				final User user = currentUserService.get();
+				notificationGroup.setUserId(user != null ? user.getId() : "anonymous");
```
bigglesandginger, line 49: Is this check still required (there may be paths that can get to this state, I have not read everything closely) but it "feels" like the group should have been created when the task was created (during the request from the user)?

Or is this here "just in case".

kbirk: Its a bit of both. Its so if the group is created from the "correct" thread, it can just inject the user. But if not, it will take it if provided. 

The issue @jryu01 saw was that we create the task from a executor thread that does not have the user scope of the request, so it has to be supplied manually.

YohannParis (approved):

PR #3373, opened 2024-04-18
fix document metadata

blanchco:
# Description

* Updating at the end of the function was causing everything to be overridden.  Removing this and now metadata will show up when uploading a PDF.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -343,7 +343,7 @@ public Future<DocumentAsset> extractPDF(final UUID documentId, final String doma
 				notificationInterface.sendFinalMessage("Extraction complete");
 
 				// return the final document
-				return documentService.updateAsset(document).orElseThrow();
+				return documentService.getAsset(documentId).orElseThrow();
```
YohannParis, line 346: We still need to update the Asset with the new metadata field, or I'm missing something?
@dvince2 @kbirk

YohannParis (changes requested):

dvince2 (approved):
What you said there makes sense :D

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
YohannParis: 🤔 that make sense now.

YohannParis (approved):

PR #3374, opened 2024-04-18
fix(model-edit): executed code is saved properly

shawnyama:
The executed code saved in model edit wasn't saved entirely. The code to be appended to the notebook was being saved. Now the entire notebook is saved.


Tom-Szendrey (approved):
Thanks for this!

PR #3375, opened 2024-04-18
feat: funman taskrunner

mwdchang:
### Summary
Setup funman validation as a taskrunner feature

### DIFF
--- a/packages/funman/core/taskrunner.py
+++ b/packages/funman/core/taskrunner.py
@@ -1,0 +1,133 @@
+# FIXME: this is a copy from mira, should consolidate
+import json
+import signal
+import sys
+import concurrent.futures
+import argparse
+import os
+import time
+import threading
+from typing import Callable
+
+
+READ_CHUNK_SIZE = 1024*1024
+
+
+class SelfDestructThread:
+    def __init__(self, self_destruct_timeout_seconds):
+        self.self_destruct_timeout_seconds = self_destruct_timeout_seconds
+        self.stop_destruct_event = threading.Event()
+        self.thread = threading.Thread(target=self._kill_after_timeout)
+
+    def start(self):
+        self.thread.start()
+
+    def stop(self):
+        self.stop_destruct_event.set()
+        self.thread.join()
+
+    def _kill_after_timeout(self):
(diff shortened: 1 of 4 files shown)


packages/funman/core/taskrunner.py
```diff
@@ -0,0 +1,133 @@
...
+    def _kill_after_timeout(self):
+        self.stop_destruct_event.wait(self.self_destruct_timeout_seconds)
+        if not self.stop_destruct_event.is_set():
+            # if for whatever reason this process is still around, do our best to self destruct
+            os.kill(
+                os.getpid(), signal.SIGKILL
+            )  # Send the SIGTERM signal to the current process
+
+
+class TaskRunnerInterface:
```
mwdchang, line 38: @kbirk this is a carbon-copy of the taskrunner.py from mira-taskrunner, any thoughts on if we want to consolidate?

packages/funman/tasks/validate_modelconfig.py
```diff
@@ -0,0 +1,108 @@
...
+    current_results.finalize_result(scenario, result)
+    print(current_results.model_dump_json(by_alias=False))
+
+    return current_results.model_dump_json(by_alias=False)
+
+
+def taskrunner_wrapper():
+    print("Taskrunner wrapper")
+    try:
+        taskrunner = TaskRunnerInterface(description="AMR to MMT")
```
YohannParis, line 71: I think the description needs to be updated

packages/funman/core/taskrunner.py
YohannParis: Something like taskrunner core for python, for julia, for c++ etc that can be importer instead of defined on each packages folder?

packages/funman/tasks/validate_modelconfig.py
mwdchang: thx

packages/funman/core/taskrunner.py
mwdchang: Just logging this into a ticket here:

https://github.com/DARPA-ASKEM/terarium/issues/3395

YohannParis (approved):

kbirk (approved):

PR #3376, opened 2024-04-18
Taskrunner intermediate progress

kbirk:
This adds a new progress pipe to be written to and forwards the output from the process, to the  taskrunner, to the hmi-server


mwdchang (approved):
Didn't test, code looks fine.

packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/service/Task.java
```diff
@@ -201,6 +259,15 @@ public byte[] readOutputWithTimeout(final int timeoutMinutes)
...
 			// we got our response
+			try {
+				final JsonNode progress = mapper.readTree((byte[]) result);
+				if (progress.has("done")) {
+					// finished reading progress
+					return null;
+				}
+			} catch (final Exception e) {
+				// do nothing
+			}
```
YohannParis, line 270: Why doing a catch if we do nothing? To not stop the thread?

PR #3377, opened 2024-04-19
TaskRunner bandaid solution

kbirk:
What happened was the mira response came back successfully, but the success response was never written to redis, so all subsequent requests assumed it was still pending and waited for it.

However the original request's future was completed. I'm a bit perplexed because the future gets completed in `onTaskResponseAllInstanceReceive`, and that method can only run if `onTaskResponseOneInstanceReceives` runs (which writes out the success response), which never did.

This code did not run: 

https://github.com/DARPA-ASKEM/terarium/blob/d3c0125daa67178606b6ae1dd3031c9d46288562/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java#L419

This code, right below it, did:

https://github.com/DARPA-ASKEM/terarium/blob/d3c0125daa67178606b6ae1dd3031c9d46288562/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java#L438

I cannot see any possible way that the code below executed without the code above it not...

I've added some more logging to help if this occurs in the future. I've also add a bandaid fix so that on the first occurance of this happening again, it will clear the cache and the tasks will not continue to block.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/AMRToMMTResponseHandler.java
```diff
@@ -1,39 +1,18 @@
...
 public class AMRToMMTResponseHandler extends TaskResponseHandler {
 	public static final String NAME = "mira_task:amr_to_mmt";
 
-	private final ObjectMapper objectMapper;
-
 	@Override
 	public String getName() {
 		return NAME;
 	}
-
```
kbirk, line 22: We don't actually run any of this code, so we don't need to override anything here

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/AMRToMMTResponseHandler.java
YohannParis: Why are we not running it? How do we return the MMT to the front-end?

kbirk: We run it synchronously, it gets the output directly and returns it to the frontend, we aren't persisting it or anything.

YohannParis: You mean via `Notifications`?

kbirk: We are persisting those. I mean we don't take the result and write it to a model or document the similar way we do for the other endpoints. It's not a stateful operation, you give input and get output, and it caches.

PR #3378, opened 2024-04-19
Client events allow multiple tabs

bigglesandginger:
Two issues were at work here, the `hmi-server` would only allow one sseEmitter per user therefore when a second tab/page was opened by the user they would constantly disconnect each other.  Secondly, the client did not block the heartbeat check well enough.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
@@ -83,8 +83,10 @@ public class ClientEventService {
 	public SseEmitter connect(final User user) {
 		final SseEmitter emitter = new SseEmitter();
 		if (!userIdToEmitter.containsKey(user.getId())) {
+			System.out.println("New user connected creating new Arraylist");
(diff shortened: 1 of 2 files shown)


dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -138,22 +137,33 @@ void onSendToAllUsersEvent(final Message message, final Channel channel) {
...
-				} catch (final IOException e) {
-					log.error("Error sending all users message to user {}", userId, e);
-				}
+			userIdToEmitter.forEach((userId, emitterList) -> {
+				List<SseEmitter> emittersToRemove = send(messageJson, emitterList, userId);
+				emittersToRemove.forEach((emitter) -> {
+					userIdToEmitter.get(userId).remove(emitter);
+					if (userIdToEmitter.get(userId).size() == 0) {
+						userIdToEmitter.remove(userId);
+					}
```
YohannParis, line 146: This could be refactored as well and called within the `send()` function.

PR #3393, opened 2024-04-22
Pyciemss Cancel button styling

Tom-Szendrey:
# Description
Updating label to be "Stop"
Moving to the left side along with the run button

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
@@ -164,7 +164,7 @@
 				:disabled="disableRunButton"
 			/>
 			<tera-pyciemss-cancel-button
-				:style="{ marginRight: 'auto' }"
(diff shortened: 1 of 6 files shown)


mwdchang:
Wouldn't it be more accurate to do `Run OR stop` as oppose to `Run AND stop`? Since we don't support running multiple simulations at the same time?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss.vue
```diff
@@ -158,13 +158,13 @@
...
 			<Button
 				outlined
-				:style="{ marginRight: 'auto' }"
 				label="Run"
 				icon="pi pi-play"
 				@click="runCalibrate"
 				:disabled="disableRunButton"
 			/>
 			<tera-pyciemss-cancel-button
+				:style="{ marginRight: 'auto' }"
```
YohannParis, line 167:
```suggestion
				:class="mr-auto"
```

YohannParis (approved):

PR #3394, opened 2024-04-22
Remove hibernate bytecode enhancements

kbirk:


packages/server/build.gradle
```diff
@@ -111,14 +111,6 @@ tasks.named('test') {
     useJUnitPlatform()
 }
 
-hibernate {
-    enhancement {
-        lazyInitialization true
```
bigglesandginger, line 116: So this should affect `@OneToMany` annotations and the like - maybe is the same as `@OneToMany(fetch = FetchType.LAZY)`... So all parent child object definitions

packages/server/build.gradle
```diff
@@ -111,14 +111,6 @@ tasks.named('test') {
...
-        dirtyTracking true
```
bigglesandginger, line 117: This brief article is as clear as mud. https://dev.to/yigi/hibernate-dirty-check-4j6m#:~:text=When%20an%20entity%20object%20is,called%20%E2%80%9CHibernate%20Dirty%20Check%E2%80%9D.

packages/server/build.gradle
```diff
@@ -111,14 +111,6 @@ tasks.named('test') {
...
-        associationManagement true
```
bigglesandginger, line 118: This seems to control objects that contain other objects and how they get written to the database.  I seriously do not understand why devs don't just learn SQL and understand how this all works instead of trusting a monstrosity of automajik code

packages/server/build.gradle
kbirk: Yeah, we explicitly use  `FetchType.EAGER` for everything which overrides this, so removing this doesn't do anything.

packages/server/build.gradle
kbirk: Yeah my understanding is that if you make changes to an object in a transaction, it will actually write them out. Honestly Id rather that _not_ be the case. I don't want changes pushed to the db unless i explicitly want to write them.

packages/server/build.gradle
kbirk: Yeah I This is the source of the extra rows. When enabled it allows you to only set one side of the relationship and it updates any other in memory references for you. When its off, you need to explicitly insert it into the other side if you want it there, or re-fetch it from the db. We've already had this logic in for project assets and now the notifications to get the tests to originally pass, so I'm not too worried it will cause issues. It doesn't affect what was written to the db, just what is already held in memory in Java.

bigglesandginger:
What about the `@ManyToOne` should we adjust those (which maybe default to `fetch=FetchType.EAGER`) and ensure they are `FetchType.LAZY`?

Honestly this yet another aspect of hidden information is just bad bad design.  What are the default fetch methods for @OneToMany, @ManyToOne, @ManyToMany, and @OneToOne - JPA does its best to not make it obvious...

YohannParis (approved):

PR #3400, opened 2024-04-22
Make event data col text

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationEvent.java
```diff
@@ -61,6 +61,7 @@ protected void onCreate() {
 	}
 
 	@Convert(converter = JsonConverter.class)
+	@Column(columnDefinition = "text")
```
bigglesandginger, line 64: This doesn't seem to change an existing table.  Either we need an `ALTER TABLE notification_event ALTER COLUMN data TYPE text;` pre Hibernate, or create a new column and simply ignore the data column

bigglesandginger (dismissed):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationEvent.java
kbirk: Another PR has the flyway migration which should be merged after this PR

YohannParis (approved):

PR #3401, opened 2024-04-22
refresh document on extractions complete

blanchco:
# Description

* Address an issue where the document asset would not refresh when extractions are finished.
* Adding subscription of the extraction event in the document asset component so that we know when its done...there does seem to be some issues to know when extractions are fully completed (using time t), but will wait to add those changes in a PR after the notification system is integrated more

Resolves #(issue)


packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -267,6 +269,9 @@ const toggleOptionsMenu = (event) => {
 watch(
 	() => props.assetId,
 	async () => {
+		// unsub and sub to avoid multiple subscriptions when switching between document assets
+		await unsubscribe(ClientEventType.Extraction, subscribeToExtraction);
+		await subscribe(ClientEventType.Extraction, subscribeToExtraction);
```
YohannParis, line 274: Then you should do the `unsubscribe` call within the `sunscribe` function as to not forget to do it in the future.

YohannParis (approved):

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -267,6 +269,9 @@ const toggleOptionsMenu = (event) => {
 watch(
 	() => props.assetId,
 	async () => {
+		// unsub and sub to avoid multiple subscriptions when switching between document assets
+		await unsubscribe(ClientEventType.Extraction, subscribeToExtraction);
```
jryu01, line 273: I think you just need to subscribe on mount and unsubscribe on unmount of the component. And then you can filter the event by asset Id in the event handlers. I'm not sure if it's necessary to sub and unsub on every assetId change.

blanchco: good idea - ill take a look into implementing this

YohannParis: I don't know if it's worth doing more right now as this will be using the notifications system in the short term

jryu01: It looks like we are handling finished event separately here which is still needed since notification system should not  care what should happen inside this component.

blanchco: ive made changes so that we subscrive onmounted and unsub when the component is unmounted - as pointed out its probably better than doing it on asset id.

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
blanchco: see comment thread below - we will unsubscribe when the component is unmounted

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
YohannParis: 🤔 Once your notification manager is done @jryu01 we could use the composable to keep track of certain notification events, isn't it?

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -308,6 +310,25 @@ onUpdated(() => {
...
 	}
 });
+
+onMounted(async () => {
+	await subscribe(ClientEventType.Extraction, subscribeToExtraction);
+});
+
+async function subscribeToExtraction(event: ClientEvent<ExtractionStatusUpdate>) {
+	if (!event.data) return;
+	if (event.data.documentId !== props.assetId) return;
```
YohannParis, line 320:
```suggestion
	if (!event.data || event.data.documentId !== props.assetId) return;
```

YohannParis (approved):

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
jryu01: The notification manger handles the events globally only for the notification stuff. If another component wants to do  some other operations based on the events from server, it needs to subscribe for the events and handle it for its own purpose. In that way we can keep components decoupled. In this case, we want to refresh the component on complete extraction event. So I think it makes sense to register its own event handler that does that.

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -308,6 +310,25 @@ onUpdated(() => {
 		emit('asset-loaded');
 	}
 });
+
+onMounted(async () => {
+	await subscribe(ClientEventType.Extraction, subscribeToExtraction);
```
jryu01, line 315: You may want to listen for `ExtractionPdf` if you are looking for pdf extraction to be done.

blanchco:
Waiting on #3412 to be merged to finish this up

PR #3413, opened 2024-04-23
Small fix for missing datatype after Enrich Dataset

asylves1:
# Description

Adding back value for missing datatype. The missing concept value will need a backend change in another ticket

image

Resolves #(issue)


Tom-Szendrey (approved):
Does this type `ColumnType` still have `dataType: DatasetColumn`?

If this is (im not certain) the only place this is being used and its not being used anymore we should remove it from DatasetColumn.java and rerun yarn generateTypes

asylves1:
> Does this type `ColumnType` still have `dataType: DatasetColumn`?
> 
> If this is (im not certain) the only place this is being used and its not being used anymore we should remove it from DatasetColumn.java and rerun yarn generateTypes

At least for the Enrich Description call the `dataType` param does not appear to be passed. Not Use about the other locations

YohannParis (approved):

PR #3418, opened 2024-04-23
import model name when adding to project

blanchco:
# Description

* Adding the model name to the project asset when adding from the explorer

### DIFF
--- a/packages/client/hmi-client/src/utils/map-project-assets.ts
+++ b/packages/client/hmi-client/src/utils/map-project-assets.ts
@@ -29,7 +29,9 @@ export const generateProjectAssetsMap = (searchAsset: string): ProjectAssetItems
 					assetId: asset.assetId.toString(),
 					assetName:
 						asset.assetName ??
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/utils/map-project-assets.ts
```diff
@@ -29,7 +29,9 @@ export const generateProjectAssetsMap = (searchAsset: string): ProjectAssetItems
 					assetId: asset.assetId.toString(),
 					assetName:
 						asset.assetName ??
-						`${asset.assetId.substring(0, 3)} - ${formatShort(asset?.updatedOn)}`,
+						`${asset.assetId.substring(0, 3)} - ${formatShort(
+							asset?.updatedOn ?? asset?.createdOn
+						)}`,
```
blanchco, line 34: this shouldn't be needed anymore - but added createdOn as a backup just in case

YohannParis: Smart!!!

PR #3425, opened 2024-04-23
fix for sharing input clearing

asylves1:
# Description

Issue: sharing input field was clearing on every input. 

Moved the clear into the if `selectedUser` if condition

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
+++ b/packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
@@ -14,8 +14,9 @@
 				:options="usersMenu"
 				optionLabel="name"
 				editable
+				showClear
 				placeholder="Add people and groups"
-				@update:model-value="(value) => addNewSelectedUser(value.id)"
+				@update:model-value="(value) => addNewSelectedUser(value?.id)"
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/widgets/share-project/tera-share-project.vue
```diff
@@ -14,8 +14,9 @@
 				:options="usersMenu"
 				optionLabel="name"
 				editable
+				showClear
 				placeholder="Add people and groups"
-				@update:model-value="(value) => addNewSelectedUser(value.id)"
+				@update:model-value="(value) => addNewSelectedUser(value?.id)"
```
mwdchang, line 19: Why would `value` be null or defined here?

asylves1: I added the primevue clear button in the input field. When the clear button is used the value can be null.
I can remove it if you want
https://primevue.org/dropdown/#clearicon

mwdchang: That sounds weird ... but okay, not going to make a big deal out of it 😄

mwdchang (approved):

PR #3426, opened 2024-04-23
feat: model configurations can be accepted into stratify and model edit

shawnyama:
# Description

The input ports on stratify and model edit accept a `modelId` or `modelConfigId`. When the drilldowns are mounted the model is fetched based on the input type. 

Before when an edge was unlinked, the type stayed at whatever input type was removed. For example if I inserted a `modelId` into `modelId|modelConfigId` port and then unlink it - the input port would then only accept `modelId`. So I updated `removeEdge` in the workflow service to support resetting the input port `type` to accept multiple types. 

https://github.com/DARPA-ASKEM/terarium/assets/28237258/99a9bbc5-0c57-4a87-839b-5603ee5a704e


YohannParis (approved):
This looks good.

packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit.vue
```diff
@@ -281,6 +282,15 @@ const buildJupyterContext = () => {
...
 const inputChangeHandler = async () => {
+	let modelId = '';
+	if (props.node.inputs[0]?.type === 'modelId') {
+		modelId = props.node.inputs[0].value?.[0];
+	} else if (props.node.inputs[0]?.type === 'modelConfigId') {
+		const modelConfiguration = await getModelConfigurationById(props.node.inputs[0].value?.[0]);
+		if (modelConfiguration) {
+			modelId = modelConfiguration.model_id;
+		}
+	}
```
mwdchang, line 293: Can we not have so much repetition? It generally makes the code cleaner, easier to maintain and easier to read.

```
const input = props.node.input[0];
if (!input) return;

... stuff ...
```

Likewise in `tera-mira-stratify.vue`

packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit.vue
```diff
@@ -116,6 +116,7 @@ import TeraModelTemplateEditor from '@/components/model-template/tera-model-temp
 import TeraNotebookJupyterInput from '@/components/llm/tera-notebook-jupyter-input.vue';
 
 import { KernelSessionManager } from '@/services/jupyter';
+import { getModelConfigurationById } from '@/services/model-configurations';
```
YohannParis, line 119: Why not create a new method `getModelFromModelConfigurationById`?

shawnyama: good point

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -174,19 +174,28 @@ export const addEdge = (
...
 	targetPort.status = WorkflowPortStatus.NOT_CONNECTED;
 	delete targetPort.label;
 
+	// Resets the type to the initial input type if it was a union type
+	const initialInput = registry.getOperation(targetNode.operationType)?.inputs?.[targetPortIndex];
+	if (initialInput?.type.includes('|')) {
+		targetPort.type = initialInput.type;
+		targetPort.label = initialInput.label;
+	}
+
```
mwdchang, line 198: I don't think this is a good use of the registry. 

Even though from a technical sense it works it is not a very cohesive idea - if we want to retain or to revert type information, it needs to be built into the operator definition themselves, rather than going through a proxy and relying on indices to match up. In the longer term this can create unwanted baggage.

PR #3437, opened 2024-04-24
Added documentation urls

jryu01:


packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -3,7 +3,9 @@
 		<div class="title-row">
 			<h4><slot /> <i v-if="props.tooltip" v-tooltip="tooltip" class="pi pi-info-circle" /></h4>
 			<slot name="inputs" />
-			<!-- <a :href="documentationUrl" rel="noopener noreferrer">Documentation</a> -->
+			<a v-if="documentationUrl" :href="documentationUrl" target="_blank" rel="noopener noreferrer"
```
YohannParis, line 6: Can you remove the target blank.
Unless specifically asked by Pascale for demo reasons links should open normally.

YohannParis (changes requested):
I would prefer that all those links were in the operator definition, where we set their name, inputs, etc. That would make more sense and easier to update.

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
jryu01: Done

YohannParis:
> Not sure if it is too late or not, wondering if it may make sense to have documentation urls as a part of WorkflowNode definition, since WorkflowNode is kind of the basic workunit that get passed around, the URL becomes more readily available without needing to explicitly import or inject into tera-drilldown params.
> 
> Any thoughts? @jryu01 @YohannParis

Another solution is too have a `Map<operatorType, URL>` saved in the `components/workflow/` folder?

jryu01:
> Not sure if it is too late or not, wondering if it may make sense to have documentation urls as a part of WorkflowNode definition, since WorkflowNode is kind of the basic workunit that get passed around, the URL becomes more readily available without needing to explicitly import or inject into tera-drilldown params.
> 
> Any thoughts? @jryu01 @YohannParis

I also like this approach and made some changes but It looks like `addNode` which creates the `WorkflowNode` is being called once when the node is created very first time(Please correct me if I'm wrong), that being said, the added `documentUrl` field is only applied to the newly created node and won't be available for existing nodes in the system and doc url won't be visible for those node. Are we fine with this?

jryu01:
> Not sure if it is too late or not, wondering if it may make sense to have documentation urls as a part of WorkflowNode definition, since WorkflowNode is kind of the basic workunit that get passed around, the URL becomes more readily available without needing to explicitly import or inject into tera-drilldown params.
> 
> Any thoughts? @jryu01 @YohannParis

@mwdchang  I also like this approach and have made some changes but It looks like `addNode` which creates the `WorkflowNode` is being called once when the node is created very first time(Please correct me if I'm wrong), that being said, the added `documentUrl` field is only applied to the newly created node and won't be available for existing nodes in the system and doc url won't be visible for those node. Are we fine with this?

mwdchang:
> @mwdchang I also like this approach and have made some changes but It looks like `addNode` which creates the `WorkflowNode` is being called once when the node is created very first time(Please correct me if I'm wrong), that being said, the added `documentUrl` field is only applied to the newly created node and won't be available for existing nodes in the system and doc url won't be visible for those node. Are we fine with this?

@jryu01 I think that should be fine, we are not yet at a stage where backwards compatibility is a concern. We can do a migration if needed.

mwdchang (approved):

mwdchang:
thanks!

PR #3442, opened 2024-04-25
3431 task ability to download configured model

### LINKED ISSUE #3431: [Task] ability to download CONFIGURED model

dvince2:
# Description

* Allowing download of configured models from configure model node.
* Allowing download of Models from the resource preview panel for Models
* Handling missing schema name in a rudimentary way when know there is a schema
* Cleaning back end types to be more reflective of whats happening with Model Configs at _present_
* Cleaning types on the front end for Model

Resolves #3431


packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -125,6 +125,24 @@ const optionsMenuItems = computed(() => [
...
 					}
 				})) ?? []
+	},
+	{
+		icon: 'pi pi-download',
+		label: 'Download',
+		command: async () => {
+			if (model.value) {
+				const data = `text/json;charset=utf-8,${encodeURIComponent(
+					JSON.stringify(model, null, 4)
```
shawnyama, line 135:
```suggestion
JSON.stringify(model.value, null, 4)
```

shawnyama: You're saving the vue ref it should be the value

YohannParis (approved):

PR #3447, opened 2024-04-25
Funman tracking n result

mwdchang:
### Summary
Proxy funman validation to use "Simulation" construct, and save final result into S3/Minio

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
@@ -8,7 +8,6 @@ import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import jakarta.annotation.PostConstruct;
-import java.util.UUID;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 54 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
```diff
@@ -73,18 +77,29 @@ public ResponseEntity<TaskResponse> createValidationRequest(@RequestBody final J
 			taskRequest.setUserId(currentUserService.get().getId());
 			taskRequest.setInput(objectMapper.writeValueAsBytes(input));
 
-			// TODO:
-			// - create Simulation for tracking
-			// - mark Simulation as pending
 			final UUID uuid = UUID.randomUUID();
+			final Simulation sim = new Simulation();
+
+			sim.setId(uuid);
```
dvince2, line 83: You shouldn't have to set the UUID like this - after creation it should come back with one

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
```diff
@@ -30,9 +30,11 @@ public class Simulation extends TerariumAsset {
...
 	@JsonAlias("execution_payload")
 	@Convert(converter = JpaConverterJson.class)
+	@Column(columnDefinition = "text")
 	private JsonNode executionPayload;
 
 	@TSOptional
+	@Column(length = 1000)
 	private String description;
 
 	@JsonAlias("result_files")
```
dvince2, line 40: did `result_files` also have its column definition or size change?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
mwdchang: will add magic annotation

PR #3452, opened 2024-04-26
change data transform language

blanchco:
# Description

* Ability to change programming languages in data transform
* fixed an issue where the ConfirmDialog was popping up multiple times when used, this should be defined at the root of the project

https://github.com/DARPA-ASKEM/terarium/assets/33158416/22e3141e-5491-4585-aad6-a4c447773eff

### DIFF
--- a/packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
+++ b/packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
@@ -59,7 +59,7 @@
 				option-label="name"
 				option-value="value"
 				class="5"
-				:disabled="!kernelState"
+				:disabled="kernelStatus !== 'idle'"
 				@change="onChangeLanguage"
 			/>
 			<Button
@@ -83,6 +83,7 @@
 			@update-kernel-status="updateKernelStatus"
 			@new-dataset-saved="onNewDatasetSaved"
 			@download-response="onDownloadResponse"
+			@update-language="(lang) => emit('update-language', lang)"
 			:notebook-session="props.notebookSession"
 		/>
 
@@ -282,8 +283,6 @@ const onChangeLanguage = (val) => {
 			};
 			const message: JupyterMessage = createMessage(messageBody);
 			kernel.sendJupyterMessage(message);
-
-			emit('update-language', val.value);
 		}
 	});
 };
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
```diff
@@ -234,11 +245,49 @@ jupyterSession.kernelChanged.connect((_context, kernelInfo) => {
...
+					language: val.value,
+					context_info: contextInfo
+				},
+				msgType: 'context_setup_request',
+				msgId: createMessageId('context_setup_request')
+			};
+			const message: JupyterMessage = createMessage(messageBody);
+			kernel.sendJupyterMessage(message);
+
+			emit('update-language', val.value);
```
mwdchang, line 286: Does ordering matter? I don't know if you have the guarantee that `kernel.sendJupyterMessage(message);` is finished at this stage.

blanchco: You're right, I'm making a change to listen to the `context_setup_response` to get the language, that should give use the language change

blanchco: @mwdchang it _should_ be good now with the latest commit

packages/client/hmi-client/src/types/common.ts
```diff
@@ -212,3 +213,25 @@ export interface NotificationItem {
...
+ * @returns {Array} An array of options for programming languages.
+ */
+export const programmingLanguageOptions = (): { name: string; value: string }[] =>
+	Object.values(ProgrammingLanguage)
+		.filter((lang) => lang !== ProgrammingLanguage.Zip)
+		.map((lang) => ({
+			name:
+				lang && `${lang[0].toUpperCase() + lang.slice(1)} (${ProgrammingLanguageVersion[lang]})`,
+			value: ProgrammingLanguageVersion[lang]
+		}));
```
mwdchang, line 237: This feels like an awkward construct, wouldn't it be easier if we start off by storing the programming languages and their versions together? Then we don't need to do these kind of data wrangling and have the slight problem of independently maintaining "languages" across two physical locations in the repo.

Just food for thought for later - no need to actionnow.

mwdchang (approved):

packages/client/hmi-client/src/types/common.ts
blanchco: might be good to use the beaker api to fetch the languages and version to use in the dropdown in the future: `https://beaker.terarium.ai/contexts`

PR #3453, opened 2024-04-26
harden(model-edit): save decomposed port positions for flattened view edits

shawnyama:
# Description

When port positions from the decomposed view aren't rendered (when we are in the flattened view) they must be saved so that the edges in the decomposed view are drawn properly as we reflect a flattened edit. This occurs specifically when we are creating a new junction. In the BEFORE video I am creating a junction for `R`. If I did this with `I` there wouldn't be a problem.  

BEFORE:

https://github.com/DARPA-ASKEM/terarium/assets/28237258/bbc0aea9-466b-4e3c-9c48-e9df9ba99f96

AFTER:

https://github.com/DARPA-ASKEM/terarium/assets/28237258/fd7e7fc3-c7d7-4a39-8d4c-84db6301bb01


packages/client/hmi-client/src/services/model-templating.ts
```diff
@@ -516,14 +517,21 @@ export function updateFlattenedTemplateInView(flattenedCanvas: ModelTemplateCanv
...
+	const id = `${templateCard.id}-${portId}`;
 
 	// Default to fallback values for port position (top right of the card)
 	let x = templateCard.x + cardWidth;
 	let y = templateCard.y;
 	// Get the position of the port element
-	const portElement = document.getElementById(`${templateCard.id}-${portId}`);
+	const portElement = potentialPortPositions
+		? potentialPortPositions.get(id)
+		: document.getElementById(id);
```
mwdchang, line 534: This is a bit sus - since `document.getElementById(id);` and `potentialPortPositions.get(id)` are returning very different objects that just so happens to have some overlap.

It would be clearer and easier to understand if you can wrap it with a helper function

```
getElementOffsetValues(document.getElementById(id))
```

mwdchang (approved):
thankx!

PR #3461, opened 2024-04-29
Ts/3398 interventions

Tom-Szendrey:
# Description
Interventions to come from an intervention object within the model config rather than a weird timeseries string
Interventions for simulating to come from the payload provided as well as whatever is within the config provided

-Can run optimization + calibration + simulation with and without interventions.
-Can see both interventions from optimization as well as the configuration within the optimization's simulation
-Can add, delete, and update the interventions in the model config

# Testing:
All the points above reran many times with pyciemss-api logs on the other screen

https://github.com/DARPA-ASKEM/terarium/assets/17088680/503581ee-2b64-4c4e-a553-2e8b584147d6


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/modelparts/ModelMetadata.java
```diff
@@ -42,9 +42,6 @@ public class ModelMetadata extends SupportAdditionalProperties implements Serial
 	@TSOptional
 	private List<JsonNode> attributes;
 
-	@TSOptional
-	private Map<String, Object> timeseries;
```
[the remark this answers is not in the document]
mwdchang: Deprecated means the feature is still supported, it looks like this is no longer the case.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -123,9 +124,27 @@ public ResponseEntity<Simulation> makeForecastRun(@RequestBody final SimulationR
...
+			if (modelInterventions != null) {
+				List<Intervention> allInterventions = request.getInterventions();
+				if (allInterventions == null) {
+					allInterventions = new ArrayList<Intervention>();
+				}
+				allInterventions.addAll(modelInterventions);
+				request.setInterventions(allInterventions);
+			}
+		} catch (IOException e) {
+			String error = "Unable to find model configuration";
```
mwdchang, line 145: What is throwing the IOException here? You don't have any I/O that I can see.

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -630,6 +654,35 @@ const isRegNet = computed(() => modelType.value === AMRSchemaNames.REGNET);
...
+			{ id: uuidv4(), name: '', timestep: 1, value: 1 }
+		];
+	}
+};
+
+const deleteIntervention = (index: number) => {
+	if (knobs.value.transientModelConfig.interventions?.[index]) {
+		knobs.value.transientModelConfig.interventions.splice(index, 1);
+	}
+};
```
mwdchang, line 676: This is a very strange setup, if you have ID then don't rely on index, if yo have index then don't rely on ID. Having both makes understanding the data structure more complicated as the rules for creation and deletion are mismatched.

packages/client/hmi-client/src/components/model/petrinet/tera-model-intervention.vue
```diff
@@ -0,0 +1,64 @@
...
+	<div class="form">
+		<div class="label-col">
+			<label>Parameter name </label>
+			<Dropdown
+				v-if="parameterOptions"
+				:options="parameterOptions"
+				v-model.lazy="name"
+				@update:model-value="updateIntervention"
+			/>
+			<InputText v-else v-model.lazy="name" @update:model-value="updateIntervention" />
```
mwdchang, line 11: I don't understand why we have a dropdown and an inputbox, under what usage scenario will parameterOPtions be null?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -123,9 +124,27 @@ public ResponseEntity<Simulation> makeForecastRun(@RequestBody final SimulationR
...
+				return ResponseEntity.notFound().build();
+			}
+			final List<Intervention> modelInterventions =
+					modelConfiguration.get().getInterventions();
+			if (modelInterventions != null) {
+				List<Intervention> allInterventions = request.getInterventions();
+				if (allInterventions == null) {
+					allInterventions = new ArrayList<Intervention>();
+				}
+				allInterventions.addAll(modelInterventions);
```
mwdchang, line 141: We should make a note here that this is going to trip people up - basically we are intercepting and changing the request, and that will make it incredibly difficult to test and to debug. The SimulationRequest object suggests that interventions should be set there, but it doesn't say anything about interventions getting changed along the way.

We should discuss and plan to straighten things out.

mwdchang:
As the next step, we need to think about validations, so we don't end up with negative time steps or something like duplicated timesteps.

packages/client/hmi-client/src/components/model/petrinet/tera-model-intervention.vue
Tom-Szendrey: Currently there is no scenario where it is null. I can remove this and make it that required

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
Tom-Szendrey: I added the comment on line 127 at the start of the try 
Would love to chat about how to make this more clear or if this is the best approach

dvince2 (approved):
If possible lets try to get some of the unknown items raised by Dan here figured out by the time `Model` gets moved to PG

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
Tom-Szendrey: Thanks for the tip on updating too much and rerendering

PR #3464, opened 2024-04-29
task[hmi-server] - server cleanup

dvince2:
# Description

* Did some legwork for future database work, bringing `description` and `name` into the parent `TerariumAsset` class, and updating all the tests for that.
* Changing the data type of `description` to be text for future proofing.
* Explicitly define string length on the two classes we've migrated so far
* Wrote needed migrations.

### DIFF
--- a/packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
+++ b/packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
@@ -168,7 +168,7 @@ const projectOptions = computed(() => [
 						// then, link and store in the project assets
 						if (datasetId) {
 							response = await useProjects().addAsset(AssetType.Dataset, datasetId, project.id);
-							assetName = selectedAsset.value.name;
+							assetName = selectedAsset.value.name ?? '';
 						}
 					} else if (isDocument(selectedAsset.value) && props.source === DocumentSource.XDD) {
 						const document = selectedAsset.value as Document;
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
@@ -28,7 +28,7 @@ public abstract class TerariumAsset implements Serializable {
 	private UUID id = UUID.randomUUID();
 
 	@TSOptional
-	@Column(length = 255)
+	@Column(length = 512)
 	@Schema(defaultValue = "Default Name")
 	private String name;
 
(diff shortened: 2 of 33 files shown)


packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
```diff
@@ -168,7 +168,7 @@ const projectOptions = computed(() => [
 						// then, link and store in the project assets
 						if (datasetId) {
 							response = await useProjects().addAsset(AssetType.Dataset, datasetId, project.id);
-							assetName = selectedAsset.value.name;
+							assetName = selectedAsset.value.name ?? '';
```
YohannParis, line 171: I wonder if instead of doing this we can force TerariumAsset to provide an empty string if `name` or `description` is null?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
```diff
@@ -28,8 +28,15 @@ public abstract class TerariumAsset implements Serializable {
 	private UUID id = UUID.randomUUID();
 
 	@TSOptional
+	@Column(length = 255)
```
YohannParis, line 31: Why the limit here? Some Document can have insanely long titles.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
```diff
@@ -73,6 +69,7 @@ public class Simulation extends TerariumAsset {
 
 	@JsonAlias("user_id")
 	@TSOptional
+	@Column(length = 255)
 	private String userId;
```
YohannParis, line 73: Isn't it a UUID as well?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
dvince2: This was always 255 - we're just explicitly defining it here for visibility. If 255 is too small though this can be upped -512?

packages/client/hmi-client/src/page/data-explorer/components/tera-search-results-list.vue
dvince2: Agreed, but from slack conversation, this is part of a much larger cleaning on these data objects that needs to be done in a separate issue: https://github.com/DARPA-ASKEM/terarium/issues/3466

dgauldie (approved):
LGTM

PR #3467, opened 2024-04-30
Number utils

blanchco:
# Description

1. Created some functions to do some conversions for the new input field:
- number -> exponential
- exponential -> number
- number -> NIST
- NIST -> number
- displaying a number (this depends on the length)

2. I have made the input for all these functions a string, since we can deal with very small or very large numbers it can lead to precision issues if we were to use the Number library to manipulate these inputs

3. UNIT TESTS!!

### DIFF
--- a/packages/client/hmi-client/src/utils/number.ts
+++ b/packages/client/hmi-client/src/utils/number.ts
@@ -1,0 +1,122 @@
+export function cleanNumberString(num: string): string {
+	// remove spaces
+	let numStr = num.replace(/ /g, '');
+
+	// Check if the string is a valid number
+	if (Number.isNaN(Number(numStr))) return 'NaN';
+
+	// Handle negative numbers
+	const isNegative = numStr.startsWith('-');
+	if (isNegative) numStr = numStr.slice(1);
+
+	// Split the number into integer and decimal parts
+	let [integerPart, decimalPart] = numStr.split('.');
+
+	// Remove leading zeros from the integer part
+	integerPart = integerPart.replace(/^0+/, '') || '0';
+
+	// Remove trailing zeros from the decimal part, if it exists
+	if (decimalPart) {
+		decimalPart = decimalPart.replace(/0+$/, '');
+	}
+
+	// Combine the integer and decimal parts back into a number
+	numStr = decimalPart ? `${integerPart}.${decimalPart}` : integerPart;
+
+	// Add the '-' sign back if the number is negative
+	if (isNegative && numStr !== '0') numStr = `-${numStr}`;
+
+	return numStr;
+}
+
+export function numberToExponential(num: string, precision: number = 3): string {
+	if (Number.isNaN(Number(num))) return Number.NaN.toString();
+	return Number(num).toExponential(precision);
+}
+
+export function exponentialToNumber(num: string): string {
+	// Split the number into base and exponent parts
+	const [base, exponent] = num.split(/[eE]/);
+	const baseStr = cleanNumberString(base);
+
+	// If there's no exponent, return the base
+	if (!exponent) return baseStr;
+
+	// Determine the sign of the number
+	const sign = Number(num) < 0 ? '-' : '';
+
+	// Remove the decimal point and negative sign from the base
+	const str = baseStr.replace('.', '').replace('-', '');
+
+	// Convert the exponent to a number and increment it
+	let mag = Number(exponent) + 1;
+
+	// If the magnitude is less than or equal to zero, move the decimal point to the left
+	if (mag <= 0) {
+		let result = `${sign}0.`;
+		while (mag++) result += '0';
+		return result + str;
+	}
+
+	// If the magnitude is greater than zero, move the decimal point to the right
+	mag -= str.length;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/utils/number.ts
```diff
@@ -0,0 +1,159 @@
+/**
+ * Cleans a number string by removing non-numeric characters.
+ *
+ * @param {string} num - The number string to clean.
+ * @returns {string} The cleaned number string.
+ */
+export function cleanNumberString(num: string): string {
```
mwdchang, line 7: Why not just use the native `parseFloat`, since what you are doing is pretty much, well, parsing a float. 

The native implementation has additional features to deal with scientific notations and other things, which yours does not.

packages/client/hmi-client/src/utils/number.ts
```diff
@@ -0,0 +1,159 @@
...
+
+	// Check if the number has a decimal point
+	if (numStr.includes('.')) {
+		const [integerPart, decimalPart] = numStr.split('.');
+		// Remove spaces from the integer part and combine with the decimal part
+		return `${integerPart.replace(/ /g, '')}.${decimalPart}`;
+	}
+	// Remove spaces from the integer part
+	return numStr.replace(/ /g, '');
+}
```
mwdchang, line 136: Another case of `parseFloat`?

```
parseFloat(`num.replace(/ /g, ''))
```

mwdchang:
With regards to **point-2,** I am a bit confused if you are referring to floating-point precision, or if you are talking about allowing arbitrary precision.

- If the former, string vs number will not solve anything, the problem is inherently baked into binary representation. Short of using a specialized library we just need to be careful and deal with it. e.g. `0.2 * 0.2 != 0.04`
- if the latter, I don't think we need to worry too much about it. Our problem domains should not have extremely large numbers or exceedingly small numbers. 

So overall it would be good if you elaborate a bit more on the design goals.

blanchco:
@mwdchang I've address these comments and am focusing on formatting rather than parsing.

Change made:
removed `cleanNumberString`
removed `numberToExponential`, we can just use `toExponential`
`exponentialToNumber` now uses toLocaleString to properly display the full number rather than the custom parsing
`numberToNist` is relatively unchanged aside from using parseFloat to clean the number input
`nistToNumber` cleans _all_ whitespaces and then parseFloat rather than the custom parsing

packages/client/hmi-client/src/utils/number.ts
```diff
@@ -0,0 +1,159 @@
...
+		return result + str;
+	}
+
+	// If the magnitude is greater than zero, move the decimal point to the right
+	mag -= str.length;
+	mag = Math.abs(mag);
+	const zeros = '0'.repeat(mag);
+	console.log(sign + str + zeros);
+	return sign + str + zeros;
+}
```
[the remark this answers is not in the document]
blanchco: rewrote this function to use parseFloat

packages/client/hmi-client/src/utils/number.ts
blanchco: got rid of this function

packages/client/hmi-client/src/utils/number.ts
blanchco: this now uses `parseFloat`. I still need to scrub all the whitespace as `parseFloat` stops parsing at the first instance of a whitespace, while a NIST number can have many whitespaces inside it

mwdchang (approved):

packages/client/hmi-client/tests/unit/utils/number.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+			expect(exponentialToNumber('-2e0')).to.equal('-2');
+		});
+
+		it('should correctly convert one exponent', () => {
+			expect(exponentialToNumber('2e1')).to.equal('20');
+			expect(exponentialToNumber('-2e1')).to.equal('-20');
+		});
+	});
+
+	describe('numberToNist', () => {
```
liunelson, line 38: Need example here and `nistToNumber` and `displayNumber` for this rule
image

>However, when there are only four digits before or after the decimal marker, we recommend that no space is required and no space should be used

For example:
`1234.123` -> `1234.123`
`1234.1234` -> `1234.1234`
`11234.123` -> `11 234.123`
`1234.12344` -> `1 234.123 44`
`11234.123` -> `11 234.123`
`11234.1234` -> `11 234.123 4`

packages/client/hmi-client/tests/unit/utils/number.spec.ts
```diff
@@ -0,0 +1,126 @@
+import { describe, expect, it } from 'vitest';
+import { exponentialToNumber, numberToNist, nistToNumber, displayNumber } from '@/utils/number';
+
+describe('number util tests', () => {
+	describe('exponentialToNumber', () => {
```
liunelson, line 5: Probably should handle this number with thin spaces and exponentials
e.g. `-1 234.567 89e+1` -> `-1.234 567 89e+4` -> `-12345.6789`

PR #3478, opened 2024-04-30
UI plumbing to hook into funman taskrunner

mwdchang:
### Summary
Reroute funman operation UIs to use funman taskrunner interfaces. This is done to avoid any kind of locking problems that can arise from funman-api rest service, since taskrunner runs every request in individual processes.

### Testing
No functional changes.

Hook up funman validation operator, run validation.


packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
```diff
@@ -72,10 +72,10 @@
 import { ref, onMounted, watch } from 'vue';
 import {
 	FunmanProcessedData,
-	getQueries,
 	processFunman,
 	renderFumanTrajectories
 } from '@/services/models/funman-service';
+import { getRunResult } from '@/services/models/simulation-service';
```
YohannParis, line 78: Why are you leveraging the simulation endpoint here?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
YohannParis, line 1: The endpoint should return a [503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) if the task runner fails.

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
mwdchang: Because it has the same mechanism as TA3 "Simulations", that there is an object that encapsulates the running process, which will also track the location(s) of the final result(s). Rather than duplicating the typing and service layer, it makes sense to just multiplex Simulation by adding a "validation" type on top of (ensemble, calibrate, optimize ... etc).

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
mwdchang: You don't know if the task runner will fail at the time the response is returned, since it runs asynchronously in the background. We can communicate errors by ways of adding error messages or error event, but success/failure is not apparent at the time of the request.

YohannParis (approved):

Issue #3476, opened 2024-04-30
Create a new input component


YohannParis:
We need a new basic input component that will handle the new label design, being able to call utilities to modify it's v-model/value.

In the future, or now, we need to display error message and status (yellow border, etc.).

Do not worry about design precision, will pass onto the Design team to ~fix~ improve it.

![Image](https://github.com/DARPA-ASKEM/terarium/assets/636801/8ded208f-0843-4c17-abbe-e8c7442aeaa6)

ngraham76:
@YohannParis  @jamiewaese-uncharted Here are specs for this input component's error state. We wouldn't have a warning (yellow) state - all we need is _valid_ and _error_.

![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/43678375-830a-4808-972e-c7ba979120cd)

ngraham76:
@YohannParis @blanchco  Slight adjustment: after talking to Jamie, the specs for the error state should be as follows (we had a couple of different variants of the error state in the designs - we'll go with this version).

![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/4364c4c6-2728-4324-a5d2-443424b62735)

PR #3483, opened 2024-04-30
custom input fields

blanchco:
# Description

* A custom input field

https://github.com/DARPA-ASKEM/terarium/assets/33158416/1a2dae76-df48-49d9-9ac4-f7db88333742


packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+	<div
+		class="input-container"
+		:class="{ error: errorMessage, disabled: disabled, focused: focused }"
+		@click="focusInput"
+	>
+		<label for="custom-input" class="input-label">{{ label }}</label>
+		<input
```
YohannParis, line 8: What about the input type?

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+			id="custom-input"
+			class="input-field"
+			ref="inputField"
+			:value="modelValue"
+			@input="updateValue"
+			:disabled="disabled"
+			@focus="onFocus"
+			@blur="onBlur"
```
YohannParis, line 16:
```suggestion
			:disabled="disabled"
			:value="modelValue"
			@input="updateValue"
			@focus="onFocus"
			@blur="onBlur"
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+		/>
+	</div>
+	<p v-if="errorMessage" class="error-message">
```
YohannParis, line 19: I think an <aside> would be better here for the error message. What do you think?

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+		<input
+			id="custom-input"
+			class="input-field"
+			ref="inputField"
+			:value="modelValue"
+			@input="updateValue"
+			:disabled="disabled"
+			@focus="onFocus"
+			@blur="onBlur"
+		/>
```
mwdchang, line 17: @shawnyama @YohannParis @blanchco @Tom-Szendrey Tagging you'all as you have seen or involved with quirky input interactions in some manner.

We should come to some consensus no how a particular field should get updated. A pattern we have seen is that - - We use some computed-value to generate the UI, which includes input fields
- The input fields have listeners on @udpate events
- The listeners recalculates the computed-value, which regenerates the UI, which rerenders the input

So there are some problems here:
1. There are a lot of rendering cycles, this can be performance bottle neck 
2. The input is super quirky, because you type a character, and the focus is lost because the element itself is rerendered. So we are virtually entering 1-character-at-a-time

So we need some consensus on either/both
- How the UI is rendered
- How the inputbox changes are propagated, ie instead of @input directly maybe we need to add some debounce mechanism, or maybe we need to consider using @blur as well when focus is intentionally moved away.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+	border-radius: var(--border-radius-small);
+	cursor: text;
+	margin-bottom: var(--gap-small);
+	transition: border-color 0.3s ease-in-out;
+
+	&.disabled {
+		opacity: 0.5;
+	}
+
+	&.focused {
```
YohannParis, line 74: You do not need JS for this:
- https://developer.mozilla.org/en-US/docs/Web/CSS/:focus
- https://developer.mozilla.org/en-US/docs/Web/CSS/:has
```suggestion
	&:has(*:focus) {
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+	&.error {
+		border-color: var(--error-border-color);
+	}
+	input {
+		min-width: 0;
+	}
+}
+
+.input-label {
+	background-color: transparent;
```
YohannParis, line 86: Why not a `none` instead?
```suggestion
	background-color: none;
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+	color: var(--text-color-secondary);
+	cursor: text;
+	padding-right: var(--gap-small);
+}
+
+.input-field {
```
YohannParis, line 92:
```suggestion
input {
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+	}
+	&.error {
+		border-color: var(--error-border-color);
+	}
+	input {
+		min-width: 0;
+	}
+}
+
+.input-label {
```
YohannParis, line 85:
```suggestion
label {
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+}
+
+.input-field {
+	text-align: right;
+	flex-grow: 1;
+	border: none;
+	background-color: transparent;
+}
+
+.error-message {
```
YohannParis, line 99:
```suggestion
input + p {
```
or just `p` there is only one in this component

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
...
+	width: 100%;
+	padding: var(--gap-xsmall) var(--gap-small);
+	background-color: #fff;
+	border: 1px solid var(--surface-border-alt);
+	border-radius: var(--border-radius-small);
+	cursor: text;
+	margin-bottom: var(--gap-small);
+	transition: border-color 0.3s ease-in-out;
+
+	&.disabled {
```
YohannParis, line 70: Same here:
- https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled
```suggestion
	&:has(*:disabled) {
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+	<div
+		class="input-container"
+		:class="{ error: errorMessage, disabled: disabled, focused: focused }"
+		@click="focusInput"
```
YohannParis, line 5: Making sure this click event is not coming from elsewhere: 
- https://vuejs.org/guide/essentials/event-handling.html#event-modifiers
```suggestion
		@click.self.stop="focusInput"
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: ya i can add that in

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: no reason, either works - ill change it to `none`

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: Changing so that I use `useAttrs()`, that way we can bind an input prop we want without explicitly defining it.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: getting rid of disabled in favor of `useAttrs()`

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: changed to `aside`

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: I've added this to the label as well to focus when that it clicked

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,89 @@
...
+	color: var(--text-color-secondary);
+	cursor: text;
+	padding-right: var(--gap-small);
+}
+
+input {
+	text-align: right;
+	flex-grow: 1;
+	border: none;
+	background-color: none;
```
shawnyama, line 76: Apply the Figtree font here? There is a `$fontFamily` variable in `_variables.scss` you just need to make it a css in variable in the root scope

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
shawnyama: Why would `has` need to be used? Can't you declare `disabled/focused` as a prop in the tag instead of putting it in the class?

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,89 @@
...
+</script>
+
+<style scoped>
+.input-container {
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	width: 100%;
+	padding: var(--gap-xsmall) var(--gap-small);
+	background-color: #fff;
```
shawnyama, line 43:
```suggestion
background-color: var(--surface-section);
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+	<div
+		class="input-container"
+		:class="{ error: errorMessage, disabled: attrs.disabled }"
```
shawnyama, line 4:
```suggestion
	:class="{ error: errorMessage  }"
```

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: looks like it was already a variable `--font-family`

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: I dont mind have a meeting around this and coming to a consensus here, perhaps whatever we do decide we can add it in a followup PR?  As for how our custom input is updated here, I more or less follow how primevue is currently doing theirs. https://github.com/primefaces/primevue/blob/master/components/lib/inputtext/InputText.vue

YohannParis:
Please update this PR with the new comment from the design team: https://github.com/DARPA-ASKEM/terarium/issues/3476#issuecomment-2088578025

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -1,52 +1,28 @@
...
-			id="custom-input"
-			class="input-field"
-			ref="inputField"
-			:value="modelValue"
-			@input="updateValue"
-			:disabled="disabled"
-			@focus="onFocus"
-			@blur="onBlur"
-		/>
+	<div class="input-container" :class="{ error: errorMessage }" @click.self.stop="focusInput">
```
YohannParis, line 2: You do not need to use a class here. You can taget `div` directly as the style is scoped.
```suggestion
	<div :class="{ error: errorMessage }" @click.self.stop="focusInput">
```

YohannParis (approved):

blanchco:
> Please update this PR with the new comment from the design team: [#3476 (comment)](https://github.com/DARPA-ASKEM/terarium/issues/3476#issuecomment-2088578025)

@YohannParis already added

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: done

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -0,0 +1,86 @@
...
+
+label {
+	background-color: none;
+	color: var(--text-color-secondary);
+	cursor: text;
+	padding-right: var(--gap-small);
+}
+
+input {
+	text-align: right;
```
YohannParis, line 70: I'm not sure this is the right way for normal text fields.
For number it makes sense and its nice.
Could you check with the design team about it?

blanchco: https://unchartedsoftware.slack.com/archives/C064P3QGVJ6/p1714588663336209?thread_ts=1714576072.852039&cid=C064P3QGVJ6

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
mwdchang: Yes, it will happen with this component. I can setup some thing to discuss.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: right aligned for numbers only

PR #3485, opened 2024-05-01
3345 task migrate code to postgres from es

### LINKED ISSUE #3345: [TASK]: Migrate `Code` to postgres from ES
Move `Code`s from ES to postgres in a simlar fashion as workflows were done in #3261

dvince2:
# Description

* Migrating the code asset from ES to PG

Resolves #3345

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
@@ -53,6 +53,7 @@ public class Code extends TerariumAsset {
 	@ManyToOne
 	@JoinColumn(name = "project_id")
 	@JsonBackReference
+	@TSOptional
 	private Project project;
 
 	@Override
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
@@ -1,62 +1,80 @@
 package software.uncharted.terarium.hmiserver.models.dataservice.code;
 
 import com.fasterxml.jackson.annotation.JsonAlias;
-import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.annotation.JsonBackReference;
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.persistence.Column;
+import jakarta.persistence.ElementCollection;
+import jakarta.persistence.Entity;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
 import java.io.Serial;
+import java.util.HashMap;
 import java.util.Map;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.experimental.Accessors;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
 import software.uncharted.terarium.hmiserver.annotations.TSModel;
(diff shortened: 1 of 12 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/code/Code.java
```diff
@@ -1,62 +1,81 @@
...
-	public enum ProgrammingLanguage {
-		PYTHON("python"),
-		R("r"),
-		Julia("julia"),
-		ZIP("zip");
+	@ManyToOne
+	@JoinColumn(name = "project_id")
+	@JsonBackReference
+	@TSOptional
+	private Project project;
```
mwdchang, line 57: @dvince2 I don't understand these annotation work, can you give a primer?

- Why do we need a project reference, this didn't exist before, and why is it optional?
- Should this be included in the clone(...)?

dvince2: So previously all of our assets existed in ES so there wasn't a link/reference between the `Project` and its assets because we couldn't do a join on the tables, so as a work around we created the stub class `ProjectAsset` which did have a link to Project. 
Now that everything is moving to PG we will eventually be getting rid of the `ProjectAsset` class as it is no longer needed since we can do a join directly between say, the a project and the code table to find all the code assets that belong to a project. 

In the Java object this shows up as a `Project` reference on the `Code` object annotated with `ManyToOne` as well as `JoinColumn`, and on `Project` we now have a `List<Code>` object with `OneToMany`.  The `JoinColumn` annotation identifies the table and column that is required to join on to populate this object (project, and id respectively). `ManyToOne` and `OneToMany` are giving us the relationship type between the two.  The two annotations starting with `@Json` on the two classes handle serialization to stop recursion; Projects contain a Code which contains a Project which contains a Code which... . `@ToString.Exclude` does the same recursion safety but for logging/printing.

Re: cloning... yes, it will. Right now `Clone` isn't being used at all, and is there for some groundwork for the ability to export/duplicate assets and projects.  Its been left off here until `clone` is written on `Project` as well as I"m not sure which of the two need to be cloned first... but my suspicion is project.

mwdchang: @dvince2 cool thanks for the explanation, it helps a lot as it wasn't clear to me which parts are finalized and which parts are pending.

Can you say a thing or two about why `code.project` is optional? With cascading delete, wouldn't we just delete or disable the Code if a project is removed? thus a Code-object should always have a project?
For better or worse, we have a somewhat liberal interpretation of optional means on the UI, so generally speaking the less optionals-types we have the deal with the better 😄

dvince2: So this another little hold over from the `ProjectAsset` era that I'm hoping to fix once we get rid of that, but, that will require changes to how the front end interacts with the server.  Currently when you want to create a `Code` asset (or any asset for that matter) the workflow is...

1) Front end makes a `POST` call to create the Code asset (no project info given), and is returned a Code asset with a UUID
2) Front end makes a `POST` call to create a new ProjectAsset, giving the UUID of the code asset and the UUID of a project.

Keeping this workflow requires us to have an empty (optional) project on Code.  Once we have all of these migrations done, I would very much like to revisit this two step process and instead have a single call to create the Code asset and add it to a project.

dvince2: In the `ProjectController.java` class I've added specific code to link the two up with a hefty TODO on what needs to happen once the migration is done

kbirk (approved):

PR #3491, opened 2024-05-01
Persist Simulation updates in pg, add to Simulation model

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
```diff
@@ -76,6 +83,12 @@ public class Simulation extends TerariumAsset {
 	@TSOptional
 	private UUID projectId; // TODO this can probably be joined to the project table soon?
 
+	@OneToMany(fetch = FetchType.EAGER, cascade = CascadeType.ALL)
+	@OrderBy("createdOn DESC")
+	@ToString.Exclude
+	@JsonManagedReference
+	private List<SimulationUpdate> updates = new ArrayList<>();
```
dvince2, line 90: Do these need to be handled in the `Clone` operation? right now these will be lost.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/SimulationUpdate.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Data
+@TSModel
+@Entity
+public class SimulationUpdate {
+
+	@Id
+	@NotNull private UUID id = UUID.randomUUID();
+
+	@NotNull private UUID simulationId;
```
dvince2, line 27: How is this working without a reference back to the owning project for hibernate? I would have expected it to look like:

```
@ManyToOne
	@JoinColumn(name = "simulation_id")
	@JsonBackReference
	private SImulation simulation;
```

kbirk: I tried to do it as a unidirectional relationship, because I dont really care for it to link back. But that has hibernate treat it as a many to many... oging to make it bidirectional. 

I hate hibernate

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
kbirk: Fixed

dvince2 (approved):

PR #3493, opened 2024-05-01
3441 feat ux design adjustments to the process manager design

asylves1:
# Description

Style changes to process manager notification
image

Resolves #(issue)


packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
```diff
@@ -76,6 +89,11 @@ const panel = ref();
 
 const togglePanel = (event) => panel.value.toggle(event);
 
+const getProjectName = () => {
+	const projectName = useProjects().activeProject.value?.name;
+	return projectName ?? '';
+};
```
YohannParis, line 95: This should `activeProject` should never return `null` it should return an empty string, to avoid making those those test here.

packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
```diff
@@ -32,10 +32,22 @@
 					{{ getTitleText(item) }}
 					<tera-asset-link :label="item.assetName" :asset-route="getAssetRoute(item)" />
 				</p>
-				<p class="msg">{{ item.msg }}</p>
+				<div class="notification-path-text msg">
+					<p>Project: {{ getProjectName() }}</p>
+					<p v-if="item.status !== 'Completed'">{{ item.msg }}</p>
```
YohannParis, line 37: are those status part of an enum type somewhere.
Also, it would be nice to have function like `isComplete` to do those test.

asylves1: added a function &`enum to test the status

packages/client/hmi-client/src/types/common.ts
```diff
@@ -200,6 +200,12 @@ export interface CompareModelsResponseType {
 	response: string;
 }
 
+export enum ItemStatusType {
+	Completed = 'Completed',
+	Failed = 'Failed',
+	Running = 'Running'
+}
+
```
YohannParis, line 208: I mean using the notification status type `Types.ts/ProgressState`

packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
```diff
@@ -103,6 +98,9 @@ const getTitleText = (item: NotificationItem) => {
 	}
 };
 
+const isStatusCorrect = (status: ItemStatusType, item: NotificationItem) =>
+	status === ItemStatusType[item.status];
+
```
YohannParis, line 103: What I meant is simpler function for readability, this is not easier to read
```suggestion
const isComplete = (item: NotificationItem) => item.status === ProgressState.Complete;
const isFailed = (item: NotificationItem) => item.status === ProgressState.Failed;
const isRunning = (item: NotificationItem) => item.status === ProgressState.Running;
```

packages/client/hmi-client/src/types/common.ts
asylves1: `ProgressState` doesn't have the value `Completed` it has `Complete` no `d`.  I would have to add that value to the enum, that feel a off since over could end up using the wrong one

asylves1: That is coming from `item.status`

YohannParis (approved):

packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
YohannParis: This still needs to be resolved

YohannParis (approved):

PR #3499, opened 2024-05-02
jackson confused between two setters

bigglesandginger:
Error seen on server `java.lang.IllegalArgumentException: Conflicting setter definitions for property "data": software.uncharted.terarium.hmiserver.models.notification.NotificationEvent#setData(java.lang.Object) vs software.uncharted.terarium.hmiserver.models.notification.NotificationEvent#setData(com.fasterxml.jackson.databind.JsonNode)`


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationEvent.java
```diff
@@ -64,11 +64,6 @@ protected void onCreate() {
 	@Column(columnDefinition = "text")
 	private JsonNode data;
 
-	public NotificationEvent setData(final JsonNode arg) {
```
kbirk, line 67: What happens now if we pass JsonNode to set data? I thought without the override it would choke on `valueToTree`

kbirk: Disregard, it works fine.

kbirk (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/notification/NotificationEvent.java
bigglesandginger: :) - I only saw the test using that function, and as tests are fake I made the test conform to the actual usage (as I saw)

PR #3501, opened 2024-05-02
Fix stack overflow

kbirk:


YohannParis (approved):
LGTM

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -688,7 +688,6 @@ export interface CiemssStatusUpdate {
     loss: number;
     progress: number;
     jobId: string;
-    dataToPersist: any;
```
dvince2, line 691: If these are super large do we want them passed up to the front end as well or should they be hit with a `@TSIgnore`? (I'm not sure the context here)

dvince2 (approved):

packages/client/hmi-client/src/types/Types.ts
kbirk: Good catch. Fixing

kbirk: Wait nvm the ignore already covered it.

dvince2: We still probably want it on here then that way the field doesn't show up on the front end, even if its empty

PR #3504, opened 2024-05-02
pyciemss simulate test notes

mwdchang:
### Summary
Test notes for PyCIEMSS Simulate operator interface.

### DIFF
--- a/testing/pyciemss-simulate.md
+++ b/testing/pyciemss-simulate.md
@@ -2,6 +2,11 @@
 Please go through __every__ step of the test scenario.\
 When blocked, an error, or a UI/UX anomaly occurs, please report which scenario and step to [\#askem-testing](https://unchartedsoftware.slack.com/archives/C06FGLXB2CE).
 
+Note: sampling combinations in pyciemss can result in numerical instability, when this happens, you can:
+- Retry the simulation, or
+- Fiddle with the paramter distribution ranges, make the intervals larger or smaller
+
+
 ### 1. Begin test
 1. Login to https://app.staging.terarium.ai using the test account
     ```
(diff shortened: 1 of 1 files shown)


testing/pyciemss-simulate.md
```diff
@@ -0,0 +1,43 @@
...
+    email: qa@test.io
+    password: askem-quality-assurance
+    ```
+2. Create, or open, project named `Q&A [Your Name] [YYMMDD]`
+
+### 2. Model setup
+1. Find a Petrinet model (e.g. SIR) and bring it onto a workflow
+2. Connect the model to a Model Configuration operation, create a configuration
+3. Ensure that the model parameters (e.g. beta, gamma) have distribution ranges
+5. Connect the model-configuration to a Simulate-PyCIEMSS operation
```
YohannParis, line 22: Please use the `data/Configured SIR.json` explicitely mentioned in the [template](https://github.com/DARPA-ASKEM/terarium/blob/main/testing/_template.md#4-about-input-file)

YohannParis (changes requested):

testing/pyciemss-simulate.md
mwdchang: That model/json will not work with CIEMSS, thus the entire section here on ranges and distributions.

mwdchang: Uploaded a sample model and updated instructions

YohannParis: Can you also add a GitHub issue bug that values needs to be change to accommodate PyCiemmss quirky behavior.

mwdchang: This is just kind of what can happen when you do sampling-based approaches - you get a weird combination that can get stuck or cause issues.

I don't think it is a bug. We can mitigate by using a super simple model with only one-variable, but that doesn't even test the UI at that point.

YohannParis: What I mean is the problem we have about sending non correct numbers to PyCiemms because they cannot handle them. Instead of making our model configuration contains 0.002 instead of 2, we should have something that do the translation into PyCiemss quirky behaviour.

PR #3518, opened 2024-05-02
3346 task migrate dataset to postgres from es

### LINKED ISSUE #3346: [TASK]: Migrate `Dataset` to postgres from ES
Move `Dataset`s from ES to postgres in a similar fashion as workflows were done in #3261. Note that datasets are searchable and must maintain an ES index

dvince2:
# Description

* Migrating datasets to postgres
* adding test cases
* simplifying creating a project.

Resolves #3346


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
YohannParis, line 1: Should we take advantage here to change those 500 into 503, as it is the DB that's messing up, not the server?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/Dataset.java
```diff
@@ -1,70 +1,129 @@
...
-	 */
+	@ManyToOne
+	@JoinColumn(name = "project_id")
+	@JsonBackReference
+	@TSOptional
+	private Project project;
+
 	@TSOptional
 	@JsonAlias("dataset_url")
+	@Column(length = 1024)
```
kbirk, line 68: Do we want to keep this when we now have the `datasetUrls` list below?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/DatasetColumn.java
```diff
@@ -16,34 +23,73 @@
...
+			clone.annotations = new ArrayList<>();
+			clone.annotations.addAll(this.annotations);
+		}
+
+		if (this.metadata != null) {
+			clone.metadata = new HashMap<>();
+			clone.metadata.putAll(this.metadata);
+		}
+
+		clone.grounding = this.grounding.clone();
```
kbirk, line 85: Should we check for null?

kbirk (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/DatasetColumn.java
dvince2: good call

PR #3542, opened 2024-05-03
Sci input field

blanchco:
# Description

* Here's the new input for sci numbers
* I wanted it to be looked through and see if the approach was good before i get back on Tuesday.  Essentially what happens is that the input gets masked into the NIST format when typing in numbers.  After we are done the blur event will convert this format back into a number.  My thoughts behind that was that when we use this input box we can simply give it a number (without the NIST formatting), show the number (NIST format) and it will spit out a number (without the NIST formatting) again so we don't have to deal with formatting/unformatting outside of the input.
* Please let me know your thoughts!! looking forward to it

https://github.com/DARPA-ASKEM/terarium/assets/33158416/018841d5-b8bc-468d-bac4-7664c12af127


packages/client/hmi-client/src/utils/number.ts
```diff
@@ -68,3 +71,58 @@ export function displayNumber(num: string): string {
 	if (digitString.length > 6) return parseFloat(num).toExponential(3);
 	return numberToNist(num);
 }
+
+export function maskToNistNumber(input: string): string {
+	// Remove any characters that are not digits, '.', '+', '-', or 'e'
```
mwdchang, line 76: This function feels like it is halfway between parsing, formatting, and validating ... which means maybe we are over-thinking this business. People don't usually go about writing their own number parsers, cause the functions are usually built-in .... so we should have a sync up on what we are really after here.

I am likely missing some hidden utilities here, but it sort of feels like I can condense the function down to just this:
```
export function maskToNistNumber(input: string): string {
  let maskedInput = input.replace(/\s/g, '');
  return numberToNist(maskedInput);
}
```

For the parsing/validation part, it will accept `4e3e2e1` which seems quite odd.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -4,18 +4,20 @@
 		<input
 			v-bind="attrs"
 			ref="inputField"
-			:value="modelValue"
+			:value="getValue()"
 			@input="updateValue"
 			:style="{ 'text-align': textAlign }"
+			@blur="unmask"
```
mwdchang, line 10: Don't have a good place to put this, but there are two things worth mentioning related to how the template is used:
1. Note `input[type=number]` automatically validates the input is actually a number, which type=sci doesn't do
2. There is not such thing as `input[type=sci]`, it isn't proper HTML, if we are rolling our own stuff it is probably `input[type=text]`

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -25,14 +27,33 @@ const emit = defineEmits(['update:modelValue']);
...
-	const value = (event.target as HTMLInputElement).value;
-	emit('update:modelValue', value);
+	const target = event.target as HTMLInputElement;
+	const value = target.value;
+
+	if (attrs.type === 'sci') {
+		// Mask the value before emitting the update
+		const maskedValue = maskToNistNumber(value);
+
+		emit('update:modelValue', maskedValue);
```
mwdchang, line 48: As briefly discussed, "mask" has a specific meaning in computer-science (usually involving bitwise operations) and it doesn't look like it applies here, if we are just formatting I think calling formatXYZ will be better for code understanding.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: I needed a unique way to identify that we are using a custom scientific number input I thought that this might be the easiest way for us to use this in other components, using text would be just simply for text fields. 

The issue with using number is that using that type will try to auto parse into a number but we needed a custom string (the number with the spaces).  As far as I know using just number cannot do this.  i.e if i typed in `12345`.  The mask would change the value to `12 345` but the number type would not recognize this as invalid.

packages/client/hmi-client/src/utils/number.ts
blanchco: I did try this - how there are many cases where this will not work since this is for input.  i.e if the user wanted decimal places they would type in `123.` and parseFloat would parse out the decimal and the input would be left at just `123` without the user ever being about to type in a decimal place.

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
blanchco: I looked a little further this morning, the term is `input masking` and is used to mask user input into certain formats https://designsystem.digital.gov/components/input-mask/.

blanchco:
Updated with some suggestions from this [thread](https://unchartedsoftware.slack.com/archives/C064P3QGVJ6/p1715101561728879)

packages/client/hmi-client/src/utils/number.ts
```diff
@@ -66,5 +70,24 @@ export function displayNumber(num: string): string {
...
 	if (digitString.length > 6) return parseFloat(num).toExponential(3);
-	return numberToNist(num);
+	return numberToNist(parseFloat(num).toString());
+}
+
+export function scrubAndParse(v) {
+	const scrubbed = v.replace(/\s/g, '');
+
+	// check if last digit is a decimal point
+	if (scrubbed[scrubbed.length - 1] === '.') return false;
```
mwdchang, line 80: Why is last character being `.` a problem? Isn't `123.` or `3.` just a floating point number?

YohannParis: I mean NIST is not clear on the subject [paragraph 10.5.2 Decimal sign or marker](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-10-more-printing-and-using-symbols-and-numbers)

I do agree that we should not parse the input until the full number is entered. A user would not enter a `.` without any value behind. Or if it does, the on blur will make it a number

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-input-field.vue
```diff
@@ -24,15 +26,59 @@ defineProps<{
 const emit = defineEmits(['update:modelValue']);
 const inputField = ref<HTMLInputElement | null>(null);
 const attrs = useAttrs();
+const error = ref('');
+const maskedValue = ref('');
+
+const textAlign = computed(() => (attrs.type === 'number' || isNistType.value ? 'right' : 'left'));
+
+const isNistType = computed(() => attrs.type === 'nist');
```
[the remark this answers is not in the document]
blanchco: nope Ill change to consts

packages/client/hmi-client/src/utils/number.ts
blanchco: from the code if the user were to type in `123.` this would evaluate as true and then call the numberToNist function which will filter out the decimal point back to `123`.

packages/client/hmi-client/src/utils/number.ts
```diff
@@ -66,5 +70,24 @@ export function displayNumber(num: string): string {
 	digitString = digitString.replace('.', '');
 
 	if (digitString.length > 6) return parseFloat(num).toExponential(3);
-	return numberToNist(num);
+	return numberToNist(parseFloat(num).toString());
+}
+
+export function scrubAndParse(v) {
```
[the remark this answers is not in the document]
blanchco: add comments and jsdoc

PR #3545, opened 2024-05-06
quick change to tile

asylves1:
# Description

Add no-wrap style to title

Resolves #(issue)


shawnyama (approved):

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -75,6 +77,12 @@ header > * {
 	color: var(--text-color-primary);
 }
 
+header .title {
```
YohannParis, line 80: just a `h4` would have been enough, no need for extra classname.

PR #3546, opened 2024-05-06
3490 Added projectId to TaskRequest / TaskResponse and Notification Group

### LINKED ISSUE #3490: Add projectId to TaskRequest / TaskResponse and NotificationGroup
Add a `UUID projectId` field to `NotificationGroup` and the `TaskRequest` / `TaskResponse` classes.

There are two sets of `TaskRequest` and `TaskResponse`, one in `packages/server` and one in `packages/taskrunner`. 

Ensure that the `projectId` is added to both sets of classes and is propagated from the `TaskRequest` to the `TaskResponse`.

The `projectId` field should also be set in all places that a `NotificationGroup` is instantiated.

jryu01:
# Description

Added projectId to TaskRequest / TaskResponse and Notification Group

Resolves #3490


packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/models/task/TaskResponse.java
```diff
@@ -16,6 +16,7 @@ public class TaskResponse implements Serializable {
 	private TaskStatus status;
 	private byte[] output;
 	private Object additionalProperties;
+	private String projectId;
```
dvince2, line 19: On both the models these should be typed `UUID` for projects to stay consistent - the typescript generated will stay as a `string` though!

dvince2 (approved):
Over all looks good - just consider changing the back end code to use `UUID` in favour of `String`

PR #3547, opened 2024-05-06
3532 task associate workflow with a project in postgres

### LINKED ISSUE #3532: [TASK] Associate workflow with a project in postgres

dvince2:
# Description

* Adding project reference to workflow, required for asset reference

Resolves #3532


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/Workflow.java
```diff
@@ -1,20 +1,20 @@
...
 import java.util.Map;
 import java.util.UUID;
-import jakarta.persistence.JoinColumn;import jakarta.persistence.ManyToOne;import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.experimental.Accessors;
 import org.hibernate.annotations.JdbcTypeCode;
 import org.hibernate.type.SqlTypes;
-import software.uncharted.terarium.hmiserver.annotations.TSOptional;import software.uncharted.terarium.hmiserver.models.TerariumAsset;
-import software.uncharted.terarium.hmiserver.models.dataservice.ObjectConverter;import software.uncharted.terarium.hmiserver.models.dataservice.project.Project;
+import software.uncharted.terarium.hmiserver.annotations.TSOptional;
```
dvince2, line 16: @YohannParis linter obliterated my imports??

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -404,6 +404,7 @@ export interface Project extends TerariumAsset {
      */
     projectAssets: ProjectAsset[];
     codeAssets: Code[];
+    workflowAssets: Workflow[];
```
mwdchang, line 407: FYI: This is going to cause a bit of confusion because the UI has a different typing for Workflow.


Side question: does this construct mean that whenever we load up a project we will automatically load its code, workflow, model... and what not? Is that going to be cost prohibitive? Or is this like a lazy load kind of thing where you only load stuff when you access them?

dvince2: On the first point, I'm wondering how we can align the two. I know that theres some weird typing things happening on the front end - i wonder if we can get away with just json blobs for the back end..

On the second point - I will investigate more, thats a good point.

mwdchang: The problem is the lack of better dynamic typing support and inferencing in Java 😛 

So you can statically type out everything if we remove the idea of "Partials" from the UI-code, but that comes at a detriment to the front-end code ... unless, you want to move all the workflow related process logic down to the Java level.

dvince2: I'm going to raise the size of `Project` going to the front end at the next hmi-server call

kbirk (approved):

PR #3550, opened 2024-05-06
Use hypersistence for json serialization

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Grounding.java
```diff
@@ -23,12 +26,14 @@ public class Grounding implements Serializable {
 	private static final long serialVersionUID = 302308407252037615L;
 
 	/** Ontological identifier per DKG */
-	@JdbcTypeCode(SqlTypes.JSON)
+	@Type(JsonType.class)
+	@Column(columnDefinition = "text")
```
dvince2, line 30: These will require a migration

dvince2: (and the other type changes will too - I won't comment spam on them :)

dvince2 (approved):
I think this looks good, but requires a migration for the change in field types.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Grounding.java
kbirk: No migration needed, as we reverted datasets and these Groundings will all be created fresh in their own table.

PR #3558, opened 2024-05-07
Llm stratify

mwdchang:
### Summary
Add llm toolbar to stratify operator for code generation

Note:
- Not all prompts work, however this template of `show me code to stratify model abc by region named xyz and zzz` had worked more consistently
- The code generated is not precisely correct, it should be "model = stratify(..." instead of "stratified_model = stratify(...". However this is outside of the scope of this PR to address.

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
@@ -192,7 +192,6 @@ const outputs = computed(() => {
 	return [];
 });
 
-const kernelReady = ref(false);
 const kernelManager = new KernelSessionManager();
 
 let editor: VAceEditorInstance['_editor'] | null;
@@ -375,7 +374,6 @@ const inputChangeHandler = async () => {
 		const jupyterContext = buildJupyterContext();
 		if (jupyterContext) {
 			await kernelManager.init('beaker_kernel', 'Beaker Kernel', buildJupyterContext());
-			kernelReady.value = true;
 		}
 	} catch (error) {
 		logger.error(`Error initializing Jupyter session: ${error}`);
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -41,8 +41,14 @@
 			</tera-drilldown-section>
 		</div>
 		<div :tabName="StratifyTabs.Notebook">
-			<tera-drilldown-section>
-				<p class="mt-3 ml-4">Code Editor - Python</p>
+			<tera-drilldown-section class="notebook-section">
+				<tera-notebook-jupyter-input
```
Tom-Szendrey, line 45: I am confused why you do not need `Suspense`

Either its not required or I am unable to click fast enough to send something before the kernelManager is built so it doesnt appear to be a problem.

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -185,6 +192,7 @@ const outputs = computed(() => {
 	return [];
 });
 
+const kernelReady = ref(false);
```
Tom-Szendrey, line 195: I see that youre setting this value, but I do see where you're using it?

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -483,6 +497,16 @@ onUnmounted(() => {
...
+.notebook-section:deep(main .notebook-toolbar),
+.notebook-section:deep(main .ai-assistant) {
+	padding-left: var(--gap-medium);
+}
+
+.notebook-section:deep(main) {
+	gap: var(--gap-small);
+	position: relative;
+}
+
```
shawnyama, line 509: Maybe write a fix me so someone can fix the css rules in the component so its more plug and play

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
mwdchang: will follow up offline.

PR #3566, opened 2024-05-07
fix: remove auto run in model edit notebook

shawnyama:
# Description

* Don't auto run notebook code (this was only happening when the Notebook tab was opened as soon as the model edit drilldown is opened)


mwdchang:
Tagging @Tom-Szendrey , I think there maybe a reason why we want to run automatically? maybe to restore to previous output's state?

PR #3578, opened 2024-05-08
Updated task service to send client notification events

jryu01:
# Description

- Updated task service to send client notification events to the client.
- Added optional notificationGroupId and projectId fields to the ClientEvent. 

Resolves #(issue)


packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/models/task/TaskResponse.java
```diff
@@ -15,6 +15,8 @@ public class TaskResponse implements Serializable {
 	private String script;
 	private TaskStatus status;
 	private byte[] output;
+	private Object type;
```
dvince2, line 18: Should this be `ClientEventType` not `Object`?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/task/TaskNotificationEventData.java
```diff
@@ -0,0 +1,36 @@
...
+
+	private TaskType taskType;
+
+	private TaskStatus status;
+
+	private String message;
+
+	private String error;
+
+	private Object data;
```
dvince2, line 24: What is `data` here? If its just JSON data this can be a `JsonNode`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/task/TaskRequest.java
```diff
@@ -13,6 +13,7 @@
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.experimental.Accessors;
+import software.uncharted.terarium.hmiserver.models.ClientEventType;
```
kbirk, line 16: We cant do this, if we import from the hmiserver it will run the entire configuration.

packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/models/task/TaskResponse.java
jryu01: This is taskrunner package and I wasn't sure if it's possible to import types/classes from the hmi server. So I set it to Object so can it can handle arbitrary object when the task messages get encoded/decoded.

kbirk: We can't import the type directly from hmi-server since it causes config issues with spring. I recall when i tried to import TaskRequest / TaskResponse from hmiserver it imported the entire app config.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
```diff
@@ -87,16 +91,19 @@ private static class TaskRequestWithId extends TaskRequest {
...
 
 		public TaskResponse createResponse(final TaskStatus status) {
 			return new TaskResponse()
 					.setId(id)
 					.setStatus(status)
+					.setType(type)
 					.setUserId(userId)
 					.setProjectId(projectId)
 					.setScript(getScript())
+					.setNotificationEventType(notificationEventType)
```
kbirk, line 106: I think a better way to deal with this is to set the `type` on the group to the `getScript` of the TaskRequest / TaskResponse. We don't need a separate ClientEventType on the taskrunner stuff.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/task/TaskNotificationEventData.java
kbirk: Why are we not just sending the TaskResponse with all these fields as we did before? Why create a near identical class?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
```diff
@@ -551,7 +581,8 @@ public TaskFuture runTaskAsync(final TaskRequest r) throws JsonProcessingExcepti
 				// create the notification group for the task
 				final NotificationGroup group = new NotificationGroup();
 				group.setId(req.getId()); // use the task id
-				group.setType(req.getType().toString());
+				group.setType(ensureNotificationEventType(req.getNotificationEventType())
```
kbirk, line 584: We should just use `group.setType(req.getScript())` here

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
```diff
@@ -431,15 +443,33 @@ private void onTaskResponseOneInstanceReceives(final Message message) {
...
 
 				log.info("Creating notification event under group id: {}", resp.getId());
 
 				notificationService.createNotificationEvent(resp.getId(), event);
+
 			} catch (final Exception e) {
 				log.error("Failed to persist notification event for for task {}", resp.getId(), e);
 			}
 
+			try {
```
kbirk, line 456: This code needs to be moved from `onTaskResponseOneInstanceReceives` method  to the `onTaskResponseAllInstanceReceive` method.

packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/models/task/TaskResponse.java
kbirk: I don't think we need an extra `"type"` fiield on the TaskRequewst / TaskResponse. The `script` field can be used as a `"type"`. We can either use it directly, or map from the `String` to the `ClientEventType` on the hmi-server.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/task/TaskNotificationEventData.java
jryu01: I initially thought of creating an object with only necessary fields just for notification handling excluding actual output data. How large output field can get from the response object? If it's neglectable, I'll just use resp since this turned out very similar to the response object as you mentioned.

packages/taskrunner/src/main/java/software/uncharted/terarium/taskrunner/models/task/TaskResponse.java
jryu01: Done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/task/TaskNotificationEventData.java
jryu01: I just updated to send the task response data.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
jryu01: Done

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extractionservice/ExtractionStatusUpdate.java
```diff
@@ -7,19 +7,12 @@
...
 	private String error;
 
-	public ExtractionStatusUpdate(
-			final UUID notificationGroupId,
-			final UUID documentId,
-			final Double t,
-			final String message,
-			final String error) {
-		this.notificationGroupId = notificationGroupId;
+	public ExtractionStatusUpdate(final UUID documentId, final Double t, final String message, final String error) {
```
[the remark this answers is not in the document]
jryu01: I'll update that in my next work, the client isn't handling the projectId yet and that needs to be updated as well.

kbirk (approved):
Looks great!

PR #3579, opened 2024-05-08
DB fixes and migrations

kbirk:


packages/server/README.md
```diff
@@ -22,3 +22,158 @@ To build the NATIVE image simply run the following from the root directory.
...
+### Defining Bi-directional Relationships
+
+For `@OneToOne` and `@OneToMany` bi-directional relationships we prefer to not use an additional table.
+
+These relationships are setup as follows:
+
+- `@OneToOne` / `@OneToMany` / `@ManyToOne` / `@ManyToMany` annotations are used to define the type of relationship between the entities.
+    - `mappedBy` defines which side "owns" the relationship. For `OneToOne` and `@OneToMany` it defines which side _does not_ contain the foreign key. If `mappedBy` is not specified, an additional table is used to store the relationship. For `@ManyToMany` relationships, it is used to define the "owner" side of the many to many.
+    - `cascade` defines what types of operations are cascaded from the owner to the other side of the relationship. `CascadeType.ALL` seems to be a safe default.
+    - `fetch` defines how the elemeents are fetched. `EAGER` fetches them immediately. `LAZY` fetches them only when accessed.
```
bigglesandginger, line 58: elemeents misspelled - I need to register that I am reading this PR :)

packages/server/README.md
```diff
@@ -22,3 +22,158 @@ To build the NATIVE image simply run the following from the root directory.
...
+```java
+@Data
+@Entity
+class Parent {
+
+    @Id
+    private UUID id = UUID.randomUUID();
+
+    @OneToMany(cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+	@JoinColumn(name = "parent_id") // <--- because this is a @OneToMany uni-directional relationship, its referencing the column on the child, not the parent.
```
bigglesandginger, line 128: it's not its

packages/server/README.md
```diff
@@ -22,3 +22,158 @@ To build the NATIVE image simply run the following from the root directory.
...
+    @Override
+    public void migrate(Context context) throws Exception {
+        try (var statement = context.getConnection().createStatement()) {
+            statement.execute("ALTER TABLE simulation ALTER COLUMN description TYPE text;");
+        }
+    }
+}
+```
+
+Only use a java based migration if it is not possible to achieve via a `.sql` one.
```
bigglesandginger, line 179: This is great!  It sheds light on a grossly obsfucating pointless layer of bad code design.

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumEntity.java
```diff
@@ -0,0 +1,63 @@
...
+	protected void onCreate() {
+		this.createdOn =
+				Timestamp.from(ZonedDateTime.now(ZoneId.systemDefault()).toInstant());
+		this.updatedOn = this.createdOn;
+	}
+
+	@TSOptional
+	@Schema(accessMode = Schema.AccessMode.READ_ONLY)
+	@Column(columnDefinition = "TIMESTAMP WITH TIME ZONE")
+	private Timestamp updatedOn;
```
dvince2, line 47: is `deletedOn` not here?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
```diff
@@ -85,7 +85,6 @@ public class Simulation extends TerariumAsset {
 
 	@OneToMany(mappedBy = "simulation", fetch = FetchType.EAGER, cascade = CascadeType.ALL)
 	@OrderBy("createdOn DESC")
-	@ToString.Exclude
```
dvince2, line 88: Is the infinite loop on `toString` no longer a problem here?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumEntity.java
kbirk: It's on TerariumAsset, I am now realizing we will need to set up triggers to cascade our soft deletes....

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/Simulation.java
kbirk: Doesn't seem to be if we use `JsonManagedReference` and `JsonBackReference`

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumEntity.java
dvince2: Right I saw it there - I'm just curious why `deletedOn` wasn't moved down to `TerariumEntity` as well

kbirk: I left it out since that logic is built into the TerariumAssetService stuff, and those nested fields don't have their own services.

PR #3580, opened 2024-05-08
feat(model-edit): better output management

shawnyama:
# Description

- Now when you open the model edit drilldown it will start at your last edited model/output rather than starting at the original model
- A default output is created which is the input 
- In order to start editing from the original model you would choose the first output (the original model). Once an edit is made to the original model a new output model is created. Any edits after that would just be applied to your new output model.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/8053bcc2-9cb7-41d1-907f-f726759064d1


jryu01:
https://github.com/DARPA-ASKEM/terarium/assets/3279838/ddc54997-958e-44b1-b9b7-4037ba32aeaa

It looks like I can't select back the original model

jryu01:
> model-edit.mov 
> It looks like I can't select back the original model

NVM I pulled the latest changes and It seems working fine. 

LGTM!

jryu01 (approved):

packages/client/hmi-client/src/components/workflow/ops/model-edit/model-edit-operation.ts
```diff
@@ -9,8 +10,8 @@ export interface ModelEditCode {
 }
 
 export interface ModelEditOperationState extends BaseState {
-	modelEditCodeHistory: ModelEditCode[];
-	hasCodeBeenRun: boolean;
+	notebookHistory: NotebookHistory[];
+	hasCodeRun: boolean;
```
YohannParis, line 14: Thank you for the better english!

PR #3604, opened 2024-05-10
Fix objectmapper

kbirk:


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/SnakeCaseRequestControllerAdvice.java
```diff
@@ -1,36 +1,44 @@
...
 
 @RestControllerAdvice
 @RequiredArgsConstructor
 @Slf4j
 public class SnakeCaseRequestControllerAdvice extends RequestBodyAdviceAdapter {
-	private final ObjectMapper snakecaseMapper;
-	private final ObjectMapper camelcaseMapper;
+
+	@Autowired
+	private ObjectMapper snakecaseMapper;
```
dvince2, line 29: how does `@Autowired` here differ in practice from initing' these ourselves or using the `@RequiredArgsConstructor`?  If autowired is the answer i think we can drop the requiredargsconstructor annotation

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/SnakeCaseRequestControllerAdvice.java
kbirk: We  want spring to inject it, because the default `ObjectMapper` is not the same as the one Spring injets, which has a bunch of random properties set, hence the difference between the snake case and default object mapper used.

kbirk: Commented as such

PR #3605, opened 2024-05-10
flag workflow to be saved on remove-node and duplicate-branch

mwdchang:
### Summary
For some reason we forgot to flag workflow to be saved on 
- operator node removal
- branch workflow

### Testing
Remove a node in a workflow, then quick navigate away, and come back, note the removed node should not reappear.


YohannParis (approved):
Nice catch!

PR #3606, opened 2024-05-13
fix cache not refresh upper and lower bound on upstream change

mwdchang:
### Summary
`requestParameter` is saved into the operator state, but when upstream configuration changes it needs to be refreshed.

### Testing
- Create a simple funman workflow, e.g. `SIR => model-config => funman`
- Select beta/gamma are parameters of interest under "additional options"
- Run funman
- Goto model-config operator, change beta/gamma upper/lower ranges
- Goto funman operator, run funman
- Compare the two results, verify that they have different output ranges based on the model configuration settings.

image


YohannParis (approved):
LGTM

PR #3607, opened 2024-05-13
Update Project to require a name.

YohannParis:
# Description

* Require a title to create/edit a project.
* Remove the need to provide a `userId` to `hmi-server`.
* Disable the submit button if the title is empty.

### DIFF
--- a/packages/client/hmi-client/src/page/project/components/tera-project-configuration-modal.vue
+++ b/packages/client/hmi-client/src/page/project/components/tera-project-configuration-modal.vue
@@ -98,7 +98,6 @@ import Textarea from 'primevue/textarea';
 import TeraModal from '@/components/widgets/tera-modal.vue';
 import Button from 'primevue/button';
 import { useProjects } from '@/composables/project';
-import useAuthStore from '@/stores/auth';
 import { cloneDeep } from 'lodash';
 import { useRouter } from 'vue-router';
 import { RouteName } from '@/router/routes';
@@ -112,20 +111,18 @@ const props = defineProps<{
 
 const emit = defineEmits(['close-modal']);
 
-const auth = useAuthStore();
 const router = useRouter();
-const userId = auth.user?.id ?? '';
 
 const title = ref(props.project?.name ?? '');
 const description = ref(props.project?.description ?? '');
 const isApplyingConfiguration = ref(false);
 
-async function createProject() {
+async function createProject(name: string, desc: string) {
 	isApplyingConfiguration.value = true;
-	const project = await useProjects().create(title.value, description.value, userId);
+	const project = await useProjects().create(name, desc);
 	if (project?.id) {
-		router.push({ name: RouteName.Project, params: { projectId: project.id } });
 		emit('close-modal');
+		await router.push({ name: RouteName.Project, params: { projectId: project.id } });
 	}
 }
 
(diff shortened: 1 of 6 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -96,20 +96,20 @@ public class ProjectController {
...
-					<li><strong>Upload stuff:</strong> Upload documents, models, code or datasets with the green button in the bottom left corner.</li>
-					<li><strong>Explore and add:</strong> Use the project selector in the top nav to switch to the Explorer where you can find documents, models and datasets that you can add to your project.</li>
-					<li><strong>Build a model:</strong> Create a model that fits just what you need.</li>
-					<li><strong>Create a workflow:</strong> Connect resources with operators so you can focus on the science and not the plumbing.</li>
-				</ul>
-			<br>
-			<p>Feel free to erase this text and make it your own.</p>
-		</div>
-		""";
+			<div>
```
dvince2, line 99: How is this block's formatting changed? i would have thought the linter would have stopped that?

packages/client/hmi-client/src/page/project/components/tera-project-configuration-modal.vue
```diff
@@ -111,20 +112,19 @@ const props = defineProps<{
...
 const title = ref(props.project?.name ?? '');
+const isTitleInvalid = computed(() => isEmpty(title.value));
 const description = ref(props.project?.description ?? '');
 const isApplyingConfiguration = ref(false);
 
-async function createProject() {
+async function createProject(name: string, desc: string) {
 	isApplyingConfiguration.value = true;
-	const project = await useProjects().create(title.value, description.value, userId);
+	const project = await useProjects().create(name, desc);
```
dvince2, line 124: We have the back end returning a `400` for a project without a name which is perfect, but, do we want to disable the 'ok' button or check for an empty name here to save us a 400 just in case (and have a nicer looking error as a result, or the inabillity to even hit OK?)

dvince2 (approved):

packages/client/hmi-client/src/page/project/components/tera-project-configuration-modal.vue
```diff
@@ -111,20 +112,19 @@ const props = defineProps<{
 
 const emit = defineEmits(['close-modal']);
 
-const auth = useAuthStore();
 const router = useRouter();
-const userId = auth.user?.id ?? '';
 
 const title = ref(props.project?.name ?? '');
+const isTitleInvalid = computed(() => isEmpty(title.value));
```
mwdchang, line 118: What do we want to enforce here? So `""` is invalid, but `" "` is okay?

Is title being empty `""` causing problems?

YohannParis: It's not, but the UI is not made for, which creates a lot of glitch.
This is a basic test, and if someone really wants to have a name equal one space, their prerogative.

packages/client/hmi-client/src/page/project/components/tera-project-configuration-modal.vue
YohannParis: I agree, but I think revamping the whole front-end for that was a bit too much for that simple PR.

PR #3608, opened 2024-05-13
Updated client notification system to handle task runner notifications

jryu01:
# Description

- Updated client notification to handle task runner notifications
- Added a handler for GoLLM model card notification. 

https://github.com/DARPA-ASKEM/terarium/assets/3279838/716b7247-df2d-452e-ab46-41e1a5248172

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/ProgressState.java
```diff
@@ -3,6 +3,8 @@
 import com.fasterxml.jackson.annotation.JsonAlias;
 
 public enum ProgressState {
+	@JsonAlias("cancelling")
```
dvince2, line 6: I think that this will require a DB migration @kbirk ?

dvince2 (commented):
I believe that modifying the enums used in our models will require a migration.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/ProgressState.java
jryu01: Is the DB migration something that can be done easily? I'm not quite familiar with the process.

packages/client/hmi-client/src/services/notificationEventHandlers.ts
```diff
@@ -3,19 +3,36 @@ import {
...
+
+const getStatusFromTaskResponse = (data: TaskResponse) => {
+	const statusMap = {
+		[TaskStatus.Queued]: ProgressState.Queued,
+		[TaskStatus.Running]: ProgressState.Running,
+		[TaskStatus.Success]: ProgressState.Complete,
+		[TaskStatus.Failed]: ProgressState.Failed,
+		[TaskStatus.Cancelled]: ProgressState.Cancelled,
+		[TaskStatus.Cancelling]: ProgressState.Cancelling
+	};
```
mwdchang, line 32: @dvince2 @jryu01 So does this mean that `ProgressState` is the canonical state we are going with? Maybe a little comment here would help?

Also it seems unnecessary to (re)create this lookup every time the function is called, can we move it outside and treat statusMap as static?

jryu01: Done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/simulation/ProgressState.java
jryu01: I placed the new enum value to the end so that it won't shuffle the existing value (e.g make sure `cancelled` was 0 and 0 is not replaced by `cancelling`) which might cause the issue with the `state` in `NotificationEvent`. Aside from that It doesn't look like we need to worry about the migration.

blanchco (approved):
I tested this out locally and it looks good. Nice job! ✅

PR #3610, opened 2024-05-13
additional sanity checks for model config

blanchco:
# Description

Resolves #(issue)


packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table.vue
```diff
@@ -343,6 +343,18 @@ const buildParameterTable = () => {
...
+					const lb = '';
+					const ub = '';
+					param.distribution = {
+						type: 'StandardUniform1',
+						parameters: {
+							minimum: lb,
+							maximum: ub
+						}
+					};
+				}
```
mwdchang, line 357: Do you have an example of model-configuration that you can share? Do we know if the model-configs produces will work downstream?

Asking because you are treating minimum and maximum as if they are strings, but they should be floats?

blanchco: I initiate them as empty strings just so that the user has the option to enter nothing in the field.  However when saving the `cleanModel` function is run which will completely remove the distribution object if there is an empty string, so i _think_ it can be used down the line properly.

Here's an example model config id: `ecfcde45-9be1-42a2-ba51-ac41a0e6c2aa`.  The gamma, I0, R0 had their distribution objects removed

mwdchang (approved):

PR #3619, opened 2024-05-14
Add permission checking to assets

bigglesandginger:
This needs help testing the front-end changes as I do not know them well enough.


YohannParis (commented):
If the goal is to pass the current projectId to ALL request, then using an Axios interceptor would be more appropriate than editing 100 files!

bigglesandginger:
Not all requests - there are many that do not require it... thoughts?

packages/client/hmi-client/src/composables/project.ts
YohannParis, line 1: That's a convoluted way of doing it, but that's Vue for you right there!

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -98,7 +98,7 @@ export async function compareModels(modelIds: string[]): Promise<CompareModelsRe
 	});
 
 	try {
-		const response = await API.get<TaskResponse>('/gollm/compare-models', {
+		const response = await API.get<TaskResponse>(`/gollm/compare-models`, {
```
YohannParis, line 101: Why this was necessary?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
```diff
@@ -313,26 +313,19 @@ public PermissionGroup getGroup(String id) {
 		return permissionGroup;
 	}
 
-	public boolean canRead(SchemaObject who, SchemaObject what) throws Exception {
+	public boolean can(SchemaObject who, Schema.Permission permission, SchemaObject what) throws Exception {
```
YohannParis, line 316: Could those method have Java Doc to understand how to read and use them?
It would be easier to use.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -158,7 +159,7 @@ public T createAsset(final T asset) throws IOException {
 	 */
 	@Override
 	@Observed(name = "function_profile")
-	public List<T> createAssets(final List<T> assets) throws IOException {
+	public List<T> createAssets(final List<T> assets, final Schema.Permission hasWritePermission) throws IOException {
```
YohannParis, line 162: I don't get it here? Why passing that `hasWritePermission` all the way from `Controller -> Xservice -> ITerariumAssetService -> TerariumAssetService` to not use it?
If the user cannot create an asset on this project, shouldn't be the `Controller` to return an 403 error?

packages/client/hmi-client/src/services/goLLM.ts
bigglesandginger: It was probably from my backing out inserting `?project-id=${projectId}` and the quotes did not change... fixing

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
bigglesandginger: It is the only way I can think of that informs the developer that they need to check for permissions on these calls, the fact that it is an unused parameter lower down is a "feature" of getting the compiler to inform the developer they have missed something vital and totally tangential.  A comment would be absolutely ignored.

bigglesandginger: I suppose I could add checks to satisfy the parameter, but they are checking for the non-absence of a permission at the wrong spot and as you say, there is not the right place to throw a 403.  Instead the `Controller` (as I have written the implementation) is first determining if the user even has the correct rights to continue before doing any additional work on the server.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
bigglesandginger: done

bigglesandginger:
Anyone have any more comments or can we push this through?

YohannParis (approved):

PR #3620, opened 2024-05-14
layout changes for edit & config drilldowns

asylves1:
# Description

Static layout change for the Edit & Config
image

image

Resolves #(issue)


shawnyama (approved):
I think defining `menuItems` for specific operations is good. I think we could inherit he default ones from a service like `/services/workflow.ts`

PR #3621, opened 2024-05-14
Fix NPE in rabbitmq handler

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -156,51 +156,56 @@ private void send(final Object message, final List<SseEmitter> emitterList, fina
...
 		if (messageJson == null) {
 			return;
 		}
+
 		final String userId = messageJson.at("/userId").asText();
-		final List<SseEmitter> emitterList = userIdToEmitter.get(userId);
 		synchronized (userIdToEmitter) {
-			send(messageJson.at("/event"), emitterList, userId);
+			final List<SseEmitter> emitterList = userIdToEmitter.get(userId);
+			if (emitterList != null) {
```
kbirk, line 188: If the `userId` was missing, or if there were no emitters for the user id, that list would result in a NPE.

bigglesandginger: Ah... that is a smart catch.  Also explains why we probably don't see it locally as it is a curious race condition

bigglesandginger (approved):

PR #3625, opened 2024-05-14
Ts/3524 task optimize box updates

Tom-Szendrey:
# Description

- Correcting watcher on run. Now you will not longer need to refresh to see output results
- Move `getOptimizedInterventions` to .ts file for shared use and update it to be able to handle both intervention types
- Allow user to save model config output which will pull optimized interventions
- move maxiter and maxfeval into the additional options section
- Remove model name and desc from state as theyre really not needed there.


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -95,11 +95,30 @@ export const OptimizeCiemssOperation: Operation = {
...
-			modelConfigDesc: '',
 			optimizeErrorMessage: { name: '', value: '', traceback: '' },
 			simulateErrorMessage: { name: '', value: '', traceback: '' }
 		};
 		return init;
 	}
 };
+
+// Get the intervention output from a given optimization run
+export async function getOptimizedInterventions(optimizeRunId: string) {
```
Tom-Szendrey, line 106: Im just correcting this now 
At the moment this is only working for 1 or the 2 types

mwdchang: I don't understand what you are saying

mwdchang (approved):

PR #3631, opened 2024-05-15
fix(model-edit): code is saved properly in output state

shawnyama:
# Description

The notebook code was only saved once the new output model was created. Any edits after that were not saved.

Now the code is saved properly.

BEFORE:

https://github.com/DARPA-ASKEM/terarium/assets/28237258/59439e70-ba39-49f3-ba77-e161f3c772ea

AFTER:

https://github.com/DARPA-ASKEM/terarium/assets/28237258/eace54c6-33c2-4638-ae15-b64dcd00f1c6

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit.vue
@@ -112,7 +112,13 @@ import { ModelEditOperationState } from './model-edit-operation';
 const props = defineProps<{
 	node: WorkflowNode<ModelEditOperationState>;
 }>();
-const emit = defineEmits(['append-output', 'update-state', 'close', 'select-output']);
+const emit = defineEmits([
+	'append-output',
+	'update-state',
+	'close',
+	'select-output',
+	'update-output-port'
+]);
 
 enum ModelEditTabs {
 	Wizard = 'Wizard',
@@ -292,6 +298,13 @@ function updateCodeState(code: string = codeText.value, hasCodeRun: boolean = tr
 // Called when the selected output is changed, component unmounts or before the window is closed
 function updateOutputModel() {
 	if (readyToSaveOutputModel.value && amr.value) {
+		// Save notebook code to output state
+		if (activeOutput.value) {
+			const updatedOutputPort = cloneDeep(activeOutput.value);
+			updatedOutputPort.state = cloneDeep(props.node.state);
+			emit('update-output-port', updatedOutputPort);
+		}
+		// Save model in backend
 		updateModel(amr.value);
 		readyToSaveOutputModel.value = false;
 	}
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit.vue
```diff
@@ -292,6 +298,13 @@ function updateCodeState(code: string = codeText.value, hasCodeRun: boolean = tr
 // Called when the selected output is changed, component unmounts or before the window is closed
 function updateOutputModel() {
 	if (readyToSaveOutputModel.value && amr.value) {
+		// Save notebook code to output state
+		if (activeOutput.value) {
+			const updatedOutputPort = cloneDeep(activeOutput.value);
+			updatedOutputPort.state = cloneDeep(props.node.state);
+			emit('update-output-port', updatedOutputPort);
+		}
```
mwdchang, line 306: Can you elaborate on the problem you are trying to solve? Maybe the steps that create said issue?

It is rather unusual to update an existing port, so `update-output-port` is more like an escape-hatch and not typically recommended.

shawnyama: The `state` property within the output port saves the state of the notebook code. This is so when we switch outputs the nodes `state` is assigned to the one in the output port.

The problem is that this notebook code is saved in the output port only once the output port is created. Any edits to the code after that are not saved anywhere.

shawnyama: Recreate the issue:
1. Edit the original model to create a new output
2. Make another edit to your new output
3. Examine the code
4. Switch to another output
5. Switch back to your new output
6. Notice that half the code is missing

mwdchang: Thanks for the explanation, did you talk with @asylves1 and @ngraham76 , this sounds very similar to the notion of `draft` that had been floating around. 

I think there are two things we need to discuss:
1. Changing an existing output, other than cosmetic changes like name/label create a rippling effect downstream, we have not discussed how to deal with that, or if that is desired. e.g. if you change stuff, then technically speaking you want to invalid downstream nodes just like when we were "switching" between different outputs.

2. The intention is not clear. If you made some changes, and move away from said change, we don't know users want to preserve it, or want to discard it. Again I think this needs a larger discussion as it impact not just the model-edit node, but across all operators.

Also if I can paraphrase into a more concrete example, and let me know if I am incorrect here as it has been a bit of time since I last used model-edit operations.

1. I start with an SIR model
2. I made it into SIRD, then I save it as a new output ==> Output-A
3. Then I further make it into SIRDH, and I save it as a new output ==> Output-B
4. I make a few more modifications while I am on Output-B
5. I switch to Output-A
6. I switch back to Output-B, and I observe the changes I made in step-4 are lost

☝️ Does that accurately describe the problem?

shawnyama: Also if you're not up to date with model edit, look at this recent update I gave it here https://github.com/DARPA-ASKEM/terarium/pull/3580

shawnyama: Added before and after videos for others to see

mwdchang (approved):
Group chat with neil/shawn/adrian - merge for now to keep the behaviour consistent, but should revisit as updating output is not a very desirable behaviour, it makes thing complicated downstream.

PR #3634, opened 2024-05-16
Add notification for taskrunner tasks (GoLLM)

jryu01:
# Description

- Added all GoLLM task runners tasks to the notification handling
- Place notification panel to be in the global app level instead of in the project level. 
- Fixed 400 errors from multiple api requests due to missing projectId when app is hard refreshed in certain pages and `activeProjectId` is not loaded. 

https://github.com/DARPA-ASKEM/terarium/assets/3279838/6aa41687-54c6-483b-bfbe-d2bf24ca4082

Resolves #(issue)


jryu01:
Waiting for this PR https://github.com/DARPA-ASKEM/terarium/pull/3619 to be merged in first as I'm expecting there are merge conflicts to be resolved.

dvince2 (approved):
I think this looks good, but you may want to wait for a second set of 👀 for the front end code

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -276,8 +276,8 @@ export const updateWorkflow = async (workflow: Workflow) => {
 };
 
 // Get
-export const getWorkflow = async (id: string) => {
-	const response = await API.get(`/workflows/${id}`);
+export const getWorkflow = async (id: string, projectId?: string) => {
+	const response = await API.get(`/workflows/${id}`, { params: { 'project-id': projectId } });
```
mwdchang, line 280: Why do we need to pass in a projectId?

packages/client/hmi-client/src/services/notificationEventHandlers.ts
```diff
@@ -141,17 +139,48 @@ export const createNotificationEventHandlers = (notificationItems: Ref<Notificat
...
+	});
+	registerHandler<TaskResponse>(ClientEventType.TaskGollmConfigureModel, (event, created) => {
+		created.supportCancel = true;
+		created.sourceName = 'Configure model';
+		created.assetId = event.data.additionalProperties.workflowId as string;
+		created.pageType = AssetType.Workflow;
+		created.nodeId = event.data.additionalProperties.nodeId as string;
+		getWorkflow(created.assetId, created.projectId).then((workflow) =>
+			Object.assign(created, { context: workflow?.name || '' })
+		);
```
mwdchang, line 164: I don't understand why we are grabbing workflow here, or how `created.assetId` is a valid workflow id??

packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
```diff
@@ -115,6 +121,9 @@ const isRunning = (item: NotificationItem) => item.status === ProgressState.Runn
 const isCancelling = (item: NotificationItem) => item.status === ProgressState.Cancelling;
 const isCancelled = (item: NotificationItem) => item.status === ProgressState.Cancelled;
 
+const getProjectName = (item: NotificationItem) =>
+	(useProjects().allProjects.value || []).find((p) => p.id === item.projectId)?.name || '';
+
```
mwdchang, line 126: Maybe we should create a map in the project composable so we don't have to do linear look up for every event-group.

packages/client/hmi-client/src/services/workflow.ts
jryu01: We are already passing active projectId for all requests that's handled by axios interceptor but when the request is called where you are out side of the project, projectId isn't available. For notification panel in the  main root page, active projectId isn't available and we are providing it manually for this method.

packages/client/hmi-client/src/services/notificationEventHandlers.ts
jryu01: When a source link is clicked in the each notification item, we are navigating user to the right asset or workflow (and operator modal) page with the corresponding project. WorkflowId is used for that purpose. The ids are passed and carried along with the request when the request is sent by the user from the specific context.

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -39,13 +39,17 @@ export async function modelCard(documentId: string): Promise<void> {
...
+	nodeId?: string
 ): Promise<TaskHandler | null> {
 	try {
 		const response = await API.get<TaskResponse>('/gollm/configure-model', {
 			params: {
 				'model-id': modelId,
-				'document-id': documentId
+				'document-id': documentId,
+				'workflow-id': workflowId,
+				'node-id': nodeId
```
mwdchang, line 52: I wonder if it would be better to create a optional-object, like tracker/tagger, than explicitly spell out all the associated Ids, cause while this works, it feels very messy since they are optionals, and maybe a bit error prone as well.

So something like this perhaps??
```
params: {
  model-id: ...,
  document-id: ...,
  tagger: { key-value-pair }
}
```


@YohannParis @dvince2 @jryu01 thoughts?

jryu01: I wasn't sure if there was a simple way to pass an arbitrary object as a parameter value with this spring annotation magic. When I first tried it didn't allow me to do so.

packages/client/hmi-client/src/services/notificationEventHandlers.ts
mwdchang: So if all you need i workflow-name, why isn't that passed into the original request? It feels like we splitting something that should be contiguous into multiple steps across multiple components.

jryu01: Actually `workflowId` is needed to get the correct page url and the workflow name is for display purpose. We could pass the name into the original request but as I mentioned in other comments, at first It seemed it wasn't very clear to have an arbitrary object as a parameter value and I didn't want to add more parameters to the api endpoints. I think that's something we can improve as we go though.

packages/client/hmi-client/src/services/workflow.ts
mwdchang: I see, so this is a hack to get the ID int the URL in order to trigger the interceptor, which at the notification component you may or may not have the active-project? 

Then why not just wait for the active project to be available before showing the notification component? 

At least we need some documentations/comments if doing a hack, because it is not clear why project_id is passed in, as it does not match the workflow-controller's API signatures.

mwdchang:
Discussed with Jaehwan/Yohann:

The major contentions here revolve around creating a summary-UI for a given event, it needs attributes like project-name, workflow-name for displaying purposes, and other identifier attributes to (recreate) a URL to link the event to. The PR here injects project-ids in some places in order to fetch the descriptive data, even though project-id is not a strictly required input.

Food for thoughts:
- Rather than fetching these additional attributes when the event arrives on the UI, the event itself can be populated with necessary information at either a) when the original request was made, or b) when the server is about to broadcast the event message.  This should clean up the project-id injection hacks, and make the UI a little leaner to not fetch larger objects like workflow.
- There is in inherent contention with now permission and notification are modelled. Permissions seem to be project-based, while event notifications are user-based
- We may want to have more thoughts on some kind of tracking context object that is a little easier to think about than a listing of optional parameters.

packages/client/hmi-client/src/services/workflow.ts
jryu01: Added comments to those methods.

packages/client/hmi-client/src/services/goLLM.ts
jryu01: As we discussed I'll think about it and implement similar to what you've mentioned in the follow up work.

packages/client/hmi-client/src/components/widgets/tera-asset-link.vue
YohannParis, line 1: What is the purpose here?

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-asset-link.vue
jryu01: I added route query props to the asset link component so that it can also take url query string value. 
It's used to navigate to a specific operator in a workflow page. For navigating to operator drilldown modal, you need `?operator={nodeId}` as a query string value on top of the workflowID as url param.

PR #3635, opened 2024-05-16
do not drag operators when clicking on the input or output ports

blanchco:
# Description

* A small fix to address the bug where dragging an operator from a label will cause wonky arrows to appear.  Now you cannot drag when clicking a label (see attached bug ticket from additional details)


mwdchang (approved):
looks fine, thx!

PR #3658, opened 2024-05-17
Add cancellation when a task timeouts on hmi-server

kbirk:


YohannParis (approved):
I never seen a try/catch within a catch before.
Learning everyday!

PR #3663, opened 2024-05-17
feat: model page update

shawnyama:
# Description

For the model page I updated the description section and added navigation links to auto scroll to certain sections on the page.

Back when we had the old `tera-asset-nav` I utilized css grid to insert it. This time I am using a more minimal approach. The old css grid stuff from before was still in `tera-asset` so I got rid of that. I also had to make some small changes to `tera-code` so it can properly fit in `tera-asset` again.

I tried to add a `scroll-padding-top` or a `scroll-margin-padding` so that when I `scrollIntoView()` and the asset header is sticky it will accommodate for that. For some reason adding the rule doesn't seem to have any effect. I wonder if we should simplify the sticky header. Maybe we could just have the header statically up there the whole time instead of having one that overlaps the content.

https://github.com/DARPA-ASKEM/terarium/assets/28237258/9e1a4a62-b41f-4ab1-83e7-dfd7bc1f8668


packages/client/hmi-client/src/components/widgets/tera-asset-nav.vue
```diff
@@ -0,0 +1,42 @@
...
+	assetContentIds: string[];
+}>();
+
+const assetContentIdsInView = ref<string[]>([]);
+
+async function scrollTo(elementId: string) {
+	const element = document.getElementById(elementId);
+	if (!element) return;
+	element.scrollIntoView({ behavior: 'smooth' });
+}
```
mwdchang, line 22: `document.getElementById` is slightly dangerous because you don't know how other folks will name things....otherwise we may want some prefix or suffix to decrease the chances of id/name collisions.

Would it be possible to pass in the section-container to tera-asset-nav, and do getElementId on that instead of the entire document?

```
const props = defineProps<{
	assetContentIds: string[];
  containerElement?: HTMLElement
}>();

... blahblahblah ...

const element = props.containerElement ? containerElement.getElementById(elemetnId) : document.getElementById(elementId);
```

mwdchang (approved):
thanks

blanchco (approved):

PR #3664, opened 2024-05-17
initial tables v2

blanchco:
# Description

* An update to the model config initial tables to the new design
* `tera-initial-tables-v2` is just a temporary name until we fully migrate over to the new design since the model page is still using the original `tera-initial-tables` component
* includes unstratified and stratified functionality
* TBD other values/other matices buttons
* TBD matrix metadata (name, description, unit, source, etc...)

https://github.com/DARPA-ASKEM/terarium/assets/33158416/2c1c1e90-030d-4d19-8da2-6934dd0f1213

https://github.com/DARPA-ASKEM/terarium/assets/33158416/55e0534e-2c97-433c-8f77-94b7fae04629


packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -113,30 +113,31 @@ export function sanityCheck(config: ModelConfiguration): string[] {
...
+	return config.configuration.metadata?.initials?.[initialId]?.name ?? '';
+}
+
+export function getInitialUnit(config: ModelConfiguration, initialId: string): string {
+	return config.configuration.metadata?.initials?.[initialId]?.unit ?? '';
+}
+
+export function getInitialDescription(config: ModelConfiguration, initialId: string): string {
+	return config.configuration.metadata?.initials?.[initialId]?.description ?? '';
+}
```
mwdchang, line 136: Note name, unit, and description do not exist on initials in the AMR structure (neither does id), so all these fields will be blanks when you create the initial configuration out of an AMR. 

The initial data structure looks like this:
```
       {
         "target": "Susceptible",
         "expression": "0.999996300000000",
         "expression_mathml": "<cn>0.99999629999999995</cn>"
       }
```

Is this what we are expecting the user to do? Or were there some misconception in the design?

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+	<!-- Stratified -->
+	<template v-if="isStratified">
+		<Accordion multiple>
+			<AccordionTab
+				v-for="[key, values] in getUnstratifiedInitials(
```
mwdchang, line 6: `getUnstratifiedInitials` is not exactly compatible with `isStratifiedModel`. The former is acting on the AMR and latter the MMT version of the AMR.

Slight preference would be to move towards the latter option, as it gives a little more freedom if we want to maintain regnet and stockflow compatibilities, since the MMT version handles all 3 but getUnstratifiedInitials only for petrinet. What you may want to look at here is collapseInitials from the mira folder


https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/model-representation/mira/mira.ts#L151C14-L167

packages/client/hmi-client/src/services/model-configurations.ts
blanchco: All of these would be available to edit in the model page according to the [figma](https://www.figma.com/design/UXZ50KQ81tdD3aj1XqVCqz/Terarium---March-%2F-April-%2F-May-2024?node-id=800-58667&t=IkwkkMdJk7yz9nH4-0) designs.

mwdchang: That is not my point. 

Is the design assuming that they will be at least partially populated? Because they will all be blank - e.g. are the designs matching reality and technical limitations? Cause it is weird to have a page of all blank fields.

blanchco: Ill bring this up in the design channel and see what they say

blanchco: here's the [slack thread](https://unchartedsoftware.slack.com/archives/C064P3QGVJ6/p1716389204519159)

blanchco: To summarize, there will be a message suggesting that metadata is missing.  Ill do this in separate [ticket](https://github.com/DARPA-ASKEM/terarium/issues/3678) though

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
```diff
@@ -0,0 +1,116 @@
...
+const props = defineProps<{
+	modelConfiguration: ModelConfiguration;
+	mmt: MiraModel;
+	mmtParams: MiraTemplateParams;
+}>();
+
+const emit = defineEmits(['update-expression', 'update-source']);
+
+const isStratified = computed(() => {
+	if (!props.mmt) return false;
```
shawnyama, line 82: I don't think there is a need to check for the mmt since it's a required prop

blanchco: true, removed this

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
```diff
@@ -0,0 +1,116 @@
...
+
+const matrixModalContext = ref({
+	isOpen: false,
+	matrixId: ''
+});
+
+function openMatrix(matrixId: string) {
+	matrixModalContext.value.isOpen = true;
+	matrixModalContext.value.matrixId = matrixId;
+}
```
shawnyama, line 91: Maybe just make the modal appear if there is a `matrixId` at all. Is there a need to have a selected `matrixId` when the modal is closed?

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -0,0 +1,91 @@
...
+		<span v-if="name" class="name">{{ '| ' + name }}</span>
+		<span v-if="unit" class="unit">({{ unit }})</span>
+		<span v-if="description" class="description">{{ description }}</span>
+	</header>
+	<main>
+		<span class="expression">
+			<tera-input
+				label="Expression"
+				:model-value="getInitialExpression(modelConfiguration, initialId)"
+				@update:model-value="emit('update-expression', { id: initialId, value: $event })"
```
YohannParis, line 13: Why are you sending an event?
It's all reactive, so if you update the modelConfiguration here, the rest will update directly?

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
```diff
@@ -0,0 +1,113 @@
...
+							<Divider type="solid" />
+						</li>
+					</ul>
+				</div>
+			</AccordionTab>
+		</Accordion>
+	</template>
+
+	<!-- Unstratified -->
+	<template v-else>
```
shawnyama, line 29: You can put the `v-if` and `v-else` in the html tags themselves since they contain the content, no need for the `<template>`

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
```diff
@@ -0,0 +1,113 @@
...
+
+const props = defineProps<{
+	modelConfiguration: ModelConfiguration;
+	mmt: MiraModel;
+	mmtParams: MiraTemplateParams;
+}>();
+
+const emit = defineEmits(['update-expression', 'update-source']);
+
+const isStratified = computed(() => isStratifiedModel(props.mmt));
```
shawnyama, line 81: See if this works without using `computed`

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -0,0 +1,91 @@
...
+
+.name {
+	padding-left: var(--gap-xsmall);
+}
+.unit {
+	padding-left: var(--gap-small);
+}
+.description {
+	padding-left: var(--gap-large);
+}
```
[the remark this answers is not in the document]
shawnyama: Also I think these should be margins

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+	<header>
+		<b>{{ initialId }}</b>
```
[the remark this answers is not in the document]
blanchco: the strong tag doesn't seem to apply any font weight at the moment, I add a global class to change this though

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
blanchco: since modelConfiguration is a prop would it not be bad practice to manipulate it with a child component?  I send the event up to the parent where modelConfiguration prop is defined to manipulate it there

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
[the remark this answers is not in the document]
blanchco: I thought you had a vendetta against these 😆 .  I'll start using these

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -150,6 +151,37 @@ export function setInitialSource(
...
+	initialId: string,
+	expression: string
+): Promise<void> {
+	const initial = getInitial(config, initialId);
+	if (initial) {
+		initial.expression = expression;
+		const mathml = (await pythonInstance.parseExpression(expression)).mathml;
+		initial.expression_mathml = mathml;
+	}
+}
```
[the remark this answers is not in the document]
blanchco: parseExpression is still a `Promise`.  I'll write the function like this:
```ts
export function setInitialExpression(
	config: ModelConfiguration,
	initialId: string,
	expression: string
): void {
	const initial = getInitial(config, initialId);
	if (!initial) return;

	pythonInstance.parseExpression(expression)
			.then((result) => {
					const mathml = result.mathml;
					initial.expression = expression;
					initial.expression_mathml = mathml;
			})
			.catch((error) => {
					// Handle error appropriately
					console.error("Error parsing expression:", error);
			});
}

File: packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue (hunk 7 above)
Comment (blanchco) [reply packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue]: yup, had to add a check the mmt.initials is not empty

File: packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue [hunk 11]
```diff
@@ -0,0 +1,91 @@
...
+	initialId: string;
+}>();
+
+const emit = defineEmits(['update-expression', 'update-source']);
+
+const name = computed(() => getInitialName(props.modelConfiguration, props.initialId));
+const unit = computed(() => getInitialUnit(props.modelConfiguration, props.initialId));
+const description = computed(() =>
+	getInitialDescription(props.modelConfiguration, props.initialId)
+);
```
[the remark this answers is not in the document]
blanchco: changed to const

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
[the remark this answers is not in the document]
blanchco: done

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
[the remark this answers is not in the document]
blanchco: i decided to use `ml-1` `ml-2` `ml-4 `respectively

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table-v2.vue
blanchco: done

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
YohannParis: ... you are right, but it's bloody annoying.

packages/client/hmi-client/src/services/model-configurations.ts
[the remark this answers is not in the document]
YohannParis: Sure either way works.
I just want to make sure all the function in this utility are not surprisingly async when we do not return anything.
They are no advantages to wait here. and it makes the API more consistant.

packages/client/hmi-client/src/assets/css/reset.css
```diff
@@ -50,3 +50,7 @@ table {
 thead {
 	border: none;
 }
+
+strong {
+	font-weight: 600;
```
YohannParis, line 55:
```suggestion
	font-weight: var(--font-weight-semibold);
```

YohannParis (approved):

PR #3675, opened 2024-05-22
Fix project asset sharing to clone assets

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -181,7 +181,8 @@ public ResponseEntity<List<Project>> getProjects(
 			final Schema.Permission permission = projectService.checkPermissionCanWrite(
 					currentUserService.get().getId(), project.getId());
 
-			// Set the user permission for the project. If we are unable to get the user permission, we remove the
+			// Set the user permission for the project. If we are unable to get the user
+			// permission, we remove the
 			// project.
```
bigglesandginger, line 186: Why is this shorter , but if it must be should we add "project" to the line above

.vscode/extensions.json
dvince2, line 1: was this meant to be checked in?

kbirk: Kind of. We really don't need the gradle plugin. No point to include it since it just creates that stupid `build` dir that breaks vscode.

bigglesandginger (approved):

PR #3677, opened 2024-05-22
Ts/3630 interventions for calibrate

Tom-Szendrey:
# Description
**PLEASE ALSO REVIEW:** https://github.com/DARPA-ASKEM/pyciemss-service/pull/88

Here we are adding the interventions from the model configuration to our calibrate and optimize requests within the HMIServer. 

This will mean when a user connects a model config with interventions to a calibrate node and hits run the HMISERVER will add the model config's interventions to the request without the front end needing to worry about it

# Testing:

1. User should be able to add interventions in a model config
2. Run a simulation - and still see interventions impacting parameters
3. Run a calibration - and now see interventions impacting parameters
4. Run an optimization - and now see interventions impacting parameters
5. Saving a model config from optimize will carry over previous interventions as well as include any new ones from the optimize itself.

image


Tom-Szendrey:
@bigglesandginger can you please ensure that I have used this `project-id` change correctly in the few java files updated

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -205,15 +205,67 @@ public ResponseEntity<JobResponse> makeCalibrateJob(@RequestBody final Calibrati
...
 	@Secured(Roles.USER)
-	public ResponseEntity<JobResponse> makeCalibrateJobCiemss(@RequestBody final CalibrationRequestCiemss request) {
+	public ResponseEntity<JobResponse> makeCalibrateJobCiemss(
+			@RequestBody final CalibrationRequestCiemss request, @RequestParam("project-id") final UUID projectId) {
+		Schema.Permission permission =
+				projectService.checkPermissionCanWrite(currentUserService.get().getId(), projectId);
+
+		// Get model config's interventions and append them to requests:
+		try {
+			final Optional<ModelConfiguration> modelConfiguration =
```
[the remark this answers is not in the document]
dvince2: (this is also applicable to the Optimize changes below)

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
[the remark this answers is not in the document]
Tom-Szendrey: The request will have a model config ID.
I am trying to get the corresponding model config which can technically fail
If I cannot find the model config ID I guess its probably better to display that on the client? I wasnt able to test this very well as when running this through the HMI im not sure when i wouldnt be able to find a model config. Maybe with the new permissions? 

I am open to any suggestions

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/CalibrationRequestCiemss.java
```diff
@@ -15,13 +18,16 @@
 // Used to kick off a calibration job in simulation-service
 public class CalibrationRequestCiemss implements Serializable {
 	@JsonAlias("model_config_id")
-	private String modelConfigId;
+	private UUID modelConfigId;
 
 	private Object extra;
```
mwdchang, line 23: We had serialization problems with Object before, should we switch to JsonNode while we are in the area? @dvince2 ?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -205,15 +205,67 @@ public ResponseEntity<JobResponse> makeCalibrateJob(@RequestBody final Calibrati
...
+				List<Intervention> allInterventions = request.getFixedStaticParameterInterventions();
+				if (allInterventions == null) {
+					allInterventions = new ArrayList<Intervention>();
+				}
+				allInterventions.addAll(modelInterventions);
+				request.setFixedStaticParameterInterventions(allInterventions);
+			}
+		} catch (IOException e) {
+			String error = "Unable to find model configuration";
+			log.error(error, e);
```
mwdchang, line 266: Strictly speaking, if you cannot find a model-configuration, you have already returned and wouldn't be here:

In above:
```
			if (modelConfiguration.isEmpty()) {
				return ResponseEntity.notFound().build();
			}
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -205,15 +205,67 @@ public ResponseEntity<JobResponse> makeCalibrateJob(@RequestBody final Calibrati
...
+				List<Intervention> allInterventions = request.getInterventions();
+				if (allInterventions == null) {
+					allInterventions = new ArrayList<Intervention>();
+				}
+				allInterventions.addAll(modelInterventions);
+				request.setInterventions(allInterventions);
+			}
+		} catch (IOException e) {
+			String error = "Unable to find model configuration";
+			log.error(error, e);
```
mwdchang, line 232: See https://github.com/DARPA-ASKEM/terarium/pull/3677/files#r1616374465

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -686,7 +687,8 @@ export interface EnsembleSimulationCiemssRequest {
 export interface OptimizeRequestCiemss {
     modelConfigId: string;
     timespan: TimeSpan;
-    interventions?: OptimizedIntervention;
+    policyInterventions?: OptimizedInterventions;
+    fixedStaticParameterInterventions?: Intervention[];
```
mwdchang, line 691: Food for thought for future improvement, It seems like we should consider consolidate the two intervention objects to make things a bit clearer and easy to work with? 

An interventions array []:
```
export interface Intervention {
    name: string;
    timestep: number;
    value: number;
}
```

And an intervention object with parallel arrays:
```
export interface OptimizedInterventions {
    selection: string;
    paramNames: string[];
    paramValues?: number[];
    startTime?: number[];
}
```

Tom-Szendrey: totally see your point. I will need to look into what changes will need to be made in the pyciemss-service for this.
This should probably be a future improvement as you suggest as it may clutter these PRs
Ive made the following to not forget:
https://github.com/DARPA-ASKEM/terarium/issues/3711

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
Tom-Szendrey: Thanks!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/CalibrationRequestCiemss.java
dvince2: The issue there was for serialization into postgres, which I don't believe these are.  However if `extra` is just a json blob we should probably just change it to JsonNode since thats what it is, and it will potentially save us a headace in the future

mwdchang (approved):

bigglesandginger (approved):
project-id stuff looks good to me

PR #3680, opened 2024-05-22
Unnecessary and Incorrect Permission check

bigglesandginger:
The initial fetch of projectIds ensures the user has READ permission for all projects, there is no need to check again (let alone incorrectly check that they have WRITE permissions - mistake).  Yes there is a potential race condition that a project between initial ids and population could change permission, but the same can be said whilst they sit on the projects page, it is a moot issue.


dvince2 (approved):

Tom-Szendrey (approved):
ty

PR #3682, opened 2024-05-22
Add api service exception for permissions

bigglesandginger:
Allow for api-service calls to bypass permissions (assume they were spawned by a valid user)


Tom-Szendrey (approved):
I have only tested this with a simulation with pyciemss but this has helped resolve that issue thank you!!

packages/server/src/main/resources/application.properties
```diff
@@ -63,6 +63,8 @@ spring.security.oauth2.client.provider.keycloak.issuer-uri=${terarium.keycloak.u
 spring.security.oauth2.client.registration.keycloak.client-id=${terarium.keycloak.client-id:app}
 spring.security.oauth2.resourceserver.jwt.issuer-uri=${terarium.keycloak.url:localhost}/realms/${terarium.keycloak.realm:Terarium}
 
+terarium.keycloak.api-service-name=api-service
```
dvince2, line 66: Is there any risk having this username here in plain text?

dvince2 (approved):

packages/server/src/main/resources/application.properties
bigglesandginger: I don't think so.  The password is in the encrypted blobs in orchestration, and this is for developer running - again orchestration can override this with whatever we deem should be the user name,  though I for one don't think of the user name being a secret, just the password... thoughts?

dvince2: ah! no that makes sense. just wanted to double check.

PR #3684, opened 2024-05-22
style: simplify loading spinner in model config

shawnyama:
# Description

I simplified the loading spinner for extracting configurations. The watering can was struggling to fit in there as the PrimeVue container for the loading area is limited. I thought there is no need to fight with PrimeVue and that a simple spinner that doesn't take up much space is more suitable anyways.

BEFORE
image

AFTER
Screenshot 2024-05-22 at 5 29 50 PM


mwdchang (approved):
sure!

YohannParis (approved):

PR #3687, opened 2024-05-23
Allow getting public non temporary assets (models)

bigglesandginger:
Idea for models


Tom-Szendrey:
With these changes I am now able to see the model preview in the explorer (a situation I am not providing a project ID as it does not exist)
If a model if private it will still show on the explorer but I will not be able to open the drilldown.
This may be dealt with in future PRs?

image

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
```diff
@@ -220,11 +222,24 @@ public ResponseEntity<Dataset> createDataset(
...
 			final Optional<Dataset> dataset = datasetService.getAsset(id, permission);
+
+			if (dataset.isEmpty()) {
+				return ResponseEntity.noContent().build();
+			}
+			// GETs not associated to a projectId cannot read non public or temporary assets
+			if (permission.equals(Schema.Permission.NONE)
+					&& !dataset.get().getPublicAsset()
+					&& !dataset.get().getTemporary()) {
+				throw new ResponseStatusException(HttpStatus.FORBIDDEN, messages.get("rebac.unauthorized-update"));
```
dvince2, line 240: I think this error message is meant to be `rebac.unauthorized-read`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -177,6 +182,12 @@ ResponseEntity<Model> getModel(
 			if (model.isEmpty()) {
 				return ResponseEntity.noContent().build();
 			}
+			// GETs not associated to a projectId cannot read non public or temporary assets
+			if (permission.equals(Schema.Permission.NONE)
+					&& !model.get().getPublicAsset()
+					&& !model.get().getTemporary()) {
+				throw new ResponseStatusException(HttpStatus.FORBIDDEN, messages.get("rebac.unauthorized-update"));
```
dvince2, line 189: Same here: `rebac.unauthorized-read`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
bigglesandginger: oh - yes, though it would appear that a 500 is returned and not a 403 but that is details

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
dvince2: err sorry, i meant the call to `messages.get` here will return

> You are not authorized to update this resource. Please contact the project owner to upgrade permissions.

YohannParis (approved):
LGTM, but I agree with @Tom-Szendrey we should update the `getAll` to only return public assets.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -177,6 +182,12 @@ ResponseEntity<Model> getModel(
 			if (model.isEmpty()) {
 				return ResponseEntity.noContent().build();
 			}
+			// GETs not associated to a projectId cannot read non public or temporary assets
+			if (permission.equals(Schema.Permission.NONE)
+					&& !model.get().getPublicAsset()
+					&& !model.get().getTemporary()) {
```
Tom-Szendrey, line 188: Does this imply that any temp asset is essentially public?

Tom-Szendrey (commented):
I suspect we also want these changes for Document.
That is another asset type that can be viewed in the explorer and not in association with a project

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
YohannParis: Nope, temporary and public are two different things.
A public asset is one that do not belong to a project and is not editable.
A temporary assets are the ones created by workflow operators and not saved by the user within the project.

The condition above reads as:
> if their is no permission on the private and permament asset.

bigglesandginger: No - I think I am implying that people without permissions can only see assets that are both Public and not temporary - do I have the logic incorrect here?  (reading again)

bigglesandginger:
Creating another task for getAll considedrations

PR #3690, opened 2024-05-23
feat: auto retrieve scroll elements for table of contents

shawnyama:
# Description

No need to specify id in PrimeVue header prop as well as a list of their names for the `tera-asset` component. I grab the text within the accordion header and inject ids into them so they can be scrolled to.

image


YohannParis (approved):
LGTM

blanchco (approved):

PR #3699, opened 2024-05-24
parameter tables v2

blanchco:
# Description

* Similar to initials table v2, this is the new parameter table v2.
* Includes stratified and unstratified functionality
* added _back_ the constant/distribution dropdown
* additional model config getters/setters

https://github.com/DARPA-ASKEM/terarium/assets/33158416/28a3bff2-e53f-4d4c-b99f-d89274b9278a

### DIFF
--- a/packages/client/graph-scaffolder/package.json
+++ b/packages/client/graph-scaffolder/package.json
@@ -46,7 +46,7 @@
     "@typescript-eslint/parser": "5.62.0",
     "copy": "0.3.2",
     "dagre": "0.8.5",
-    "esbuild": "0.21.1",
+    "esbuild": "0.21.4",
     "esbuild-jest": "0.5.0",
     "eslint": "8.57.0",
     "eslint-config-esnext": "4.1.0",
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
@@ -222,9 +222,8 @@ public class DatasetController {
 	public ResponseEntity<Dataset> getDataset(
 			@PathVariable("id") final UUID id,
 			@RequestParam(name = "project-id", required = false) final UUID projectId) {
-		final Schema.Permission permission = projectId == null
-				? Schema.Permission.NONE
-				: projectService.checkPermissionCanRead(currentUserService.get().getId(), projectId);
+		final Schema.Permission permission = projectService.checkPermissionCanReadOrNone(
+				currentUserService.get().getId(), projectId);
 
 		try {
 			final Optional<Dataset> dataset = datasetService.getAsset(id, permission);
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DocumentController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DocumentController.java
@@ -248,9 +248,8 @@ public class DocumentController {
 	public ResponseEntity<DocumentAsset> getDocument(
 			@PathVariable("id") final UUID id,
 			@RequestParam(name = "project-id", required = false) final UUID projectId) {
-		final Schema.Permission permission = projectId == null
-				? Schema.Permission.NONE
-				: projectService.checkPermissionCanRead(currentUserService.get().getId(), projectId);
+		final Schema.Permission permission = projectService.checkPermissionCanReadOrNone(
+				currentUserService.get().getId(), projectId);
 
 		final Optional<DocumentAsset> document = documentAssetService.getAsset(id, permission);
 		if (document.isEmpty()) {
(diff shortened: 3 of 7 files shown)


packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -249,6 +265,63 @@ export function setParameterSource(
...
+		case DistributionType.Uniform:
+			parameter.distribution = {
+				type: DistributionType.Uniform,
+				parameters: {
+					minimum: 0,
+					maximum: 0
+				}
+			};
+			break;
+		default:
```
blanchco, line 305: Im doing some additional tests here on how this value/distribution should be populated when changing the dropdown:
https://github.com/DARPA-ASKEM/terarium/issues/3700

packages/client/hmi-client/src/types/common.ts
```diff
@@ -236,3 +236,19 @@ export const programmingLanguageOptions = (): { name: string; value: string }[]
...
+export const DistributionTypeLabel: { [key in DistributionType]: string } = {
+	[DistributionType.Constant]: 'Constant',
+	[DistributionType.Uniform]: 'Uniform'
+};
+
+export const distributionTypeOptions = (): { name: string; value: string }[] =>
+	Object.values(DistributionType).map((dist) => ({
+		name: DistributionTypeLabel[dist],
+		value: dist
+	}));
```
mwdchang, line 254: May want to consider having these distribution stuff in a separate file, as none of these are strictly speaking, type declarations anyway.

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -249,6 +265,63 @@ export function setParameterSource(
...
+	parameterId: string,
+	parameters: { [index: string]: any }
+): void {
+	const parameter = getParameter(config, parameterId);
+	if (parameter && parameter.distribution) {
+		parameter.distribution.parameters = {
+			...parameter.distribution.parameters,
+			...parameters
+		};
+	}
```
mwdchang, line 322: Not sure what the intention is here, but this doesn't look like it works, as you are doing a disjunctive join. For example:

```
p = { mean: 0, std: 0.2 }
p2 = { max: 10, min: 3 }
p = { ...p, ...p2 }
```

We get for p = `{mean: 0, std: 0.2, max: 10, min: 3}` which does not properly describe anything...

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -0,0 +1,132 @@
...
+	'update-distribution',
+	'update-source',
+	'select-distribution'
+]);
+
+const name = getParameterName(props.modelConfiguration, props.parameterId);
+const unit = getParameterUnit(props.modelConfiguration, props.parameterId);
+const description = getParameterDescription(props.modelConfiguration, props.parameterId);
+
+const sourceOpen = ref(false);
```
shawnyama, line 106:
```suggestion
const isSourceOpen = ref(false);
```

packages/client/hmi-client/src/services/model-configurations.ts
blanchco: It works in its current state 😄, but its definitely hacky. Ill rethink this implementation

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -249,6 +265,63 @@ export function setParameterSource(
...
+}
+
+// sets the parameter distribution parameters (i.e. minimum and maximum for uniform distribution)
+export function setParameterDistributionParameters(
+	config: ModelConfiguration,
+	parameterId: string,
+	parameters: { [index: string]: any }
+): void {
+	const parameter = getParameter(config, parameterId);
+	if (parameter && parameter.distribution) {
```
shawnyama, line 317:
```suggestion
if (parameter?.distribution) {
```

shawnyama (approved):

packages/client/hmi-client/src/types/common.ts
blanchco: put it in a file called distribution under `services/distribution.ts`

packages/client/hmi-client/src/services/model-configurations.ts
blanchco: i've might a slight adjustment, the function will now update the keys that currently exist in the distribution object from the given parameters array. i.e

```
p = { min: 0, max: 2}
p2 = { min: 1, std: 8 }
```

Now p will = `{ min: 1, max: 2 }`

This function will be strictly for updating the distribution parameters that are already given in the initialized distribution and not add additional parameters that are not needed.

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -197,22 +198,40 @@ export function getParameter(
...
+			parameters: {
+				value: parameter?.value
+			}
+		};
 	}
+
+	if (parameter.distribution.type === 'Uniform1')
+		parameter.distribution.type = DistributionType.Uniform;
+
+	return parameter.distribution;
```
mwdchang, line 234: I think we need a few minor tweaks here.

In one instance, you are returning a new copy (constant-distribution). 
In another you are returning a modified reference of the config object.

So there are two problems here:
- It is inconsistent
- Your getter changes the underlying object

I am guess you probably want to do something like this?
```
const r = clone(parameter.distribution);
if (r.distribution.type === ....) r.distribution.type = ....
return r
```

blanchco: 👍 it does not return a reference now

mwdchang (approved):

PR #3701, opened 2024-05-27
Simulation forecast does not have project id set

bigglesandginger:
# Description

Resolves #(issue)


blanchco (approved):

dvince2 (approved):
Does the projectId need to be in SimulationRequest? should it also be manually injected here?

bigglesandginger:
> Does the projectId need to be in SimulationRequest? should it also be manually injected here?

A question for someone else, but a reasonable question

PR #3710, opened 2024-05-27
Asset cloning and project export / import

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -523,6 +526,79 @@ public ResponseEntity<Project> updateProject(
...
+						responseCode = "200",
+						description = "The project export",
+						content = {
+							@Content(
+									mediaType = "application/json",
+									schema =
+											@io.swagger.v3.oas.annotations.media.Schema(
+													implementation = ProjectExport.class))
+						}),
+				@ApiResponse(
```
bigglesandginger, line 579: Shouldn't be possible to get a 403 on this function (I think) as all Users have the right to create a project

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -523,6 +526,79 @@ public ResponseEntity<Project> updateProject(
...
+						responseCode = "403",
+						description = "The current user does not have read privileges to this project",
+						content = @Content),
+				@ApiResponse(responseCode = "404", description = "Project could not be found", content = @Content),
```
bigglesandginger, line 583: Also no 404

kbirk: Fixed

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ITerariumAssetService.java
YohannParis, line 1: those `hasWritePermission` and `hasReadPermission` are never used, why?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -523,6 +526,74 @@ public ResponseEntity<Project> updateProject(
...
+						description = "An error occurred when trying to communicate with either the postgres or spicedb"
+								+ " databases",
+						content = @Content)
+			})
+	@GetMapping("/export/{id}")
+	@Secured(Roles.USER)
+	public ResponseEntity<ProjectExport> exportProject(@PathVariable("id") final UUID id) {
+		projectService.checkPermissionCanRead(currentUserService.get().getId(), id);
+		try {
+			return ResponseEntity.ok(cloneService.exportProject(id));
```
YohannParis, line 558: I do not understand where the resulting file from ProjectExport.clone is, or what is is?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ITerariumAssetService.java
bigglesandginger: By design - they are there to remind developers to check that the user has Permission.  Making the the functions require the permission forces the developer to acknowledge that they have either done the work required or are choosing to ignore it, unlike any other way I can think of which involves developers having to remember to read a document attached to the underside of a bottle of milk in the fridge.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
bigglesandginger: I don't think Kevin has written the UI to use the export feature, so in theory whatever call makes this request will receive a massive JSON blob, what they chose to do with it is up to the receiver - that is my reading of the code

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
kbirk: Yeah that's basically it, you will get a giant JSON blob containing everything needed to import. We can play around with the exact format, maybe use a .zip file instead rather than a .json file.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TerariumAssetCloneService.java
```diff
@@ -0,0 +1,225 @@
...
+			// create the asset
+			final TerariumAsset asset =
+					(TerariumAsset) terariumAssetService.createAsset(assetExport.getAsset(), Schema.Permission.WRITE);
+
+			// upload the files
+			for (final Map.Entry<String, FileExport> entry :
+					assetExport.getFiles().entrySet()) {
+				final String fileName = entry.getKey();
+				final FileExport fileExport = entry.getValue();
+				terariumAssetService.uploadFile(
```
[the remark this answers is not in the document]
kbirk: Good idea. For now l say lets keep it single threaded and super simple just so its easier to debug any issues that arise in the shortterm.

PR #3716, opened 2024-05-28
Show Abstract text in Document view

asylves1:
# Description
The Abstract text from pdf documents was not being extracted from the cosmos zip file.

image

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -201,7 +202,26 @@ public Future<DocumentAsset> extractPDF(
...
+						final byte[] bytes = zipEntryToBytes(zipInputStream);
+
+						fileMap.put(filenameNoExt, bytes);
+						if (entry != null && entry.getName().toLowerCase().endsWith(".json")) {
+							ObjectMapper objectMapper = new ObjectMapper();
+
+							JsonNode rootNode = objectMapper.readTree(bytes);
+							if (rootNode.isArray()) {
+								ArrayNode arrayNode = (ArrayNode) rootNode;
+								String target = "Abstract";
```
dvince2, line 214: As a java formatting convention, these sort of definitions are usually declared as static constants at the class level rather than inline like this.  Something like:

```
private static final String TARGET = "Abstract";
private static final String PARAM = "detect_cls";
```

packages/client/hmi-client/src/components/documents/tera-document-asset.vue
```diff
@@ -308,7 +308,7 @@ watch(
 	{ immediate: true }
 );
 
-const formattedAbstract = computed<string>(() => document.value?.description ?? '');
```
dvince2, line 311: We still have the `description` value on document. What was being put here before?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
```diff
@@ -60,6 +60,10 @@ public class DocumentAsset extends TerariumAsset {
 	@JoinColumn(name = "grounding_id")
 	private Grounding grounding;
 
+	@TSOptional
+	@Column(columnDefinition = "text")
+	private String documentAbstract;
```
YohannParis, line 65:
```suggestion
	private String abstract;
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -70,6 +71,11 @@ public class ExtractionService {
 	private final ProvenanceService provenanceService;
 	private final CurrentUserService currentUserService;
 
+	// Used to get the Abstract text from PDF
+	private static final String NODE_TARGET = "Abstract";
+	private static final String NODE_PARAM = "detect_cls";
+	private static final String NODE_CONTENT = "content";
```
YohannParis, line 77: I don't understand the choice of name here `NODE_XXX`?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
dvince2: This suggestion won't work - `abstract` is a reserved keyword in Java

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
asylves1: As we discussed, I will remove the target & param and leave the content definition

YohannParis (approved):

Issue #3719, opened 2024-05-28
Add `workflowId` to `NotebookSession` and create proper migration from `name`


dvince2:
`NotebookSession`s contain a reference to a workflow via a UUID id field. HOWEVER for reasons i can't fully comprehend this ID was stashed into the String field of `name` on the parent `TerariumAsset` rather than being its own field.
This task is to...

- [ ] add a `workflowId` to `NotebookSession`
- [ ] Update the front end to use `workflowId` instead of `name`
- [ ] Create a migration to move the workflow ID's from `name` into `workflowId`. This will likely look like:

```
UPDATE notebook_session SET workflow_id=name
UPDATE notebook_session SET name=NULL
```

### IMPORTANT
This issue is currently blocked by #3717 .  This issue MUST be completed AND must be deployed to both staging and production before this task can be merged safely.

PR #3722, opened 2024-05-28
3717 task migrate notebooksession from es to pg

### LINKED ISSUE #3717: [TASK]: Migrate `NotebookSession` from ES to PG

dvince2:
# Description

* Migrating NotebookSessions from ES to PG
* Severing NotebookSession from project assets
* For reasons beyond my comprehension the front end is using the `TerariumAsset`'s `name` which is a String to store the owning Workflow's UUID. This issue needs to be tackled *after* the initial migration is done on staging and prod. A task has been opened here: #3719 

Resolves #3717


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -88,7 +87,7 @@ ResponseEntity<List<NotebookSession>> getNotebookSessions(
 	/**
 	 * Create session and return its ID
 	 *
-	 * @param request session to create and projectId
+	 * @param session session to create and projectId
```
YohannParis, line 90: Why the double `session`?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
dvince2: To match the method definition - at some point someone changed the parameter name from `request` to `session`.  Github doesn't do the nice formatting but the first word after param is the variable name, then the next string is the description: 
image

kbirk (approved):

bigglesandginger (approved):

PR #3725, opened 2024-05-29
nist to number fix

blanchco:
# Description

* I want this function to return a number, not a string.
* This was causing our nist input to return a string instead of a number

### DIFF
--- a/packages/client/hmi-client/src/utils/number.ts
+++ b/packages/client/hmi-client/src/utils/number.ts
@@ -49,8 +49,7 @@ export function numberToNist(num: string) {
 export function nistToNumber(numStr: string): number {
 	// Remove any spaces from the formatted number
 	numStr = numStr.replace(/\s/g, '');
-	const num = parseFloat(numStr);
-	return num;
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/utils/number.ts
```diff
@@ -46,11 +46,11 @@ export function numberToNist(num: string) {
...
  */
-export function nistToNumber(num: string): string {
+export function nistToNumber(numStr: string): number {
 	// Remove any spaces from the formatted number
-	let numStr = num.replace(/\s/g, '');
-	numStr = parseFloat(numStr).toString();
-	return numStr;
+	numStr = numStr.replace(/\s/g, '');
+	const num = parseFloat(numStr);
+	return num;
```
YohannParis, line 53:
```suggestion
	return parseFloat(numStr);
```

YohannParis (approved):

YohannParis (approved):

mwdchang (approved):

PR #3727, opened 2024-05-29
Config matrix copy

mwdchang:
### Summary
Just an experiment with copy-n-paste buffer to be able to populate a configuration matrix faster.

### Testing
Copy and paste this into testing interface `http://localhost:8080/clipboard`

```
,I_unvax_0_9,I_unvax_10_19
S_unvax_0_9,999,777
S_unvax_10_19,888,666
```

The top-left parts of the matrix should be replaced with the given values above

image


packages/client/hmi-client/src/temp/Clipboard.vue
```diff
@@ -0,0 +1,131 @@
...
+ **/
+const updateByMatrixCSV = async (
+	model: Model,
+	paramLocationMap: Map<string, TemplateSummary[]>,
+	matrixType: string,
+	text: string
+) => {
+	// Parse
+	const rows = text.split('\n');
+	const columns = rows[0].split(',');
```
YohannParis, line 57: If the first row is not the column's names, what would happen? Should we test for that or it's too much?

packages/client/hmi-client/src/temp/Clipboard.vue
```diff
@@ -0,0 +1,131 @@
...
+			paramLocationMap.forEach((v, k) => {
+				if (matrixType === 'subjectControllers') {
+					if (rowLabel === v[0].subject && colLabel === v[0].controllers.join('-')) {
+						updateParameter(model, k, 'value', +data[j]);
+					}
+				} else if (matrixType === 'subjectOutcome') {
+					// TODO
+				} else if (matrixType === 'outcomeControllers') {
+					// TODO
+				}
```
YohannParis, line 74: What is the differences in matrixType in terms of `updateParameter` uses?

packages/client/hmi-client/src/temp/Clipboard.vue
```diff
@@ -0,0 +1,131 @@
...
+};
+
+onMounted(async () => {
+	model.value = sirJSON as Model;
+	const response = await getMMT(model.value);
+	mmt.value = response.mmt;
+	mmtParams.value = response.template_params;
+	ready.value = true;
+
+	document.onpaste = processPasteEvent;
```
YohannParis, line 125: Why not `addEventListener`? This could be a source of bug if another method overwrite the `document.onpaste`.

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
```diff
@@ -109,16 +109,17 @@ const props = defineProps<{
...
 
 const emit = defineEmits(['update-cell-value']);
 
 const matrixTypes = ['subjectOutcome', 'subjectControllers', 'outcomeControllers', 'other'];
-const matrixType = ref('subjectOutcome');
+const matrixType = ref(props.matrixType || 'subjectOutcome');
 const matrixMap = ref<any>();
 const matrix = ref<any>([]);
 
-const currentMatrixtype = ref('');
```
shawnyama, line 121: What does `currentMatrixType` do differently from `matrixType`?

packages/client/hmi-client/src/temp/Clipboard.vue
```diff
@@ -0,0 +1,131 @@
...
+					// TODO
+				}
+			});
+		}
+	}
+};
+
+const csvStringProcessor = async (item: DataTransferItem) => {
+	if (item.kind !== 'string') return;
+	if (item.type !== 'text/plain') return;
```
shawnyama, line 82:
```suggestion
        if (item.kind !== 'string' || item.type !== 'text/plain') return;
```

packages/client/hmi-client/src/temp/Clipboard.vue
shawnyama, line 74: switch case could be nicer here

shawnyama (approved):

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
```diff
@@ -109,16 +109,17 @@ const props = defineProps<{
 	id: string;
 	stratifiedMatrixType: StratifiedMatrix;
 	shouldEval: boolean;
+	matrixType?: string;
```
Tom-Szendrey, line 112: Will this break if I use matrixType = "example"? 
If it has to be one of the types in `matrixTypes` would it be worth throwing into an enum?

packages/client/hmi-client/src/temp/Clipboard.vue
```diff
@@ -0,0 +1,131 @@
...
+		mmt.value = response.mmt;
+		mmtParams.value = response.template_params;
+	});
+};
+
+const processPasteEvent = async (event: ClipboardEvent) => {
+	const clipboardData = event.clipboardData;
+	if (!clipboardData) return;
+	const item = clipboardData.items[0];
+	await csvStringProcessor(item);
```
Tom-Szendrey, line 115: Minor:
The `csvStringProcessor` could return `item` to then allowing this `processPasteEvent` to handle the `updateByMatrix` call and `getMMT` logic  
This would take out a layer which would make reading it a bit easier and debugging a bit easier

I am not sure the plans for `csvStringProcessor` though. If it is going to be used more often than just here leave it as is 

Rather than
```
func a() 
  call b()

func c() 
  call a()
```
```
func c()
  call a()
  call b()  
```

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
mwdchang: If I recall, this was introduced because of the refreshing problems in updating model-configuration, I needed an extra variable to keep track across component-refreshes.

packages/client/hmi-client/src/temp/Clipboard.vue
mwdchang: why is that?

packages/client/hmi-client/src/temp/Clipboard.vue
mwdchang: For some reason this does not work, it seems like part of the paste mechanism is async, though the documentation says otherwise. But I need to do everything via loops and callbacks, I can't return things cleanly.

packages/client/hmi-client/src/temp/Clipboard.vue
mwdchang: it will break, we will need to craft different parsers for different types of CSV-formats

packages/client/hmi-client/src/temp/Clipboard.vue
shawnyama: Since you're just checking for the same variable, doesn't really matter I guess.

packages/client/hmi-client/src/temp/Clipboard.vue
mwdchang: done

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
mwdchang: Not worth it right now - if this copy/paste idea moves further we shall revisit.

Issue #3705, opened 2024-05-27
[BUG]: Uploading dataset not working


Tom-Szendrey:
When uploading a new dataset to the project we are getting an object that has multiple file names and no columns section filled in
This occurs when we only upload a singular file

Sample on staging:
https://app.staging.terarium.ai/projects/d7e690d5-ec81-4b01-8e97-295863454dfe/dataset/1501aea6-ddda-4b4d-a8c4-c8e7da18dd9b
![image](https://github.com/DARPA-ASKEM/terarium/assets/17088680/463cb5d4-52f5-430c-8a99-e099718c2193)

Sample object locally with the following file:
[simResult.csv](https://github.com/DARPA-ASKEM/terarium/files/15457529/simResult.csv)
![image](https://github.com/DARPA-ASKEM/terarium/assets/17088680/c6147e48-b4ca-4168-8240-5da7d9013c10)

```
{
    "id": "ca081a8c-bc2f-4c2d-9070-2bc79b18e635",
    "createdOn": "2024-05-27T14:59:13.498+00:00",
    "updatedOn": "2024-05-27T14:59:14.377+00:00",
    "name": "simResult",
    "description": "simResult.csv",
    "fileNames": [
        "simResult.csv",
        "simResult.csv"
    ],
    "temporary": false,
    "publicAsset": false,
    "userId": "7421f1cd-b9aa-4e26-be86-a4cefc3acc7e",
    "datasetUrls": [],
    "columns": []
}
```

asylves1:
Talked to @dvince2 about the values missing from the columns. It appears to be an issue with database. Re-assigning it to @dvince2 .

PR #3729, opened 2024-05-29
Dataset filenames repeating

asylves1:
# Description

Small fix to stop filenames from duplicating

image

image

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
@@ -775,7 +775,7 @@ public class DatasetController {
 				// add the filename to existing file names
 				if (updatedDataset.get().getFileNames() == null) {
 					updatedDataset.get().setFileNames(new ArrayList<>(List.of(filename)));
-				} else {
+				} else if (!updatedDataset.get().getFileNames().contains(filename)) {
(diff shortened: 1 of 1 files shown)


Tom-Szendrey:
This does not resolve the underlying issue:
1) We still do not have column information in our dataset? 
image

2) Now for some reason the port on our dataset has a value of null? :O 
image

Tom-Szendrey (changes requested):
This prevents the dataset node from having a correct value in the node's output
This will prevent the user from using this dataset in the workflow

Please double check this maybe my testing included something incorrect?

asylves1:
I'm seeing that issue on main with the workflow on the main branch

dvince2:
> This does not resolve the underlying issue:
> 
>     1. We still do not have column information in our dataset?
@asylves1 and i spoke this morning and this is a deeper database issue. I now have #3705 assigned to me in the sprint and either myself or @kbirk will tackle it

asylves1:
The issue with the columns is a database one that will be handled by @dvince2

Tom-Szendrey (approved):
LGTM

Tom-Szendrey:
> This does not resolve the underlying issue:
> 
> 1. We still do not have column information in our dataset?
> 
> image
> 2. Now for some reason the port on our dataset has a value of null? :O
> 
> image

This is not a you problem. Thank you for pointing that out

YohannParis (approved):

PR #3732, opened 2024-05-29
3595 notebook summary - model stratify operator Part 1

jryu01:
# Description

Summary generation for model stratification operator.

- Added `createNotebook` function to the operation interface so that each operator can implement its own function to produce the corresponding notebook payload object for the summary request.
- Summary request for a operator is being sent at the workflow level when an operation output is appended to the node. 

I'll work on the summary UI and interaction on the follow up PR 

Resolves #(issue)


packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -36,9 +36,14 @@
...
 					<slot name="header-actions" />
 					<tera-operator-annotation
+						v-if="activeOutputSummary === undefined"
 						:state="node.state"
 						@update-state="(state: any) => emit('update-state', state)"
 					/>
+					<!-- TODO: Make this as a summary component -->
+					<div v-else>
+						{{ activeOutputSummary === '' ? 'Generating AI summary...' : activeOutputSummary }}
+					</div>
```
mwdchang, line 46: These are two separate concept, they are not, I don't think they are meant to replace one another
- State.annotation: The notes made on the operator
- WorkflowOutput.summary: The summary/note of the output

So you can have one State.annotation but many Worlfowoutput.summary.

I think we should revert the changes here for now, and get clarifications from Pacale and the design team on what they want.

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/stratify-mira-operation.ts
```diff
@@ -70,5 +72,16 @@ export const StratifyMiraOperation: Operation = {
...
+	createNotebook: async (
+		_node: WorkflowNode<StratifyOperationStateMira>,
+		outputPort: WorkflowOutput<StratifyOperationStateMira>
+	) => {
+		const modelIdToLoad = outputPort.value?.[0];
+		const outputModel = await getModel(modelIdToLoad);
+		const code = outputPort.state?.strataCodeHistory?.[0].code ?? '';
+		// TODO: Add llm query and thought to the notebook
+		const notebook = createNotebookFromCode(code, 'python3', { 'application/json': outputModel });
+		return notebook;
```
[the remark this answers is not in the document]
jryu01: Done

mwdchang (approved):

mwdchang:
I think we should start to think about externalizing output summaries into their own table/structures rather than having them inside the workflow. They would be easier to wrangle if want them as asynchronous tasks, and less clogging on the workflow object which can get fairly large as it is.

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -178,6 +179,7 @@ const kernelManager = new KernelSessionManager();
 
 let editor: VAceEditorInstance['_editor'] | null;
 const codeText = ref('');
+const llmQuery = ref('');
```
Tom-Szendrey, line 182: Is it worth throwing `codeText`, `llmQuery`, and `llmThoughts` into one object? 
May be cleaner to read as theyre all individual parts of the same `StratifyCode` type.

If you dont think it will be easier to follow, or it adds complexity in a way I am not seeing no pressure

jryu01:
> I think we should start to think about externalizing output summaries into their own table/structures rather than having them inside the workflow. They would be easier to wrangle if want them as asynchronous tasks, and less clogging on the workflow object which can get fairly large as it is.

Yeah I think it would be useful to have it as external table so that we can easily keep track of the status of the async task. But isn't it also useful to have summary baked inside the state object? Without it, we would need to fetch the summary for every single node and I wonder if that's too much overhead as well.

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -389,10 +395,26 @@ function appendOutput(
 	node.outputs = node.outputs.filter((d) => d.value);
 
 	selectOutput(node, uuid);
-
+	generateSummary(node, outputPort);
```
Tom-Szendrey, line 398: Because `generateSummary` is calling `updateOutputPort` I think there may be some race condition here when other nodes emit `update-output-port`

I will need to dig a bit deeper to be sure as its been a little bit since ive read this section

mwdchang:
> > I think we should start to think about externalizing output summaries into their own table/structures rather than having them inside the workflow. They would be easier to wrangle if want them as asynchronous tasks, and less clogging on the workflow object which can get fairly large as it is.
> 
> @mwdchang Yeah I think it would be useful to have it as external table so that we can easily keep track of the status of the async task. But isn't it also useful to have summary baked inside the state object? Without it, we would need to fetch the summary for every single node in the workflow and I wonder if that's too much overhead as well.

No not necessarily - the workflow object is continuously fetched and updated in the background, it automatically buffers up changes and save/refetch every 7,8 seconds. Where I think we'd only fetch the summaries for a given node when you drilldown into it. So overall we will have more API calls, but potentially the data throughput should be lower, and it is easier to deal with if the summary generation process is longer running (e.g. if we were to remove the 1000-character truncation on the output).

jryu01:
> > > I think we should start to think about externalizing output summaries into their own table/structures rather than having them inside the workflow. They would be easier to wrangle if want them as asynchronous tasks, and less clogging on the workflow object which can get fairly large as it is.
> > 
> > 
> > @mwdchang Yeah I think it would be useful to have it as external table so that we can easily keep track of the status of the async task. But isn't it also useful to have summary baked inside the state object? Without it, we would need to fetch the summary for every single node in the workflow and I wonder if that's too much overhead as well.
> 
> No not necessarily - the workflow object is continuously fetched and updated in the background, it automatically buffers up changes and save/refetch every 7,8 seconds. Where I think we'd only fetch the summaries for a given node when you drilldown into it. So overall we will have more API calls, but potentially the data throughput should be lower, and it is easier to deal with if the summary generation process is longer running (e.g. if we were to remove the 1000-character truncation on the output).



> > > I think we should start to think about externalizing output summaries into their own table/structures rather than having them inside the workflow. They would be easier to wrangle if want them as asynchronous tasks, and less clogging on the workflow object which can get fairly large as it is.
> > 
> > 
> > @mwdchang Yeah I think it would be useful to have it as external table so that we can easily keep track of the status of the async task. But isn't it also useful to have summary baked inside the state object? Without it, we would need to fetch the summary for every single node in the workflow and I wonder if that's too much overhead as well.
> 
> No not necessarily - the workflow object is continuously fetched and updated in the background, it automatically buffers up changes and save/refetch every 7,8 seconds. Where I think we'd only fetch the summaries for a given node when you drilldown into it. So overall we will have more API calls, but potentially the data throughput should be lower, and it is easier to deal with if the summary generation process is longer running (e.g. if we were to remove the 1000-character truncation on the output).

Yeah If we are fetching summary only when we drilldown into the node, I think that should be fine but I'm a bit concerned because It looks like we want to display the AI generated summary for the selected output on the operator node itself as well (according to the design)

packages/client/hmi-client/src/services/beaker.ts
```diff
@@ -0,0 +1,11 @@
+import axios, { AxiosHeaders } from 'axios';
+
+const BeakerAPI = axios.create({
+	baseURL: '/beaker',
+	headers: new AxiosHeaders()
+});
+
+export const summarizeNotebook = async (notebook: any) => {
+	const { data } = await BeakerAPI.post('/summary', { notebook });
```
Tom-Szendrey, line 9: I am not familiar with this endpoint, does it accept just `any` and tries its best no matter what?
If it has a loose type can you add a comment?
If it has a strict type can you type it?

mwdchang:
> Yeah If we are fetching summary only when we drilldown into the node, I think that should be fine but I'm a bit concerned because It looks like we want to display the AI generated summary for the selected output on the operator node itself as well (according to the design)

@jryu01 you need to look at the design sketches as general directions and not the absolute truth. They are often created before the fundamental theories are established and proven to be sound - so usually 2-3 iterations are needed as we discover new things.

packages/client/hmi-client/src/services/beaker.ts
mwdchang: FYI this is not on our end:

https://github.com/jataware/beaker-kernel/blob/886b2b3913ca1460f0301a5cd97cbcf15de609bc/beaker_kernel/server/main.py#L217-L221

jryu01:
> @jryu01 you need to look at the design sketches as general directions and not the absolute truth. They are often created before the fundamental theories are established and proven to be sound - so usually 2-3 iterations are needed as we discover new things.

if you think we need further discussion on the design for the output summary at the node level, it would be better to reach a general consensus on the topic first I guess.

packages/client/hmi-client/src/services/beaker.ts
jryu01: I'll add a comment on that. It's from Jataware's beaker repo https://github.com/jataware/beaker-kernel/blob/886b2b3913ca1460f0301a5cd97cbcf15de609bc/beaker-ts/src/notebook.ts#L405

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
jryu01: Yeah makes sense. I'll look in to it and address them on my follow up PR if that's better.

mwdchang:
> @mwdchang if you think we need further discussion on the design for the output summary at the node level, it would be better to reach a general consensus on the topic first I guess. (I personally don't have a strong opinion on this 😅 )

I think there are still unclear points on how the operator-annotation interacts with the output-summaries, and that impacts how things are fetched.

I think for now the more important part is to actually prove that we can generate succinct, and useful summaries across a variety of operators. And to do that, we may need to craft our own prompts instead using the generalized notebook summarization. @jryu01

Tom-Szendrey (approved):
Approving - I believe all questions and concerns have been either answered or made into comments to still be addressed

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
mwdchang: The usage of update-output-port is frowned upon, I think we will probably shift base and store the summary outside as per other thread to side-step these issues altogether as per discussions in other threads.

PR #3733, opened 2024-05-29
move menu save option to a button

asylves1:
# Description

On the Model Config the save option should not be in the menu. It should be a button under it in the drill down section.
image

Resolves #(issue)


shawnyama (dismissed):

asylves1:
forgot to switch it over

shawnyama (approved):

PR #3734, opened 2024-05-29
add uncertainty values

blanchco:
# Description

* Add uniform distributions to constants with uncertainties
* The changes seem pretty big for that attached ticket so I'll do the tooltips in another [ticket](https://github.com/DARPA-ASKEM/terarium/issues/3735)

https://github.com/DARPA-ASKEM/terarium/assets/33158416/83286e14-cf10-40f8-a48d-e9331e45f05d

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
@@ -1,90 +1,92 @@
 <template>
-	<header>
-		<strong>{{ parameterId }}</strong>
-		<span v-if="name" class="ml-1">{{ '| ' + name }}</span>
-		<span v-if="unit" class="ml-2">({{ unit }})</span>
-		<span v-if="description" class="ml-4">{{ description }}</span>
-	</header>
-	<main>
-		<span class="flex gap-2">
-			<Dropdown
-				:model-value="getParameterDistribution(modelConfiguration, parameterId).type"
-				@change="
-					emit('update-parameter', {
-						id: parameterId,
-						distribution: formatPayloadFromTypeChange($event.value)
-					})
-				"
-				option-label="name"
-				option-value="value"
-				:options="distributionTypeOptions()"
-			/>
+	<div class="flex flex-column flex-1">
+		<header>
+			<strong>{{ parameterId }}</strong>
+			<span v-if="name" class="ml-1">{{ '| ' + name }}</span>
+			<span v-if="unit" class="ml-2">({{ unit }})</span>
+			<span v-if="description" class="ml-4">{{ description }}</span>
+		</header>
+		<main>
+			<span class="flex gap-2">
+				<Dropdown
+					:model-value="getParameterDistribution(modelConfiguration, parameterId).type"
+					@change="
+						emit('update-parameter', {
+							id: parameterId,
+							distribution: formatPayloadFromTypeChange($event.value)
+						})
+					"
+					option-label="name"
+					option-value="value"
+					:options="distributionTypeOptions()"
+				/>
 
-			<!-- Constant -->
-			<tera-input
-				v-if="
-					getParameterDistribution(modelConfiguration, parameterId).type ===
-					DistributionType.Constant
-				"
-				label="Constant"
-				type="nist"
-				:model-value="getParameterDistribution(modelConfiguration, parameterId)?.parameters.value"
-				@update:model-value="
-					emit('update-parameter', {
-						id: parameterId,
-						distribution: formatPayloadFromParameterChange({ value: $event })
-					})
-				"
+				<!-- Constant -->
+				<tera-input
+					v-if="
+						getParameterDistribution(modelConfiguration, parameterId).type ===
+						DistributionType.Constant
+					"
+					label="Constant"
+					type="nist"
+					:model-value="getParameterDistribution(modelConfiguration, parameterId)?.parameters.value"
+					@update:model-value="
+						emit('update-parameter', {
+							id: parameterId,
+							distribution: formatPayloadFromParameterChange({ value: $event })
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -1,90 +1,92 @@
...
-				"
-			>
-				<tera-input
-					label="Min"
-					type="nist"
-					:model-value="
-						getParameterDistribution(modelConfiguration, parameterId)?.parameters.minimum
-					"
-					@update:model-value="
+	<div class="flex flex-column flex-1">
```
blanchco, line 2: Just surrounded this with a div for styling with checkbox

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -154,26 +154,16 @@
...
-							<p class="empty-section">No parameters found.</p>
-						</section>
-					</AccordionTab>
+				</Accordion>
+				<tera-parameter-table-v2
+					v-if="!isEmpty(knobs.transientModelConfig) && !isEmpty(mmt.parameters)"
+					:model-configuration="knobs.transientModelConfig"
+					:mmt="mmt"
+					:mmt-params="mmtParams"
+					@update-parameters="setParameterDistributions(knobs.transientModelConfig, $event)"
```
blanchco, line 163: can now update 1 or more parameters at a time

YohannParis: Can we select only the available constant

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
```diff
@@ -49,41 +99,99 @@
...
+const uncertaintyType = ref(DistributionType.Uniform);
+const uncertaintyPercentage = ref(10);
+const selectedParameters = ref<string[]>([]);
+
+const numParameters = computed(() => Object.keys(props.mmt.parameters).length);
+
 const matrixModalId = ref('');
+
+const onAddUncertainty = () => {
+	const selected = Object.keys(props.mmt.parameters);
```
YohannParis, line 154: You should filter to select only the constant and not the distributions.

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
YohannParis, line 1: I don't know how I feel about this <template>... After this PR we can have a look and see how this can be cleaned up. Nothing wrong in terms of engineering. But while we have the opportunity to make things clean and tidy before the next hachathon, I would like this keystone part of the application to the best it can be.

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
blanchco: easily yes, will do

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
```diff
@@ -49,41 +99,99 @@
...
+				id: paramId,
+				distribution: {
+					type: uncertaintyType.value,
+					parameters: {
+						minimum:
+							parameter.parameters.value -
+							(parameter.parameters.value * uncertaintyPercentage.value) / 100,
+						maximum:
+							parameter.parameters.value +
+							(parameter.parameters.value * uncertaintyPercentage.value) / 100
```
mwdchang, line 184: Some comments on the `onUpdateDistributions` function:

- The naming convention here is confusing and can make the logic harder to follow, for example `distributions` and `distribution` are different types of objects, where one would expect them to be the same. On top of that "distributions" get feed into "parameters" slot in setParameterDistributions ... 🤯 

- I think the math parts are more readable if we don't repeat so much

```
const v = parameter.parameters.value;
const delta = (parameter.parameters.value * uncertaintyPercentage.value) / 100;
...
minimum: v - delta,
maximum: v + delta
```

- More of a nitpick and I am not totally sure, but I think in most place we specify percentage between `[0-1]` rather than `[0-100]`.

blanchco: I'll make some changes in the morning 😁

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
blanchco: Do you mean about the accordion inside the component?  I couldn't really figure a better way of following the design faithfully without putting it in the component, but ya I don't like it either.  Anyways if we move those buttons and filters outside the accordion header it should be easy to put the accordion back into the parent

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
blanchco: this has been updated, we can now only select constants and not distributions

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table-v2.vue
blanchco: I've renamed everything so that its clearer (i hope) and fixed up the math

mwdchang: cool thx

mwdchang (approved):
Note I only looked at the service-level functions and their connections. They look okay to me.

YohannParis (approved):

PR #3736, opened 2024-05-29
correcting use of save-asset for models

Tom-Szendrey:
# Description
Previously we could not save the stratified model as a new model correctly.

Here you should be able to now stratify a model and use the dropdown to save a stratified model.
This should then allow you to go to the model page and see model equations ect

image

image


shawnyama (approved):
♥️

YohannParis (approved):
LGTM 🎉

PR #3740, opened 2024-05-30
Fix dataset cols

kbirk:
We weren't setting the `dataset` field on the columns, so they had no `dataset_id` stored. I hate hibernate

Fixes https://github.com/DARPA-ASKEM/terarium/issues/3705


dvince2 (approved):
Thanks for the help on this!!!

mwdchang (approved):

PR #3742, opened 2024-05-30
feat: explicit saving for model edit

shawnyama:
# Description

- Before new outputs were created when an edit was done to the original model. Now new outputs are created when the Save button is clicked
- This is like a first pass, in the next pass I'll handle drafts (see if there is some sort of design for that)
- The top bar with the decomposed/flat and save button can be clicked through (only the buttons are clickable)


Tom-Szendrey (approved):
TY

YohannParis (approved):

PR #3743, opened 2024-05-30
feat: stratified initials and parameter metadata on model page

shawnyama:
# Description

- Added base variables for initials and parameters that are stratified
- You can toggle the base variables to view the stratified variables
- As usual, base initials are edited in the model metadata - base parameters are not edited in the `model.parameters` as they don't exist there, for base parameters I am creating them in the metadata similarly to how initials are edited
- Updated stratified matrix components to have a read only mode, the read only mode doesn't show values - just variables
- You can open stratified matrices under base parameters to see how they relate to the initials (no need for initial matrices)
- Got rid of an unused prop `openValueConfig` in `tera-stratified-matrix-modal`

https://github.com/DARPA-ASKEM/terarium/assets/28237258/fe7e05a5-6b6b-47a5-8426-c69d4f400afb


mwdchang (changes requested):
The elephant in the room, is that `metadata` is not a fully recognized field on the program.  Before going too far down the rabbit hole of using a data structure that everyone else ignores, what are our plan and strategy to consolidate "metadata" after

- Model edit operations
- Model stratification operations

I don't think the outputs of these operations will have the original metadata intact.

packages/client/hmi-client/src/components/model/tera-parameters-metadata.vue
```diff
@@ -1,5 +1,49 @@
...
-	<ul>
+	<template v-if="isStratified">
+		<ul>
+			<li
+				v-for="([baseParameterId, values], index) in collapsedParameters.entries()"
+				:key="baseParameterId"
+			>
+				<template v-if="values.length > 1">
+					<tera-parameter-metadata-entry
+						:parameter="model.metadata?.parameters?.[baseParameterId] ?? { id: baseParameterId }"
```
mwdchang, line 10: I see quite a few of thees conditional assignment and emit patterns that I have a concern with.

```
<something
  :parameter = "model.foobar.blah[xyz] ?? { id: "abc" }"
  @update-parameter = update(xyz) >
</something>
```

This is a bit difficult to discern exactly what is getting updated, and I don't know why we need to have an alternate path `?? {id "abc"}`

I also see in quite a few places you are doing `parameters.find(...)`, this not very efficient. And it may be the case that we should consider preprocessing the data into a more template-friendly data structure rather than using it as-is.

shawnyama: For that case I am updating the base parameter. The base parameter doesn't exist in the model so I have to create it in the metadata. So first it is `{ id: baseParameterId }` and once `@update-parameter` is ran it'll be referenced in the metadata as `model.metadata?.parameters?.[baseParameterId]`

shawnyama: I thought specifying the update type would be a pain because I'd be passing too many emits

mwdchang: I think I need a refresher on why are we creating phantom variables that do not exist.....this might be a chicken-and-egg problem, let's maybe sync up.

YohannParis:
> The elephant in the room, is that `metadata` is not a fully recognized field on the program. Before going too far down the rabbit hole of using a data structure that everyone else ignores, what are our plan and strategy to consolidate "metadata" after
> 
>     * Model edit operations
> 
>     * Model stratification operations
> 
> 
> I don't think the outputs of these operations will have the original metadata intact.

This is the same problem we had with _model configuration_. We should not keep them as AMR and try to fix those issue with the rest of the program.
We need to think about a way for us to preserve `metadata` after a model edit or stratification for the UI. We have been talking for more than a year about this problem with TA2... and no one wants to solve it.
I propose that we create a model merging function to deal with those basic issues of receiving edited AMR that removes or destroy existing data.

packages/client/hmi-client/src/components/model/tera-parameters-metadata.vue
```diff
@@ -1,5 +1,49 @@
 <template>
-	<ul>
+	<template v-if="isStratified">
+		<ul>
+			<li
+				v-for="([baseParameterId, values], index) in collapsedParameters.entries()"
```
mwdchang, line 5: It would be good if we can do a bit of preprocessing ahead of time so we can avoid complicated logic in the templates, as currently it is a bit hard to follow.

I would suggest we process `collapsedParameters.entries()` into something like 
```
[
  { baseParameterId, values, isVirtualParameter },
  ...
]
```

Then in the template we can simplify:

```
<li v-for="(obj, idx) in computedList">
  <!-- doesnt exist in model -->
  <template v-if="obj.isVirtual">
    <tera-parameter-metadata-entry ... >
  </template>
  <!-- exist in model -->
  <template v-else>
    <tera-parameter-metadata-entry ... >
  </template>
  <!-- children --->
  <template v-if="obj.values.length > 1">
    <tera-parameter-metadata-entry ... />
  </tempplate>
</li>
```

shawnyama: applied to parameters, doing initials in another branch

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table.vue
blanchco, line 1: I think we can remove tera-parameter-table and tera-initial-table components, they aren't being used anywhere.

YohannParis (approved):

YohannParis (approved):

shawnyama:
Sorry, this PR looks huge now because I deleted the old `tera-initial-table` and `tera-parameter-table` and renamed the v2 components.

blanchco (approved):

mwdchang (approved):

PR #3747, opened 2024-05-31
model config tooltips

blanchco:
# Description

* Just adding a tooltip to describe the distribution type
* Was having some formatting some issues with `tera-tooltip` so I couldn't use a custom component for the dropdown options.  Im using the primvue tooltip directive until we have some time to rework the tooltip.

https://github.com/DARPA-ASKEM/terarium/assets/33158416/c4adf786-a62d-4549-9784-8f383e36b3c7

### DIFF
--- a/packages/client/hmi-client/src/services/distribution.ts
+++ b/packages/client/hmi-client/src/services/distribution.ts
@@ -8,7 +8,13 @@ export const DistributionTypeLabel: { [key in DistributionType]: string } = {
 	[DistributionType.Uniform]: 'Uniform'
 };
 
-export const distributionTypeOptions = (): { name: string; value: string }[] =>
+export const DistributionTypeDescription: { [key in DistributionType]: string } = {
+	[DistributionType.Constant]: 'value is the constant value.',
+	[DistributionType.Uniform]:
+		'low is the lower range (inclusive), high is the upper range (exclusive).'
+};
+
+export const distributionTypeOptions = (): { name: string; value: DistributionType }[] =>
 	Object.values(DistributionType).map((dist) => ({
 		name: DistributionTypeLabel[dist],
 		value: dist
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/distribution.ts
```diff
@@ -8,7 +8,13 @@ export const DistributionTypeLabel: { [key in DistributionType]: string } = {
 	[DistributionType.Uniform]: 'Uniform'
 };
 
-export const distributionTypeOptions = (): { name: string; value: string }[] =>
+export const DistributionTypeDescription: { [key in DistributionType]: string } = {
+	[DistributionType.Constant]: 'value is the constant value.',
+	[DistributionType.Uniform]:
+		'low is the lower range (inclusive), high is the upper range (exclusive).'
+};
```
mwdchang, line 15: I am wondering if these are really needed here. In my head our target users will already have some notions of epidemiology, differential-equations, and running/evaluating models. So they have some notion of distributions, not to mention these are pretty basic things in the scientific/mathematic world.

So there might end up being things that cause minor nuisance rather than offer any particular help. 

As an aside, where did you get the tooltip-text for Uniform? Cause I don't think that is true.

blanchco: Just got it from the [link](https://pytorch.org/docs/stable/distributions.html#uniform) in the attached ticket the @ngraham76 made, Im not sure if its true either...but I can always change the text if its not.  

Im not sure whose call it would be to keep the tooltips or not, but we can just let the ticket sit in the meantime...

mwdchang: Lol the definitions are slightly different.
- Ontology has defined `[x, y]` range, see https://sites.google.com/site/probonto/downloads
- While pytorch defined `[x, y)` range

blanchco: maybe Ill get rid of the inclusive/exclusive text for now 😄

blanchco:
Closing for now.  Tooltips will probably be more useful when we support more distributions

PR #3749, opened 2024-05-31
3668 link devs to openstack

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/resources/application.properties
```diff
@@ -1,7 +1,7 @@
 ########################################################################################################################
 # Database configuration
 ########################################################################################################################
-spring.datasource.url=jdbc:postgresql://postgres.staging.terarium.ai:5432/terarium
+spring.datasource.url=jdbc:postgresql://10.64.22.49:5432/terarium
```
YohannParis, line 4: Why does PostGres do not have its dev url?

YohannParis (approved):
LGTM, soon we should be able to send those error messages to Mick

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/askem/RebacUser.java
```diff
@@ -9,44 +9,45 @@
...
-	private boolean serviceUser;
+	private final boolean serviceUser;
 
-	public RebacUser(String id, ReBACService reBACService) {
+	public RebacUser(final String id, final ReBACService reBACService) {
 		super(id);
 		this.reBACService = reBACService;
 
-		serviceUser = reBACService.isServiceUser(id);
+		serviceUser = ReBACService.isServiceUser(id);
```
bigglesandginger, line 20: Why did this case change? The local variable is "reBACService"

bigglesandginger:
I appreciate the message stuff and yet more finals but we have to stop overloading PRs from one concept to many

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/askem/RebacUser.java
dvince2: This looks like the call is being done statically on ReBACService the class... is that right?

dvince2 (approved):

packages/server/src/main/resources/application.properties
dgauldie: doesn't need it. ES and keycloak needed it because of SSL certs, but MQ and PG do not

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/askem/RebacUser.java
dgauldie: yeah

dgauldie:
> I appreciate the message stuff and yet more finals but we have to stop overloading PRs from one concept to many

yeah, that's my bad... I was working on 2 branches at once and merged the wrong files into the wrong branch... Instead of reverting and fixing things, I just merged the 2 branches together... Sorry... I passed the pain onto you guys

packages/server/src/main/resources/application.properties
YohannParis: That makes sense, even if it disturb my need to organise everything neatly.

PR #3760, opened 2024-06-04
adding optimize doc explaining our interface

Tom-Szendrey:
If this does not give you a full enough picture please let me know.

Some confusing parts to focus on to be sure they are understandable:
-maxiter
-maxfeval

### DIFF
--- a/docs/workflow operations/Optimize Doc.md
+++ b/docs/workflow operations/Optimize Doc.md
@@ -42,12 +42,12 @@ For example:
 `Step Size` (Not exposed):
     This is not currently exposed through the API.
     The step size for each iteration is 30% of the longest euclidean distance between the lower and upper bounds of a given parameter.
-    For example if we provide beta is [0.5, 1.5] then our step/hop size is a uniform random float between 0.0 and 0.3.
+    For example if we provide some parameter `beta` is [0.5, 1.5] then our step/hop size is a uniform random float between 0.0 and 0.3.
 
 `Minimized` (Settings)
     This is a boolean that is being used to determine if we are looking for a minima or a maxima.
-    :minimized = true, when I went to smallest values for beta that satisfy my constraint(s)
-    :minimized = false, when I want the largest values for beta satisfy my constraint(s)
+    :minimized = true, when I went to smallest values for some parameter that satisfy my constraint(s)
+    :minimized = false, when I want the largest values for some parameter that satisfy my constraint(s)
 
 `Intervention type`
     How you want to intervene on the parameter(s) you have selected.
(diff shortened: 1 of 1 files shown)


YohannParis (approved):
Great job!

docs/workflow operations/Optimize Doc.md
```diff
@@ -0,0 +1,107 @@
...
+## Goal
+Used to find the minimal or maximum value to a given parameter value or time that a QoI (Quality of Interest) constraint does not exceed the provided threshold.
+
+We can ask questions to try to find out a good start time for an intervention:
+
+I want to turn on "masking" (beta set to a smaller value ~0.1) 
+but I want to know how late I can do it while making the number of infected never exceed a specified number.
+
+We can use questions to try to find out parameter values:
+I want to turn on "quarantine" in day 30. How strong/effective (high beta needs to be) will this quarantine need to be in order to keep the number of infected lower than a (specified threshold) number?
```
mwdchang, line 11: Maybe rephrase here,  as well as the similar copies below. We don't have context into what `beta` refers to here, so you want to spell out things in full.

E.g. "I want to turn on quarantine on day 30 for 90 days, represented by the variable xyz, how strong/weak will xyz need to be in order to keep the number of infected below ...."

Tom-Szendrey: Thank you that is a good point. Especially if we are going to feed this documentation through llm it is best to be explicit and provide the full scope in one place rather than assume we understand beta here is a parameter that has some impact on infected

mwdchang:
FYI I don't know if the formatting you have in the markdown is intended to drive the rendering of the markdown, the actual render will not respect any of the indentation levels that you have created.

docs/workflow operations/Optimize Doc.md
```diff
@@ -0,0 +1,107 @@
+# Optimize:
+## Goal
+Used to find the minimal or maximum value to a given parameter value or time that a QoI (Quality of Interest) constraint does not exceed the provided threshold.
```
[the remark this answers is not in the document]
bigglesandginger: This is a very cumbersome sentence.  Would something like this work better?

`Used to find the limit that does not exceed the threshold of a Quality of Interest constraint for a given parameter.`

docs/workflow operations/Optimize Doc.md
```diff
@@ -0,0 +1,107 @@
...
+
+We can ask questions to try to find out a good start time for an intervention:
+
+I want to turn on "masking" (beta set to a smaller value ~0.1)
```
bigglesandginger, line 7: If this is a quote (question being asked) it should have quotes around it, otherwise the mixing of plural and singular pronouns is just confusing.

mwdchang (approved):

PR #3761, opened 2024-06-04
Model diagram wheel disable

shawnyama:
# Description

- Enable zoom on graph renderer only when the Ctrl key is held. Ctrl for zoom, let me know if another key would be better
- This removes the need for a unlock to pan/zoom button
- Panning is always enabled
- Made some slight padding/cursor adjustments to make the diagram look nice

https://github.com/DARPA-ASKEM/terarium/assets/28237258/76fae188-6c75-470e-9b0d-5bd9a76aff10


Tom-Szendrey (approved):

packages/client/graph-scaffolder/src/core/renderer.ts
```diff
@@ -324,6 +326,14 @@ export abstract class Renderer<V, E> extends EventEmitter {
 				.style('pointer-events', 'none');
 		}
 
+		window.addEventListener('keydown', (event) => {
+			if (event.key === 'z') this.isZoomable = true;
+		});
+
+		window.addEventListener('keyup', (event) => {
+			if (event.key === 'z') this.isZoomable = false;
+		});
```
mwdchang, line 335: - I know we discussed this as a possible solution, but please check with the design team or Pascale if this is the desirable behaviour

- Rather than hardwiring `z` and use a isZoomable flag. You may want to consider having a zoom-modifier key as a parameter. If the modifier key is null or "" than everything is normal, if the modifier key is specified than you need to hold that key to zoom

- Rather than latching on to windows, you might be able to chec for the modifier keys in the zoom event and do early exit. This will be a bit cleaner, since windows is a global object, and also note here you never cleaned up things in here. 

- Treat graph-scaffolder as a library, if you change the default everyone suffers. So ideally we would just pass in the intentioned behaviour rather than forcing everyone else to opt-out.

YohannParis (approved):

mwdchang (approved):

PR #3767, opened 2024-06-05
Missing metadata card

blanchco:
# Description

* Add a dismissible information card when any data is missing from model or parameter
* I've now based name, description, and units on the entry based off the model.  This allowed model changes to be shown here

Screenshot 2024-06-05 at 9 42 55 AM

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
@@ -30,28 +30,28 @@
 </template>
 
 <script setup lang="ts">
-import { ModelConfiguration } from '@/types/Types';
-import {
-	getInitialName,
-	getInitialDescription,
-	getInitialUnit,
-	getInitialExpression,
-	getInitialSource
-} from '@/services/model-configurations';
+import { Model, ModelConfiguration } from '@/types/Types';
+import { getInitialExpression, getInitialSource } from '@/services/model-configurations';
 import TeraInput from '@/components/widgets/tera-input.vue';
 import { ref } from 'vue';
 import Button from 'primevue/button';
+import {
+	getInitialDescription,
+	getInitialName,
+	getInitialUnits
+} from '@/model-representation/service';
 
 const props = defineProps<{
+	model: Model;
 	modelConfiguration: ModelConfiguration;
 	initialId: string;
 }>();
 
 const emit = defineEmits(['update-expression', 'update-source']);
 
-const name = getInitialName(props.modelConfiguration, props.initialId);
-const unit = getInitialUnit(props.modelConfiguration, props.initialId);
-const description = getInitialDescription(props.modelConfiguration, props.initialId);
+const name = getInitialName(props.model, props.initialId);
+const unit = getInitialUnits(props.model, props.initialId);
+const description = getInitialDescription(props.model, props.initialId);
 
 const sourceOpen = ref(false);
 
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
@@ -86,6 +86,7 @@ import Button from 'primevue/button';
 import Dropdown from 'primevue/dropdown';
 import { DistributionType, distributionTypeOptions } from '@/services/distribution';
 import { getParameter } from '@/model-representation/service';
+import { cloneDeep } from 'lodash';
 
 const props = defineProps<{
 	model: Model;
@@ -112,10 +113,7 @@ function getSourceLabel(initialId) {
 }
 
 function formatPayloadFromParameterChange(parameters) {
-	const parameterDistribution = getParameterDistribution(
-		props.modelConfiguration,
-		props.parameterId
-	);
+	const parameterDistribution = cloneDeep(distribution.value);
 	Object.keys(parameters).forEach((key) => {
 		if (!parameterDistribution) return;
 		if (key in parameterDistribution.parameters) {
@@ -127,11 +125,7 @@ function formatPayloadFromParameterChange(parameters) {
 }
 
 function formatPayloadFromTypeChange(type: DistributionType) {
-	const parameterDistribution = getParameterDistribution(
-		props.modelConfiguration,
-		props.parameterId
-	);
-
+	const parameterDistribution = cloneDeep(distribution.value);
 	parameterDistribution.type = type;
 	parameterDistribution.parameters = {};
 	return parameterDistribution;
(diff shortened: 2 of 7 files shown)


packages/client/hmi-client/src/model-representation/service.ts
```diff
@@ -365,3 +365,17 @@ export function cleanModel(model: Model): void {
...
+
+	const initialsCheck = initials.some((i) => {
+		const initialMetadata = getInitialMetadata(model, i.target);
+		return !initialMetadata?.name || !initialMetadata?.description || !initialMetadata?.units;
+	});
+
+	const parametersCheck = parameters.some((p) => !p.name || !p.description || !p.units?.expression);
+
+	return initialsCheck || parametersCheck;
+}
```
blanchco, line 381: checking initials and parameters for the field that are shown on model config

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -112,48 +112,27 @@
 					<AccordionTab header="Diagram">
 						<tera-model-diagram v-if="model" :model="model" :is-editable="false" />
 					</AccordionTab>
-					<template v-if="isPetriNet || isStockFlow">
```
blanchco, line 115: Im removing stockflow and regnet for now as we are designing for petrinet at the moment and this also wasnt a great way to do it.

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -38,18 +35,11 @@
 					"
 				/>
 				<!-- Uniform Distribution -->
-				<template
-					v-if="
-						getParameterDistribution(modelConfiguration, parameterId).type ===
-						DistributionType.Uniform
-					"
-				>
+				<template v-if="distribution.type === DistributionType.Uniform">
```
YohannParis, line 38: Should those be their own component?

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -90,30 +78,30 @@
...
 	parameterId: string;
 }>();
 
 const emit = defineEmits(['update-parameter', 'update-source']);
 
-const name = getParameterName(props.modelConfiguration, props.parameterId);
-const unit = getParameterUnit(props.modelConfiguration, props.parameterId);
-const description = getParameterDescription(props.modelConfiguration, props.parameterId);
+const name = getParameter(props.model, props.parameterId)?.name;
+const unit = getParameter(props.model, props.parameterId)?.units?.expression;
```
shawnyama, line 99:
```suggestion
const units = getParameter(props.model, props.parameterId)?.units?.expression;
```

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -124,23 +112,29 @@ function getSourceLabel(initialId) {
 }
 
 function formatPayloadFromParameterChange(parameters) {
-	const distribution = getParameterDistribution(props.modelConfiguration, props.parameterId);
+	const parameterDistribution = getParameterDistribution(
```
shawnyama, line 115: Is there a need to specify that its a `parameterDistribution` if we already in the `tera-parameter-entry` component?

packages/client/hmi-client/src/model-representation/service.ts
shawnyama: fyi you'll probably have to edit this later since @mwdchang told me that for initials I should be getting the `name`, `description` etc from `model.states` instead of storing everything in metadata

shawnyama (approved):

packages/client/hmi-client/src/model-representation/service.ts
shawnyama: Actually I can edit this once I do this with states

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
blanchco: I just needed a new name cause I created that `distribution` computed value and the linter was whining. I figured `parameterDistribution` would be okay instead

packages/client/hmi-client/src/model-representation/service.ts
blanchco: 🙏

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -30,28 +30,28 @@
...
 import Button from 'primevue/button';
+import {
+	getInitialDescription,
+	getInitialName,
+	getInitialUnits
+} from '@/model-representation/service';
 
 const props = defineProps<{
+	model: Model;
 	modelConfiguration: ModelConfiguration;
```
Tom-Szendrey, line 46: why do things like initial name, unit, description all come from model
and initial source, and expression comes from the configuration?

Can each configuration have its own unique source but not its own unique description?

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -30,28 +30,28 @@
...
 import { ref } from 'vue';
 import Button from 'primevue/button';
+import {
+	getInitialDescription,
+	getInitialName,
+	getInitialUnits
+} from '@/model-representation/service';
 
 const props = defineProps<{
+	model: Model;
```
Tom-Szendrey, line 45: If given the model configuration already, why do you also need to be given the model?
Im not sure if i am misunderstanding the new structure

blanchco: The exercise to revamp ModelConfiguration is to make a structured way to store any related metadata to the ModelConfiguration.  Its a better solution than shoving everything into the `metadata` key and overall to make it easier for us to work with a ModelConfiguration without explicitly sticking to the AMR data structure.

The use case here for needing Model values is for the fields (name, description, unit), is cause they are not stored on the ModelConfiguration, they are taken from the Model.  These fields are unique to the Model not ModelConfiguration.

Tom-Szendrey: after quick chat:
model config has a model ID not a model.
We do not want to make GET calls

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
blanchco: I could clone it 😄

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
blanchco: thats correct they cannot have unique names, units, descriptions so those come from Model. but can have unique sources and expressions 😄

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
blanchco: as discussed might be better when we support more than 2 distribution types 😄

PR #3768, opened 2024-06-05
add error messages for gollm controller

dgauldie:
# Description

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
@@ -1,5 +1,6 @@
 package software.uncharted.terarium.hmiserver.controller.gollm;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.Content;
@@ -10,6 +11,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 import java.util.regex.Matcher;
 import lombok.Data;
 import lombok.RequiredArgsConstructor;
@@ -46,6 +49,7 @@ import software.uncharted.terarium.hmiserver.service.tasks.ConfigureModelRespons
 import software.uncharted.terarium.hmiserver.service.tasks.ModelCardResponseHandler;
 import software.uncharted.terarium.hmiserver.service.tasks.TaskService;
 import software.uncharted.terarium.hmiserver.service.tasks.TaskService.TaskMode;
+import software.uncharted.terarium.hmiserver.utils.Messages;
 import software.uncharted.terarium.hmiserver.utils.rebac.Schema;
 
 @RequestMapping("/gollm")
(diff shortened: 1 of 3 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -187,58 +211,73 @@ public ResponseEntity<TaskResponse> createConfigureModelTask(
...
 
-			// Grab the document
-			final Optional<DocumentAsset> document = documentAssetService.getAsset(documentId, permission);
-			if (document.isEmpty()) {
-				throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Document not found");
-			}
+		// make sure there is text in the document
+		if (document.get().getText() == null || document.get().getText().isEmpty()) {
+			log.warn("Document {} has no text to send", documentId);
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("document.extraction.not-done"));
```
dvince2, line 223: I know that this was a `404` before, but technically is that right? I think that this may [technically](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/424) be a `424`.

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
YohannParis: Do you mean `422 Unprocessable Entity`? I don't see how a `424 Failed Dependency` explain the issue.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -187,58 +211,73 @@ public ResponseEntity<TaskResponse> createConfigureModelTask(
 		final Schema.Permission permission =
 				projectService.checkPermissionCanRead(currentUserService.get().getId(), projectId);
 
-		try {
+		// Grab the document
+		final Optional<DocumentAsset> document = documentAssetService.getAsset(documentId, permission);
+		if (document.isEmpty()) {
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("document.not-found"));
+		}
```
YohannParis, line 191: Why the `try/catch` is removed? the `getAsset` take cares of it?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dgauldie: "Regular web servers will normally not return this status code." this is from the link you sent

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dgauldie: I guess... getAsset does not throw exceptions. However, that is another task I want to go through. Services should throw better errors as well. But that is not in the scope of this initial work.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dvince2: Are you calling HMI-Server just a regular web server ;)?

PR #3777, opened 2024-06-06
Use read perm for gets

kbirk:


bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -242,7 +242,7 @@ ResponseEntity<NotebookSession> cloneNotebookSession(
 			@PathVariable("id") final UUID id,
 			@RequestParam(name = "project-id", required = false) final UUID projectId) {
 		final Schema.Permission permission =
-				projectService.checkPermissionCanWrite(currentUserService.get().getId(), projectId);
+				projectService.checkPermissionCanRead(currentUserService.get().getId(), projectId);
```
dgauldie, line 245: what is the difference between get and clone? do we save the new session?

dgauldie (approved):
LGTM

PR #3783, opened 2024-06-06
Stop model config from saving with no name

asylves1:
# Description

Name is now required to save model config
image

Resolves #(issue)


mwdchang (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -24,7 +24,12 @@
 					/>
 				</template>
 				<template #header-controls-right>
-					<Button class="mr-3" label="Save" @click="() => createConfiguration(false)" />
+					<Button
+						class="mr-3"
+						:disabled="isSaveDisabled"
+						label="Save"
+						@click="() => createConfiguration(false)"
```
YohannParis, line 31:
```suggestion
						@click="() => createConfiguration()"
```

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -613,7 +618,7 @@ const downloadConfiguredModel = async () => {
 };
 
 const createConfiguration = async (force: boolean = false) => {
-	if (!model.value) return;
+	if (!model.value || !props.node.state?.transientModelConfig?.name) return;
```
YohannParis, line 621: Is this test the same than with `isSaveDisabled`?
If yed, why not use it to return early?

YohannParis (approved):

PR #3802, opened 2024-06-07
Spike/update model configuration

dvince2:
# Description

* All existing model configuration objects/code have been moved into "legacy" objects and endpoints for this transition, to allow for a seamless light switch without a long running hard to maintain branch
* New model configuration class
* New model configuration services, repositories
* External services that use model configs will be updated to use new legacy endpoints:
* https://github.com/DARPA-ASKEM/askem-beaker/pull/136
* https://github.com/DARPA-ASKEM/sciml-service/pull/178
* https://github.com/DARPA-ASKEM/pyciemss-service/pull/89

Resolves #(issue)


packages/client/hmi-client/src/types/Types.ts
```diff
@@ -229,6 +229,38 @@ export interface ModelFramework extends TerariumAssetThatSupportsAdditionalPrope
     semantics: string;
 }
 
+export interface InitialSemantic extends Semantic {
+    target: string;
+    expression: string;
+    modelConfiguration: ModelConfiguration;
+}
```
mwdchang, line 236: I don't quite understand how the Java types are resolved here. 

Are you saying that in Javascript land we will have the full model-configuration object embedded into every children objects?

Also when we have `expression` we would normally also have `expression_mathml`, is this missed or are they somewhere else?

blanchco: For the `expression_mathml` point, I think initially I had an idea in my head of us being able to convert the expression to a mathml statement in the back end when we create an AMR from a ModelConfiguration.  

If that's not possible, maybe we have to add that key in for the InitialSemantic.

mwdchang: @blanchco can you elaborate a bit more? What library do you have in mind to do the mathml conversion and do we know what kind of coverage we would get?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+	@OneToMany(mappedBy = "modelConfiguration", fetch = FetchType.EAGER, cascade = CascadeType.ALL)
+	@Schema(accessMode = Schema.AccessMode.READ_ONLY)
+	@JsonManagedReference
+	private List<ParameterSemantic> parameterSemanticList;
+
+	@TSIgnore
+	@OneToMany(mappedBy = "modelConfiguration", fetch = FetchType.EAGER, cascade = CascadeType.ALL)
+	@Schema(accessMode = Schema.AccessMode.READ_ONLY)
+	@JsonManagedReference
+	private List<InitialSemantic> initialSemanticList;
```
mwdchang, line 48: I may have missed a chat here. So we have these three `List<blah>` members that seem to be used as proxies to populate "values".

- Why are we doing this?

So then the UI is working with `ModelConfiguration.values` which is a map, but the API, according to swagger, is expecting the ModelConfiguration to be specified in terms of these three xyzSemanticList.

Can someone maybe shed some light here?

blanchco:
Might be cause its late on a Friday, I checked this branch out and have been having troubles creating a new ModelConfiguration (the new one).  Perhaps we can chat Monday?

packages/client/hmi-client/src/types/Types.ts
blanchco: I haven't looked into any libraries haha, might be be just to stick the `expression_mathml` key in there for now

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
```diff
@@ -27,137 +27,245 @@
...
-import software.uncharted.terarium.hmiserver.utils.rebac.Schema;
+import software.uncharted.terarium.hmiserver.utils.Messages;
+import software.uncharted.terarium.hmiserver.utils.rebac.Schema.Permission;
 
 @RequestMapping("/model-configurations")
 @RestController
 @RequiredArgsConstructor
 @Slf4j
-@Transactional
 public class ModelConfigurationController {
```
mwdchang, line 47: @dvince2 So some observations:

### In general
Can you provide some examples of how CRUD would work with the new data structure and API-end points?


#### posting a new model-config
It seems like when creating a new model-config, we need to pass in the list-form of the Semantics, initailSemanticList, parameterSemanticList ... etc. This isn't readily available in the UI, because the payload had been transformed to the map-version under ModelConfiguration.values. 

So posting this worked:
```
{
  "name": "Default Name",
  "description": "Default Description",
  "fileNames": [
    "string"
  ],
  "temporary": true,
  "publicAsset": true,
  "calibrationRunId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "modelId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "initialSemanticList": [
    {
      "source": "string",
      "type": "initial",
      "target": "abc",
      "expression": "3.0"
    }
  ]
}
```

but something like this fails

```
{
  "name": "Default Name",
  "description": "Default Description",
  "fileNames": [
    "string"
  ],
  "temporary": true,
  "publicAsset": true,
  "calibrationRunId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "modelId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "values": {
    "abc": {
      "source": "string",
      "type": "initial",
      "target": "abc",
      "expression": "3.0"
    }
  }
}
```

#### get
I can't seem to GET model-configurations, they result in 401s

packages/client/hmi-client/src/types/Types.ts
mwdchang: @blanchco Downstream processes use mathml, we cannot just leave it as a dummy field.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+@Entity
+public class ModelConfiguration extends TerariumAsset {
+	private UUID calibrationRunId;
+	private UUID modelId;
+
+	@Transient
+	private Map<String, Semantic> values;
+
+	@TSIgnore
+	@OneToMany(mappedBy = "modelConfiguration", fetch = FetchType.EAGER, cascade = CascadeType.ALL)
```
mwdchang, line 33: I think we may need to have a think about indices, the way the PG is setup, where you have a model-configuration table and a semantics table via a foreign key.

The number of entries in the Semantics table will greatly out-number the number of model-configurations ... say by a factor of 15 to 20. And the indices don't look like they are setup properly:

```
Indexes:
    "semantic_pkey" PRIMARY KEY, btree (id)
Check constraints:
    "semantic_type_check" CHECK (type >= 0 AND type <= 2)
Foreign-key constraints:
    "fkjx7n8qnbr2wnt3ar4j7gp8tkx" FOREIGN KEY (model_configuration_id) REFERENCES model_configuration(id)
Access method: heap
```

YohannParis (commented):
Is the upload of a model file creating a model and a model configuration done in another PR?

packages/client/hmi-client/src/types/Types.ts
YohannParis: I thought they all used SimPy for expression!?
Then, indeed we need to add that field back.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -255,13 +255,13 @@ public ResponseEntity<JobResponse> makeOptimizeJobCiemss(
...
-					allInterventions = new ArrayList<Intervention>();
+					allInterventions = new ArrayList<>();
 				}
 				allInterventions.addAll(modelInterventions);
 				request.setFixedStaticParameterInterventions(allInterventions);
 			}
-		} catch (IOException e) {
-			String error = "Server error has occured while fetching the model configuration";
+		} catch (final IOException e) {
+			final String error = "Server error has occured while fetching the model configuration";
```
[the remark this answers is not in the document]
dvince2: Oh! nothing here was changed - just the class type so thats why i never saw these

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -430,13 +430,14 @@ ResponseEntity<Model> createModel(
 												@ArraySchema(
 														schema =
 																@io.swagger.v3.oas.annotations.media.Schema(
-																		implementation = ModelConfiguration.class)))),
+																		implementation =
+																				ModelConfigurationLegacy.class)))),
 				@ApiResponse(
 						responseCode = "500",
 						description = "There was an issue retrieving configurations from the data store",
```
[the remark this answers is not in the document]
dvince2: Similar to SImulationRequestController, there wasn't a change here outside of the class rename and it felt out of scope to move these in this PR

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ModelConfigurationService.java
```diff
@@ -1,34 +1,26 @@
...
-	@Observed(name = "function_profile")
-	protected String getAssetIndex() {
-		return elasticConfig.getModelConfigurationIndex();
-	}
-
-	public Integer uploadFile(final UUID uuid, final String filename, final ContentType contentType, final byte[] data)
-			throws IOException {
-		throw new UnsupportedOperationException("Unimplemented");
+	protected String getAssetPath() {
+		throw new UnsupportedOperationException("Model Configurations are not stored in S3");
```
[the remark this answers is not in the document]
dvince2: We have that in a few places - basically all of the Terarium Assets have this method stub and assets that don't use it throw this (not that we'd ever hit it). Models is the same, etc

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -416,7 +416,7 @@ ResponseEntity<Model> createModel(
 		}
 	}
 
-	@GetMapping("/{id}/model-configurations")
+	@GetMapping("/{id}/model-configurations-legacy")
```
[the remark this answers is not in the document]
dvince2: All the new stuff is on the offically named `model-configurations` endpoint, with all the old code going to `legacy` for now

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
dvince2: So the plan for the class is to have the `values` map. However this won't serialize correctly unless its all just one giant JSON blob, so the individual lists of types are needed to maintain the foreign-key relationship between the model config and the various Semantic types.  The lists were @TSIgnored to keep them hidden from the front end and instead use the map
If we want to use the lists throughout though that would make things easier!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
dvince2: I've resolved the weirdness around the lists/maps... there are now just the lists and ive confirmed they're working.

Unfortunately like most of the other API endpoints you ned to provide a project ID you have permissions to

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
mwdchang: We may have sufficient data, after migration, that we will end up with semantic table(s) with rows in the tens of thousands. So it can be bottleneck as is, since we bring out "suggested configurations" the scan will happen numerous times when we open up the model-config drilldown.

Given how often data migrations had annoyed us  in the past - I think it would be good to at least get some metrics and amend stuff where we can, to minimize things shifting around down the road.

mwdchang: I did some perf tests, with around 50k entries devoted to configurations (initials + params + observables) we are still okay-ish without indices.

With my tests, joining model_config to semantics
- without index is about 15-20 milliseconds
- with index is about 2-3 milliseconds

Will leave it with @dvince2 if this is something we should pursue now.

mwdchang (approved):

PR #3811, opened 2024-06-10
Function to create an AMR model structure from model configuration

blanchco:
# Description

* create AMR from config

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
```diff
@@ -342,6 +350,58 @@ public ResponseEntity<ResponseDeleted> deleteModelConfiguration(
...
+		for (final ParameterSemantic configParameter : configParameters) {
+			configParameterMap.put(configParameter.getReferenceId(), configParameter);
+		}
+
+		// Iterate through the list of ModelParameter objects
+		for (final ModelParameter modelParameter : modelParameters) {
+			// Look up the corresponding ConfigParameter in the map
+			final ParameterSemantic matchingConfigParameter = configParameterMap.get(modelParameter.getId());
+			if (matchingConfigParameter != null) {
+				// I'm not sure if we want the set the id, name, description here since we have that on model already
```
mwdchang, line 366: I don't think you want to set name/description. In fact one might question why you'd have that in the data structure ...

blanchco: we're getting rid of it those fields, wrote this last week before those changes were made

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
```diff
@@ -201,10 +207,12 @@ public ResponseEntity<Model> getConfiguredModel(
 			if (model.isEmpty()) {
 				throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("model.not-found"));
 			}
+			updateModelInitials(
+					model.get().getInitials(), modelConfiguration.get().getInitialSemanticList());
+			updateModelParameters(
+					model.get().getParameters(), modelConfiguration.get().getParameterSemanticList());
```
mwdchang, line 213: Strictly speaking, you want to do in order of
- parameters
- initials
- observables

Because initials depends on parameters and observables depends on initials and parameters. Also you seem to be missing observables.

blanchco: Sure Ill go in that order.  We weren't using observables in the model config drilldown on the front end so far which is why I hadn't created it, but I can add that in.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ObservableSemantic.java
```diff
@@ -22,12 +24,12 @@ public class ObservableSemantic extends Semantic {
...
 	private String referenceId;
 
-	@Type(JsonType.class)
-	@Column(columnDefinition = "json")
-	private State[] states;
+	private List<String> states;
 
 	private String expression;
 
+	private String expressionMathml;
```
mwdchang, line 31: I think you will need to sort out camel/snake transforms. I am not exactly sure how we are doing this now so pinging @dvince2

dvince2: Is this being read by an external service? or are they sending us snake-case?

blanchco: I think this should be fine, the new model config is all for our internal use so we can name these pretty much whatever we want and map to the correct values when we create an AMR from config.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
```diff
@@ -195,10 +206,12 @@ public ResponseEntity<Model> getConfiguredModel(
...
 			}
-
-			// TODO: Apply configuration to the Model here. Should this use ModelConfigurationLegacy??
-
-			throw new ResponseStatusException(HttpStatus.NOT_IMPLEMENTED);
+			updateModelParameters(
+					model.get().getParameters(), modelConfiguration.get().getParameterSemanticList());
+			updateModelInitials(
+					model.get().getInitials(), modelConfiguration.get().getInitialSemanticList());
+			updateModelObservables(model.get().getObservables(), modelConfiguration.get().getObservableSemanticList());
```
YohannParis, line 213: Why are we using `update` instead of `set` here?
A `model-configuration` should match the model that is tied to the `modelId` so there is no reasons to do some matching of information here? Or I am missing something?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
blanchco: Ill rename the function to set instead of update.

This jist of these functions are to set the semantics in an AMR structure from the model it is derived from.  The model semantics would have additional properties that we do not get from a model configuration (i.e. name, description, concept...).  I am just simply setting the certain properties that we have in the model configuration to the model (value, distribution, expression...).

mwdchang: @YohannParis fyi it isn't super clear that the config will match the model exactly, you may have extra observables and I don't think we have made it clear where observables should be defined. May be worth to clear up after the 1st around.

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelConfigurationController.java
YohannParis: @blanchco in that case you are right, what ever name you decide is good with me. No worries.
@mwdchang the Observable are set on the Model operator, and are port of the model object, no?

PR #3815, opened 2024-06-10
workflow house cleaning June

mwdchang:
### Summary
- Remove unused/unimplemented buttons: reset zoom and reset layout
- Remove send to front/back that are unimplemented
- Clean up auto-output propagation


packages/client/hmi-client/src/components/operator/tera-operator-header.vue
YohannParis, line 1: Is it coming back in the future, or can we remove this feature cleanly?

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -41,6 +41,7 @@
 				</div>
 				<Menu ref="optionsMenu" :model="optionsMenuItems" :popup="true" />
 				<div class="button-group">
+					<!--
```
YohannParis, line 44: Why not remove the code, like the functions below?

YohannParis (approved):
LGTM

Tom-Szendrey (approved):
Ty for the house cleaning

packages/client/hmi-client/src/components/operator/tera-operator-header.vue
mwdchang: Neil ask them to be hidden, but yeah good point, might as well. Easy enough to bring things back.

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -260,6 +260,12 @@ export function getPortLabel({ label, type, isOptional }: WorkflowPort) {
 	return portLabel;
 }
 
+// Checker for resource-operators (e.g. model, dataset) that automatically create an output
+// without needing to "run" the operator because we can drag them onto the canvas
+export function canPropagateResource(outputs: WorkflowOutput<any>[]) {
+	return _.isEmpty(outputs) || (outputs.length === 1 && !outputs[0].value);
+}
```
shawnyama, line 267: This pattern is used in non resource operators as well. The input is typically the first output we want to propagate. I use it in model-edit

shawnyama (approved):

packages/client/hmi-client/src/services/workflow.ts
mwdchang: I don't see any append-output-port in `tera-model-edit-node`

shawnyama: It's in the drilldown

mwdchang: Gotcha, I would say that is kind of a special case then - that you would propagate when opening the drilldown. Will keep a note on it and revisit.

PR #3816, opened 2024-06-10
Fix migration pagination logic

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/DataMigrationESToPG.java
```diff
@@ -149,7 +149,8 @@ Pair<Integer, Integer> migrateFromEsToPg(final ElasticsearchService elasticServi
 					}
 				}
 
-				if (Objects.equals(lastId, lastPagesLastId) || assets.size() < PAGE_SIZE) {
+				if (Objects.equals(lastId, lastPagesLastId)
+						|| resp.hits().hits().size() < PAGE_SIZE) {
```
dvince2, line 153: Does there need to be any null safety here?
if resp, or either call to .hits fails this would blow up

dvince2 (approved):

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/DataMigrationESToPG.java
kbirk: We already read the .hits().hits() directly. They seem to not ever be null even if no results come in.

PR #3825, opened 2024-06-11
Task runner log: check for null type and log

Tom-Szendrey:
# Description
Add a logger for when we do not have an event type found. 
This will help us update our enums accordingly and have a bit more information when this scenario happens

```
2024-06-11 10:38:19 [WARN] Event type not found for task: mira_task:amr_to_mmt [software.uncharted.terarium.hmiserver.service.tasks.TaskNotificationEventTypes:21]
2024-06-11 10:38:19 [INFO] Sending client event with type TASK_UNDEFINED_EVENT for task befe548e-adf3-4614-8eec-26378b0f9b4f  [software.uncharted.terarium.hmiserver.service.tasks.TaskService:452]
```

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskNotificationEventTypes.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskNotificationEventTypes.java
@@ -18,7 +18,8 @@ public class TaskNotificationEventTypes {
 		final ClientEventType eventType = clientEventTypes.get(taskName);
 		if (eventType == null){
 			log.warn("Event type not found for task: " + taskName);
+			return ClientEventType.TASK_UNDEFINED_EVENT;
 		}
-		return eventType == null ? ClientEventType.TASK_UNDEFINED_EVENT : eventType;
+		return eventType;
 	}
 }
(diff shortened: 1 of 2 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskNotificationEventTypes.java
```diff
@@ -14,6 +16,9 @@ public class TaskNotificationEventTypes {
 
 	public static ClientEventType getTypeFor(String taskName) {
 		final ClientEventType eventType = clientEventTypes.get(taskName);
+		if (eventType == null) {
+			log.warn("Event type not found for task: " + taskName);
+		}
 		return eventType == null ? ClientEventType.TASK_UNDEFINED_EVENT : eventType;
```
YohannParis, line 22: While you do the test, include the next line within 😄 
```suggestion
		if (eventType == null) {
			log.warn("Event type not found for task: " + taskName);
			return ClientEventType.TASK_UNDEFINED_EVENT
		}
		return eventType;
```

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskNotificationEventTypes.java
Tom-Szendrey: what a thinker

PR #3829, opened 2024-06-11
3764 research how we can save calibrated configurations from a dill file

### LINKED ISSUE #3764: research how we can save calibrated configurations from a dill file

dvince2:
# Description

* With the simulation ID we can access the dill file to pass on

Resolves #3764


YohannParis (commented):
That comment is *chef's kiss*

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
```diff
@@ -23,6 +23,9 @@ public class ModelConfiguration extends TerariumAsset {
 	private UUID calibrationRunId;
 	private UUID modelId;
 
+	/** This is "simulation" in the sense of our POJO. It actually corresponds to a pyciemss calibration */
+	private UUID simulationId;
```
mwdchang, line 27: Probably optional and/or tsoptional? Doe does it already default to null in hibernate land?

dvince2: oh good catch - yes it does

mwdchang (approved):

PR #3830, opened 2024-06-12
fix: Read states for intial metadata

shawnyama:
# Description

States metadata are modified on the model page, instead of initial metadata


blanchco (commented):
Since we are using state instead of initials for this data, the model configuration page will have to be changed according to use the state's name, description, unit etc

packages/client/hmi-client/src/components/model/tera-model-semantic-tables.vue
shawnyama, line 1: @mwdchang this has the `onUnmounted` that doesn't run once you close the model resource drilldown

mwdchang: checking it out ...

blanchco (approved):

YohannParis (approved):

PR #3831, opened 2024-06-12
Fix: Description from gollmcard to be saved and used existing functionality

Tom-Szendrey:
# Description
When enriching a model the mit card `card` was being added to the model's metadata. However the `gollmCard` was never being added.
This means that this ta4 tab would not have the information to display thus showing the `card` instead of the `gollmCard` at all times

https://github.com/DARPA-ASKEM/terarium/assets/17088680/345bd680-4332-4543-b878-545e3bf5fd66


blanchco (approved):
👍

kbirk (approved):

PR #3839, opened 2024-06-12
update client event service based on PanTera updates

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -63,28 +62,48 @@ public static class UserClientEvent<T> implements Serializable {
...
 	 * Connects a user to the SSE service
 	 *
 	 * @param user the user to connect
 	 * @return the emitter to send messages to the user
 	 */
 	public SseEmitter connect(final User user) {
-		final SseEmitter emitter = new SseEmitter();
-		if (!userIdToEmitter.containsKey(user.getId())) {
-			userIdToEmitter.put(user.getId(), new ArrayList<>());
+		synchronized (userIdToEmitters) {
```
bigglesandginger, line 72: Why are we synchonizing on the ConcurrentHashMap - the whole point of using a ConcurrentHashMap is that it will handle the locking for us

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -63,28 +62,48 @@ public static class UserClientEvent<T> implements Serializable {
...
+		}
+	}
+
+	/**
+	 * Disconnects a user from the sse service. Completes all emitters for the user and removes them
+	 *
+	 * @param user the user to disconnect
+	 */
+	public void disconnect(final User user) {
+		synchronized (userIdToEmitters) {
```
bigglesandginger, line 96: again should not be required

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -63,28 +62,48 @@ public static class UserClientEvent<T> implements Serializable {
...
+				emitter.complete();
+				synchronized (userIdToEmitters) {
+					emitters.remove(emitter);
+					if (emitters.isEmpty()) {
+						userIdToEmitters.remove(user.getId());
+					}
+				}
+			});
+			emitters.add(emitter);
+			userIdToEmitters.put(user.getId(), emitters);
```
bigglesandginger, line 85: We should probably be using a more thread safe data model instead of the ArrayList internally, perhaps the ConcurrentLinkedQueue<> would be a smarter choice within the ConcurrentHashMap<>.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -125,42 +144,38 @@ public <T> void sendToUser(final ClientEvent<T> event, final String userId) {
...
-				log.error("Error sending all users message to user {}", userId, e);
-			}
-		});
-
-		synchronized (userIdToEmitter) {
-			emittersToRemove.forEach((emitter) -> {
-				userIdToEmitter.get(userId).remove(emitter);
-				if (userIdToEmitter.get(userId).size() == 0) {
-					userIdToEmitter.remove(userId);
+			userIdToEmitters.forEach((userId, emitterList) -> {
```
bigglesandginger, line 164: To be consistent this should be using the synchronize

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
```diff
@@ -171,24 +186,41 @@ private void send(final Object message, final List<SseEmitter> emitterList, fina
...
 
-		final String userId = messageJson.at("/userId").asText();
-		synchronized (userIdToEmitter) {
-			final List<SseEmitter> emitterList = userIdToEmitter.get(userId);
+		synchronized (userIdToEmitters) {
+			final String userId = messageJson.at("/userId").asText();
+			final List<SseEmitter> emitterList = userIdToEmitters.get(userId);
 			if (emitterList != null) {
-				send(messageJson.at("/event"), emitterList, userId);
+				final Set<SseEmitter> emittersToRemove = new HashSet<>();
```
bigglesandginger, line 209: Again to be consistent this should be using synchronize (either all should or all should not)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
dgauldie: fixed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
dgauldie: fixed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
dgauldie: fixed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
dgauldie: fixed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ClientEventService.java
dgauldie: fixed

bigglesandginger (approved):

bigglesandginger:
This should be better, also I realise in my review I missed the anonymous functions inside the emitters which, if going with the locking approach, should also have been locking as access to a concurrent variable should be managed with locks.  Though personally I dislike the Java 0.x synchronize method and prefer Java 1.6 Locks as they allow for differentation between read and write and more complicated senarios.  Synchronize is a total hack of overloading every single object in Javaand should never have been allowed into the language... hurrumph.

YohannParis (approved):

PR #3877, opened 2024-06-14
Added notification support for simulate operations

jryu01:
# Description

https://github.com/DARPA-ASKEM/terarium/assets/3279838/ecbba90f-5e14-4436-8712-ec5eb0303ae3

https://github.com/DARPA-ASKEM/terarium/assets/3279838/b1b01a7d-090d-4350-8a85-80012eb2d6c7

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -135,7 +136,13 @@ watch(
...
 					inferred_parameters: id
 				},
-				engine: 'ciemss'
+				engine: 'ciemss',
+				metadata: {
+					workflowId: props.node.workflowId,
+					workflowName: useProjects().getAssetName(props.node.workflowId),
+					nodeId: props.node.id,
+					nodeName: props.node.displayName
+				}
```
mwdchang, line 145: I am wondering if it will be better to have the metadata as a sibling rather than as a part of the simulation request. that is to say, rather than:

```
function makeForecastJobCiemss(simulationParam: SimulationRequest)
```

we have this as the functional signature

```
function makeForecastJobCiemss(simulationParam: SimulationRequest, metadata?: any)
```

This should make things a bit cleaner in my opinion:
- You don't need to add objects in, only to take it out later a second later because downstream has schema checks or whatnot
- Even if we can pass metadata down, we do not leak implementation detail to services that do not care about workflow-node-ids
- Less rippling effects due to typing changes

jryu01: Yeah I think that's ideal but I haven't find a simple enough way to include the metadata object as a separate request param with `RequestParam` annotation. So I included it in the request body object. I'll look a bit more to see if theres's a work around.

jryu01: Or we can just restructure the request body like following and use `@RequestBody` without additional request param
```
{
  metadata: any
  simulationRequest: SimulationRequest
}
```

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -123,6 +124,12 @@ watch(
 				extra: {
 					num_samples: props.node.state.extra.numIterations,
 					inferred_parameters: id
+				},
+				metadata: {
+					workflowId: props.node.workflowId,
+					workflowName: useProjects().getAssetName(props.node.workflowId),
+					nodeId: props.node.id,
+					nodeName: props.node.displayName
```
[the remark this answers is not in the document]
jryu01: Done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/notification/NotificationGroupInstance.java
```diff
@@ -107,10 +117,22 @@ public void sendMessage(final String msg) {
...
 
+	public void sendMessage(final String msg, ProgressState state) {
+		sendNotification(msg, null, estimateT(), state);
+	}
+
 	public void sendFinalMessage(final String msg) {
 		sendNotification(msg, null, 1.0, ProgressState.COMPLETE);
 	}
 
+	public void sendFinalMessage(final String msg, ProgressState state) {
```
mwdchang, line 128: Stylistically seems like `ProgressState` should be final as well? ditto for the sendMessage method.

jryu01: Good catch! Fixed it.

YohannParis (approved):

PR #3885, opened 2024-06-14
add default config if doesnt exist

blanchco:
# Description
* Fixes an issue where default config doesnt show up in model config node.  This is because the default configuration is created when opening the model node.  I now create one when opening the model configuration node.
* this is a temporary fix, I plan on creating a default config in the backend when a model is created in the ModelConfigurationLegacy -> ModelConfiguration upgrade


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -691,8 +692,14 @@ const initialize = async () => {
 	const state = props.node.state;
 	const modelId = props.node.inputs[0].value?.[0];
 	if (!modelId) return;
-	fetchConfigurations(modelId);
+	await fetchConfigurations(modelId);
+
 	model.value = await getModel(modelId);
+	if (suggestedConfigurationContext.value.tableData.length === 0 && model.value) {
+		await addDefaultConfiguration(model.value);
+		fetchConfigurations(modelId);
```
YohannParis, line 700: could this be a then() call instead of an await?

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
blanchco: it could be, does the team like promise chaining better now? I like async/await

YohannParis: It's not about liking one or the other. It's about, do we need to block the main thread for this?

blanchco: ahh i see, ya .then would work better here

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config.vue
```diff
@@ -691,8 +692,15 @@ const initialize = async () => {
...
 	if (!modelId) return;
-	fetchConfigurations(modelId);
+	await fetchConfigurations(modelId);
+
 	model.value = await getModel(modelId);
+	if (suggestedConfigurationContext.value.tableData.length === 0 && model.value) {
+		addDefaultConfiguration(model.value).then(() => {
+			fetchConfigurations(modelId);
+		});
+	}
```
mwdchang, line 702: Please mark the section(s) clearly if the solution is meant to be temporary, so other readers would know this is, and whether or not to build on it.

mwdchang (approved):

PR #3889, opened 2024-06-14
fix drilldown styles

blanchco:
# Description

* With the recent changes to the model some css changes broke scrolling on some drilldown components.  This should fix it.
* surrounding `<tera-model>` with `<tera-drilldown-section>` allows unmount to be called

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model/tera-model-workflow-wrapper.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model/tera-model-workflow-wrapper.vue
@@ -4,12 +4,14 @@
 		@on-close-clicked="emit('close')"
 		@update-state="(state: any) => emit('update-state', state)"
 	>
-		<tera-model
-			v-if="node.state.modelId"
-			:asset-id="node.state.modelId"
-			@new-model-configuration="refreshModelNode"
-			@update-model-configuration="refreshModelNode"
-		/>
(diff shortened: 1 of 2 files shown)


shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/model/tera-model-workflow-wrapper.vue
```diff
@@ -4,12 +4,14 @@
...
-			@update-model-configuration="refreshModelNode"
-		/>
+		<tera-drilldown-section>
+			<tera-model
+				v-if="node.state.modelId"
+				:asset-id="node.state.modelId"
+				@new-model-configuration="refreshModelNode"
+				@update-model-configuration="refreshModelNode"
+			/>
+		</tera-drilldown-section>
```
mwdchang, line 14: @shawnyama @blanchco we changed this recently to remove the parent-element to workaround the weird unmount event-propagation issue. Is this re-introducing the problem by adding in the tera-drilldown-section?

blanchco: I tested it out beforehand, since `<tera-drilldown-section>` is a component the lifecycle functions work for some reason 🤷 rather than if it were an HTML element

YohannParis (approved):

PR #3893, opened 2024-06-17
hide unused operators

mwdchang:
### Summary
Hide operators that are no longer in active use
- model transformer
- decapodes
- model-coupling

### Notes
Existing workflows with the operators mentioned above may fail.


Tom-Szendrey (approved):
Im not sure of the projects we want to keep in prod.
If they contain workflows with these operators will they fail to open?

YohannParis (approved):
Great, could you remove the SciML for now too?

PR #3895, opened 2024-06-17
Fix local staging configs

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -346,18 +350,21 @@ public ResponseEntity<TaskResponse> createConfigFromDatasetTask(
 						datasets.add(datasetText.get().replaceAll("(?<!\\\\)\\n", Matcher.quoteReplacement("\\\\n")));
 					}
 				} catch (final Exception e) {
+					log.warn("Unable to fetch dataset files", e);
```
dvince2, line 353: Also these three new errors feel like `error`s as well

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -501,17 +510,17 @@ public ResponseEntity<TaskResponse> createCompareModelsTask(
 		try {
 			resp = taskService.runTask(mode, req);
 		} catch (final JsonProcessingException e) {
-			log.error("Unable to serialize input: {}", e.getMessage());
+			log.error("Unable to serialize input", e);
 			throw new ResponseStatusException(
 					HttpStatus.INTERNAL_SERVER_ERROR, messages.get("task.gollm.json-processing"));
 		} catch (final TimeoutException e) {
-			log.error("Timeout while waiting for task response: {}", e.getMessage());
+			log.warn("Timeout while waiting for task response", e);
```
dvince2, line 517: ditto for these and a bunch more

dvince2 (approved):
The configs themselves are good.

There are a bunch of error message changes in this PR, and a lot of `warnings` that should be `errors` as they're a failure case that kick back to the client

YohannParis (approved):

PR #3925, opened 2024-06-19
3812 Added GoLLM summary endpoint

### LINKED ISSUE #3812: [TASK]: Create an operator summarization endpoint
### Summary
Add a summarization feature for workflow operators. this will allow us to better experiment with prompting and summarization of workflows.

This has two parts across two different github repos:

#### Augmenting a summarization tool to GoLLL (https://github.com/DARPA-ASKEM/gollm)
Right now GoLLM is serving specific features aimed at AMR and datasets, we want to have a more generic completion task that is closer to how one would work with a vanilla LLM model.  Per Julian it would look very close to this existing function:

https://github.com/DARPA-ASKEM/GoLLM/blob/468f7863d0d63f005d6dd1564cfadaecdbe04180/gollm/openai/tool_utils.py#L83

We probably do not need to split into query and chunk sections, since the "query" is always to summarize the text.

#### Add a taskrunner to invoke said tool
- Add a summarization wrapper taskrunner python script for the above
- Add an API endpoint to invoke the summarization taskrunner

jryu01:
# Description

Created GoLlm `generate-summary` endpoint

Resolves #3812


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -517,6 +522,67 @@ public ResponseEntity<TaskResponse> createCompareModelsTask(
...
+		final TaskResponse resp;
+		try {
+			resp = taskService.runTask(mode, req);
+		} catch (final JsonProcessingException e) {
+			log.error("Unable to serialize input: {}", e.getMessage());
+			throw new ResponseStatusException(
+					HttpStatus.INTERNAL_SERVER_ERROR, messages.get("task.gollm.json-processing"));
+		} catch (final TimeoutException e) {
+			log.error("Timeout while waiting for task response: {}", e.getMessage());
+			throw new ResponseStatusException(HttpStatus.SERVICE_UNAVAILABLE, messages.get("task.gollm.timeout"));
```
dvince2, line 573: A few of these statuses are missing from the above swagger api docs - 503, 422

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/GenerateSummaryHandler.java
```diff
@@ -0,0 +1,16 @@
+package software.uncharted.terarium.hmiserver.service.tasks;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+// @Slf4j
+public class GenerateSummaryHandler extends TaskResponseHandler {
+	public static final String NAME = "gollm_task:generate_summary";
```
dvince2, line 10: pedantically, this can be `private` :D

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/GenerateSummaryHandler.java
```diff
@@ -0,0 +1,16 @@
+package software.uncharted.terarium.hmiserver.service.tasks;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+// @Slf4j
+public class GenerateSummaryHandler extends TaskResponseHandler {
```
mwdchang, line 9: Just a reminder here that we will want to have a discussion about the handler logic here as we will want to track the results coming back in Postgres table (e.g. uniqueId => summary_text).

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/GenerateSummaryHandler.java
jryu01: It looks like it needs to be public since it's used like `GenerateSummaryHandler.Name` without creating an instance of it in GoLLM Controller.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
jryu01: I just updated the swagger docs

PR #3929, opened 2024-06-20
Task: optimize update Forecast should run twice.

Tom-Szendrey:
# Description
Updating the state to allow for two forecast runs. 
This will be later utilized in the output graphs.
In the meantime everything should continue to use the post optimized forecast to prevent functional changes.

Also add on-focus-out to tera-input


packages/client/hmi-client/src/components/widgets/tera-input.vue
```diff
@@ -28,7 +29,7 @@ const props = defineProps<{
 	type?: InputTypeHTMLAttribute | 'nist';
 }>();
 
-const emit = defineEmits(['update:modelValue']);
+const emit = defineEmits(['update:modelValue', 'onFocusOut']);
```
shawnyama, line 32:
```suggestion
const emit = defineEmits(['update:modelValue', 'on-focus-out']);
```

shawnyama (approved):
lgtm

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -51,8 +51,10 @@ export interface OptimizeCiemssOperationState extends BaseState {
 	// Charts + Outputs:
 	chartConfigs: string[][];
 	inProgressOptimizeId: string;
-	inProgressForecastId: string;
-	forecastRunId: string;
+	inProgressPreForecastId: string;
+	preForecastRunId: string;
+	inProgressPostForecastId: string;
+	postForecastRunId: string;
```
mwdchang, line 57: Your choice of names may be a bit confusing? Isn't it more like [pre-optimize, post-optimize], instead of [pre-forecast, post-forecast]?

packages/client/hmi-client/src/components/widgets/tera-input.vue
```diff
@@ -82,7 +83,7 @@ onMounted(() => {
 const unmask = () => {
 	// convert back to a number when finished
 	if (isNistType && !getErrorMessage.value) {
-		emit('update:modelValue', nistToNumber(maskedValue.value));
+		emit('update:model-value', nistToNumber(maskedValue.value));
```
mwdchang, line 86: Do we need to change all other places that use tera-input?

Tom-Szendrey: Others were already referring to this as `model-value` I checked!

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
Tom-Szendrey: hmm yes but they are both IDs for forecasts.
[pre-optimized-forecast] is a bit wordy.
Maybe firstForecast, and secondForecast? Im not certain but am open to suggestions!

mwdchang: Oh I think I am reading them as:

```
preForecast-runId
postForecast-runId
```
but you are implying
```
pre-forecastRunId
post-forecastRunId
```

Tom-Szendrey: Correct

mwdchang (approved):

PR #3930, opened 2024-06-20
improved mira controller error messages

dgauldie:
# Description

Resolves #(issue)


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -262,21 +315,14 @@ public ResponseEntity<List<DKG>> search(
...
-			final int status = e.status() >= 400 ? e.status() : 500;
-			log.error(error, e);
-			throw new ResponseStatusException(org.springframework.http.HttpStatus.valueOf(status), error);
-		} catch (final Exception e) {
-			log.error("Unable to fetch DKG", e);
-			return ResponseEntity.internalServerError().build();
+			throw handleMiraFeignException(e, "concepts", "query", q, "mira.concepts.bad-query");
 		}
+
+		return new ResponseEntity(response.getBody(), response.getStatusCode());
```
dvince2, line 325: Is there any risk here of returning a 500 or something with the body? I suppose that wouldnt matter?

dvince2 (approved):
A few `log.error`s were downgraded to `log.warn`, but still throw an exception and are a failure case. We may want these to be errors? but over all good.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
dgauldie: a 500 should be thrown as a feignException. If we get to this point, we have a 2xx response. I didn't want to convert to a 200 and lose that.

dgauldie:
> A few `log.error`s were downgraded to `log.warn`, but still throw an exception and are a failure case. We may want these to be errors? but over all good.

I was kinda going with 4xx and 503 errors are warnings... as they are still failures of the controllers but associated with problems of the client and of other services. 500 errors are errors.

PR #3934, opened 2024-06-20
feat: observables in model diagram

shawnyama:
# Description
- A new `observableSummary` object is used to create observable nodes and edges in `mira.ts`

### graph.ts
- In `graph.ts` I first set nodes and edges for the states and transitions (skipping observables)
- Once they are set I can determine where the "right" of the graph is so I can start setting the positions of the observable nodes
- Then I draw custom edges from the observable nodes to the states that they are attached to
- I also did some cleanups in `graph.ts` such as using `forEach` instead of `for of`

### petrinet-renderer.ts
- added observable nodes/edges graphics here
- added `mouse-enter` and `mouse-leave` events to focus on observable nodes/edges when the node is hovered on

https://github.com/DARPA-ASKEM/terarium/assets/28237258/562043ef-8548-4abf-a2f6-6f12b03c0ebb


YohannParis (approved):

packages/client/hmi-client/src/model-representation/petrinet/petrinet-renderer.ts
```diff
@@ -9,16 +10,13 @@ export interface NodeData {
 	expression?: string;
 	strataType?: string;
 	isStratified?: boolean;
+	value?: any;
```
mwdchang, line 13: Are you using this `value?: any` you introduced? Seems to be no?

shawnyama: I am using it in the `mode-mouse-enter` here. Where I extract the references. I'll adjust that, there doesn't seem to be a need for an entire value property

shawnyama: should be more clear now

mwdchang (approved):

PR #3958, opened 2024-06-21
decouple-read-+-write-interactions

asylves1:
# Description

When the user is on the workflow, for some of the model nodes a GET and POST is called together. I have added a condition check to ensure that the user has edit rights to allow the POST call.

Resolves #(issue)


packages/client/hmi-client/src/composables/project.ts
```diff
@@ -250,6 +250,14 @@ export function useProjects() {
 		return created;
 	}
 
+	function hasEditPermission() {
+		const project = useProjects().activeProject.value;
+		if (project != null) {
+			return ['creator', 'writer'].includes(project.userPermission ?? '');
+		}
+		return false;
+	}
```
mwdchang, line 259: The usage of this function will cause various things to silently fail in the application - probably a better approach is to prevent users from doing things they are not able to do. Let's check in with the design team to see where they land in terms of this permission-related interactions.

In the meantime, if it isn't too noisy, you may want to consider putting a console.warn here.

mwdchang (approved):

packages/client/hmi-client/src/composables/project.ts
asylves1: added warning message

PR #3959, opened 2024-06-21
New project thumbnail

YohannParis:
# Description

* Remove P5 generated image
* Added five new thumbnails images from the design-team
* Updated the Java model of Project to inlcude a `thumbnail` string
* Updated the hmi-client to offer the thumbnail
* Updated the hmi-client services


shawnyama (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -265,6 +272,7 @@ public ResponseEntity<List<Project>> getProjects(
 			}
 		});
 
+		log.info("[getAllProject] End {}", System.currentTimeMillis());
```
mwdchang, line 275: What is up with all the `log.info` timing stuff? Are they for a different task?

YohannParis: Like I said in the description: this is to found out were the issue is with getAllProjects. I will remove them soon.

mwdchang (approved):

PR #3960, opened 2024-06-21
Fix model alignment

kbirk:
This is if we are keeping it. We can just remove it as well.


packages/client/hmi-client/src/components/widgets/tera-related-documents.vue
```diff
@@ -201,13 +200,7 @@ const sendForExtractions = async () => {
...
 	if (props.assetType === AssetType.Model && selectedResourceId) {
 		await extractVariables(selectedResourceId, [props.assetId]);
-		const isAligned = await alignModel(props.assetId, selectedResourceId);
-		if (isAligned) {
-			logger.success('Model aligned after variable extraction.');
-			emit('enriched');
-		} else {
-			logger.warn('Model was not aligned after variable extraction. Please try again.');
-		}
+		emit('enriched');
```
kbirk, line 203: I dunno if this should even be here since that call is async and just queues the work

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -440,7 +440,7 @@ private DocumentAsset runVariableExtraction(
 				for (final UUID modelId : modelIds) {
 					notificationInterface.sendMessage("Attempting to align models for model: " + modelId);
 					try {
-						alignAMR(documentId, modelId, hasWritePermission).get();
```
kbirk, line 443: This was running async and not included in the transaction, so the attributes in the metadata would not be visible.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -212,7 +212,8 @@ public Future<DocumentAsset> extractPDF(
 							final ObjectMapper objectMapper = new ObjectMapper();
 
 							final JsonNode rootNode = objectMapper.readTree(bytes);
-							if (rootNode instanceof final ArrayNode arrayNode) {
+							if (rootNode instanceof ArrayNode) {
```
dvince2, line 215: Just an FYI - java 17 has some neat improvements here that let you get rid of the casting statement and just do something like:



```
if (rootNode instanceof ArrayNode arrayNode){
   for (final JsonNode record : arrayNode)
   
   ...
}
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -498,84 +499,96 @@ public Future<Model> alignAMR(
...
 						error + ": " + e.getMessage());
-			} catch (final Exception e) {
-				final String error = "SKEMA link_amr request from document: " + documentId + " failed.";
-				log.error(error, e);
-				notificationInterface.sendError(error + " — " + e.getMessage());
-				throw new ResponseStatusException(org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR, error);
 			}
-		});
+			if (!res.getStatusCode().is2xxSuccessful()) {
+				throw new ResponseStatusException(res.getStatusCode(), "Unable to link AMR file");
```
dvince2, line 558: In the past some of the TA3 services have returned -1 as an HTTP error code. I see there is safety for this above, but, we may want it here too.

dvince2 (approved):

PR #3967, opened 2024-06-24
Update workflow summarization

mwdchang:
### Summary
For the most part, this PR adds/updates the user interface plumbings to handle summary generation for operator outputs:
- Remove the operator-output summary from the drilldown/drilldown-header
- Update tera-operator-output-summary to use the new summary object
- Example summary generation with simulate-ciemss

image

### Testing
Only works with simulate-sciemss
- Run a simulation, it will generate a summary after simulation is complete
- You can edit and overwrite the summary text


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -105,20 +108,46 @@ const chartProxy = chartActionsProxy(props.node, (state: SimulateCiemssOperation
...
+- timespan: ${JSON.stringify(state.currentTimespan)}
+
+The output has these metrics at the start:
+- ${JSON.stringify(start)}
+
+The output has these metrics at the end:
+- ${JSON.stringify(end)}
+
+Provide a summary in 100 words or less.
+	`;
```
YohannParis, line 138: Why this is done on the front-end?

packages/client/hmi-client/src/components/operator/tera-operator-output-summary.vue
```diff
@@ -3,98 +3,129 @@
...
-	{ immediate: true, deep: true }
+	{ immediate: true }
 );
 </script>
 
 <style scoped>
 section {
 	display: flex;
-	flex: 1;
+	/* flex: 1; */
```
YohannParis, line 128:
```suggestion
```

packages/client/hmi-client/src/components/operator/tera-operator-output-summary.vue
```diff
@@ -3,98 +3,129 @@
...
-	() => activeOutput?.value?.summaryHasBeenEdited !== true && activeOutput?.value?.summary === ''
-);
-const isGenerated = computed(
-	() =>
-		activeOutput?.value?.summaryHasBeenEdited !== true &&
-		(activeOutput?.value?.summary?.length ?? 0) > 0
-);
+/**
+ * It can take some time (30-40 seconds) for generated summary to appear
+ * */
```
YohannParis, line 77: Because the summary is done via the gollm-taskrunner, why not using the notifications?

YohannParis (approved):

packages/client/hmi-client/src/components/operator/tera-operator-output-summary.vue
mwdchang: Notification should be the end-product of some long running process, it should not be used as a bridge. For example I think the following model would be incorrect:

```
Start generating summary ==> [Notification bus] ==> Update summary
```

This can introduce many problems like a self perpetuating loop, and a bit harder to tracking thigns.

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
mwdchang: Because the API-side does not know about operators, and that was sort of intentional by design.

The original conception of the workflow operators was that you can extend it, eg adding a new operator, without needing to modify the core Terarium's Java code, and that the new operator is self-sufficient in someway. 

So you can do something like:
```
import { Workflow } from '@terarium/workflow`
```
and you can start building your own thing, the API provides some structure for nodes/edges but does not dictate the logic.

TLDR: the hope was to make this reusable and domain agnostic, to that end we have succeeded in some areas and kind of murky in others. But we have been following this pattern so far and haven't had any major problems yet.

packages/client/hmi-client/src/components/operator/tera-operator-output-summary.vue
jryu01: Instead of polling, you can listen for the SEE event for the task response update which isn't necessarily same as the notification. 
So it would be like
```
After generating summary, 
1. Try fetching summary ==> if not ready, listen for SSE event ==> on success event, fetch summary object again ( or use output from the task response)  
2.  Try fetching summary ==> if ready, just use the summary. 
```
But I think in this case, ^ requires more logic with a bit more complexity than just doing simple polling.

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
YohannParis: That makes complete sense!

jryu01 (approved):
LGTM

packages/client/hmi-client/src/components/operator/tera-operator-output-summary.vue
mwdchang: @jryu01 you may have just volunteered yourself

There are a number of edge cases on top of what you had mentioned, technically speaking you can invoke the next run before the first run had finished generating a summary, so you end up tracking multiple signals. 
- Invoke run ... call it A
- A's summary is generating
- Invoke 2nd run, call it B
- B's summary is generating
- B summary finishes
- A summary finishes

PR #3968, opened 2024-06-24
Model Configuration Sidebar

blanchco:
# Description
* adding a sidebar to the drilldown
* Added some simple usage from the design as a first pass of the sidebar such as using a config, deleting a config, downloading a config
* Some more advanced design is not yet implemented (default values, calibrated models, sorting) but I figure this is a good first pass

https://github.com/DARPA-ASKEM/terarium/assets/33158416/d0f7ae79-894e-470b-8a3d-1a1e9a17b5df


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-configuration-item.vue
```diff
@@ -0,0 +1,110 @@
...
+	}
+]);
+
+const toggleContextMenu = (event) => {
+	contextMenu.value.toggle(event);
+};
+
+const onDeleteConfiguration = () => {
+	confirm.require({
+		message: 'Are you sure you want to delete this configuration?',
```
mwdchang, line 71: Displaying the model-configuration's name would be nice here, for confirmation purposes.

packages/client/hmi-client/src/components/widgets/tera-slider.vue
```diff
@@ -60,6 +60,9 @@ const sidePanelTabStyle = computed(
 </script>
 
 <style scoped>
+aside {
+	position: relative;
+}
```
mwdchang, line 65: Have we checked if there are any adverse impact of this change? tera-slider is used in tera-slider-panel, which is used across several places in the application.

blanchco: Yup it should all be working as normal - I made a small change to the `tera-preview-panel` since it was breaking there, but it should all be good now.

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-configuration-item.vue
blanchco: done 👍

mwdchang (approved):

dvince2 (approved):

PR #3970, opened 2024-06-24
PR feedback

dvince2:
# Description

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SummaryController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SummaryController.java
@@ -51,9 +51,9 @@ public class SummaryController {
 
 	final CurrentUserService currentUserService;
 
-	@GetMapping
+	@PostMapping("/search")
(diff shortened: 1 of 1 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SummaryController.java
```diff
@@ -51,9 +51,9 @@ public class SummaryController {
 
 	final CurrentUserService currentUserService;
 
-	@GetMapping
+	@PostMapping("/search")
```
YohannParis, line 54: How is that a search, this does not accept any query?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SummaryController.java
dvince2: Its taking in a list of IDs to search for in the body. @mwdchang and I landed on `search` for the operation

mwdchang (approved):

PR #3980, opened 2024-06-25
fix precision

blanchco:
# Description
* Fixes an issue where JS math would lead to many precision points, this may be an issue for very small numbers.
* See attached bug ticket for issue

https://github.com/DARPA-ASKEM/terarium/assets/33158416/668a5f85-6b2b-464f-b058-d4fec5a14148

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table.vue
@@ -209,9 +209,11 @@ const onUpdateDistributions = () => {
 				id: paramId,
 				distribution: {
 					type: uncertaintyType.value,
+					// A way to get around the floating point precision issue is to set a fixed number of decimal places and parse as a float
+					// This will be an issue for adding uncertainty to very small numbers, but I think 8 decimal points should do
 					parameters: {
-						minimum: v - delta,
-						maximum: v + delta
+						minimum: parseFloat((v - delta).toFixed(8)),
+						maximum: parseFloat((v + delta).toFixed(8))
 					}
 				}
 			};


packages/client/hmi-client/src/components/model/petrinet/tera-parameter-table.vue
```diff
@@ -209,9 +209,11 @@ const onUpdateDistributions = () => {
 				id: paramId,
 				distribution: {
 					type: uncertaintyType.value,
+					// A way to get around the floating point precision issue is to set a fixed number of decimal places and parse as a float
+					// This will be an issue for adding uncertainty to very small numbers, but I think 8 decimal points should do
 					parameters: {
-						minimum: v - delta,
-						maximum: v + delta
+						minimum: parseFloat((v - delta).toFixed(8)),
+						maximum: parseFloat((v + delta).toFixed(8))
```
mwdchang, line 216: I am not sure what is the scope here, because floating point precision issues will happen all over the place, and across almost all languages that follow IEEE-754 will exhibit this symptom. So not just distributions. 

Almost all the operators except for TA1 can result in numbers like:
- 3.000000000000000001
- 2.999999999999999998

YohannParis: Should we save the original value and the distribution instead of min, max, and only display those value in the UI as needed, and figure out how TA3 wants to consume them?

Like 
```ts
parameters: {
    value: v,
    delta: delta,
}
```

Then on the front-end:
```ts
const minimum = parseFloat((v - delta).toFixed(8))
const maximum = parseFloat((v + delta).toFixed(8))
```

YohannParis (approved):

PR #3983, opened 2024-06-25
fix for dragging bug

asylves1:
# Description

Unable to drag select text from add note input field on a node. This causes the note to be dragged

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -85,6 +85,7 @@
 					@remove-operator="(event) => removeNode(event)"
 					@duplicate-branch="duplicateBranch(node.id)"
 					@remove-edges="removeEdges"
+					@editing-note="editingNote"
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -301,6 +302,11 @@ const toggleOptionsMenu = (event) => {
 	optionsMenu.value.toggle(event);
 };
 
+const isEditingNote = ref(false);
+const editingNote = (isEditing) => {
+	isEditingNote.value = isEditingNote.value ? isEditing : false;
```
YohannParis, line 307: What is this test? Why not?
```suggestion
	isEditingNote.value = isEditing ?? false;
```

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
asylves1: code changed

shawnyama (approved):

PR #3984, opened 2024-06-25
create default config on node connection

blanchco:
# Description

* Creates a default config if there are no configs present on connect a model to a model config node


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -27,6 +29,23 @@ const emit = defineEmits(['open-drilldown', 'append-input-port', 'update-state']
 const modelInput = props.node.inputs.find((input) => input.type === 'modelId');
 const isModelInputConnected = computed(() => modelInput?.status === WorkflowPortStatus.CONNECTED);
 
+watch(
+	() => modelInput?.value,
+	async () => {
+		if (modelInput?.value) {
+			const modelId = modelInput.value[0];
+			const modelConfigurations = await getModelConfigurations(modelId);
```
dvince2, line 37: As something sort of unrelated we may want to rename this function - i just did a deep dive in the code to look at `getModelConfigurations` because I assumed it was a model config endpoint call. Perhaps something like `getModelConfigurationsForModel` or something?

dvince2 (approved):

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -52,6 +52,14 @@ export const getAsConfiguredModel = async (
 	return response?.data ?? null;
 };
 
+export const postAsConfiguredModel = async (model: Model): Promise<ModelConfiguration> => {
+	const response = await API.post<ModelConfiguration>(
+		`model-configurations/as-configured-model/`,
```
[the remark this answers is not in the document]
blanchco: this endpoint creates a configuration from a Model.  If no name and description are passed it will be a default config

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
blanchco: ya I can definitely rename

packages/client/hmi-client/src/services/model-configurations.ts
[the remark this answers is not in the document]
YohannParis: I meant, if the AMR is not configured?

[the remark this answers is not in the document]
blanchco: I might not be fully understanding, by an AMR that is not configured would that mean that there are no values on the base AMR?  In that case we default the values to constant 0's.  Here's the [function](https://github.com/DARPA-ASKEM/terarium/blob/2ec9faadcab92176fab1c921e00d707839ddeb77/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ModelConfigurationService.java#L73) if you wanna look

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -27,6 +29,23 @@ const emit = defineEmits(['open-drilldown', 'append-input-port', 'update-state']
...
+			if (isEmpty(modelConfigurations)) {
+				// Create a model configuration if it does not exist
+				const model = await getModel(modelId);
+				if (model) {
+					postAsConfiguredModel(model);
+				}
+			}
+		}
+	}
+);
```
mwdchang, line 47: I don't think this watcher is working properly, cause it is watching a non-reactive variable?

blanchco: IDK how it was working in the first place to be honest

blanchco: should be okay now...

mwdchang (approved):

PR #3988, opened 2024-06-26
fix: model edit initialization

shawnyama:
# Description

- Model edit drilldown appears properly
- Beaker timeout errors are avoided by not generating a summary for the output

image


Tom-Szendrey (approved):
Fine enough.

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -131,6 +131,9 @@ const handleTabChange = (event: TabViewChangeEvent) => {
 	selectedViewIndex.value = event.index;
 };
 
+const selectedTab = computed(() => views.value[selectedViewIndex.value]);
+defineExpose({ selectedTab });
```
mwdchang, line 135: Can you elaborate a bit why this defineExpose is desired here? 

This pattern kind of breaks away from the notion of a one-way dataflow, normally we would like to limit the exposure of the children to their parent, as it generally makes things harder to reason with and to reuse components.

mwdchang (approved):

PR #4010, opened 2024-06-27
Copy edits for error messages

mecrouch:
Copy edits to Terarium error messages easier to understand. Below are some suggestions for adding to and maintaining these going forward.

## General tips
- Use informal conversational language wherever possible.
- Use contractions ("can't" instead of "cannot").
- Use active voice instead of passive (instead of "the service **was terminated**…" try "**We terminated** the service…").
- Avoid software development jargon that users may not understand ("UUID," "curies").
- Don't introduce backend terms that users never see in the UI. For example, avoid terms like "artifacts" and "assets" if users only see "resources" in the UI.
- Consider whether it's necessary to name a backend service that users never see. Instead of referencing "SKEMA," determine whether it would be clearer to refer to our "model service."

## Terms

<dl>
<dt>Curie</dt>
<dd>Not typically familiar to an end user. Replace with "ID" or "unique ID."</dd>

<dt>Error</dt>
<dd>Avoid negative terms. Simply state the results of the error. Instead of "An error occurred while trying to set project permissions…" try "We couldn't set project permissions…"</dd>

<dt>Fail</dt>
<dd>Avoid negative terms. Simply state the results of the failure. Instead of "the service failed to perform the task…" try "the service couldn't perform the task…"</dd>

<dt>Please</dt>
<dd>Only use "please" when we're asking the user to take an action as a result of a system error. If the error is a result of a user action (for example, trying to upload a file that's too big), state the corrective action without saying "please."</dd>

<dt>Provide</dt>
<dd>Avoid this verb when describing user actions. Be more descriptive about the action that triggered the error message, such as "selected" or "entered."</dd>

<dt>Return</dt>
<dd>Avoid this verb when describing search results or application outputs. Instead of "Our service didn't return any results…" try "Our service didn't find any results."</dd>

<dt>Request</dt>
<dd>Avoid this verb when describing user actions. Be more descriptive about the action that triggered the error message, such as "selected" or "entered."</dd>

<dt>Supply</dt>
<dd>Avoid this verb when describing user actions. Be more descriptive about the action that triggered the error message, such as "selected" or "entered."</dd>

<dt>Terarium</dt>
<dd>Don't repeat the name of the application in error messages; users know what tool they're using. Replace with first person plural (instead of "Terarium's model service is unavailable" try "Our model service is unavailable.")</dd>

<dt>Unable</dt>
<dd>Instead of saying we "were unable to…," simply state "we couldn't…"</dd>

<dt>UUID</dt>
<dd>Not typically familiar to an end user. Replace with "ID" or "unique ID."</dd>
</dl>

### DIFF
--- a/packages/server/src/main/resources/messages.properties
+++ b/packages/server/src/main/resources/messages.properties
@@ -1,88 +1,88 @@
 generic.unknown = Oops! Something went wrong. Please try again later.
-generic.invalid-uuid = The ID you supplied is not a valid UUID. Please ensure that the ID is correct.
-generic.io-error.read = An error occurred while trying to read data. Please try again later.
-generic.io-error.write = An error occurred while trying to write data. Please try again later.
+generic.invalid-uuid = The ID you selected isn't valid. Please verify that the ID is correct and try again.
+generic.io-error.read = We couldn't read your data. Please try again later.
+generic.io-error.write = We couldn't write your data. Please try again later.
 
-artifact.not-found = We were unable to find the requested artifact. Please ensure that the artifact ID you supplied is correct or contact support.
-artifact.unable-to-delete = We were unable to delete the artifact asset. Please try again later.
-artifact.unable-to-update = We were unable to update the artifact asset. Please try again later.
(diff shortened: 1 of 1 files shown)


dvince2 (approved):
Nice!

dgauldie (approved):
LGTM

PR #4020, opened 2024-06-28
updated error messages for funman and sciml

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/TransformController.java
```diff
@@ -21,21 +28,28 @@
 public class TransformController {
 
 	final SkemaRustProxy skemaProxy;
-
 	final SimulationServiceProxy simulationServiceProxy;
+	final Messages messages;
 
 	@PostMapping("/mathml-to-acset")
```
mwdchang, line 34: Aside: I don't think we really touch ACSET-related matters anymore right? We may be able to deprecate or even remove this altogether.

dvince2 (approved):

mwdchang (approved):

PR #4021, opened 2024-06-28
Generate embeddings as part of model / document creation.

kbirk:
This PR includes the following:

Anytime a `model card` task completes successfully, if the document was public and non-temporary, it generates embeddings and indexes them. Therefore when running pdf extraction, it will automatically create the semantic search vectors for the document if it is publically searchable.

When creating / updating a public + non-temporary model, the amr is used to generate search embeddings. If model alignment is run with a document, the documents gollmCard, if present, will be used to generate new embeddings.


dvince2 (approved):
just one question about the logic

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -342,8 +346,24 @@ ResponseEntity<Model> updateModel(
...
+
+			if (updatedModel.getPublicAsset() && !updatedModel.getTemporary()) {
+				try {
+					final String amr = objectMapper.writeValueAsString(updatedModel);
+					final TerariumAssetEmbeddings embeddings = embeddingService.generateEmbeddings(amr);
+					modelService.uploadEmbeddings(model.getId(), embeddings, permission);
+				} catch (final Exception e) {
+					log.warn("Unable to update embeddings for model " + model.getId(), e);
+				}
+			}
```
mwdchang, line 364: Note we may have a lot of small updates from the user-interface from things that @shawnyama is doing (updating metadata and sources). We probably want a debounce-mechanism at the UI side of things otherwise this may become expensive.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/search/SearchByAssetTypeController.java
```diff
@@ -114,32 +109,12 @@ public ResponseEntity<List<JsonNode>> searchByAssetType(
...
-				final byte[] outputBytes = resp.getOutput();
-				final JsonNode output = objectMapper.readTree(outputBytes);
+				final TerariumAssetEmbeddings embeddings = embeddingService.generateEmbeddings(text);
 
-				final EmbeddingsResponse embeddingResp = objectMapper.convertValue(output, EmbeddingsResponse.class);
-
-				final List<Float> vector = embeddingResp.getResponse();
+				final List<Float> vector = Arrays.stream(
+								embeddings.getEmbeddings().get(0).getVector())
+						.mapToObj(d -> (float) d)
```
mwdchang, line 116: Is this intentional? Elsewhere embeddings are declared as `double`s

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -575,6 +576,25 @@ public Model runAlignAMR(
 					ProvenanceType.DOCUMENT);
 			provenanceService.createProvenance(provenance);
 
+			// update model embeddings
+			if (model.getPublicAsset() && !model.getTemporary()) {
```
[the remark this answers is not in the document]
kbirk: Whoops

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/search/SearchByAssetTypeController.java
kbirk: Yeah, the task returns doubles, but the java ES API wants floats...

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
kbirk: Good to know. I could just disable it and only generate them on model alignment instead?

mwdchang: Nah, we should go okay for now. 
More of a anchor/reminder here to chat when Shawn is back.

mwdchang (approved):

PR #4024, opened 2024-07-02
Fix compare models gollm request failure

### LINKED ISSUE #3941: [BUG]: Compare Models fails on the manual testing scenario
- Scenario: https://github.com/DARPA-ASKEM/terarium/blob/main/testing/manual/compare-two-models.md
- Link to the project: https://app.staging.terarium.ai/projects/296b2313-ee16-4c0f-a081-c5458831e1a3/workflow/24d9ab8a-03e9-47b9-9219-0378d2b0e4bc?operator=aa201c99-584f-4e14-80bb-dbb9a01143a9

4-3 and 5-5 fails from the above test scenario. 
Also following error occurs during step 4

Screenshot 2024-06-20 at 2 24 41 PM

**Describe the issue**

jryu01:
# Description

Fixes the compare models gollm request failure. 

https://github.com/DARPA-ASKEM/terarium/assets/3279838/dfd6b096-0ff8-4027-ac8a-3a378c1f1354

**Note:** It looks like we got rid of the event subscription api end point from the task runner (in this PR https://github.com/DARPA-ASKEM/terarium/pull/3913/files#). But we are still relying on it for GoLLM tasks in the client side and it results errors or unexpected behaviours. I updated this compare models operator to use the global event source (ClientEventService) for handling the event but we probably need to update other components that's using GoLlm tasks as well. 

Resolves #3941


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/CompareModelsResponseHandler.java
```diff
@@ -20,8 +20,8 @@ public String getName() {
 
 	@Data
 	public static class Input {
-		@JsonProperty("model_cards")
-		List<String> modelCards;
+		@JsonProperty("cards")
```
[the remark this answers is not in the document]
jryu01: I just removed the annotation

mwdchang (approved):
Looks okay!

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison.vue
```diff
@@ -191,13 +194,21 @@ const sampleAgentQuestions = [
...
 const llmThoughts = ref<any[]>([]);
 const isKernelReady = ref(false);
 const modelsToCompare = ref<Model[]>([]);
 const contextLanguage = ref<string>('python3');
 
+const llmAnswer = computed(() => {
+	if (!compareModelsTaskOutput.value) return '';
+	const str = b64DecodeUnicode(compareModelsTaskOutput.value);
+	const parsedValue = JSON.parse(str) as CompareModelsResponseType;
+	return parsedValue.response;
```
YohannParis, line 209: Why is this not done in line 338 when we read the `taskRed.output`?

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -100,47 +98,15 @@ export async function compareModels(
...
-					closeConnection();
-					logger.warn(`Failed to Compare Models with ids: ${modelIds.join(',')}`);
-					reject({ response: '' });
-				}
-				if (data.status === TaskStatus.Success) {
-					closeConnection();
-
-					// data.output is a base64 encoded json object. We decode it and return the json object.
-					const str = b64DecodeUnicode(data.output);
-					resolve(JSON.parse(str));
```
YohannParis, line 135: Why removing it from here and add it to the component itself?

jryu01: Because the gollm request is async and it just returns the status object without the output and the event listener for listening Success event is now registered on the component.

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison.vue
jryu01: We assign the output in line 345 (and also 338) as well. So instead of duplicating parsing logics in two places, we just do once here.

YohannParis: Yes, but in a computed seems a bit much no?

PR #4034, opened 2024-07-03
If project is public, force all assets to be public

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetServiceWithoutSearch.java
```diff
@@ -187,6 +206,10 @@ public Optional<T> updateAsset(final T asset, final Schema.Permission hasWritePe
 			throw new IllegalArgumentException("Cannot update a non-temporary asset to be temporary");
 		}
 
+		if (projectService.isProjectPublic(projectId)) {
+			asset.setPublicAsset(true);
```
bigglesandginger, line 205: How does this get set to false if the project is no longer public?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -960,25 +789,13 @@ public ResponseEntity<JsonNode> removeProjectGroupPermissions(
...
+	@ApiResponses(value = {
+			@ApiResponse(responseCode = "200", description = "Project visibility has been updated", content = {
+					@Content(mediaType = "application/json", schema = @io.swagger.v3.oas.annotations.media.Schema(implementation = UUID.class))
+			}),
+			@ApiResponse(responseCode = "304", description = "The current user does not have privileges to modify this project.", content = @Content),
+			@ApiResponse(responseCode = "500", description = "An error occurred verifying permissions", content = @Content)
+	})
 	@PutMapping("/set-public/{id}/{isPublic}")
 	@Secured(Roles.USER)
 	public ResponseEntity<JsonNode> makeProjectPublic(
```
bigglesandginger: I would have thought this would propagate the isAssetPublic logic through to its children, either setting to true or false as appropriate

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetServiceWithoutSearch.java
kbirk: Doh

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
kbirk: Oh lord so the flag doesn't even get set on the underlying project. Okay I'll inject some code here too.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/DocumentAssetService.java
```diff
@@ -48,4 +62,55 @@ protected String getAssetIndex() {
 	public String getAssetAlias() {
 		return elasticConfig.getDocumentAlias();
 	}
+
+	@Override
+	@Observed(name = "function_profile")
+	public DocumentAsset createAsset(
```
dvince2, line 68: If this is just calling `super` can it be removed?

packages/server/src/main/resources/application-local.properties
```diff
@@ -29,6 +29,9 @@ terarium.elasticsearch.auth-enabled=false
 ########################################################################################################################
 # Keycloak configuration
 ########################################################################################################################
+terarium.keycloak.url=https://keycloak.staging.terarium.ai
```
dvince2, line 32: This looks like an accident!

bigglesandginger (approved):

YohannParis (approved):
this is insane, why not found the project id from the asset itself?

bigglesandginger:
> this is insane, why not found the project id from the asset itself?



> this is insane, why not found the project id from the asset itself?

What is insane, the old concept that is not being fixed here, or something Kevin did?

PR #4038, opened 2024-07-04
switch threshold and value, use states

blanchco:
# Description

* Use states instead of parameters for target in dynamic interventions as per this [thread](https://unchartedsoftware.slack.com/archives/C03S02YH3D5/p1720047460929369?thread_ts=1720042002.213279&cid=C03S02YH3D5)
* switched threshold and value around
* added intervention semantic type (variable, parameter) selection

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
@@ -226,8 +226,8 @@ const onRemoveStaticIntervention = (index: number) => {
 const onAddNewStaticIntervention = () => {
 	const intervention = cloneDeep(props.intervention);
 	intervention.staticInterventions.push({
-		threshold: Number.NaN,
-		value: Number.NaN
+		threshold: 0,
+		value: 0
 	});
 	emit('update', intervention);
 };
@@ -237,8 +237,8 @@ const onInterventionTypeChange = (value: string) => {
 	if (value === 'static') {
 		intervention.staticInterventions = [
 			{
-				threshold: Number.NaN,
-				value: Number.NaN
+				threshold: 0,
+				value: 0
 			}
 		];
 		intervention.dynamicInterventions = [];
@@ -246,8 +246,8 @@ const onInterventionTypeChange = (value: string) => {
 		intervention.staticInterventions = [];
 		intervention.dynamicInterventions = [
 			{
-				threshold: Number.NaN,
-				value: Number.NaN,
+				threshold: 0,
+				value: 0,
 				parameter: props.stateOptions[0].value,
 				isGreaterThan: true
 			}
--- a/packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
@@ -59,7 +59,7 @@
 							type="nist"
 							:model-value="intervention.staticInterventions[0].threshold"
 							@update:model-value="(val) => onUpdateThreshold(val, 0)"
-							placeholder="time step"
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
```diff
@@ -62,15 +62,15 @@
 						<div class="flex align-items-center pt-2 pb-2 gap-2">
 							<tera-input
 								type="number"
-								:model-value="i.threshold"
-								@update:model-value="(val) => onUpdateThreshold(val, index)"
+								:model-value="i.value"
+								@update:model-value="(val) => onUpdateValue(val, index)"
```
mwdchang, line 66: Just noticed this, in normal circumstances we don't want the `v-for` to key on the iterating index unless the list itself is static. When you key on the index of the loop, you will get into various vue-cache issues if the list is allowed to shrink or to grow, which I think will happen in this case.

Also note what we discussed about maintainability cause you need to do this in multiple places 😄

blanchco: Is there a standard we have if we don't have a unique indicator on the data object?  All StaticInterventions just have this structure: 
```
{
    value: number
    threshold: number
}
```
Normally Im using the index as a fallback if there isn't a unique way to identify the object.  Should we store ids on the backend, or even preprocess the data with ids?

mwdchang: I think you can either:
- reassign the array reference when modifications happens, eg: don't modify stuff in place, or
- create a composite key of "value:threshold"

blanchco:
converting to draft to address comments from demo

packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
blanchco: @mwdchang I think it should be reassigning the array reference on update and delete now with the `key fixes?` commit. 

I tried doing the composite key but since some of the keys could be the same it was screwing up the UI

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
```diff
@@ -11,6 +11,7 @@ public class Intervention {
 
 	private String name;
 	private String appliedTo;
+	private InterventionSemanticType type;
```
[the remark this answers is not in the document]
blanchco: I needed it for the dropdown for whether the `appliedTo` type is a variable or a parameter.   Can there be a case where we have variable 'S' and parameter 'S'?  In that case we couldn't identify between the 2.

^ if thats not the case I can do some funky filtering on the front end 😄 

Screenshot 2024-07-04 at 1 17 43 PM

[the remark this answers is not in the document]
YohannParis: I see your point, fair enough

packages/client/hmi-client/src/components/workflow/ops/interventions/tera-intervention-card.vue
mwdchang: okay, so long as we have tested non-trivial insert/delete/update combinations.

blanchco: ya, Ive tested it out with a bunch of card and different intervention combos and it seems to be working fine

YohannParis (approved):

YohannParis (approved):

dvince2 (approved):

PR #4064, opened 2024-07-08
fix: AMR JSONs from MIRA can fail to load into Terarium due to mismat…

### LINKED ISSUE #4060: [BUG]: AMR JSONs from MIRA can fail to load into Terarium due to mismatched annotations schema
**Describe the bug**
When you upload a AMR JSON file generated by MIRA, Terarium can fail to load into Terarium since MIRA and Terarium use different JSON schema for model metadata.annotations

MIRA's annotations schema is here:
https://github.com/gyorilab/mira/blob/main/mira/metamodel/schema.json

**To Reproduce**

You can see this bug when you use what MIRA generates:
```
    "metadata": {
        "annotations": {
            "authors": [
                {
                    "name": "Nelson Liu"
                },
            ],
```

Terarium expects this instead:
```
    "metadata": {
        "annotations": {
            "authors": ["Nelson Liu"],
```

**Expected behavior**

We should use MIRA annotations schema.

More details and example here:
https://github.com/liunelson/mira/blob/main/docs/model_metadata_annotation.md?plain=1

dvince2:
# Description

Updating the model author metadata to match the spec.  This *looks* like it should need a migration, but, because of the malformed metadata `authors` is only stored as `[]` so it works! Not to mention the upcoming data wipe...

Resolves #4060

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-model-card.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-model-card.vue
@@ -74,7 +74,8 @@ const card = computed(() => {
 });
 
 const authors = computed(() => {
-	const authorsArray = props.model?.metadata?.annotations?.authors ?? [];
+	const authorsArray =
+		props.model?.metadata?.annotations?.authors?.map((author) => author.name) ?? [];
 
 	if (card.value?.authorAuthor) authorsArray.unshift(card.value?.authorAuthor);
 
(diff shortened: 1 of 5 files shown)


YohannParis (approved):

mwdchang (approved):
We need to keep an eye on `header.description`, I am not exactly sure what format is coming back. It was causing a problem for Holly last time because it was a `string[]` instead of a `string`

PR #4065, opened 2024-07-08
Task: removing outdated TODO comments.

Tom-Szendrey:
# Description

Removing outdated comments.
If we want we can create an issue to move all of the intervention logic from pyciemss-service to HMI. But this will involve more than just this todo comment and should also involve moving pyciemss-service  into a pure service.


dvince2 (approved):

YohannParis (approved):
Love it

PR #4067, opened 2024-07-08
Fix NPE in model config task handler

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ConfigureModelResponseHandler.java
```diff
@@ -91,7 +92,7 @@ public TaskResponse onSuccess(final TaskResponse resp) {
 				}
 
 				// Map the initials values to the model
-				final ArrayNode gollmExtractionsInitials = objectMapper.createArrayNode();
+				// final ArrayNode gollmExtractionsInitials = objectMapper.createArrayNode();
```
YohannParis, line 95: Why not removing the line directly?

YohannParis (approved):

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ConfigureModelResponseHandler.java
kbirk: True, fixed

Issue #3894, opened 2024-06-17
[FEAT]: UX Design: Model / Stratified Model updates


ngraham76:
The **Model** and **Stratified Model** pages have a number of design updates. 

The Figma designs are here:
https://www.figma.com/design/EoihrRoNNbuOAJtudKqp7L/Terarium---June-2024?node-id=374-12947&t=IYmeWIrCXnmMWNDP-1

Feel free to reach out with any questions or concerns. Thanks!

### **Descriptions and Reset / Save buttons**

**Current**
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/c5d4a176-317a-4e2b-a2ed-14ce5750a89c)

**Proposed**
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/8c9091cd-1c70-41fc-a3db-8b343ff4b0e7)
Remove the Provenance section, this will now be part of the Description.
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/b4e56ef3-77cd-4416-b534-67046ec31f68)

**Editing the description**
When the user clicks the pencil icon button next to the Description section title, they enter edit mode - an editable text area with formatting controls:
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/9f7d2ea7-7582-4b78-8763-58779de8a695)

**Reset, Save as and Save**
`Reset` - reverts all changes made back to the previous save.
`Save as...` - saves the model as a new model. A confirmation modal allows the user to name the new model. An LLM-suggested name appears in the field by default.
`Save` - overwrites the current model with all changes made. A confirmation modal appears, allowing the user to confirm they want to overwrite the model.
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/a572bc9f-7631-4ce0-9075-b532426b3344)

### **Initial Variables / Parameters (Model)**

**Current**
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/5f9ce981-3019-47a9-b493-cfbb1d49cb6c)

**Proposed**
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/bb06c937-2497-45cb-b32b-ba13b57fb186)

### **Initial Variables / Parameters (Stratified Model)**

**Current**
![image](https://github.com/DARPA-ASKEM/terarium/assets/141876103/7819d159-fdd6-44c7-8df9-3007274c782b)

**Proposed**
[...]

PR #4070, opened 2024-07-08
Update model drilldown and remove output dropdown from assets

YohannParis:
Part of the big task #3894 

* rename `tera-model-workflow-wrapper` to `tera-model-drilldown` to match other operator
* remove the output dropdown if of asset type `Model, Dataset, Code, Document` which only have one output.


packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -21,8 +21,9 @@
 								<tera-operator-port-icon v-if="input.type" :portType="input.type" />
 							</template>
 						</Chip>
-					</div>
+					</aside>
 					<tera-output-dropdown
+						class="ml-auto"
```
mwdchang, line 26: what does `ml-auto` and what kind of tests are needed to verify that the embedded operators are okay?

YohannParis: * `ml-auto` is `margin-left: auto` via _Tailwind CSS_.
* To test the operator you can just open an operator of each type mentioned above.

mwdchang (approved):

shawnyama (approved):

PR #4086, opened 2024-07-09
fix "decimal" tera-input type

mwdchang:
### Summary
"decimal" is an invalid option that gets routed to produce string-type, which then create problems for downstream consumers.


Tom-Szendrey (approved):
nice catch!
What does nist stand for? I see its our addition

blanchco (approved):

PR #4107, opened 2024-07-11
Optimize success and intervention charts

blanchco:
# Description

* Constraint Bar charts
* Interventions line charts (before lines only)
* These are by no means final...labels and styling can be changed

TODO: after optimize data for interventions chart
TODO: simulation plots (before and after) perhaps using Dan's charts

https://github.com/DARPA-ASKEM/terarium/assets/33158416/64a5500a-a918-46eb-a54d-b3ca9e6adfbd


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss.vue
```diff
@@ -654,6 +673,23 @@ const setOutputValues = async () => {
 	optimizationResult.value = optimzationResult;
 };
 
+const preProcessedInterventionsData = computed<
```
blanchco, line 676: I want to group and format the data for the interventions chart here...interventions such as beta can be created in separate cards, so I want to group them all together to be used in 1 chart

mwdchang (changes requested):
We are likely going to tweak these charts several times, and they likely need to adapt to different circumstances, for example you can imagine the charts need be rendered as:
- small preview with no interactions
- fullsize in the operator drilldown with interactions
- fullsize but not interactive in the report, whatever that may be

So I am thinking we might as well bite the bullet now, and have these spec generations in vanilla javascript/typescript rather than as computed properties of the Vue-framework.  
This would make it easier to customize things and make this feature more portable; plus, your template-sections are effectively pass-throughs calling "vega-chart".


I would imagine something like this

```
<vega-chart  :visualization-spec="interventionChartSpec" />
...
<vega-chart  :visualization-spec="optimizeChartSpec" />

```

Where

```
const interventionChartSpec = createInterventionChart(data, blah blah config, .....)
const optimizeChartSpec = createOptimizeChart(data, blah blah config, .....)
```

blanchco:
> We are likely going to tweak these charts several times, and they likely need to adapt to different circumstances, for example you can imagine the charts need be rendered as:
> 
> * small preview with no interactions
> * fullsize in the operator drilldown with interactions
> * fullsize but not interactive in the report, whatever that may be
> 
> So I am thinking we might as well bite the bullet now, and have these spec generations in vanilla javascript/typescript rather than as computed properties of the Vue-framework. This would make it easier to customize things and make this feature more portable; plus, your template-sections are effectively pass-throughs calling "vega-chart" and they don't add any additional elements per se.
> 
> I would imagine something like this
> 
> ```
> <vega-chart  :visualization-spec="interventionChartSpec" />
> ...
> <vega-chart  :visualization-spec="optimizeChartSpec" />
> ```
> 
> Where
> 
> ```
> const interventionChartSpec = createInterventionChart(data, blah blah config, .....)
> const optimizeChartSpec = createOptimizeChart(data, blah blah config, .....)
> ```

@mwdchang I've moved the logic and spec creation into the `optimize.ts` file and removed the chart components, is this more of what you were thinking of?

YohannParis (approved):

mwdchang (approved):

PR #4115, opened 2024-07-12
Updsate the zipfile format to prevent choking on json parsing

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/AssetExportDeserializer.java
```diff
@@ -14,21 +14,21 @@
...
 
 		final String assetTypeStr = node.get("type").asText();
 		final AssetType assetType = AssetType.getAssetType(assetTypeStr, mapper);
 
-		TerariumAsset asset = mapper.treeToValue(node.get("asset"), assetType.getAssetClass());
+		final TerariumAsset asset = mapper.treeToValue(node.get("asset"), assetType.getAssetClass());
 
 		Map<String, FileExport> files = new HashMap<>();
-		if (node.has("files")) {
+		if (node.has("files") && !node.get("files").isNull()) {
```
YohannParis, line 27: What an incongrus test. How can node.files exist but return null?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/AssetExportDeserializer.java
bigglesandginger: Should that be an OR and not an AND (like ContentTypeDeserializer L23)?

bigglesandginger: Wait the logic in the two is totally confusing :). NOTs abound... .

YohannParis: _ChatGPT_ says to use `if (node.hasNonNull("files")) {` instead.

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/AssetExportDeserializer.java
kbirk: I don't understand why it worked before, but it is returning a `NullNode` entry now, makes the `has` return true, but you get an exception if you try to use the entry

kbirk: I like the `hasNonNull` ill use that instead

bigglesandginger:
Javascript, making its namesake parent garbage through association too.  Amazing, well done whoever coded the JSON libraries.

PR #4116, opened 2024-07-12
refactor: generalize metadata editing interface

shawnyama:
# Description

* Layout for initial and parameter metadata editing are the same components: `tera-variables-metadata` and `tera-variable-metadata-entry`
* `tera-parameters-metadata` and `tera-states-metadata` just pass there unique data into the above components
* their data is transformed into a `ModelVariable` type which helps make everything compatible and easy to work with in the UI
* This is going to make writing the new interface for observables, transitions and time less painful 
* Sorry for all the old commit messages I need to fix my main history


packages/client/hmi-client/src/components/model/tera-parameters-metadata.vue
```diff
@@ -145,76 +42,45 @@ const emit = defineEmits(['update-parameter']);
...
+			const baseParameter = isParent ? { id } : parameters.value.find((p) => p.id === id);
+			const base = isParent
+				? { id: baseParameter?.id ?? '' }
+				: {
+						id: baseParameter?.id ?? '',
+						name: baseParameter?.name ?? '',
+						description: baseParameter?.description ?? '',
+						grounding: baseParameter?.grounding,
+						unitExpression: baseParameter?.units?.expression ?? ''
+					};
```
mwdchang, line 79: This kind of pattern is rather hard to read, is everything actually nullable or is this just because of typescript?

```
const blah = foobar.find(....)

return {
   a: blah?.a ?? '',
   b: blah?.b ?? ''
}
```

YohannParis (approved):

packages/client/hmi-client/src/components/model/tera-parameters-metadata.vue
shawnyama: Cleaned up the assignment here as well as in `tera-states-metadata`.

mwdchang (approved):
Did read super closely, seems to be okay.

PR #4117, opened 2024-07-12
fix collapse initials

blanchco:
# Description

* Fixes an edge cases issue where initials have an underscore AND that are not stratified the initial is renamed to something unrecognized.
* The fix is if there is only 1 child initial when we collapse the initials, we rename the root to the child initial which is recognized

Example:
We have `Cumulative_cases` which is an initial target, but this is not a stratified initial:
before fix collapseInitials:
```
{"Cumulative" => Array(1)}
key: "Cumulative"
value: ['Cumulative_cases']
```

On the model config node it will look for `Cumulative` to edit which does not exist

after fix :
```
{"Cumulative_cases" => Array(1)}
key: "Cumulative_cases"
value: ['Cumulative_cases']
```
On the model config node it will look for `Cumulative_cases` to edit which does exist

An example where this was causing an issue:
[SEIRHD vacc model for LA County t0 = 10_28_2021.json](https://github.com/user-attachments/files/16197543/SEIRHD.vacc.model.for.LA.County.t0.10_28_2021.json)

### DIFF
--- a/packages/client/hmi-client/src/model-representation/mira/mira.ts
+++ b/packages/client/hmi-client/src/model-representation/mira/mira.ts
@@ -166,6 +166,17 @@ export const collapseInitials = (miraModel: MiraModel) => {
 			map.set(rootName, [name]);
 		}
 	}
+
+	// when a key only has one child, we will rename the key of the root to the child.
+	// this is to avoid renaming when a single entry key has an underscore
+	[...map.keys()].forEach((key) => {
+		const newKey = map.get(key)?.[0];
+		if (newKey && map.get(key)?.length === 1 && newKey !== key) {
+			map.set(newKey, [newKey]);
+			map.delete(key);
+		}
+	});
+
 	return map;
 };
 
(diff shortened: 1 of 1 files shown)


mwdchang:
@blanchco can you provide a couple of examples so we can get a sense if we need a more generalized solution?

Also Observables might exhibit the same issue as well, if we can lets address it at the same time. https://github.com/DARPA-ASKEM/terarium/blob/a103cb2f07ad9e6c903f5248dbfc4a3f4798fc56/packages/client/hmi-client/src/model-representation/mira/mira.ts#L287-L294

YohannParis (approved):
Thank you!

shawnyama (approved):

blanchco:
> @blanchco can you provide a couple of examples so we can get a sense if we need a more generalized solution?
> 
> Also Observables might exhibit the same issue as well, if we can lets address it at the same time.
> 
> https://github.com/DARPA-ASKEM/terarium/blob/a103cb2f07ad9e6c903f5248dbfc4a3f4798fc56/packages/client/hmi-client/src/model-representation/mira/mira.ts#L287-L294

@mwdchang Observable's should be handled as well -  it checks the initials list to see if we should split the reference name or not

PR #4124, opened 2024-07-15
feat: editing for observables, transitions and time 

shawnyama:
# Description

* The new variables are successfully edited here
* Saving is now done only if you click the Save button, the old autosave is commented out for now

https://github.com/user-attachments/assets/e8537d84-7be7-45de-bca0-6bd7f6c093a7


packages/client/hmi-client/src/components/model/variables/tera-variable-entry.vue
```diff
@@ -1,18 +1,33 @@
...
 			:model-value="variable.unitExpression ?? ''"
-			@update:model-value="$emit('update-variable', { key: 'unitExpression', value: $event })"
+			@update:model-value="
+				($event) => {
+					const value = $event.replace(/[\s.]+/g, '');
+					$emit('update-variable', { key: 'unitExpression', value });
+				}
+			"
 			:disabled="disabledInputs?.includes('unitExpression')"
+			@focusout="($event) => ($event.target.value = $event.target.value.replace(/[\s.]+/g, ''))"
```
mwdchang, line 30: - Let's try to keep the template section clean, you are doing too much data wrangling here
- If MIRA is having problem parsing spaces, that is issue that should be raised and discussed before we try to work around it

packages/client/hmi-client/src/components/model/tera-model-variables.vue
```diff
@@ -1,19 +1,22 @@
...
 		@update-model="$emit('update-model', $event)"
-		@update-state="onUpdateState"
-		@update-parameter="onUpdateParameter"
+		@update-state="(e: any) => onUpdate('state', e)"
+		@update-parameter="(e: any) => onUpdate('parameter', e)"
+		@update-observable="(e: any) => onUpdate('observable', e)"
+		@update-transition="(e: any) => onUpdate('transition', e)"
+		@update-time="(e: any) => onUpdate('time', e)"
 	/>
 </template>
```
mwdchang, line 15: I am not crazy about the naming here, between
- tera-model
- tera-model-variables
- tera-variables
- tera-variable-entry
- tera-parameters

And that the terms variables and parameters are kind of exchangeable....

I find it hard to keep track of what is what, can you brief on the roles they play? Let's see if we can have a better naming scheme.

shawnyama: In my mind a variable counts as a state, transition, parameter, observable. Would there be a better name to generally describe these?

packages/client/hmi-client/src/components/model/variables/tera-transitions.vue
```diff
@@ -0,0 +1,92 @@
...
+const transitionsList = computed<
+	{
+		base: ModelVariable;
+		children: ModelVariable[];
+		isParent: boolean;
+	}[]
+>(() =>
+	Array.from(collapsedTemplates.value.matrixMap.keys())
+		.flat()
+		.map((templateId) => {
```
mwdchang, line 51: What is the point of the Array and flat here? Would it be simpler to directly
```
[...collapsedTemplates.value.matrixMap.keys()].map((templateId) => { ......
```

packages/client/hmi-client/src/components/model/variables/tera-variable-entry.vue
shawnyama: Yeah the `amr_to_mmt` doesn't work if there is a space in the unit expression. I'll remove the string edit. How would it handle spaces though, would they still be there? I think the UI should reflect what's in MIRA whether it removes the space or replaces it with a `_`

mwdchang: We can keep the space truncation hack for now. 

But we should find and examine what the error actually is, because you never know if it is just a symptom of a deeper problem. Basically if other performer's stuff error out, the error should always be raised for awareness and triaging.

Do you have a couple of examples of the strings you had tried and failed? And what does the the stack trace look like? Let's take the conversation off this PR and into our slack channel.

packages/client/hmi-client/src/components/model/tera-model-variables.vue
mwdchang: I would describe these (state, transition ... etc) as components or parts of a petrinet-model ... or alternatively components of a compartmental-model in a hazy kind of way. 

Specifically:
- "transition" describe relationships between multiple "state"
- "state" points to a "parameter"
- "observable" contains "state" or "parameter" with some relationship bindings
- "parameter" are values or distributions

shawnyama: I think `model-parts` sounds good - there a modelparts folder in the server. `model-components` could mixed up with a vue component.

packages/client/hmi-client/src/components/model/variables/tera-transitions.vue
YohannParis: Or better a flatMap()

shawnyama: I actually gotta check why it's like this, I forgot

mwdchang: > Or better a flatMap()

`keys()` are already flat

shawnyama: removed them, looks like I blindly copied from `tera-initial-table` and `tera-parameter-table` (which I fixed too)

YohannParis (approved):

mwdchang (approved):

PR #4125, opened 2024-07-15
legend and formatting

mwdchang:
### Summary
- Adding legend for drilldown, adding translation lookup to format variable names. 
- x-axis for pyciemss-simulate

image


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -74,12 +77,24 @@ export const createForecastChart = (
...
 	yaxis.title = options.yAxisTitle;
 
+	const translationMap = options.translationMap;
+	let labelExpr = '';
+	if (translationMap) {
+		Object.keys(translationMap).forEach((key) => {
+			labelExpr += `datum.value === '${key}' ? '${translationMap[key]}' : `;
+		});
+		labelExpr += " 'other'";
+	}
```
mwdchang, line 87: This is really more of an artifact from running pyciemss, but it seems like it would be useful to have a translation map for variables/columns.

YohannParis (approved):

blanchco (approved):

PR #4143, opened 2024-07-16
Change riskTolerance to be a negative of alpha for pyciemss-service

YohannParis:
# Description

* We want to display the risk tolerance the opposite of how `alpha` is being sent to `pyciemss-service`


Tom-Szendrey (approved):
getting rid of my import organization pains me deeply. but is fair enough

YohannParis:
> getting rid of my import organization pains me deeply. but is fair enough

Unfortunately it was not being respected by anyone... so it just made things more complex

mwdchang (approved):

PR #4145, opened 2024-07-16
fix observable mapping

mwdchang:
### Summary
In `getRunResultCiemss` we establish how PyCIEMSS result should map back to the model, our mapping logic seem to be out-of-date for observable variables, which cause the ground-truth not not appear in calibrate operation.

This PR should fix the logic, but also mark getRunResultCiemss as deprecated as we start to replace the chart.js charts with Vegalite, which generally wants a more simplified data format that is closer to the raw produced data.

image

### Testing
- Do a calibrate operation, the mapping should map an observed variable in the model to the grounding data
- When calibrate finishes, chart the observed variable, should see both the forecast-values and ground-truth-values

### DIFF
--- a/packages/client/hmi-client/src/services/models/simulation-service.ts
+++ b/packages/client/hmi-client/src/services/models/simulation-service.ts
@@ -147,6 +147,8 @@ export async function getCalibrateBlobURL(runId: string) {
 	return resp.data.url;
 }
 
+// @deprecated - The notion of RunResult is a outdated with introduction of Vegalite charts
+// that use a more barebone setups closwer to the raw data
 export async function getRunResultCiemss(runId: string, filename = 'result.csv') {
 	const resultCsv = await getRunResult(runId, filename);
 	const csvData = csvParse(resultCsv);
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -147,6 +147,8 @@ export async function getCalibrateBlobURL(runId: string) {
 	return resp.data.url;
 }
 
+// @deprecated - The notion of RunResult is a outdated with introduction of Vegalite charts
+// that use a more barebone setups closwer to the raw data
```
YohannParis, line 151:
```suggestion
// that use a more barebone setup closer to the raw data
```

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -171,20 +173,31 @@ export async function getRunResultCiemss(runId: string, filename = 'result.csv')
...
-				outputRowRunResults[keyName] = inputRow[key];
+				runConfigs[newKey].push(Number(inputRow[key]));
+				return;
+			}
+			if (key.endsWith('_state')) {
+				const newKey = key.replace(/_state$/, '');
+				outputRowRunResults[newKey] = inputRow[key];
+				if (!runConfigs[newKey]) {
+					runConfigs[newKey] = [];
+				}
```
YohannParis, line 190: Should this test be done before those `if`s?

YohannParis (approved):

packages/client/hmi-client/src/services/models/simulation-service.ts
mwdchang: nope, each if block generates a different 'newKey'

blanchco (approved):
Lgtm

PR #4150, opened 2024-07-16
4149 constraint drop down should not have observables these are broken in pyciemss at the moment

### LINKED ISSUE #4149: constraint drop down should not have observables - these are broken in pyciemss at the moment

YohannParis:
# Description

* Rename ContraintGroup to appropriate `Criterion`
* Remove the observables from the dropdown
* Clean up few things

Resolves #4149

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
@@ -104,8 +104,8 @@ export const blankInterventionPolicyGroup: InterventionPolicyGroupForm = {
 	}
 };
 
-export const defaultConstraintGroup: ConstraintGroup = {
-	name: 'Constraint',
+export const defaultCriterionGroup: ConstraintGroup = {
+	name: 'Criterion',
 	qoiMethod: ContextMethods.max,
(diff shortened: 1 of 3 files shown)


Tom-Szendrey (approved):
ty

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -57,7 +57,7 @@ export interface OptimizeCiemssOperationState extends BaseState {
 	interventionPolicyId: string; // Used to determine if we need to reset the InterventionPolicyGroupForm.
 	interventionPolicyGroups: InterventionPolicyGroupForm[];
 	// Constraints:
-	constraintGroups: ConstraintGroup[];
+	constraintGroups: Criterion[];
```
mwdchang, line 60: Are we going all the way and rename the field as well? e.g. it still says "state.constraintGroups"

YohannParis: It would be great to do so. But the PR is quite big already, and I thought stopping here will be good for now.
I want to keep doing the cleaning, maybe in another PR soo.

blanchco (approved):

PR #4151, opened 2024-07-16
Optimize - utilize interventions in simulation runs

Tom-Szendrey:
# Description

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
@@ -1,6 +1,8 @@
 import { Operation, WorkflowOperationTypes, BaseState } from '@/types/workflow';
-import { Intervention, InterventionSemanticType } from '@/types/Types';
+import { Intervention, InterventionSemanticType, InterventionPolicy } from '@/types/Types';
 import { getRunResult, getSimulation } from '@/services/models/simulation-service';
+import { getModelIdFromModelConfigurationId } from '@/services/model-configurations';
+import { createInterventionPolicy } from '@/services/intervention-policy';
 
 const DOCUMENTATION_URL =
 	'https://github.com/ciemss/pyciemss/blob/main/pyciemss/interfaces.py#L747';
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -151,10 +152,14 @@ watch(
 		const response = await pollResult(optId);
 		if (response.state === PollerState.Done) {
 			// Start 2nd simulation to get sample simulation from dill
-			const simulationIntervetions = await getOptimizedInterventions(optId);
+			const newInterventionResponse = await createInterventionPolicyFromOptimize(
+				modelConfigId.value ?? '',
+				optId
+			);
```
mwdchang, line 158: At this stage, something catastrophic would've happened if you don't actually have `modelConfigId.value` right? 

Also it doesn't seem to make sense to create interventionPolicy with the blank model-config id??

If this is the case it would be better to phrase this as the follow if possible:
```
modelConfigId.value as string
```

It better indicates the intent here, as well as the situational awareness of the variables/states.

Tom-Szendrey: Fair point will do

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -166,42 +168,75 @@ export async function getOptimizedInterventions(optimizeRunId: string) {
...
+	const modelId = await getModelIdFromModelConfigurationId(modelConfigId);
+	const optimizedInterventions = await getOptimizedInterventions(optimizeRunId);
+
+	const newIntervention: InterventionPolicy = {
+		modelId,
+		interventions: optimizedInterventions
+	};
+	const newInterventionResponse: InterventionPolicy =
+		await createInterventionPolicy(newIntervention);
+	return newInterventionResponse;
```
mwdchang, line 241: We are not dealing with a "Response" object per se, would be better to just call it "newInterventionPolicy" or something.

mwdchang: Tangent, you are pulling in a good amount of service-level logic into the this operator file, be careful not to create circular dependencies.

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -106,14 +106,15 @@ const startForecast = async (simulationIntervetions) => {
...
-	// if (simulationIntervetions) {
-	// 	simulationPayload.interventions = simulationIntervetions;
-	// }
+
+	if (optimizedInterventions) {
+		// Use the intervention policy ID provided.
+		simulationPayload.policyInterventionId = optimizedInterventions.id;
+	} else {
+		// Use the input interventions provided
+		const inputIntervention = props.node.inputs?.[2]?.value?.[0];
```
mwdchang, line 115: I don't think `props.nodes.inputs` can be null/undefined here according to the definition, do you mind doing a quick look around and clean up?

Tom-Szendrey: Thank you

mwdchang (approved):

YohannParis (approved):

PR #4152, opened 2024-07-16
Default intial guess for intervention policy

YohannParis:
# Description

* Rename from Interventions to Intervention Policy
* Use a unique blank intervention object
* Replace the default value to be initial guess
* Rename the action button on Intervention Policy to `Save` instead of `Run`


Tom-Szendrey (approved):

mwdchang (approved):
Looks fine. But note this is a breaking change because the operator's registration name has changed.

PR #4157, opened 2024-07-16
Jamie style simulate charts

jamiewaese-uncharted:
Yes, I touched a lot of files here. I don't think I broke anything. It looks good, no?

image

image


mwdchang (approved):
We are starting to accumulated a lot of repeated structures that should be factored out, which I am planning to do in the near future.

PR #4158, opened 2024-07-16
Intervention created from FE needs snake -> camel caes

Tom-Szendrey:
# Description
This `getOptimizedInterventions` is kindve ugly. We should consider moving it later but while everything is in flux this fix should allow us to leave this alone for a bit.


mwdchang:
%s/@mwdchang /@dvince2  /g

YohannParis (approved):
Unless it cannot be done before the demo, but this should be done in the back-end @dvince2. We have an annotation for that purpose.

dvince2:
> Unless it cannot be done before the demo, but this should be done in the back-end @dvince2. We have an annotation for that purpose.

This is coming in as a `JsonNode` which isn't easily converted to camel case. I'm sure its possible but it would be something new to figure out.

dvince2 (approved):

PR #4172, opened 2024-07-17
Support for Gradle 9

YohannParis:
# Description

Fix the depracated features for Gradle 9.

https://docs.gradle.org/8.8/userguide/upgrading_version_8.html#java_convention_deprecation

### DIFF
--- a/packages/server/build.gradle
+++ b/packages/server/build.gradle
@@ -33,7 +33,10 @@ apply plugin: 'com.diffplug.spotless'
 
 group = 'software.uncharted'
 version = '1.0.0-SNAPSHOT'
-sourceCompatibility = '17'
+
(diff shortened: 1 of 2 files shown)


packages/server/build.gradle
```diff
@@ -33,7 +33,10 @@ apply plugin: 'com.diffplug.spotless'
 
 group = 'software.uncharted'
 version = '1.0.0-SNAPSHOT'
-sourceCompatibility = '17'
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_17
```
bigglesandginger, line 38: We don't seem to need this anymore - running with Java 22 and removed this and the server seems to be running fine

PR #4173, opened 2024-07-17
Ts/optimize tweaks

Tom-Szendrey:
# Description
-Do not reverse alpha
-Default alpha to 95
-correct the drop down option/InterventionTypes
-correct how bounds are being generated.
-Intervention start time label should like be: Intervention Start time (or where its bolded if that is cleaner)

# Testing:
image

YAY
https://github.com/DARPA-ASKEM/program-milestones/blob/74-july-2024-scenario/Monthly_Scenarios_CIEMSS/July_2024_Scenario/July_2024_Scenario_Optimization_without_Calibration.ipynb


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
```diff
@@ -7,9 +7,9 @@ import { createInterventionPolicy } from '@/services/intervention-policy';
 const DOCUMENTATION_URL =
 	'https://github.com/ciemss/pyciemss/blob/main/pyciemss/interfaces.py#L747';
 
-export enum InterventionTypes {
-	paramValue = 'param_value', // provide a parameter value to get a better start time.
-	startTime = 'start_time' // provide a statr time to get a better parameter value.
+export enum OptimizationTypes {
+	getStartTime = 'start_time', // provide a parameter value to get a better start time.
+	getParamValue = 'param_value' // provide a statr time to get a better parameter value.
```
mwdchang, line 12: We do not normally use verbs as enumeration names, is there a problem if we do "StartTime", "ParamValue"?

mwdchang (approved):

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-ciemss-operation.ts
Tom-Szendrey: I wasnt aware of that norm
I can update to the following if it is easier to read?
```
OptimizationInterventionObjective{ 
   startTime ...
   paramValue ...
}
```

PR #4179, opened 2024-07-18
intervention card style fix

blanchco:
# Description

* Fixes an issue where if interventions list was too short it would wrap to the right side as seen in this [thread](https://unchartedsoftware.slack.com/archives/C06FGLXB2CE/p1721308943409219)
* I just surrounded `<ul>` 100% width

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
@@ -68,35 +68,37 @@
 						.
 					</template>
 
-					<ul v-if="intervention.staticInterventions.length > 1" class="flex-1">
-						<li v-for="(i, index) in intervention.staticInterventions" :key="index">
-							<div class="flex align-items-center pt-2 pb-2 gap-2">
-								<tera-input
-									type="nist"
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
```diff
@@ -68,35 +68,37 @@
...
-									icon="pi pi-times"
-									text
-									@click="onRemoveStaticIntervention(index)"
-								/>
-							</div>
-							<Divider />
-						</li>
-					</ul>
+					<div class="w-full">
+						<ul v-if="intervention.staticInterventions.length > 1" class="flex-1">
```
shawnyama, line 72: did you try putting the `w-full` in the `ul`?

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
blanchco: ahhh I had to move the flex-1 down to the `li` element, less changes now 😄

YohannParis (approved):

PR #4181, opened 2024-07-18
feat: string expression to latex expression conversion

shawnyama:
# Description

* String expressions can now be converted to latex so they can be rendered nicely in `<katex-element>`
* Did some back and forth with copilot until I got something I liked

- **Before:**
image
- **After:**
image


YohannParis (approved):
Great job!

packages/client/hmi-client/src/services/model.ts
```diff
@@ -182,3 +182,27 @@ export function isInitial(obj: Initial | ModelParameter | null): obj is Initial
...
+	// latexExpression = latexExpression.replace(/\*/g, ' ');
+
+	// Convert ^ to LaTeX superscript notation
+	latexExpression = latexExpression.replace(/\^([a-zA-Z0-9]+)/g, '^{$1}');
+
+	// Detect and convert fractions a/b to \frac{a}{b}
+	latexExpression = latexExpression.replace(/([a-zA-Z0-9]+)\/([a-zA-Z0-9]+)/g, '\\frac{$1}{$2}');
+
+	return latexExpression;
+}
```
mwdchang, line 208: This is the sort of function that would love a few unit test cases, not only to test but to demonstrate functionality. 

For example it is a bit unusual to see that we need superscript `^` conversion here ... what is the use-case?

shawnyama: Assuming exponents a written as `x^2` in the string format, I'd have to convert it to `x^{2}` for latex

mwdchang (approved):

PR #4188, opened 2024-07-18
feat: edit concepts in model page

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -35,7 +35,7 @@
 			placeholder="Select a concept"
 			icon="pi pi-search"
 			:disabled="disabledInputs?.includes('concept')"
-			:model-value="''"
+			v-model="query"
 		/>
 		<katex-element
 			class="expression"
@@ -54,19 +54,33 @@
 </template>
 
 <script setup lang="ts">
+import { ref, watch } from 'vue';
 import TeraInput from '@/components/widgets/tera-input.vue';
 import type { ModelPartItem } from '@/types/Model';
-
-// import { getCurieFromGroundingIdentifier, getNameOfCurieCached } from '@/services/concept';
+import type { DKG } from '@/types/Types';
+import {
+	// getCurieFromGroundingIdentifier,
+	// getNameOfCurieCached,
+	searchCuriesEntities
+} from '@/services/concept';
 // getNameOfCurieCached(
 // 		new Map<string, string>(),
 // 		getCurieFromGroundingIdentifier(grounding.identifiers)
 // 	)
 
-defineProps<{
+const props = defineProps<{
 	item: ModelPartItem;
 	disabledInputs?: string[];
 }>();
+
+if (!props.disabledInputs) console.log(props.item);
+
+const query = ref('');
+const results = ref<DKG[]>([]);
+
+watch(query, async (value) => {
+	results.value = await searchCuriesEntities(value);
+});
 
 defineEmits(['update-item']);
 </script>
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -30,12 +30,21 @@
 			@focusout="($event) => ($event.target.value = $event.target.value.replace(/[\s.]+/g, ''))"
 		/>
 		<!--TODO: Add support for editing concepts-->
-		<tera-input
+		<!-- <tera-input
 			label="Concept"
 			placeholder="Select a concept"
 			icon="pi pi-search"
 			:disabled="disabledInputs?.includes('concept')"
-			v-model="query"
+			v-model="query"@complete="$emit('update-item', { key: 'concept', value: $event.value?.curie })"
+		/> -->
+		<AutoComplete
+			label="Concept"
+			placeholder="Select a concept"
+			v-model="q"
+			@suggestions="results"
+			@complete="searchConcepts"
+			option-label="name"
+			:disabled="disabledInputs?.includes('concept')"
 		/>
 		<katex-element
 			class="expression"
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/concept.ts
```diff
@@ -83,9 +83,11 @@ async function getEntitySimilarity(
 	}
 }
 
-const getNameOfCurieCached = (cache: Map<string, string>, curie: string): string => {
+const getNameOfCurieCached = async (curie: string, cache: Map<string, string> = new Map<string, string>()) => {
```
dvince2, line 86: This is still returning a `string` - why remove the typing?

dvince2, line 86: This is still returning a `string` -why remove the typing?

packages/client/hmi-client/src/model-representation/service.ts
```diff
@@ -295,10 +298,9 @@ export function getInitialUnits(model: Model, target: string): string {
 	return getInitialMetadata(model, target)?.units?.expression ?? '';
 }
 
-export function getInitialConcept(model: Model, target: string): string {
+export async function getInitialConcept(model: Model, target: string) {
```
dvince2, line 301: This is still returning a `string` -why remove the typing?

packages/client/hmi-client/src/components/model/petrinet/tera-petrinet-tables.vue
```diff
@@ -106,7 +106,7 @@ const states = computed<State[]>(() => props.model?.model?.states ?? []);
 const transitions = computed<Transition[]>(() =>
 	props.model.model.transitions?.map((transition: Transition) => ({
 		...transition,
-		name: transition.name ?? transition?.properties?.name ?? transition.id,
+		name: transition.name ?? transition.properties?.name ?? transition.id,
```
dvince2, line 109: Is this correct? You're checking the status of `transition.name` and then return `transition.properties.name` without checking it?

dvince2 (approved):
Just one question on returning the right property `name`

packages/client/hmi-client/src/components/model/petrinet/tera-petrinet-tables.vue
shawnyama: `transition.properties` can be `undefined` so there wasn't a need for the first `?`. What's happening here is assigning a name to edit in the UI. The order of these variables are just the priority that I'd assign the default name to be when you'd start editing.

dvince2: err sorry... i was just wondering why we're checking `transition.name` only to then turn around and use `transition.properties?.name`

packages/client/hmi-client/src/services/concept.ts
YohannParis, line 1: Good catch on my silly mistake here!

packages/client/hmi-client/src/components/widgets/tera-input.vue
```diff
@@ -136,7 +136,7 @@ main {
 	flex: 1 1 0;
 	justify-content: space-between;
 	align-items: center;
-	padding: var(--gap-xsmall) var(--gap-small);
+	padding: var(--gap-1) var(--gap-2);
```
YohannParis, line 139: 🤩

YohannParis (approved):

packages/client/hmi-client/src/services/concept.ts
```diff
@@ -83,9 +83,14 @@ async function getEntitySimilarity(
 	}
 }
 
-const getNameOfCurieCached = (cache: Map<string, string>, curie: string): string => {
+const getNameOfCurieCached = async (
+	curie: string,
+	cache: Map<string, string> = new Map<string, string>()
+): Promise<string> => {
```
mwdchang, line 89: This is a rather confusing design pattern, you are calling this function without the cache functional parameter, you then instantiate the cache on the fly and it goes out of the scope when the function finishes.

So effective the changes you have made are not using cache at all. So either:
- we should create a cache-store at the service level, and we don't pass in a cache for the function invocation (which you don't seem to be doing anyway), or
- we don't use cache

Have a slight preference for option one. 

If you want to see a client side caching implementation,  we did it in the world modeler project, and there is an example here on how to hook things up:

https://github.com/uncharted-causemos/causemos/blob/56c57f0891d8d97051a34f411bedecf850922b0e/client/src/services/outputdata-service.ts#L38-L41

packages/client/hmi-client/src/components/model/petrinet/tera-petrinet-tables.vue
shawnyama: A lot transitions have their `properties.name` filled while the `name` is empty. So starting on that makes sense in a lot of cases.

packages/client/hmi-client/src/components/dataset/tera-dataset-overview-table.vue
```diff
@@ -114,30 +88,36 @@ function updateDataset(index: number, field: string, value: any) {
...
+					id: col.name,
+					name: formatName(col.name),
+					description: col.description,
+					concept: await getNameOfCurieCached(getCurieFromGroundingIdentifier(identifier)),
+					conceptURL: getCurieUrl(getCurieFromGroundingIdentifier(identifier)),
+					unit: col.metadata?.unit,
+					dataType: col.metadata?.column_stats?.type,
+					stats: col.metadata?.column_stats,
+					column: col
+				};
```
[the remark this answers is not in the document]
shawnyama: The concept property is awaiting for a value so I have to wait for those all to resolve

```	concept: await getNameOfCurieCached(getCurieFromGroundingIdentifier(identifier)),```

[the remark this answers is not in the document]
mwdchang: My bad, it is hard to read, so maybe that is more of the issue here than the actual logic of the code. But let's deal with style another day.

mwdchang (approved):

PR #4200, opened 2024-07-19
support operator output with multiple types

mwdchang:
### Summary
Initial work to support operator output that can contain more than one type. For example calibrate run create a dataset + config. The way this is handled is to piggy back on the current notion where we "accept" one or more types.

```
type: 'datasetId | modelConfigId`
```

and the value of the output being an object keyed by the type:
```
value = [{
  datasetId: 'abcdefg',
  modelConfigId: 'aaaaabbbbbb'
}]
```

I largely removed the "allowMultiple" idea, as it isn't been used from what I can gather and it creates a boatload of edge cases that we would need to deal with.

### Testing
In unit tests
```
yarn workspace hmi-client run test workflow
```

### DIFF
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -109,7 +109,7 @@ export const addNode = (
 };
 
 /**
- * Create an eddge between two nodes, and transfer the value from
+* Create an edge between two nodes, and transfer the value from
  * source-node's output port to target-node's input port
  *
  * The terms are used as depicted in the following schematics
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -171,7 +171,7 @@ export const addEdge = (
 	const allowedInputTypes = targetInputPort.type.split('|');
 	const intersectionTypes = _.intersection(outputTypes, allowedInputTypes);
 
-	// Not supported if there are more than oen matches
+	// Not supported if there are more than one matche
 	if (intersectionTypes.length > 1) {
 		console.error(`Ambiguous matching types [${outputTypes}] to [${allowedInputTypes}]`);
 		return;
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -171,7 +171,7 @@ export const addEdge = (
 	const allowedInputTypes = targetInputPort.type.split('|');
 	const intersectionTypes = _.intersection(outputTypes, allowedInputTypes);
 
-	// Not supported if there are more than one matche
+  // Not supported if there are more than one match
 	if (intersectionTypes.length > 1) {
 		console.error(`Ambiguous matching types [${outputTypes}] to [${allowedInputTypes}]`);
 		return;
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -173,7 +173,7 @@ export const addEdge = (
 
 	// Not supported if there are more than oen matches
 	if (intersectionTypes.length > 1) {
-		console.error(`Ambiguous matching types ${intersectionTypes}`);
+		console.error(`Ambiguous matching types [${outputTypes}] to [${allowedInputTypes}]`);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -135,26 +167,38 @@ export const addEdge = (
...
 		(!targetInputPort.acceptMultiple && targetInputPort.status === WorkflowPortStatus.CONNECTED)
 	) {
 		return;
 	}
 
 	// Transfer data value/reference
-	if (targetInputPort.acceptMultiple && targetInputPort.value && sourceOutputPort.value) {
-		targetInputPort.label = `${sourceOutputPort.label},${targetInputPort.label}`;
-		targetInputPort.value = [...targetInputPort.value, ...sourceOutputPort.value];
+	targetInputPort.label = sourceOutputPort.label;
```
mwdchang, line 189: The port labelling part will likely need a second pass, not exactly sure what we want to show if a port has multiple types.

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -135,26 +167,38 @@ export const addEdge = (
 	if (existingEdge) return;
 
 	// Check if type is compatible
-	const allowedTypes = targetInputPort.type.split('|');
+	const outputTypes = sourceOutputPort.type.split('|');
+	const allowedInputTypes = targetInputPort.type.split('|');
```
YohannParis, line 171: Should those be trimmed of whitespaces?

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -135,26 +167,38 @@ export const addEdge = (
...
+	const intersectionTypes = _.intersection(outputTypes, allowedInputTypes);
+
+	// Not supported if there are more than oen matches
```
YohannParis, line 174:
```suggestion
	// Not supported if there are more than one matche
```

packages/client/hmi-client/src/services/workflow.ts
YohannParis: Maybe a simple ` or `.

YohannParis (approved):
LGTM

packages/client/hmi-client/src/services/workflow.ts
mwdchang: yep, thx, will fix

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -108,6 +108,38 @@ export const addNode = (
 	wf.nodes.push(node);
 };
 
+/**
+ * Create an eddge between two nodes, and transfer the value from
```
shawnyama, line 112:
```suggestion
* Create an edge between two nodes, and transfer the value from
```

packages/client/hmi-client/src/services/workflow.ts
shawnyama: There's a `getPortLabel` function in this file that does this

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -108,6 +108,38 @@ export const addNode = (
...
+ *
+ * The target-input can accept multiple types, but this is more of an exception
+ * rather than the norm. This is resolved by checking one of the accepted type
+ * matches the provided type.
+ *
+ * The second case is when the source produce multiple types, a similar check, but
+ * on the reverse side is performed.
+ *
+ *
+ * Ambiguity arises when the source provide multiple types and the target accepts multiple
```
shawnyama, line 136:
```suggestion
* Ambiguity arises when the source provides multiple types and the target accepts multiple
```

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -135,26 +167,38 @@ export const addEdge = (
 	if (existingEdge) return;
 
 	// Check if type is compatible
-	const allowedTypes = targetInputPort.type.split('|');
+	const outputTypes = sourceOutputPort.type.split('|');
+	const allowedInputTypes = targetInputPort.type.split('|');
+	const intersectionTypes = _.intersection(outputTypes, allowedInputTypes);
+
+	// Not supported if there are more than one matche
```
shawnyama, line 174:
```suggestion
  // Not supported if there are more than one match
```

shawnyama (approved):

packages/client/hmi-client/src/services/workflow.ts
mwdchang: getPortLabel happens "after" we set the port label if I recall - this here happens at a lower level.

PR #4209, opened 2024-07-22
FIX: Optimize setInterventionPolicyGroups re-init correctly.

Tom-Szendrey:
# Description
-Allow user to disconnect an intervention and correctly update the optimize intervention groups.

Previously we were checking is state of input ID === input ID.
This will never work, state is updated when input is updated.

#Testing
https://github.com/user-attachments/assets/320e2d30-7953-4785-b238-262bbc508350


mwdchang (approved):

mwdchang:
Talked to Tom on an aside, reminder to mark necessary sections as FIXMES.

PR #4214, opened 2024-07-22
Added histogram chart

### LINKED ISSUE #4195: [TASK]: Histogram charts
Create a function that generates Vegalite specs to show histograms. 

We should check if we can using Vegalite's built in transforms for the binning:
https://vega.github.io/vega-lite/docs/bin.html

Consider a function signature:
```
createHIstogram(data, options)
```

Where options might look like this, that defines the overall width/height of the chart, and what fields out of the data should be used to generate the histogram
```
{
  width,
  height,
  variables: [ 
    {
       field: "var1",
       label: "before",
       width: number
    },
    {
       field: "var1",
       label: "before",
       width: number
    }
  ]
}
```
 
and data would look something like

```
[
   { var1: number, var2: number, var3: number ... },
   { var1: number, var2: number, var3: number ... },
   ...
]
```

Example Design sketch: 

Image

jryu01:
# Description
Screenshot 2024-07-22 at 3 31 13 PM

Resolves #4195

### DIFF
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -86,17 +86,23 @@ export const createHistogramChart = (dataset: Record<string, any>[], options: Hi
 			domain: opts.variables.map((v) => v.label ?? v.field),
 			range: opts.variables.map((v) => v.color)
 		};
+		const bin = { maxbins: maxBins };
+		const aggregate = 'count';
 		return opts.variables.map((varOption) => ({
 			mark: { type: 'bar', width: varOption.width, tooltip: true },
 			encoding: {
-				x: { bin: { maxbins: maxBins }, field: varOption.field, axis: xaxis, scale: { padding: xPadding } },
-				y: { aggregate: 'count', axis: yaxis },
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -24,6 +24,96 @@ export interface ForecastChartLayer {
...
+		height: options.height,
+		autosize: { type: 'fit' },
+		data: { values: dataset },
+		layer: createLayers(options),
+		config: {
+			font: globalFont
+		}
+	};
+	return spec;
+};
```
mwdchang, line 115: Do we have tooltip capabilities here?  May be good if we can show the actual quantities?

YohannParis (approved):

packages/client/hmi-client/src/services/charts.ts
jryu01: I'll add the tooltip

jryu01: Screenshot 2024-07-24 at 11 20 03 AM
Screenshot 2024-07-24 at 11 19 58 AM

mwdchang (approved):

PR #4217, opened 2024-07-22
Changed optimize chart labels

### LINKED ISSUE #4207: [FEAT]:
Improve success criteria chart on the optimize operator:

Chart axes labels:
 * y axis: Min value of [variable] at all timepoints
 * x axis: Number of samples

Legend:
* orange: failing
* green: passing

Screenshot 2024-07-22 at 11 13 44

jryu01:
# Description

Screenshot 2024-07-22 at 6 59 54 PM

Resolves #4207

### DIFF
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -357,7 +357,7 @@ export function createOptimizeChart(
 					y: {
 						field: 'start',
 						type: 'quantitative',
-						title: `${isMinimized ? 'Max' : 'Min'} value of ${targetVariable} at all times`
+						title: `${isMinimized ? 'Max' : 'Min'} value of ${targetVariable} at all timepoints`
 					},
 					y2: { field: 'end' },
 					x: {
@@ -370,7 +370,7 @@ export function createOptimizeChart(
 						field: 'tag',
 						type: 'nominal',
 						scale: {
-							domain: ['out', 'in'],
+							domain: ['failing', 'passing'],
 							range: ['#FFAB00', '#1B8073']
 						}
 					}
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -370,7 +370,7 @@ export function createOptimizeChart(
 						field: 'tag',
 						type: 'nominal',
 						scale: {
-							domain: ['out', 'in'],
+							domain: ['failing', 'passing'],
```
mwdchang, line 373: I am not sure if you can mess around with the domain like that, because `in` and `out` have some special meanings in @blanchco 's logic if I recall

jryu01: hm.. I think you are right. I'll take a look at if there are other ways.

blanchco: @jryu01 you may need to use something like this in the transform array:

```
{
        calculate: "datum.tag === 'out' ? 'failing' : 'passing'", // Add this line
        as: 'tagLabel' // Use a new field name to store the mapped value
}
```

and then use tagLabel as the new field name

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -372,7 +372,8 @@ export function createOptimizeChart(
 						scale: {
 							domain: ['out', 'in'],
 							range: ['#FFAB00', '#1B8073']
-						}
+						},
+						legend: { labelExpr: 'datum.value === "in" ? "passing" : "failing"' }
```
blanchco, line 376: Nice!

blanchco (approved):

YohannParis (approved):

Tom-Szendrey (approved):

PR #4219, opened 2024-07-22
fix tera input

blanchco:
# Description

* fix an issue where nist types were being parsed as strings, since update was being called twice
* can parse decimals now with `isTextContainingOnlyDigits`

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-input.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-input.vue
@@ -21,7 +21,7 @@
 </template>
 
 <script setup lang="ts">
-import { nistToNumber, numberToNist, scrubAndParse } from '@/utils/number';
+import { nistToNumber, nistToString, numberToNist, scrubAndParse } from '@/utils/number';
 import { isEmpty, isString } from 'lodash';
 import { CSSProperties, InputTypeHTMLAttribute, computed, onMounted, ref, watch } from 'vue';
 
@@ -77,7 +77,7 @@ function isTextContainingOnlyDigits(value: string | number | undefined): boolean
 }
 
 function getValue() {
-	if (isNistType || isTextContainingOnlyDigits(maskedValue.value)) return maskedValue.value;
+	if (isNistType || isTextContainingOnlyDigits(props.modelValue)) return maskedValue.value;
 	return props.modelValue;
 }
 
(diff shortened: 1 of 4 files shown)


shawnyama (approved):
Does this fix that issue that I had with passing a string instead of a number to pyciemss in model config?

blanchco:
> Does this fix that issue that I had with passing a string instead of a number to pyciemss in model config?

Yes this should fix that, there was an issue updating the input twice when we tried to support the NIST number on all input types

blanchco:
converting back to draft..having too many issues

blanchco:
latest changes should work...

packages/client/hmi-client/src/components/widgets/tera-input.vue
```diff
@@ -122,10 +122,8 @@ const unmask = () => {
 	// convert back to a number when finished
 	if (isNistType && !getErrorMessage.value) {
 		emit('update:model-value', nistToNumber(maskedValue.value));
-	}
-
-	if (isTextContainingOnlyDigits(maskedValue.value)) {
-		emit('update:model-value', nistToString(maskedValue.value));
+	} else if (isTextContainingOnlyDigits(maskedValue.value)) {
+		emit('update:model-value', nistToNumber(maskedValue.value));
```
YohannParis, line 126: Why the reversal, this is needed for expressions?
And if you have to remove it, you should remove the function as well.

mwdchang (changes requested):
Given how we have come to rely on this component, it would be good to have some unit tests that actually tests realistic scenarios of what we would enter as input and what we would expect. Then the next time someone tweaks this we have some assurance that there wouldn't be regressions and such things.


We kind of have a component under `tests/component/navbar.spec.ts` that sets up the barebone of things.

packages/client/hmi-client/src/components/widgets/tera-input.vue
blanchco: The component isn't really working as expected at the moment so I was trying some things. Im going to look into this a bit more...this component is not in a great state at the moment

YohannParis: Okay, because doing this might create some issue on the model page

blanchco:
Okay going to take this back and make the component a little more reliable

blanchco:
no longer needed

PR #4224, opened 2024-07-23
ta3 (simulation/calibration/optimization)result cache

mwdchang:
### Summary
Simple FIFO cache to reduce the number of API requests we make to fetch TA3 results.

In a nutshell we have a tendency to fetch the same set of things multiple times; for example forecast-result is used for both the preview-component and the full drilldown-component. Having an application-level cache means we make less API requests.

### Testing
Run simulate or calibrate, there shouldn't be any errors. 

The actual performance gain of the cache is hard to observe, because the other requests (e.g. fetching Dataset) overshadows the improvements made here. But roughly it takes around 400ms to fetch simulation results + light processing, with cache this drops to around 30-40ms.


YohannParis:
I do not understand why we are not caching on the back-end and provide proper caching tag on the request so that the browser would handle that itself?

mwdchang:
> I do not understand why we are not caching on the back-end and provide proper caching tag on the request so that the browser would handle that itself?

@YohannParis you are talking about two separate things, but we definitely could do one or both of them

- caching on the backend, sure
- caching tags is actually caching on the front end, this is a bit more complicated because the browser doesn't always follow your instructions and generally doesn't like to cache responses past some threshold size

The benefit of this client-side caching, is that it is easy to implement and you can more closely tailor it to the user's actions, this is also possible at the backend but your cache-tables/cache-maps will probably become more complicated (e.g. if you want to do per user or per session cache).

dvince2 (approved):

PR #4230, opened 2024-07-23
Add PG indexes

kbirk:
Unfortunately there isn't support in hibernate for partial indixes. Ideally we would index on `IS NULL` for `deletedOn` to speed those up.

We can put those in flyway, but due to how we setup the `baselineVersion` stuff, it would skip those.


YohannParis (approved):
LGTM

dgauldie (approved):

PR #4233, opened 2024-07-23
legends-on-node-charts

blanchco:
# Description

* These legends are a pain in the ass to work with...😢 
* Did my best here - essentially compacted the legends so that they can be seen in the node view
* The drilldown chart view should be the same as before

Screenshot 2024-07-23 at 11 43 06 AM

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -84,17 +84,20 @@ export const createForecastChart = (
 		labelExpr += " 'other'";
 	}
 
+	const isCompact = options.width < 200;
+
 	const legendProperties = {
 		title: null,
 		padding: { value: 0 },
 		strokeColor: null,
 		orient: 'top',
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -456,22 +478,14 @@ function createStatisticLayer(
...
-					? {
-							title: null,
-							orient: 'top',
-							direction: 'horizontal',
-							labelExpr: isPreStatistic
-								? 'datum.value ? "Before Optimization" : ""'
-								: 'datum.value ? "After Optimization" : ""'
-						}
-					: null
+				// we only want to render a legend for the first statistic layer
```
blanchco, line 487: only render the first statistic layer legend so we can vertically align it.  the legend otherwise doesnt play nice...

YohannParis (approved):

shawnyama (approved):
seems legit

PR #4235, opened 2024-07-23
Ts/3433 speed quality button

dvince2:
# Description

* Simulation presets.

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -266,6 +305,18 @@ const chartProxy = chartActionsProxy(props.node, (state: SimulateCiemssOperation
 	emit('update-state', state);
 });
 
+const setPresetValues = (data: PresetTypes) => {
+	console.log(data);
```
Tom-Szendrey, line 309: no need for this console log

Tom-Szendrey (approved):
IMO the labels should maybe be debug and preset 
-Debug as it should only be used for this 
-Preset as a generic wording
What "Normal" means can vary :P 
But this is more of a design thing than anything and a super minor change in future if we want

mwdchang:
> IMO the labels should maybe be debug and preset -Debug as it should only be used for this -Preset as a generic wording What "Normal" means can vary :P But this is more of a design thing than anything and a super minor change in future if we want

Yep this is the issue here, we have fast/debug and not fast/debug ... so there is really only one preset but we are modelling it as two.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss.vue
```diff
@@ -255,6 +282,18 @@ const outputs = computed(() => {
...
 	return [];
 });
+
+const presetType = computed(() => {
+	if (numSamples.value === speedValues.numSamples && method.value === speedValues.method) {
+		return PresetTypes.Fast;
+	}
+	if (numSamples.value === qualityValues.numSamples && method.value === qualityValues.method) {
+		return PresetTypes.Normal;
+	}
```
mwdchang, line 292: There is only "Fast" mode, everything that is not Fast is Normal

mwdchang (approved):

PR #4237, opened 2024-07-23
4062 add other values styling

asylves1:
# Description

image

image

Resolves #(issue)


packages/client/hmi-client/src/components/model/petrinet/tera-initial-other-value-modal.vue
```diff
@@ -1,89 +1,85 @@
 <template>
 	<tera-modal>
-		<section>
```
asylves1, line 3: removed unneeded section tag

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-other-value-modal.vue
```diff
@@ -213,18 +242,27 @@ const onCustomSelectionChange = (val) => {
 	}
 };
 
+function getValueString(value) {
+	if (!value) {
+		return '';
+	}
+	return displayNumber(value);
```
YohannParis, line 249:
```suggestion
	return value ? displayNumber(value) : '';
```

YohannParis (approved):

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-other-value-modal.vue
```diff
@@ -33,30 +35,36 @@
...
 							{{ data.referenceId }}
 						</template>
+						<template v-if="col.field === 'range'">
+							<tera-line-graphic
+								:id="data.index"
+								:max-value="maxValue"
+								:min-value="minValue"
+								:distribution="data.distribution"
+							/>
+						</template>
```
shawnyama, line 48: No need to wrap this in a template

packages/client/hmi-client/src/components/model/petrinet/tera-line-graphic.vue
```diff
@@ -0,0 +1,67 @@
...
+const circlePoint = ref<number>(0);
+const range = ref<Array<number>>([1, 100]);
+
+const max: number = props.maxValue;
+const min: number = props.minValue;
+
+const isUniformType = () => props.distribution.type === DistributionType.Uniform;
+
+onMounted(() => {
+	parentWidth.value = document.getElementById(graphicId)?.getBoundingClientRect().width ?? 0;
```
shawnyama, line 46: maybe parentWidth doesnt need to be a ref?

packages/client/hmi-client/src/components/model/petrinet/tera-line-graphic.vue
```diff
@@ -0,0 +1,67 @@
...
+	parentHeight.value = document.getElementById(graphicId)?.getBoundingClientRect().height ?? 0;
```
mwdchang, line 47: Generally speaking, don't use `id` attribute in components that are meant to be reused.

Moreover don't use document-level query selectors, cause you never know what you are gonna get. 

If you need to get the width/height you can do it through a element-ref in vue, it would also be a bit more idiomatic with the other things in the code base.

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-other-value-modal.vue
```diff
@@ -142,17 +151,37 @@ import Column from 'primevue/column';
...
+	id: any;
 	otherValueList: any;
 }>();
 
-const otherValueList = ref(props.otherValueList);
+const otherValueList = ref(
+	props.otherValueList.map((value, index) => {
+		value.index = index + 1;
+		return value;
+	})
```
mwdchang, line 165: Generally speaking it is good practice to not use mutators in `map` functions.

Also would prefer to not change things coming in through `props`

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-other-value-modal.vue
```diff
@@ -142,17 +151,37 @@ import Column from 'primevue/column';
...
+otherValueList.value.forEach((element) => {
+	if (element.distribution?.parameters?.maximum) {
+		maxValues.push(element.distribution?.parameters?.maximum);
+	}
+	if (element.distribution?.parameters?.minimum) {
+		minValues.push(element.distribution?.parameters?.minimum);
+	}
+});
+const minValue: number = min(minValues) ?? 0;
+const maxValue: number = max(maxValues) ?? 0;
```
mwdchang, line 179: This part is a bit faulty from a technical/completness sense, because you have a predicate conditioned on max/min, but they are numbers and can very well be 0 and disrupt your truthy checks

So for example if my maximums were `-2, -1, 0` the true max should be `0` but I think your algorithm will return me `-1`.

Also if you are sticking to D3, it has a method called `extent` where it figures out the max/min on your behalf, and you don't need to worry too much of what the values are in the process:

```
> d3.extent([1, 2, null, 4, 5, "blue", "green"])

< (2) [1, 5]
```

mwdchang (changes requested):
See comments, there are a couple of potential bugs

packages/client/hmi-client/src/components/model/petrinet/tera-line-graphic.vue
```diff
@@ -0,0 +1,65 @@
...
+<script setup lang="ts">
+import * as d3 from 'd3';
+import { onMounted, ref } from 'vue';
+import { DistributionType } from '@/services/distribution';
+
+const props = defineProps<{
+	maxValue: any;
+	minValue: any;
+	distribution: any;
+}>();
```
mwdchang, line 29: Is it possible to tighten up the props minValue, maxValue, distribution? 

Below you are treating them as if they have types so can we just do the typings here?

mwdchang (approved):

packages/client/hmi-client/src/components/model/petrinet/tera-line-graphic.vue
asylves1: will do

PR #4238, opened 2024-07-23
4208 show units on the chart for optimize drilldown

jryu01:
# Description

Show units on the axis for the optimize drilldown charts

Screenshot 2024-07-23 at 1 50 54 PM

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -584,6 +584,7 @@ const initialize = async () => {
 	const modelConfigurationId = props.node.inputs[0].value?.[0];
 	if (!modelConfigurationId) return;
 	modelConfiguration.value = await getModelConfigurationById(modelConfigurationId);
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -574,16 +584,15 @@ const initialize = async () => {
 	const modelConfigurationId = props.node.inputs[0].value?.[0];
 	if (!modelConfigurationId) return;
 	modelConfiguration.value = await getModelConfigurationById(modelConfigurationId);
-	const model = await getAsConfiguredModel(modelConfiguration.value);
+	model.value = await getAsConfiguredModel(modelConfiguration.value);
```
mwdchang, line 587: Can we add a FIXME here? It seems like we are getting model information, per @dvince2 this is not the right way to do it even though it may look like it is working currently.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
jryu01: I added the comment.

mwdchang (approved):

PR #4241, opened 2024-07-23
Add model from config id

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/ModelRepository.java
YohannParis, line 1: lol, are we bypassing hibernate now?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/ModelRepository.java
kbirk: The only way to do this was to, because we don't define it as an explicit relationship inside hibernate :P

mwdchang (approved):

PR #4243, opened 2024-07-23
Display variable name and unit on model config operator

YohannParis:
# Description

* Clean up unused function/method/constant
* Change where we get intials metadata, from the metadata properties to the model object


shawnyama (approved):
🧹

PR #4244, opened 2024-07-23
digging out model metadata for simulate/calibrate charts

mwdchang:
### Summary
Tidy up calibrate and simulate charts with more canonical title, x/y axis labels. 

TODO: Still missing is the units of selected variables, this is less straight forward as everything in the model is treated as a variable, so it can be State, Parameter, Observable, Time, and other things that pyciemss adds during the simulation. 

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
@@ -31,12 +31,13 @@ import {
 } from '@/services/models/simulation-service';
 import { Poller, PollerState } from '@/api/api';
 import { logger } from '@/utils/logger';
-import { chartActionsProxy, nodeOutputLabel } from '@/components/workflow/util';
+import { chartActionsProxy, nodeOutputLabel, getModelByModelConfigurationId } from '@/components/workflow/util';
 
 import type { WorkflowNode } from '@/types/workflow';
 import { createLLMSummary } from '@/services/summary-service';
 import { createForecastChart } from '@/services/charts';
 import VegaChart from '@/components/widgets/VegaChart.vue';
+import { Model } from '@/types/Types';
 import { SimulateCiemssOperationState, SimulateCiemssOperation } from './simulate-ciemss-operation';
 
 const props = defineProps<{
@@ -44,6 +45,7 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits(['open-drilldown', 'update-state', 'append-output']);
+const model = ref<Model | null>(null);
 const runResults = ref<{ [runId: string]: DataArray }>({});
 const runResultsSummary = ref<{ [runId: string]: DataArray }>({});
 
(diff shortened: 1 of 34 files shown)


packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -4,9 +4,43 @@ import type { CsvAsset, TimeSpan } from '@/types/Types';
...
+import { getModelConfigurationById } from '@/services/model-configurations';
+import { getModel } from '@/services/model';
+import { Model } from '@/types/Types';
+
+// FIXME: replace with single API call
+export const getModelByModelConfigurationId = async (id: string) => {
+	const modelConfig = await getModelConfigurationById(id);
+	const model = await getModel(modelConfig.modelId as string);
+	return model;
+};
```
YohannParis, line 16: You can use the new endpoint `/models/from-model-configuration/{model-config-id}` now.

packages/client/hmi-client/src/page/data-explorer/components/tera-asset-card.vue
```diff
@@ -201,7 +201,9 @@ const title = computed(() => {
 	let value = '';
 	if (props.resourceType === ResourceType.XDD) {
 		value =
-			props.source === DocumentSource.XDD ? (props.asset as Document).title : (props.asset as DocumentAsset).name ?? '';
+			props.source === DocumentSource.XDD
+				? (props.asset as Document).title
+				: ((props.asset as DocumentAsset).name ?? '');
```
mwdchang, line 206: I don't know what happened here, the formatter has a mind of its own

packages/client/hmi-client/src/components/workflow/util.ts
mwdchang: done thx

packages/client/hmi-client/src/page/data-explorer/components/tera-asset-card.vue
YohannParis: Nope, I updated Prettier from `3.3.2` to `3.3.3` and that is what happened...

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -146,6 +149,7 @@ const preparedCharts = computed(() => {
 	Object.keys(pyciemssMap).forEach((key) => {
 		reverseMap[`${pyciemssMap[key]}_mean`] = key;
 	});
+	const xAxisTitle = model.value?.semantics?.ode.time.units?.expression;
```
shawnyama, line 152: I think this should be computed since the `model` could change. Maybe not needed in the drilldown

shawnyama (approved):

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
mwdchang: Noted, but I think that is going to be a general problem with our hand-wavy way of updating things.

The node-component has no means of knowing when the model itself has changed, when all it has is an model-configuration-id (not even a model-id).

PR #4247, opened 2024-07-23
Fixing wrong model configuration selection

dvince2:
# Description

* After creating a new model configuration, we weren't properly selecting it from the the list of model configs. As a result the output and the download would be the last selected/created config even though it *looked* like it shouldn't be.
* In addition I've found more casting exceptions because of the presence of `Object` in the back end. This should be `Number` but that could break our python services down stream, and requires a migration and I just want to get it in.
* Lastly, we should return the new model we've created ;)

Resolves #noissue

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
@@ -517,6 +517,8 @@ const createConfiguration = async () => {
 		return;
 	}
 
+	knobs.value.transientModelConfig = cloneDeep(data);
+	state.transientModelConfig = knobs.value.transientModelConfig;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -517,6 +517,8 @@ const createConfiguration = async () => {
 		return;
 	}
 
+	knobs.value.transientModelConfig = cloneDeep(data);
+	state.transientModelConfig = knobs.value.transientModelConfig;
```
YohannParis, line 521: You can make it weird with just a one liner
```suggestion
	state.transientModelConfig = knobs.value.transientModelConfig = cloneDeep(data);
```

YohannParis (approved):

shawnyama (approved):

PR #4249, opened 2024-07-23
Show units on the chart for the simulation and calibration drilldown

### LINKED ISSUE #4208: [FEAT]: Display units on charts
All charts should display the units on their axis when dealing with model variables.

Screenshot 2024-07-22 at 11 16 09

jryu01:
# Description

**Simulation drilldown charts**
Screenshot 2024-07-23 at 4 54 47 PM

**Calibration drilldown charts**
Screenshot 2024-07-23 at 3 39 16 PM

Resolves #4208


packages/client/hmi-client/src/services/model.ts
```diff
@@ -165,6 +165,16 @@ export async function getModelEquation(model: Model): Promise<string> {
 	return latex ?? '';
 }
 
+export const getModelVariableUnits = (model: Model) => {
```
shawnyama, line 168: Chatted with @mwdchang about calling states, parameters, transitions, etc as `variables`, we settled for calling them `parts`. `variables` could be mixed with `parameters`. So `getUnitsFromModelParts`

shawnyama (approved):

packages/client/hmi-client/src/services/model.ts
jryu01: Done

packages/client/hmi-client/src/services/model.ts
```diff
@@ -173,6 +173,16 @@ export async function getModelEquation(model: Model): Promise<string> {
...
 
+export const getUnitsFromModelParts = (model: Model) => {
+	const modelVariableUnits: { [key: string]: string } = {
+		_time: model?.semantics?.ode?.time?.units?.expression || ''
+	};
+	[...(model?.model.states ?? []), ...(model?.semantics?.ode?.parameters ?? [])].forEach((v) => {
+		modelVariableUnits[v.id] = v.units?.expression || '';
+	});
+	return modelVariableUnits;
+};
```
mwdchang, line 184: If we do it this way, to pre-calculate the units per variable - we also need to compute for
- observables (which has no expression but their children do), and
- time

YohannParis (approved):

packages/client/hmi-client/src/services/model.ts
jryu01: I updated it so that it grabs the unit for the observable from the first state in `observables.states`. I wonder if there are cases where the units from states are different and how we should handle in such case.

PR #4258, opened 2024-07-24
better success criteria charts

blanchco:
# Description

* Improving the success criteria chart based on @liunelson's feedback (see attached ticket)
* binning is now done by vegalite rather than our own function

Below:
Screenshot 2024-07-24 at 11 25 22 AM

Above:

Screenshot 2024-07-24 at 11 41 44 AM


packages/client/hmi-client/tests/unit/utils/chart.spec.ts
blanchco, line 1: We no longer do the binning here it is done by vegalite so the function has been removed

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -253,19 +253,8 @@
...
-											:visualization-spec="
-												createOptimizeChart(
-													riskResults[knobs.postForecastRunId],
-													constraint.targetVariable,
-													constraint.threshold,
-													constraint.isMinimized
-												)
-											"
-										/>
+									<li v-for="(_constraint, key) in node.state.constraintGroups" :key="`success_criteria_${key}`">
```
Tom-Szendrey, line 256: Whats the point of the :key being a static string concat with the index?

shawnyama (approved):
lgtm

Tom-Szendrey (approved):
Thanks

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
blanchco: I thought it was screwing with other list objects with the same key but thats not even an issue 😆. Its not needed, Ill make it just `key`

PR #4261, opened 2024-07-24
style: clean up slider and carousel

shawnyama:
# Description

* I noticed an random chevron was placed in the slider (in the Data Explorer) and then I figured out it was from the carousel in a `tera-asset-card`
* we are getting rid of carousels from the asset cards but in case the carousel is used anywhere else in the future this would be helpful
* I am commenting out the chevrons for now since I am not coming across carousels int the data explorer anymore..., we'll worry about them once we use them again
* I couldn't resist writing a more readable `tera-slider`

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-carousel.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-carousel.vue
@@ -20,13 +20,14 @@
 				<li v-if="isNumeric && itemCount > 5">(+{{ itemCount }})</li>
 			</ul>
 		</nav>
+		<!-- FIXME: If we are going to use this consider making the chevron navigations work (they should positioned properly)
 		<Button
 			v-if="itemCount > 1"
 			text
 			severity="secondary"
 			icon="pi pi-chevron-left"
 			@click.stop="move(currentPage - 1)"
-			class="go-back"
+			class="back"
 		/>
 		<Button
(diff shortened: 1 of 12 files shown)


packages/client/hmi-client/src/components/widgets/tera-carousel.vue
```diff
@@ -20,29 +20,30 @@
 				<li v-if="isNumeric && itemCount > 5">(+{{ itemCount }})</li>
 			</ul>
 		</nav>
+		<!-- FIXME: If we are going to use this consider making the chevron navigations work (they should positioned properly)
```
blanchco, line 23: can we just get rid of all the commented out stuff 😄

blanchco (approved):

packages/client/hmi-client/src/components/widgets/tera-carousel.vue
shawnyama: Nah we may wanna add it back later. I just didn't fix it because I can't find a carousel in the app anymore! We'll worry about adding them back once we need em.

packages/client/hmi-client/src/components/widgets/tera-slider-panel.vue
YohannParis, line 1: You made my day!

YohannParis (approved):

PR #4266, opened 2024-07-25
Calibrate model config

mwdchang:
### Summary
Create model-config (pointing to the calibrated dill file) in calibrate, instead of a special calibrate-simulation-id

image

### Testing
Run calibrate, and attach the output to simulation, run simulation and check simulation result to see that they fall within the expected range.


Tom-Szendrey:
Optimize drilldown shares the same `inferred_parameters` logic as simulate does.
If that is to be done in a follow up PR or you want to throw that my way no problem

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -273,11 +274,24 @@ watch(
 			`;
 			const summaryResponse = await createLLMSummary(prompt);
 
-			const portLabel = props.node.inputs[0].label;
+			const parentModel = await getModelByModelConfigurationId(modelConfigId.value as string);
+			const calibratedModelConfig: ModelConfiguration = {
+				name: 'Calibrated configuration',
+				description: 'Calibrated configuration',
+				simulationId: state.calibrationId,
```
Tom-Szendrey, line 281: Given that we have two uses for the word Simulation (Object + forecast)
and that this is always expected to be a calibration's run id, is there a reason why we are not calling it calibrationId?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -273,11 +274,24 @@ watch(
 			`;
 			const summaryResponse = await createLLMSummary(prompt);
 
-			const portLabel = props.node.inputs[0].label;
+			const parentModel = await getModelByModelConfigurationId(modelConfigId.value as string);
+			const calibratedModelConfig: ModelConfiguration = {
+				name: 'Calibrated configuration',
+				description: 'Calibrated configuration',
```
YohannParis, line 280: I think those should use the same name as the original modelConfiguration but with a prepend like `Calibrated`.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
mwdchang: The "new" model configuration object is a bit vague - it has both simulatonId and calibrationId, I asked which one should be used for calibrated-model-configs, I got back "simulationId" ... CC @blanchco

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
mwdchang: sure

PR #4267, opened 2024-07-25
optimize uses forecast chart

blanchco:
# Description

* the motivation behind this PR was so that the optimize charts can stay consistent style wise rather than optimize constantly catching up to the styles of the other forecast charts
* optimize now uses forecast charts
* added annotations layer to forecast chart, you can essentially add any number of layers here needed for annotations (I needed a rule and a text layer)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -366,6 +366,7 @@ import AccordionTab from 'primevue/accordiontab';
 import { createSuccessCriteriaChart, createForecastChart, createInterventionChartMarkers } from '@/services/charts';
 import VegaChart from '@/components/widgets/VegaChart.vue';
 import MultiSelect from 'primevue/multiselect';
+import { mergeResults, renameFnGenerator } from '@/components/workflow/ops/calibrate-ciemss/calibrate-utils';
 import teraOptimizeCriterionGroupForm from './tera-optimize-criterion-group-form.vue';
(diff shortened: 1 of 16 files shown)


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -146,6 +141,7 @@ export const createForecastChart = (
 	samplingLayer: ForecastChartLayer | null,
 	statisticsLayer: ForecastChartLayer | null,
 	groundTruthLayer: ForecastChartLayer | null,
+	annotationLayers: any[],
```
blanchco, line 144: This can be an array of any vegalite spec...ideally keep it to annotations only.  I needed 2 layers for a rule layer and a text layer

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -371,6 +377,7 @@ import {
 	OptimizeCiemssOperationState,
 	OptimizationInterventionObjective
 } from './optimize-ciemss-operation';
+import { mergeResults, renameFnGenerator } from '../calibrate-ciemss/calibrate-utils';
```
YohannParis, line 380: Please do not use relative path, could you use path with an `@`

packages/client/hmi-client/src/services/charts.ts
mwdchang: I am hesitant to make things too open ended, because then you might as well extract the json and just do things after-the-fact, e.g:

```
const blah = createForecastChart(....)
blah.layers.push({
});
```

blanchco: That makes sense, maybe Ill keep the function as is and do what you said here

mwdchang (approved):

YohannParis (approved):

PR #4270, opened 2024-07-25
feat: initialize default config once input is connected instead of once drilldown is opened

shawnyama:
# Description

* Append the default configuration to the output without the need to open the drilldown
* Renamed `tera-intervention-policy-operation.ts` to `intervention-policy-operation.ts` (wanted to do this in my intervention PR but forgot)

https://github.com/user-attachments/assets/0ef4c052-c27a-4a64-bf23-b6c1aa0d85e4


YohannParis (approved):
Thank you!

packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
```diff
@@ -14,6 +15,17 @@ export interface ModelConfigOperationState extends BaseState {
...
+export const blankModelConfig: ModelConfiguration = {
+	id: '',
+	modelId: '',
+	name: '',
+	description: '',
+	calibrationRunId: '',
+	observableSemanticList: [],
+	parameterSemanticList: [],
+	initialSemanticList: []
+};
```
mwdchang, line 27: We are actually using `ModelConfiguration.simulationId` and not the "calibrationRunId"

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -12,59 +12,61 @@
...
+			if (configs[0].id) {
+				const config = configs[0];
+				state.transientModelConfig = config;
+				emit('update-state', state);
+				emit('append-output', {
+					type: ModelConfigOperation.outputs[0].type,
+					label: config.name,
+					value: config.id,
+					isSelected: false,
+					state: props.node.state
```
mwdchang, line 61: This is not the same state object that you just modified (which is a cloned copy), I am guessing you want to emit the cloned version.

mwdchang (approved):

PR #4274, opened 2024-07-25
auto focus modal input and wrap tera-modal in `<teleport>` by default

shawnyama:
# Description

* First take a look at `tera-modal.vue` - I focus on the first input box available once it mounts
* I also decided to wrap `tera-modal` with `<teleport>` by default so there isn't a need to do so every time you use it
* _Me removing all those old `teleport` tags makes this PR look huge_ 
* Removed some old/unused classes that were passed to `tera-modal` and made passing a specified width to the modal easier
* Similar to `tera-modal` I wrapped `tera-drilldown` with `teleport`

https://github.com/user-attachments/assets/15347678-7a23-45c5-9960-91382cf5534a


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -144,19 +144,17 @@
...
+		v-if="dialogIsOpened && currentActiveNode"
+		:is="registry.getDrilldown(currentActiveNode.operationType)"
+		:node="currentActiveNode"
+		@append-output="(event: any) => appendOutput(currentActiveNode, event)"
+		@update-state="(event: any) => updateWorkflowNodeState(currentActiveNode, event)"
+		@update-status="(event: any) => updateWorkflowNodeStatus(currentActiveNode, event)"
+		@select-output="(event: any) => selectOutput(currentActiveNode, event)"
+		@close="addOperatorToRoute(null)"
+		@update-output-port="(event: any) => updateOutputPort(currentActiveNode, event)"
+	/>
```
mwdchang, line 157: I am not exactly sure what is the ramification here, the drilldowns were originally designed such that they can be opened into a new tab/window. Now that is not entirely true anymore because some operators have dependencies on the node to kick things off or to set a flag, but nevertheless I am wondering if we still want to be able to "open in a new tab" and what this might look like? @YohannParis

shawnyama: This change assumes the drilldown will always be a modal, (I wrapped `tera-drilldown` with a `<teleport>`). But if we still want to open the drilldown in a new window then this change gets in the way of that.

packages/client/hmi-client/src/components/navbar/tera-navbar.vue
```diff
@@ -45,123 +45,118 @@
...
-						<Button size="large" class="p-button-secondary" outlined @click="isEvaluationScenarioModalVisible = false"
-							>Close</Button
-						>
-					</div>
-					<div class="align-self-center">
-						<p>Runtime {{ evaluationScenarioRuntimeString }}</p>
+		<tera-modal
+			v-if="isEvaluationScenarioModalVisible"
+			@modal-mask-clicked="isEvaluationScenarioModalVisible = false"
+			class="w-7"
```
YohannParis, line 51: 🥇

packages/client/hmi-client/src/components/widgets/tera-modal.vue
```diff
@@ -29,9 +31,15 @@
  * Clicking outside the modal will emit the event modalMaskClicked.
  * @example
  * <modal @modal-mask-clicked="closeModal"></modal>
+ *
+ * To easily specify modal width add a tailwind width class by passing it as a prop. eg. `class="w-4"`
+ *
+ * Due to this component being wrapped in a <teleport> tag, :deep() classes specified
+ * in the parent's <style scoped> won't be inherited since this component spawns as a child of the <body>.
+ * :deep() only effects child tags and obviously the <body> tag is not a child of the parent.
```
YohannParis, line 39: Great comment

packages/client/hmi-client/src/components/widgets/tera-modal.vue
```diff
@@ -54,6 +62,14 @@ onMounted(() => {
 			detail: { id: element.id ?? '' }
 		});
 		element.dispatchEvent(event);
+
+		// Focus the first input element in the default slot
+		nextTick(() => {
+			const input = element.querySelector('section.content input') ?? element.querySelector('section.content textarea');
+			if (input) {
+				(input as HTMLElement).focus();
+			}
```
YohannParis, line 71: Why not use a ref here?

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
YohannParis: Yes, we still want to open drilldown on new windows, that would really help the UX, specially once we implement the arrows to move up and down the worflow

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-modal.vue
shawnyama: A ref is actually used but I guess it's not as obvious due to the variable name, I shall fix

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
shawnyama: Ok I'll revert this

PR #4285, opened 2024-07-26
fix: model config beaker initialization

blanchco:
# Description

* fixes an issues with model configuration node where beaker was always be initialized twice causing some messy logs


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -628,10 +628,6 @@ const resetConfiguration = () => {
 	});
 };
 
-onMounted(async () => {
```
blanchco, line 631: Since we have `immediate:true` on the watcher, `initialize()` was always being called _twice_.  Now initialize will always be called when the component is created and whenever then output changes _one_ time

YohannParis (approved):

mwdchang (approved):
ok!

PR #4286, opened 2024-07-26
feat: preview model parts

shawnyama:
# Description

- Use `featureConfig` to specify if we are previewing data

Notes:
- If there is no expression or description the gaps from the CSSGrid are still there causing a bit of a larger margin at the bottom of the list items, tried to fix it with a computed style but it was being weird - will handle later

image
image


packages/client/hmi-client/src/components/widgets/tera-input.vue
```diff
@@ -35,6 +39,8 @@ const props = defineProps<{
 	type?: InputTypeHTMLAttribute | 'nist';
 	placeholder?: string;
 	autoWidth?: boolean;
+	featureConfig?: FeatureConfig;
+	previewTag?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span';
```
mwdchang, line 43: We should have a chat about what it is exactly we want to do? It isn't too clear from the PR description. It is possible we are putting the abstractions in the wrong places for the sake of avoid repeating similar code, but it has other costs.

I feel like we should strive to keep the basic components simple and not overload it with different modalities, as it is easy to induce errors in other places. Also @blanchco is refactoring (splitting) this component into two separate things ... so there is that ...

shawnyama: Okay sure I'll revert that

blanchco (approved):
👍

YohannParis (approved):

Issue #4033, opened 2024-07-03
[TASK]: House cleaning


mwdchang:
### To remove/repace
- [ ] #4294

### No longer supported or in active use
We should keep these around for now, mark them as deprecated and just hide them from the user interface.
- [x] CodeOperator
- [x] Sciml Simulate
- [x] Sciml Calibrate

### Triage and candidates for removal
Check if the following are being used, if not remove from both API and UI
- [x] #4287

PR #4288, opened 2024-07-26
task: remove mathml to acset

blanchco:
# Description

* these endpoints are not being used anymore so I have removed them.
* a part of House cleaning #4033

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/TransformController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/TransformController.java
@@ -31,28 +31,6 @@ public class TransformController {
 	final SimulationServiceProxy simulationServiceProxy;
 	final Messages messages;
 
-	@PostMapping("/mathml-to-acset")
-	@Secured(Roles.USER)
-	public ResponseEntity<JsonNode> mathML2ACSet(@RequestBody final List<String> list) {
-		try {
-			return skemaProxy.convertMathML2ACSet(list);
-		} catch (final FeignException error) {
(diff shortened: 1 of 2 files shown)


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/TransformController.java
```diff
@@ -31,28 +31,6 @@ public class TransformController {
 	final SimulationServiceProxy simulationServiceProxy;
 	final Messages messages;
 
-	@PostMapping("/mathml-to-acset")
-	@Secured(Roles.USER)
-	public ResponseEntity<JsonNode> mathML2ACSet(@RequestBody final List<String> list) {
-		try {
-			return skemaProxy.convertMathML2ACSet(list);
```
dvince2, line 38: If this is unused it can come out of `SkemaRustProxy.java` as well

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/TransformController.java
```diff
@@ -31,28 +31,6 @@ public class TransformController {
 	final SimulationServiceProxy simulationServiceProxy;
 	final Messages messages;
 
-	@PostMapping("/mathml-to-acset")
```
dvince2, line 34: The member variable `skemaProxy` can also be removed

blanchco: ahh yes done!

PR #4291, opened 2024-07-26
Populate simulation results to fix progress ordering

kbirk:


mwdchang (approved):
What are these `@Observed(name = "function_profile")` things?

YohannParis (approved):
Do you have Node and Yarn installed?
It looks like the Prettier linter for the backend is not running.

YohannParis:
> What are these `@Observed(name = "function_profile")` things?

Those are for the Grafana Dashboard that @dgauldie set up.

dvince2 (approved):

PR #4300, opened 2024-07-29
use calibrated model-config

mwdchang:
### Summary
Use calibrated model-config instead of the special calibratedSimulationId

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -713,9 +713,7 @@ const runOptimize = async () => {
 	};
 
 	// InferredParameters is to link a calibration run to this optimize call.
-	if (modelConfiguration.value) {
-		optimizePayload.extra.inferredParameters = modelConfiguration.value.simulationId;
-	}
+	optimizePayload.extra.inferredParameters = modelConfiguration.value.simulationId;
 
 	const optResult = await makeOptimizeJobCiemss(optimizePayload, nodeMetadata(props.node));
 	const state = _.cloneDeep(props.node.state);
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -714,8 +713,8 @@ const runOptimize = async () => {
 	};
 
 	// InferredParameters is to link a calibration run to this optimize call.
-	if (inferredParameters.value) {
-		optimizePayload.extra.inferredParameters = inferredParameters.value[0];
+	if (modelConfiguration.value) {
```
Tom-Szendrey, line 716: We already check for modelConfiguration at the start of this function
We should be able to just add this to the payload when we define it a few lines above if this is the case

What do we send for a non-calibrated model? Has this been tested?
pyciemss-service expects None 
https://github.com/DARPA-ASKEM/pyciemss-service/blob/main/service/models/converters.py#L10-L24

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -714,8 +713,8 @@ const runOptimize = async () => {
...
+		optimizePayload.extra.inferredParameters = modelConfiguration.value.simulationId;
```
Tom-Szendrey, line 717: What is sent if this simulationId has not been filled in by running a calibration?
pyciemss-service will accept this as an optional field and will provide None if it is not filled in:

https://github.com/DARPA-ASKEM/pyciemss-service/blob/main/service/models/operations/optimize.py

https://github.com/DARPA-ASKEM/pyciemss-service/blob/main/service/utils/tds.py#L192-L194

Tom-Szendrey (approved):
Thank you for this

blanchco (approved):

YohannParis (approved):

PR #4302, opened 2024-07-29
separate input components

blanchco:
# Description

* separate out number and text inputs from tera-input component
* `tera-input-text` formats text as nist onBlur when the input is numbers only
* `tera-input-number` formats text as nist whilst user is typing and will always output `number` type


packages/client/hmi-client/src/components/widgets/tera-input-number.vue
blanchco, line 1: Some slight adjustments here, this code is mostly from when `tera-input` was stable regarding number inputs.  Now it _only_ handles `nist` numbers

YohannParis (approved):

packages/client/hmi-client/src/assets/css/theme/designer/components/input/_terainput.scss
blanchco, line 1: thoughts on this? the components share the same styles

packages/client/hmi-client/src/components/widgets/tera-input-text.vue
```diff
@@ -0,0 +1,96 @@
...
+	<aside v-if="getErrorMessage"><i class="pi pi-exclamation-circle" /> {{ getErrorMessage }}</aside>
+</template>
+
+<script setup lang="ts">
+import { numberToNist } from '@/utils/number';
+import { isNumber, isString, toNumber, isNaN } from 'lodash';
+import { CSSProperties, computed, ref } from 'vue';
+
+const props = defineProps<{
+	modelValue: string | number | undefined;
```
mwdchang, line 30: For both input-text and input-number ... if the idea of the PR is to clamp down the typing, why is the `modelValue` so loosey-goosey?

blanchco: Fair, I've changed the type of `modelValue` to `number | undefined` in `tera-input-number`.

packages/client/hmi-client/src/components/widgets/tera-input-text.vue
```diff
@@ -0,0 +1,96 @@
...
+
+const updateValue = (event: Event) => {
+	const target = event.target as HTMLInputElement;
+	const value = target.value;
+
+	console.log(value);
+	emit('update:model-value', value);
+};
+
+const onBlur = () => {
```
YohannParis, line 93: Tera-input-number should have this as well. Formatting only onblur

mwdchang (approved):
looks okay, thanks

packages/client/hmi-client/src/components/widgets/tera-input-text.vue
blanchco: Done

PR #4313, opened 2024-07-30
fix yaxis label extraction expression

mwdchang:
### Summary
`_.uniq`  causing a problem splitting a single variable string into a character-array

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
@@ -147,10 +147,7 @@ const preparedCharts = computed(() => {
 				height: 120,
 				legend: true,
 				xAxisTitle: modelVarUnits.value._time || 'Time',
-				yAxisTitle:
-					_.uniq(modelVarUnits.value[variable])
-						.filter((v) => !!v)
-						.join(',') || '',
+				yAxisTitle: modelVarUnits.value[variable] || '',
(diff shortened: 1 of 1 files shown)


shawnyama (approved):

blanchco (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -147,10 +147,7 @@ const preparedCharts = computed(() => {
 				height: 120,
 				legend: true,
 				xAxisTitle: modelVarUnits.value._time || 'Time',
-				yAxisTitle:
-					_.uniq(modelVarUnits.value[variable])
-						.filter((v) => !!v)
-						.join(',') || '',
+				yAxisTitle: modelVarUnits.value[variable] || '',
```
YohannParis, line 150: Why not using a Set?
```js
Array.from(New Set(modelVarUnits.value[variable])).filter((v) => !!v).join(',') || ''
```

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
mwdchang: Optimized chart is a bit different, in that it is always single-variable. I was lazy and copy-pasted the array expression we use elsewhere but that just made things unnecessarily complicated so ...

PR #4319, opened 2024-07-30
Ensure result files are exported correctly.

kbirk:
Still need to test.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
@@ -69,14 +69,14 @@ public class SimulationService extends TerariumAssetServiceWithoutSearch<Simulat
 	) {
 		final Simulation simulation = getAsset(simulationId, hasReadPermission).orElseThrow();
 
-		final SimulationUpdate updated = simulationUpdateRepository.save(update);
+		update.setSimulation(simulation);
+		final SimulationUpdate created = simulationUpdateRepository.save(update);
 
-		updated.setSimulation(simulation);
-		simulation.getUpdates().add(updated);
+		simulation.getUpdates().add(created);
 
 		repository.save(simulation);
 
-		return updated;
+		return created;
 	}
 
 	@Observed(name = "function_profile")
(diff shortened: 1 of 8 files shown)


packages/server/src/main/resources/application.properties
```diff
@@ -8,6 +8,7 @@ spring.datasource.initialize=false
 spring.jpa.hibernate.ddl-auto=none
 spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 spring.flyway.enabled=false
+spring.datasource.hikari.maximum-pool-size=64
```
bigglesandginger, line 11: This will be overridden by Orchestration - and I really do think we are doing something very incorrectly - but it is hidden in Hibernatese and thus so difficult to wend the thread

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
```diff
@@ -84,4 +93,41 @@ public void copySimulationResultToDataset(final Simulation simulation, final Dat
 			}
 		}
 	}
+
+	@Observed(name = "function_profile")
+	public Map<String, FileExport> exportAssetFiles(final UUID simId, final Schema.Permission hasReadPermission)
```
bigglesandginger, line 98: This looks like the money, will it be imported correctly - go into S3 somehow?

packages/server/src/main/resources/application.properties
kbirk: Yeah this is just for local dev mainly. The issue we ran into was with workflows that would spawn 100+ individual  requests, a majority of which are async taskrunner requests that each write to the db multiple times. So with only default 10 connections, causes everything to timeout.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
kbirk: Yes, I change the format of how we import the `FileExport` class, so that i can specify the fullpath, since result files do into a diff key pattern than filenames.

YohannParis (approved):

packages/server/src/main/resources/application.properties
bigglesandginger: Each task runner holds a connection for the entire duration!?!?!?!?!!?!?!?!?!?!?!?!?

bigglesandginger (approved):

PR #4322, opened 2024-07-30
Updating optimize to have presets

dvince2:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -6,7 +6,7 @@
 		@on-close-clicked="emit('close')"
 		@update-state="(state: any) => emit('update-state', state)"
 	>
-		<section :tabName="OptimizeTabs.Wizard" class="ml-4 mr-2 pt-3">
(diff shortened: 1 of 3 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -572,6 +599,35 @@ const formatJsonValue = (value) => {
...
+	maxiter: 0,
+	maxfeval: 1
+};
+
+const qualityValues = {
+	numSamplesToSimModel: 100,
+	method: CiemssMethodOptions.dopri5,
+	maxiter: 5,
+	maxfeval: 25
+};
```
mwdchang, line 629: May want to consider adding `Object.freeze` to the presets to lower the risk of someone accidentally changing things

```
const speedValues = Object.freeze({
  blahblah: 123
})
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

dvince2: oh neat! I'm going to bring this to the other two as well.

mwdchang (approved):

PR #4326, opened 2024-07-30
Add cache control headers for some static payloads

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DocumentController.java
```diff
@@ -516,7 +519,9 @@ public ResponseEntity<byte[]> downloadDocument(
 	) {
 		try {
 			final Optional<byte[]> bytes = documentAssetService.fetchFileAsBytes(id, filename);
-			return bytes.map(ResponseEntity::ok).orElse(null);
+
+			final CacheControl cacheControl = CacheControl.maxAge(24, TimeUnit.HOURS).cachePublic();
+			return ResponseEntity.ok().cacheControl(cacheControl).body(bytes.orElse(null));
```
YohannParis, line 524: Should we cache something that could be null? Artifact and Simulation controllers do not respond with OK if it's null.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/S3Service.java
```diff
@@ -526,21 +533,24 @@ public ResponseEntity<StreamingResponseBody> getDownloadStream(final String buck
 			}
 		};
 
+		final CacheControl cacheControl = CacheControl.maxAge(24, TimeUnit.HOURS).cachePublic();
```
YohannParis, line 536: Should we put that `24` in the application.properties @dvince2?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/S3Service.java
dvince2: I'm always a fan of getting rid of magic numbers - either as a static final at the top or in application.properties

dvince2 (approved):

PR #4329, opened 2024-07-31
Enable menu button on hover over output port

asylves1:
# Description

image

Will be adding the dynamic options base on node in the next PR
Resolves #(issue)


packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+	<section ref="operatorMenu" :style="style">
```
YohannParis, line 2: I would prefer an `nav` in that case.
```suggestion
	<nav ref="operatorMenu" :style="style">
```

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,77 @@
...
+		<Button
+			v-if="node"
```
YohannParis, line 4: Why the `v-if="node"`? The component should not be visible if there is no node right?

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,77 @@
...
+	}
+]);
+
+function showMenu(event) {
+	menu.value.show(event);
+	isMenuShowing.value = true;
+}
+</script>
+<style scoped>
+.button {
```
YohannParis, line 66: You do not need a class, just use `button`, the style is scoped here.

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,77 @@
...
+<script setup lang="ts">
+import { ref } from 'vue';
+import type { WorkflowNode } from '@/types/workflow';
+import Menu from 'primevue/menu';
+import Button from 'primevue/button';
+
+defineProps<{
+	node: WorkflowNode<any> | null;
+	style: { top: string; left: string };
+	currentPortPosition: any;
```
YohannParis, line 31: I'm wondering if you should leave this to the parent on how to display the component.
Try using relative absolute positioning with the parent instead.

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,77 @@
...
+			icon="pi pi-plus"
+			:style="{
+				top: `${currentPortPosition.y - 18}px`,
+				left: `${currentPortPosition.x + 40}px`
+			}"
+			aria-haspopup="true"
+			aria-controls="overlay_menu"
+			severity="secondary"
+		/>
+		<Menu ref="menu" id="overlay_menu" :model="menuItems" :popup="true" />
```
YohannParis, line 19: This should display the main menu items from the Workflow but filtered by types.
We can have a chat about it if you want more details.

asylves1: Ya, that is the plan with the next PR. 
Was just looking into the `WorkflowRegistry()` in `tera-workflow.vue` any suggestions are welcome :)

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -89,6 +89,18 @@
...
+				:node="nodeWithOpenedMenu"
+				:style="{
+					top: `${currentPortPosition.y - 18}px`,
+					left: `${currentPortPosition.x + 40}px`
+				}"
+				@mouseenter="() => (isMenufocused = true)"
+				@mouseleave="() => (isMenufocused = false)"
+				@focus="() => (isMenufocused = true)"
+				@blur="() => (isMenufocused = false)"
+			/>
```
mwdchang, line 103: This should not be a first-class item on the canvas, it makes things more complicated to manage the more floating-things we have.

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,74 @@
...
+
+defineProps<{
+	node: WorkflowNode<any> | null;
+	style: { top: string; left: string };
+}>();
+
+const openMenuButton = ref<HTMLElement>();
+
+const isMenuShowing = ref<boolean>(false);
+const operatorMenu = ref<HTMLElement>();
```
shawnyama, line 30: `openMenuButton` and `operatorMenu` look unused

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+	<nav ref="operatorMenu" :style="style">
```
shawnyama, line 2: the `nav` may not be necessary, just apply the style to the button and it should work

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -272,6 +285,10 @@ const optionsMenuItems = ref([
 	}
 ]);
 
+const isMenufocused = ref<boolean>(false);
```
shawnyama, line 288:
```suggestion
const isMenuFocused = ref<boolean>(false);
```

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -678,15 +696,26 @@ function onCanvasClick() {
 	if (isCreatingNewEdge.value) {
 		cancelNewEdge();
 	}
+	hideMenuButton();
+}
+
+function hideMenuButton() {
+	if (!isMenufocused.value) {
+		isOutputMenuShown.value = false;
```
shawnyama, line 704: Maybe we don't need, `isOutputMenuShown`. Instead you can set `nodeWithOpenedMenu` to be null here and just show the menu if `nodeWithOpenedMenu` exists

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
shawnyama: Yeah, should be in `tera-operator` I assumed I was reading that file actually

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -0,0 +1,70 @@
...
+</template>
+<script setup lang="ts">
+import { ref } from 'vue';
+import type { WorkflowNode } from '@/types/workflow';
+import Menu from 'primevue/menu';
+import Button from 'primevue/button';
+
+defineProps<{
+	node: WorkflowNode<any> | null;
+	style: { top: string; left: string };
```
YohannParis, line 24: Again, the style should be done on the parent component, there is not reason to have those here as props. Or I misunderstanding something.

shawnyama: `tera-canvas-item` has the same issue too

YohannParis (approved):

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
asylves1: using the nav element for button and menu element to handle the mouse over events

PR #4332, opened 2024-07-31
primevue text inputs to tera text inputs pt1

blanchco:
# Description

* Primevue text input -> tera-input-text pt1 (more to come!)
* adds some more events to tera-text-input


packages/client/hmi-client/src/components/widgets/tera-input-text.vue
```diff
@@ -11,20 +11,22 @@
 				type="text"
 				:value="displayValue()"
 				@click.stop
-				@focus="isFocused = true"
+				@focus="onFocus"
 				@focusout="$emit('focusout', $event)"
 				@blur="onBlur"
 				@input="updateValue"
+				@keyup="$emit('keyup', $event)"
+				@keypress="$emit('keypress', $event)"
```
mwdchang, line 19: Maybe we should forward-port to use the change event in tera-input, rather than back-port tera-input to support these events?

blanchco: As I replace the primevue text components - Im just trying not too disturb what we currently have.  There are some primevue components which listen to these events, Im just sending them up (keyup.esc, keyup.enter).  Perhaps once I've moved over all the components it might be worth it to re-evaluate whether we should have these events?

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -376,7 +371,7 @@ header > section > button {
 	padding-left: 0.5rem;
 }
 
-.p-inputtext {
+.tera-input {
```
shawnyama, line 374: where is this class used?

blanchco: thats the class used for all `tera-inputs-text`/`tera-input-number` components. This just adds that min-width to that tera input rather than the old primevue input

packages/client/hmi-client/src/components/widgets/tera-input-number.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tera-input" :label="label" :title="title">
+	<div class="tera-input">
```
shawnyama, line 2: remove `tera-input` class 😉

shawnyama (approved):

packages/client/hmi-client/src/components/widgets/tera-input-number.vue
blanchco: its needed!

YohannParis (approved):

PR #4336, opened 2024-07-31
collaboration capable workflow - first pass

mwdchang:
### Summary
Version 0 of workflow collaboration feature.  Not we assume that people will not modify the same exact node at the same time.

- The changes add/augment the Java classes for WorkflowNode and WorkflowEdge, adding in version number and isDeleted tombstone flag. Note currently this is only for API end - the UI side is still using their own types. Consolidation will come later.

- Add in logic to handle entire-workflow update, this introduces tombstoning logic, meaning that we do not delete nodes/edges, but rather just mark them as deleted.

- Update UI workflow service to account for tombstoning

### Testing
You need two terarium browser sessions, A and B
- A: create workflow, add model operation, add model-config operation, connect them
- B: goto the same workflow, verify you can see model => model-config
- A: add another operation, say simulate, wait for 7-8 seconds for workflow write to occur or check network tab for workflow-POST
- B: remove the model operation, wait for 7-8 seconds for workflow write to occur or check network tab for workflow-POST
- A: refresh, verify that we still have simulated operation, verify that model-operation is removed
- B: refresh, verify that model operation is deleted, plus that there is a new simulate operation


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -240,11 +241,18 @@ export const removeEdge = (wf: Workflow, id: string) => {
 		targetPort.type = targetPort.originalType;
 	}
 
+	// Tombstone
+	const edge = wf.edges.find((e) => e.id === id);
+	if (edge) {
+		console.log('tombstoning edge');
```
YohannParis, line 247:
```suggestion
```

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -261,18 +269,25 @@ export const removeNode = (wf: Workflow, id: string) => {
 		removeEdge(wf, edgeId);
 	});
 
+	// Tombstone
+	const node = wf.nodes.find((n) => n.id === id);
+	if (node) {
+		console.log('tombstoning node');
```
YohannParis, line 275:
```suggestion
```

YohannParis (approved):

jryu01 (approved):
LGTM

dvince2 (approved):

PR #4344, opened 2024-08-01
model config fix

blanchco:
# Description

* fix default config creation when there are no configs

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
@@ -43,13 +43,13 @@ watch(
 
 		if (modelId) {
 			let configs = await getModelConfigurationsForModel(modelId);
-			if (!configs[0].id) {
+			if (!configs[0]?.id) {
 				const model = await getModel(modelId);
 				if (model) await postAsConfiguredModel(model); // Create a model configuration if it does not exist
 				configs = await getModelConfigurationsForModel(modelId);
 			}
 			// Auto append output
-			if (configs[0].id) {
+			if (configs[0]?.id) {
 				const config = configs[0];
 				state.transientModelConfig = config;
 				emit('update-state', state);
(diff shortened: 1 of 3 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -43,13 +43,13 @@ watch(
...
 			let configs = await getModelConfigurationsForModel(modelId);
-			if (!configs[0].id) {
+			if (!configs[0]?.id) {
 				const model = await getModel(modelId);
 				if (model) await postAsConfiguredModel(model); // Create a model configuration if it does not exist
 				configs = await getModelConfigurationsForModel(modelId);
 			}
 			// Auto append output
-			if (configs[0].id) {
+			if (configs[0]?.id) {
```
mwdchang, line 52: These logical checks are a bit janky, like, what makes getModelConfigurationsForModel returns the result such that the first-element IS the default config? 

Let's patch this logic up since we are not in a super hurry, e.g. if we need API support let's just add that.

blanchco: Sure, Ill take a look and maybe see returning model configs in order by oldest date (we can assume the oldest date is the default config since its the first created).

packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/ModelConfigRepository.java
```diff
@@ -9,5 +9,8 @@
 
 @Repository
 public interface ModelConfigRepository extends PSCrudSoftDeleteRepository<ModelConfiguration, UUID> {
-	List<ModelConfiguration> findByModelIdAndDeletedOnIsNullAndTemporaryFalse(UUID modelId, Pageable pageable);
+	List<ModelConfiguration> findByModelIdAndDeletedOnIsNullAndTemporaryFalseOrderByCreatedOnAsc(
```
blanchco, line 12: theseNamesGetPrettyCrazy

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
blanchco: @mwdchang I've updated the API to ensure that we get the configs in order of createdBy (oldest -> newest).  Let me know what you think.

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/ModelConfigRepository.java
YohannParis: They are _SQL_ like method for hibernates to do the right request.

YohannParis (approved):

PR #4350, opened 2024-08-01
fix POST /model-configuration endpoint logic

mwdchang:
### Summary
`POST /model-configuration` is janky due to a handful of issues, while the object is created in the DB, it is missing bits and pieces, and at times will pull back incorrect data.

- Fix a chicken-and-egg problem problem with model-configuration-id, this was causing children to be orphaned
- Yes, `InferredParameterSemantic` is exactly the same as `ParameterSemantic`, hibernate shoves children objects into the same table (via the dtype column), so you cannot have `List<Foo> a, List<Foo> b`, the lists need to be of different class types otherwise it will end up fetching the wrong thing
- As a consequence of the above, we need to do data migration - took this opportunity to remove the extra calibration-run-id column
- Fix places where we missed setting a few needed fields


YohannParis (approved):
LGTM

dvince2 (approved):

PR #4351, opened 2024-08-01
Move primevue to version 3.53.0

YohannParis:
# Description

* Update PrimeVue to `3.53.0`
* Big change on how the theme is setup on version `3.48.0`
  * a lot of component design is moved from the component CSS to the theme CSS 
  * use the theme we previously copied as our new base
  * random fixes for the update
* Move each type of CSS import into a different [CSS Layers](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer)
* Move _Katex_ and _JupyterLab_ CSS into the main CSS file, and put them on their own layers


packages/client/hmi-client/package.json
```diff
@@ -63,7 +63,7 @@
 		"pixi.js": "7.0.5",
 		"primeflex": "3.3.1",
 		"primeicons": "7.0.0",
-		"primevue": "3.43.0",
+		"primevue": "3.53.0",
```
dvince2, line 66: I just want to confirm that despite the issue saying `4` its intentional to do 3.53 :D

dvince2 (approved):

packages/client/hmi-client/package.json
YohannParis: Yes, changed, moving to v4 is a big task and needs a lot of changes.
But I succeeded to move to 3.53.0, just finishing removing some weird focus bright green and it will be good to go.

packages/client/hmi-client/package.json
```diff
@@ -34,9 +34,6 @@
 		"@lumino/commands": "2.1.3",
 		"@lumino/coreutils": "2.1.2",
 		"@lumino/widgets": "2.1.0",
-		"@uncharted.software/design-tokens": "1.1.0",
-		"@uncharted.software/facets-core": "3.2.6",
-		"@uncharted.software/facets-plugins": "3.2.5",
```
mwdchang, line 39: Some of these libs are still referenced in the source code, for example
```
https://github.com/search?q=repo%3ADARPA-ASKEM%2Fterarium+facets-core&type=code
```

Great to clean things up, but a bit harder to track things when we do things piecemeal - what is the plan for facets?

YohannParis: Brought back. There is a task to remove the Data Explorer.

blanchco (approved):
Looks good!

mwdchang:
@YohannParis a couple of the more noticeable changes that I noticed.....will leave it with you if they are as intended or candidates for tweaks

| Before  | After  |
| ------------- | ------------- |
| image  | image  |
| image | image  |

mwdchang (approved):

PR #4352, opened 2024-08-02
remove operation-type enum

mwdchang:
### Summary
Enums are useful for capturing controlled a vocabulary, things that do not change very often or are static in nature.

Since we are adding and removing operators at a fairly quick cadence this is not an appropriate data structure, especially if we want to model this in the DB which will likely mean a data-migration every time we add or remove operators.

In additional, while the keys are unique,  there is no guarantee on the value so there isn't really a safe guarantee.

Change this to a frozen-object, and the datatype of operator name to just a string.

### Testing
I removed a few references to operators that are deprecated, otherwise no functional changes, app should work as before.

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
@@ -122,12 +122,14 @@ const selectedOutputId = computed(() => {
 const outputOptions = computed(() => {
 	// We do not display output selection for Asset operators
 	if (
-		[
-			WorkflowOperationTypes.MODEL,
-			WorkflowOperationTypes.DATASET,
-			WorkflowOperationTypes.DOCUMENT,
-			WorkflowOperationTypes.CODE
-		].includes(props.node.operationType)
+		(
+			[
+				WorkflowOperationTypes.MODEL,
+				WorkflowOperationTypes.DATASET,
+				WorkflowOperationTypes.DOCUMENT,
+				WorkflowOperationTypes.CODE
+			] as string[]
+		).includes(props.node.operationType)
 	) {
 		return null;
 	}
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-julia/calibrate-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-julia/calibrate-operation.ts
@@ -1,4 +1,4 @@
-import { Operation, WorkflowOperationTypes, BaseState } from '@/types/workflow';
+import { Operation, BaseState } from '@/types/workflow';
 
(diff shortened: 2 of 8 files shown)


YohannParis (approved):
I do not understand what this resolves, but it looks good to me.

mwdchang:
> I do not understand what this resolves, but it looks good to me.

@YohannParis Having an enum here does not help anything, having an enum does mean a higher development cost - especially if we model this into Java/Hibernate combination, where hibernate will create a table-constraint based on the enum.  For example we have an enum representing Operation-status, if you look at the "simulation" table:

```
Check constraints:
    "simulation_engine_check" CHECK (engine::text = ANY (ARRAY['SCIML'::character varying, 'CIEMSS'::character varying]::text[]))
    "simulation_status_check" CHECK (status::text = ANY (ARRAY['CANCELLED'::character varying, 'COMPLETE'::character varying, 'ERROR'::character varying, 'FAILED'::character varying, 'QUEUED'::character varying, 'RETRIEVING'::character varying, 'RUNNING'::character varying]::text[]))
```

A similar check would be generated for operator-type, and we'd have to update the constraint as often as we update our operators.

YohannParis:
That is what I thought, but because this enum was not saved on the back-end and just a front-end tool, I was confused.
Would you like we stop using ENUM for operators and things not 100% defined?

> > I do not understand what this resolves, but it looks good to me.
> 
> @YohannParis Having an enum here does not help anything, having an enum does mean a higher development cost - especially if we model this into Java/Hibernate combination, where hibernate will create a table-constraint based on the enum. For example we have an enum representing simulation-status, if you look at the "simulation" table:
> 
> ```
> Check constraints:
>     "simulation_engine_check" CHECK (engine::text = ANY (ARRAY['SCIML'::character varying, 'CIEMSS'::character varying]::text[]))
>     "simulation_status_check" CHECK (status::text = ANY (ARRAY['CANCELLED'::character varying, 'COMPLETE'::character varying, 'ERROR'::character varying, 'FAILED'::character varying, 'QUEUED'::character varying, 'RETRIEVING'::character varying, 'RUNNING'::character varying]::text[]))
> ```
> 
> A similar check would be generated for operator-type, and we'd have to update the constraint as often as we update our operators.

shawnyama (approved):

dvince2 (approved):

mwdchang:
> That is what I thought, but because this enum was not saved on the back-end and just a front-end tool, I was confused.
> Would you like we stop using ENUM for operators and things not 100% defined?

For me, it isn't about how well objects are defined, but what we are using enums for. You wouldn't normally have an enum for team members for example if you are modelling an engineering project.

```
TeamMember.Shawn
TeamMember.Derek
...
```

Because things change, and you generally want enum to be fixed and stable(e.g. Status.fail, Status.in_progress, Status.success ).

PR #4354, opened 2024-08-02
Add options for node menu

asylves1:
# Description

image

Resolves #(issue)


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -634,3 +634,46 @@ export const branchWorkflow = (wf: Workflow, nodeId: string) => {
...
+function getPortOptions(operationMap: Map<string, any>, portKey: string, getOptionsFn: Function) {
+	const portOptions = new Map<string, Array<string>>();
+
+	operationMap.forEach((value, key) => {
+		const port: Array<any> = value?.[portKey] ?? [];
+		port.forEach((option) => getOptionsFn(option.type, key, portOptions, value?.displayName));
+	});
+
+	return portOptions;
+}
```
mwdchang, line 668: I think you are overcomplicating things with these mutable functions and callbacks, as you have `any` in a handful of places it is harder to follow what is going on.

Since we only have two types of ports, we may consider just spelling it out rather than going the indirect route.

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -439,6 +443,21 @@ const addOperatorToWorkflow: Function =
 		workflowDirty = true;
 	};
 
+function onMenuSelection(selection, menuNode) {
```
mwdchang, line 446: Consider rename and add typing information.

```
onMenuSelection(operatorType: string, node: WorkflowNode<any)
```

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
```diff
@@ -9,35 +9,45 @@
...
-	<Menu ref="menu" id="overlay_menu" :model="menuItems" :popup="true" />
+	<Menu
+		ref="menu"
+		id="overlay_menu"
+		:model="menuItems"
+		:popup="true"
+		@mouseenter="emit('menu-mouseenter')"
+		@mouseleave="emit('menu-mouseleave')"
+		@focus="() => {}"
+		@blur="() => {}"
```
mwdchang, line 20: Why are focus/blur specified as no-ops?

asylves1: the type check was asking for them when using `mouseenter` & `mouseleave`.

YohannParis: Because that's what they should be used for on an input, the `mouseenter` and `mouseleave` is a last resort.

mwdchang:
Need to unique the candidate items

image

asylves1:
> Need to unique the candidate items, eg there are two "compare models" in the menu items.
> 
> image

Push fix

packages/client/hmi-client/src/components/operator/tera-operator-menu.vue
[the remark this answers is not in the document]
asylves1: removed

mwdchang:
I took some liberties to clean up a few things
- There were some redundant checks that were not necessary, the data structure calls for the fields to be present so the algorithm can be made simpler
- Renamed a few things, namely input/output, key/value which are a little vague

mwdchang (approved):

PR #4359, opened 2024-08-02
First pass of using panteras search widget

YohannParis:
# Description

Resolves #(issue)


packages/client/hmi-client/src/utils/debounce-util.ts
YohannParis, line 1: How is this better than [Lodash.debounce()](https://lodash.com/docs/4.17.15#debounce)?

YohannParis:
@dvince2 I think we can delete this PR and associated branch?

PR #4366, opened 2024-08-06
Refactored forecast chart

jryu01:
# Description

Previously, a separate top level tooltip layer used the same dataset as the statistic layer, duplicating data and doubling the number of data points in the chart spec. This change refactors the chart by placing the main line and tooltip layers under a single top layer object containing the dataset. Also, this allows additional sublayer objects, such as annotations to be added to the chart.
Example: 
```
// Now we have 
layer: [
  // statistical layer
  {
    dataset: data,
    layer: [
      { line layer },
      { tooltip layer },
     .... 
    ]
  }
]

// Instead of
layer: [
  {
    // statistical layer
    dataset: data,
     .... 
  },
  {
    // overlay tooltip layer
    dataset: data,
     .... 
  }
  ...
]
```

Resolves #(issue)


YohannParis (approved):
LGTM

mwdchang:
Sorry, I should have mentioned this before @jryu01 - the tooltip layer is somewhat temporary, it was something that Jamie put together pretty quickly but not exactly what we want to have.

The ideal "toolip" would be a rule-based tooltip, rather than the currently point/cursor based tooltip. A rule-based tooltip will have a better options for selection heuristics - whereas right now we artificially inflate the stroke width to make things "easier" to select. 
But moreover, the tooltip wouldn't be restricted to just the statistical data, but others like values from ground-truth layer as well. (ground-truth and stat layers are not necessarily aligned ... so we also haven't done this part to merge them together).

mwdchang (approved):

PR #4369, opened 2024-08-06
refactor stratify related components , fix equations to model saving

shawnyama:
# Description

* During my transition matrix investigation I did some cleanups/fixes
* Fixed saving model in eqautions to model drilldown


YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
```diff
@@ -105,11 +105,7 @@
 			</tera-drilldown-section>
 			<tera-drilldown-preview>
 				<section v-if="selectedModel">
```
blanchco, line 107: is the `<section>` needed here?

blanchco (approved):

mwdchang (approved):
@shawnyama next time put the fix into a different PR, it makes things easier to track and revert if needed.

PR #4370, opened 2024-08-06
parameter with strata suffixes

mwdchang:
### Summary
Support for `param_renaming_uses_strata_names` parameter in MIRA stratify:
- Change the request to include said paraemter
- Change how we parse/collapse parameters in stratified models

Note there appears to be a problem in MIRA with stratify where it replicates template names, this can mess up the model because there may be multiple expressions applied to a single transition.

### Testing
Stratify a model including parameters, eg. stratify SIR with ages=young,old applied to S and beta. 
Verify that the new beta parameters have young/old suffixes instead of numbers


shawnyama (dismissed):

packages/client/hmi-client/src/model-representation/mira/mira.ts
```diff
@@ -214,6 +207,7 @@ export const collapseTemplates = (miraModel: MiraModel) => {
 
 	// 1. Roll back to "original name" by trimming off modifiers
 	miraModel.templates.forEach((t) => {
+		console.log('..', t.name);
```
YohannParis, line 210: Do we need this?

YohannParis (approved):

packages/client/hmi-client/src/model-representation/mira/mira.ts
mwdchang: nope

blanchco (approved):

PR #4375, opened 2024-08-07
save dataset on enter press

blanchco:
# Description

* Save dataset when pressing enter on the dataset transformer
* Took the opportunity to replace the primevue dialog with our own here as well


Tom-Szendrey (approved):
LGTM
Thanks

YohannParis (approved):

PR #4377, opened 2024-08-07
feat: calibrated configruration view

shawnyama:
# Description

- I identify a configuration as calibrated if there are `inferredParameters` (which turns on a read only configuration)
- Added type, mean and std values to parameters
- Show concept to reflect designs
image


packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -15,71 +15,82 @@
...
-		</main>
-		<footer v-if="isSourceOpen" class="mb-2">
-			<tera-input-text
-				placeholder="Add a source"
-				:model-value="getParameterSource(modelConfiguration, parameterId)"
-				@update:model-value="emit('update-source', { id: parameterId, value: $event })"
-			/>
-		</footer>
+			<span class="std">
+				<label>STD</label> {{ displayNumber(inferredDistribution?.parameters?.stddev.toString()) }}
```
mwdchang, line 91: Typically standard deviation is abbreviated "stddev" or "sd" .... std is something else.

packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -1,34 +1,40 @@
...
+					<label class="ml-auto">Concept</label>
+					<span class="ml-1">{{ concept }}</span>
+				</template>
 			</div>
-			<span v-if="description" class="ml-4">{{ description }}</span>
+			<span v-if="description" class="description">{{ description }}</span>
 		</header>
-		<main>
-			<span class="expression">
+		<template v-if="isEmpty(modelConfiguration.inferredParameterList)">
```
mwdchang, line 18: FYI we will need to revisit this, strictly speaking you still have initial values, they just shouldn't be editable.

shawnyama: I am just hiding the expression and some buttons, should I just show a read only expression?

mwdchang: I want to clear up with TA3 what exactly do they use when we send them a calibrated model-config.

So for example in SIR model the dill file has some functor representation of the parameters beta, gamma, S0, I0, R0.
Say our S has expression `S0 - 10` ... it is unclear to me whether it makes sense or not to change it to say `1000` as that basically would void the calibration, but technically there's nothing wrong with doing that.

Moreover, currently for the calibrated config `initialSemanticList` is empty so also not entirely sure what we want to do there.

mwdchang (approved):

YohannParis (approved):

PR #4380, opened 2024-08-07
ability to drag n drop edges in the workflow node

blanchco:
# Description

* Ability to drag n drop edges in the workflow (clicking on the ports still works as well)
* Check out the branch and let me know what you think...I couldn't see any issues while testing it out 😄


mwdchang (approved):
okay!

YohannParis (approved):
Good job!

PR #4387, opened 2024-08-08
update logic for checking if there are any changes on workflow node s…

mwdchang:
### Summary
Update the logic and add a higher-level wrapper `isWorkflowNodeDirty`

The potential usage is as follows, create a computed property where WorkflowNode is available: in the operator-node or operator-drilldown for example:

```
const isDirty = computed(() => isWorkflowNodeDirty(props.node));

...

 <Button
         :disabled="!isDirty"
         label="Run something"
         size="small"
         icon="pi pi-play"
/>
```

Then, assuming the operator is coded to workflow guideline, this will tell you if there are any unsaved changes on the workflow-node.

### Testing
Covered in unit test case
```
yarn run test workflow
```


mwdchang:
@brittanykondo

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -517,18 +517,39 @@ export async function generateSummary(
 }
 
 // Check if the current-state matches that of the output-state.
-// Note operatorState subsumes the keys of the outputState
-export const isOperatorStateInSync = (operatorState: Record<string, any>, outputState: Record<string, any>) => {
+//
+// Note operatorState subsumes the keys of the outputState, thus if
+// all of outputState[x] matches operatorState[x] it is considered
+// to be in sync.
+export const isOperatorStateInSync = (
```
YohannParis, line 524: This doesn't need to be a `const` as it is self contain. Keeping it a `function` will be easier to read than an arrow function, and to manipulate in a potential refactoring.

YohannParis (approved):
Why not use a composable for this?

packages/client/hmi-client/src/services/workflow.ts
mwdchang: Can you provide some examples of why arrow functions are harder to refactor or manipulate?

YohannParis: Because you have to check that they have no side-effect, as a function does not have this issue.

mwdchang: @YohannParis If we have a particular preference we want to try to adhere to as a team I am all on board with that. But this is not a valid argument, I can easily have side effect regardless of whether I use arrow-functions or the classic function declaration. 

```
const list = []

export const doSomething = () => {
  list.push(123)
}

export function doSomething() {
  list.push(123)
}
```

These two are functionally (excuse the pun) equivalent, they both have side effects of modifying another data structure that is not in their own respective closure.

blanchco (approved):
cool!

PR #4414, opened 2024-08-09
4276 parameter distribution charts

### LINKED ISSUE #4276: Add parameter distribution charts to calibrate operator output panel
### Summary
We have created prototype for show comparison of distribution, how it is time to deploy it into the application.

Image

This is an initial step towards the design, I think we will built out all the base pieces first, and then we can rearrange them to match the design sketches below. If you have a better suggestion or approach let's chat and work out the details.

[figma-link](https://www.figma.com/design/EoihrRoNNbuOAJtudKqp7L/Terarium---June-%2F-July-2024?node-id=1-26505&t=p9aBDRmTusZ9cX0y-0)

### Steps
1. We need to extend Calibrate operator's state to, similar to how `chartConfigs` we need to either augment this data structure or create a new config data structure to indicate the parameters we want to see.

2. Technically speaking we have the data already calculated, it is the output of the `mergeResults` function and you have access to the before/after fields. We should keep an eye on the memory usage because these datasets can get pretty large and we may need to do additional processing to trim down the unneeded fields.

3. Similar to the existing forecast-chart, use "tera-chart-control" to select the parameters

4. Add a data table showing the mean/variance. I think we may be able to extend the existing histogram chart, see maybe

```
https://vega.github.io/editor/#/examples/vega-lite/brush_table
```
or if that is annoying to do, we can just do it outside with a vue/html table.
 
image

### Misc
It is unclear what the calibration-node will show, we will work this out during the sprint.

jryu01:
# Description

https://github.com/user-attachments/assets/8149edb2-7264-4839-9ed4-2e3b8435aaa9

- Added histogram chart and the mean/variance table for the parameters
- Added `selectedParameters` to the node state for selecting parameters for the distribution charts
- Replaced `chartConfigs` with single array `selectedVariables` for selecting model variables for the forecast charts
- Updated UI to Add/Remove charts by multi selection from `tera-chart-control`

Resolves #4276


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -153,20 +153,56 @@
...
+												<td>{{ preparedDistributionCharts[param].stat.mean[0] }}</td>
+												<td>{{ preparedDistributionCharts[param].stat.mean[1] }}</td>
+											</tr>
+											<tr>
+												<th scope="row">Variance</th>
+												<td>{{ preparedDistributionCharts[param].stat.variance[0] }}</td>
+												<td>{{ preparedDistributionCharts[param].stat.variance[1] }}</td>
+											</tr>
+										</tbody>
+									</table>
```
mwdchang, line 191: FYI: No action needed for now, there might be a chance that we need to move the table "into" the chart itself so it can be exported along with the graph. The idea is to better support the notions of reproducibility and transparency.

packages/client/hmi-client/src/services/model.ts
```diff
@@ -189,6 +189,20 @@ export const getUnitsFromModelParts = (model: Model) => {
...
+	const typeMapping: { [key: string]: string } = {};
+	[...(model?.model.states ?? [])].forEach((v) => {
+		typeMapping[v.id] = 'state';
+	});
+	[...(model?.semantics?.ode?.parameters ?? [])].forEach((v) => {
+		typeMapping[v.id] = 'parameter';
+	});
+	(model?.semantics?.ode?.observables || []).forEach((o) => {
+		typeMapping[o.id] = 'observable';
+	});
```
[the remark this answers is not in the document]
jryu01: yeap that's right. I fixed it.

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: From my experience, drawing table with vegalite was limited and tricky. But I think we can still explore more if there are any other ways or work arounds moving forward.

YohannParis (approved):

PR #4415, opened 2024-08-09
Add role gobbling error

bigglesandginger:
# Description

For some reason the generic which should be reported Exception and return a 500, and RelationshipAlreadyExistsException which should return a 304 (not modified) were being conflated together (and not reporting the errors)

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/permissions/UsersController.java
```diff
@@ -83,7 +86,10 @@ public ResponseEntity<Void> addRoleToUser(
 				publicGroup.setPermissionRelationships(who, Schema.Relationship.MEMBER.toString());
 			}
 			return reBACService.addRoleToUser(roleName, userId);
-		} catch (final Exception | RelationshipAlreadyExistsException e) {
+		} catch (final RelationshipAlreadyExistsException e) {
+			return ResponseEntity.status(304).build(); // Not Modified
```
dvince2, line 90: We could also return `HttpStatus.NOT_MODIFIED` here to avoid the `304` hard coded

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/permissions/UsersController.java
bigglesandginger: I had tried but failed to find it - but I guess I was using the wrong HttpStatus (apparently there are 3 that come up immediately... ugh)

YohannParis (approved):

PR #4420, opened 2024-08-09
Cache rebac

bigglesandginger:
# Description

Add caching to rebac in a hope to balance the hope that calling SpiceDB and Keycloak is a serious bottleneck against the need to refresh the data from the primary source as users get added to the system outside of the app.

Caffeine cache is used as it provides TTL (time to live) details which regular Spring annotation caching does not (bizarre and the reasons given are weak at best).

Resolves #(issue)


dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
```diff
@@ -77,12 +84,51 @@ public class ReBACService {
...
+
+		SchemaObject who;
+		Schema.Permission permission;
+		SchemaObject what;
+
+		CacheKey(SchemaObject who, Schema.Permission permission, SchemaObject what) {
+			this.who = who;
+			this.permission = permission;
+			this.what = what;
+		}
```
YohannParis, line 103: Is this the contructor of the class?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
```diff
@@ -77,12 +84,51 @@ public class ReBACService {
...
+		@Override
+		public int hashCode() {
+			return Objects.hash(who, permission, what);
+		}
+	}
+
+	private final Cache<CacheKey, Boolean> permissionCache = Caffeine.newBuilder()
+		.expireAfterWrite(5, TimeUnit.MINUTES)
+		.recordStats()
+		.removalListener((Object key, Object value, RemovalCause cause) -> log.info("Key {} was removed {}", key, cause))
```
YohannParis, line 123: Should this be a the `debug` level and not `info`?
Would thise be everywhere in the logs?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
bigglesandginger: Yes that is a constructor (the only one for this class)

Issue #4228, opened 2024-07-23
[TASK]: Remove temporary from the TerariumAsset


YohannParis:
**Describe the task**
The goal of `isTemporary` was to decern an asset in a project or not. But now we have the ProjectAsset relationship.

**Task Items**
- [ ] Remove the `temporary` property of `TerariumAsset`
- [ ] Remove all related uses.

YohannParis:
irrelevant because of #4440

PR #4422, opened 2024-08-12
Update Simulate Output 

Tom-Szendrey:
# Description
-Rename simulate-ciemss drilldown to keep up with naming.
-Change simulate-ciemss state to include "forecastId" + rename inProgressForecastId
-Simulate node now outputs a dataset ID. This dataset is a temporary asset.

# Note
This is displaying the forecast's results not the output dataset's results.
Im not sure if this is a concern as the dataset created is from the forecast. 
**Please feel free to discuss on this PR.**


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -126,15 +128,23 @@ Provide a summary in 100 words or less.
 
 	const summaryResponse = await createLLMSummary(prompt);
 
+	const datasetName = `Forecast run ${runId}`;
+	const projectId = useProjects().activeProject.value?.id ?? '';
+	const datasetResult = await createTemporaryDatasetFromSimulationResult(projectId, runId, datasetName);
+	if (!datasetResult) {
+		logger.error('Error creating dataset from simulation result.');
+		return;
+	}
```
mwdchang, line 137: We have something called `createDatasetFromSimulationResult` - I am not familiar with it, but have we looked to see if we can use it or extend the feature to adapt to this use case? Cause at the surface it seems like you are doing very similar things.

https://github.com/DARPA-ASKEM/terarium/blob/412a24f7b5fc14d98dde66d72e21d3de111d6bde/packages/client/hmi-client/src/services/dataset.ts#L273-L276

Tom-Szendrey: I can add a flag to `createDatasetFromSimulationResult` for temporary 
And change the return type from boolean -> Dataset
as well as update the use of it in `saveDataset` to still have that boolean logic there

Good call give me a moment

packages/client/hmi-client/src/services/dataset.ts
```diff
@@ -252,27 +252,32 @@ async function createNewDatasetFromFile(
...
-): Promise<boolean> {
+	datasetName: string | null,
+	isTemporary?: boolean
+): Promise<Dataset | null> {
+	if (isTemporary === undefined) isTemporary = false;
 	try {
-		const response: AxiosResponse<Response> = await API.post(
-			`/simulations/${simulationId}/add-result-as-dataset-to-project/${projectId}?dataset-name=${datasetName}`
+		const response: AxiosResponse<Dataset> = await API.post(
+			`/simulations/${simulationId}/create-result-as-dataset/${projectId}?dataset-name=${datasetName}&is-temporary=${isTemporary}`
```
YohannParis, line 266: We have an issue #4228 to remove the `isTemporary`.
So I'm not sure we need to do this, here.

Tom-Szendrey: It was my understanding that when we run a forecast with a node we should create a _temporary_ dataset asset.
If the user then wants to save this asset into their project they would use this menu:
image

Which will create a copy of the dataset asset, marked as not temporary, and add this new asset to the project.

Do you have a quick minute to chat about this?

Tom-Szendrey: Thanks Yohann!

YohannParis (approved):

mwdchang (approved):
thanks

PR #4423, opened 2024-08-12
add petri-amr semantic check

mwdchang:
### Summary
AMR can be syntactically valid, but semantically flawed. This has created a lot of problems - we don't detect them early enough and then the errors downstream are cryptic and hard to trace. This PR adds a function to perform some semantic checks so we can leverage it to catch errors/warnings early on.

The function returns a listing of notable observations, with a type field indicating error and possible problems:

e.g.
```
[
  { type: 'error', content: 'states need to match initials' },
  { type: 'error', content: 'Vßaccinated has no initial' },
  { type: 'warn', content: 'Vßaccinated has no initial.expression' }
]
```

Note only the capability is added here, the check still need to be deployed/leveraged into the application itself.

### Testing
```
 yarn workspace hmi-client run test service
```


YohannParis (approved):
Let's implement it on the model operator to check user inputs are accurate.
Also on upload of a model before saving it to the back-end.

shawnyama (approved):

PR #4424, opened 2024-08-12
Update Calibrate Output Type

Tom-Szendrey:
# Description
-Update calibrate to output both a model config as well as a dataset.

image


shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -396,13 +399,26 @@ watch(
 				initialSemanticList: _.cloneDeep(baseConfig.initialSemanticList),
 				inferredParameterList: inferredParameters
 			};
+
 			const modelConfigResponse = await createModelConfiguration(calibratedModelConfig);
+			const datasetName = `Forecast run ${state.forecastId}`;
+			const projectId = useProjects().activeProject.value?.id ?? '';
+			const datasetResult = await createDatasetFromSimulationResult(projectId, state.forecastId, datasetName, false);
+			if (!datasetResult) {
+				logger.error('Error creating dataset from simulation result.');
```
YohannParis, line 408: You should check what kind of error we get back from the endpoint, because if it's an 4XX or 5XX error, we already display a red message to the user, so another one will not make it better.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
Tom-Szendrey: Removing log entirely as its already in the service file for us 
Good point

PR #4429, opened 2024-08-12
Update ensemble simulate output

Tom-Szendrey:
# Description
-Rename both ensemble's drilldowns to follow naming
-Update ensemble simulate state to include forecastId + rename inProgressForecastId to match
-Update simulate ensemble output type.


YohannParis (approved):
Nice work

PR #4441, opened 2024-08-13
Workflow wrapper class

mwdchang:
### Summary
Wrap workflow data structure inside WorkflowWrapper class to make accidental data corruption less likely.  There are a lot of changes but basically they are of the form:

From

```
blah = workflowService.doSomething(wf, param1, param2, ...)
```

To

```
blah wf.doSomething(param1, param2)
```

If you use the WorkflowWrapper then generally speaking you don't need to deal with isDeleted and version flags.


dvince2 (approved):
handy!

shawnyama (approved):

Tom-Szendrey (approved):
Thank you!!!

PR #4446, opened 2024-08-13
Add project search index with permission mirroring

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1120,6 +1130,9 @@ public ResponseEntity<JsonNode> updateProjectUserPermissions(
 
 			final RebacProject what = new RebacProject(projectId, reBACService);
 			final RebacUser who = new RebacUser(userId, reBACService);
+
+			// no need to update the search perms since we cant never reduce below a read.
```
dgauldie, line 1134: double negative

packages/server/src/main/resources/static/es/index-templates/tds_2.0_project_index_template.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "vector": {
+              "type": "dense_vector",
+              "element_type": "float",
+              "dims": 1536
+            },
+            "embeddingId": {
+              "type": "keyword"
+            },
+            "span": {
```
mwdchang, line 61: What are `embeddingId` and `span`?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
kbirk: I've never not heard something that doesn't say that this isn't what you don't mean, or isn't it?

packages/server/src/main/resources/static/es/index-templates/tds_2.0_project_index_template.json
kbirk: The embedding generation can only take an input of a fixed size. So for large inputs we chunk them. Then the span a two element array containing the start / stop index in the source that the embedding references. 

So the id + span allows us the option to known exactly which chunk of text the lookup matches.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -0,0 +1,205 @@
...
+		elasticService.deleteIndex(index);
+	}
+
+	public void indexProject(final Project project) throws IOException {
+		final ProjectDocument doc = new ProjectDocument();
+		doc.setUserId(project.getUserId());
+		doc.setPublicAsset(project.getPublicAsset());
+		doc.setCreatedOn(new Timestamp(System.currentTimeMillis()));
+		doc.setUpdatedOn(new Timestamp(System.currentTimeMillis()));
+		doc.setPermissionJoin(PermissionJoin.CreateProjectJoin());
```
bigglesandginger, line 125: So this is the initial "permission" for a project, but it doesn't have a user assigned?  Why isn't the creator added to the list?

bigglesandginger:
The use of the word `permission` I think should be renamed to `canRead` or something like that as it does not track the Permission enum and instead is simply a list of users that can know of the project

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -0,0 +1,205 @@
...
+		return elasticService.getIndexFromAlias(getAlias());
+	}
+
+	@Data
+	public static class PermissionJoin {
+
+		private String name = null;
+		private UUID parent = null;
+
+		static PermissionJoin CreateProjectJoin() {
```
bigglesandginger, line 50: should this not be `createProjectJoin()` as it is not the constructor for PermissionJoin

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -0,0 +1,205 @@
...
+			final PermissionJoin join = new PermissionJoin();
+			join.setName("project");
+			join.setParent(null);
+			return join;
+		}
+
+		static PermissionJoin CreatePermissionJoin(final UUID projectId) {
```
bigglesandginger, line 57: should this not be `createPermissionJoin()` as it is not the constructor for PermissionJoin

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -0,0 +1,205 @@
...
+		elasticService.indexWithRouting(getAlias(), project.getId().toString(), doc, routing);
+	}
+
+	public void updateProject(final Project project) throws IOException {
+		final ProjectDocument doc = new ProjectDocument();
+		doc.setUserId(project.getUserId());
+		doc.setPublicAsset(project.getPublicAsset());
+		doc.setUpdatedOn(new Timestamp(System.currentTimeMillis()));
+		doc.setPermissionJoin(PermissionJoin.CreateProjectJoin());
+
```
bigglesandginger, line 138: Will this not erase the existing `PermissionJoin` thing?

bigglesandginger: Or, I don't understand this data structure.  It appears that there is one PermissionJoin for a project and yet it changes to be either a specific user via `addProjectPermission` or reset to whatever `CreateProjectJoin()` creates...

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
kbirk: That is the permission "join field" declaring that the document is a `"project"` document, and a parent in the parent-child relationship.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
kbirk: The two types of documents share the same index. So the "parent", which is the project, has a join field declaring it a project parent type. And the "child", which is the user read permission, has a join field declaring it a child doc. The `userId` field is overrided, in projects it is the owner of the project, and in a permission it is the userId of the person who has read permission.

kbirk: The structure of join fields is not very intuitive. They aren't separate docs in different indices, they are in the same index, and have the same mapping. The implicit `name` and `parent` property of the join field are used to define the relationship

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/elasticsearch/ElasticsearchService.java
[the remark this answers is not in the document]
kbirk: My editorconfig is the default that is in the project :(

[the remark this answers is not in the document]
YohannParis: _C'est la vie !_

packages/server/src/main/resources/static/es/index-templates/tds_2.0_project_index_template.json
mwdchang: gotcha, thanks!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
bigglesandginger: Well based on the created_on fiasco... I suspect this is where this code will fail

bigglesandginger (approved):
As we do not have groups, this should work fine (assuming PermissionJoin works as advertised).  If we ever add groups/sub-groups then this will need a good sanding, but the concept as I take it should be solid.

PR #4447, opened 2024-08-13
4330 replace loss chart with vega chart

### LINKED ISSUE #4330: [TASK]: Replace calibrate loss chart with vegalite chart
The current loss chart was quickly put together and is a bit out of place with the themes and designs of terarium, we should look at replacing it with the other charts we have.

We have `createForecastChart` function in services/chart.js that should be a sufficient replacement for the loss chart. I think we can just plug in the "summary-layer" and that should do the trick.

jryu01:
# Description

- Replaced the loss chart with vega forecast chart leveraging streaming data

https://github.com/user-attachments/assets/e688f672-9bd9-4a6a-8cc6-be1ef5c44451

Resolves #4330


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -326,7 +326,8 @@ export function createForecastChart(
 	samplingLayer: ForecastChartLayer | null,
 	statisticsLayer: ForecastChartLayer | null,
 	groundTruthLayer: ForecastChartLayer | null,
-	options: ForecastChartOptions
+	options: ForecastChartOptions,
+	summaryLayer?: ForecastChartLayer | null
```
mwdchang, line 330: Would it be possible to reuse the "statisticsLayer" or the "samplingLayer" rather than creating a new layer definition? Is it a lot of work? 

It just keeps things tighter - because there are a lot of similarities in terms of how people define statistics and summary. e.g. average, max, variation ... are all statistical summarizations of some data 😄 

As an aside, would be better to keep "options" parameter as the last to make the function a little easier to reason with.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -156,18 +156,18 @@ const preparedCharts = computed(() => {
...
 			{
-				dataset: resultSummary,
+				data: resultSummary,
 				variables: [`${pyciemssMap[variable]}_mean:pre`, `${pyciemssMap[variable]}_mean`],
 				timeField: 'timepoint_id'
 			},
 			{
-				dataset: groundTruth,
+				data: groundTruth,
 				variables: datasetVariables,
```
mwdchang, line 171: Sorry, don't know where to put it, but there is also a loss-chart here in the calibrate node that we should replace. Search for `drilldownLossPlot`

packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -4,7 +4,7 @@
 			{{ renderErrorMessage }}
 		</p>
 		<div ref="vegaContainer"></div>
-		<footer>
+		<footer v-if="$slots.footer">
```
YohannParis, line 7: Nice improvment!
I wish the slot was an attribute and not an element.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
jryu01: Done!

mwdchang (approved):

PR #4449, opened 2024-08-13
Latex font fix

shawnyama:
# Description

image
image


mwdchang:
@YohannParis you added `cursive, serif` attributes? Are they needed for a specific purpose?

@shawnyama not sure what is the actual fix here, it may be that we are applying the right font to the wrong places?

YohannParis (approved):

mwdchang (approved):

PR #4451, opened 2024-08-13
amr-errors in dev interface, more amr check criteria

mwdchang:
### Summary
Upgrade engineering interface to show semantic petrinet amr errors and warnings.  We can use this page as a first stop to check if the AMR will cause potential problems down stream.

image

### Testing
Housed under
```
http://localhost:8080/amr-petri-test
```


YohannParis (approved):

Tom-Szendrey (approved):
Nice!

PR #4457, opened 2024-08-14
New output section for ta3 operators

Tom-Szendrey:
# Description
We want to add the output settings section from optimize to calibrate, and simulate as well. 
Here I have made it its own component and used it in all 3 of these operators.

Calibrate and Optimize have previously moved away from the proxyChart stuff. 
Simulate does with this PR.
Im wondering if we should try to clean up a lot of the shared code for `preparedCharts` in these nodes or not in future PRs

image


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -1,13 +1,10 @@
...
-		<template v-if="selectedRunId && runResults[selectedRunId]">
-			<vega-chart
-				v-for="(_config, index) of props.node.state.chartConfigs"
-				:key="index"
-				:are-embed-actions-visible="false"
-				:visualization-spec="preparedCharts[index]"
-			/>
-		</template>
+		<div v-if="runResults && selectedRunId">
+			<template v-for="(_, index) of node.state.selectedSimulationVariables" :key="index">
```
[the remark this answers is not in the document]
Tom-Szendrey: No reason. I will clean that up for each file i did that

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-output-settings.vue
```diff
@@ -0,0 +1,119 @@
...
+const props = defineProps<{
+	successDisplayChartsCheckbox?: boolean;
+	summaryCheckbox?: boolean;
+	interventionsDisplayChartsCheckbox?: boolean;
+	simulationDisplayChartsCheckbox?: boolean;
+	selectedSimulationVariables?: string[];
+	selectedInterventionVariables?: string[];
+	simulationChartOptions: string[];
+	interventionsOptions: string[];
+}>();
```
mwdchang, line 98: Do we have the design sketches you can either link to? or post a few screen grabs here? 

I am wondering if we should assess this a bit more systematically ... I am thinking if we can pass in the `<State>` into these component and it figures it out? or if each operator should define some kind of emit-function of the things it wants to be configured. This component right now feels a bit like a free-for-all and not entire sure what is what.

mwdchang: Chatted with Tom off thread:

There is a pretty high chance we'd need to overhaul the output settings panel and how things work:
- Some of the design elements are problematic, not really up-to-date with the latest thinking, or are some what contradictory. For example we'd need to think about how these settings are going to work with reporting features
- Likewise, the "chartConfig" is getting rejigged to take into consideration annotations, this will impact how we would pass things back and forth between the operator drilldown and the settings panel as they wouldn't be just variable string arrays
- It would still be good to assess the situation in a more systematic manner. It may be nice, if possible, for each operator to provide a generator function of its output-settings, and how the settings should be displayed.

mwdchang:
### Simulate
So simulate operator is somewhat of an outlier, is that the variables are configured as `string[][]` rather than `string[]` to allow the selection of multiple variables per chart. For example I want to see states S and I on the same chart to see how they compare. It looks like it has been restricted to univariate on this PR.

On staging environment:

image

Tom-Szendrey:
Clearly needs more design discussion and work. 
No point in moving this fwd for all of the reasons Dan has mentioned.

PR #4458, opened 2024-08-14
fix: notebook creates new config

shawnyama:
# Description

* once code is run a new config is saved
* use `saveCodeToState()` from notebook service and adjust `model-config-operation.ts` properties to match
* removed updating the state in functions since it's updated in the watcher anyways
* fixed bug with @blanchco in the backend where the previous config becomes empty after you save a new one

https://github.com/user-attachments/assets/96453f29-664b-467d-ac55-3deef00ae62c

Resolves #(issue)


YohannParis (approved):
Nice cleaning up

mwdchang (approved):

PR #4459, opened 2024-08-14
model config Only show confirmation dialog if config has changes

asylves1:
# Description

Only show confirmation dialog when there are unsaved changes
image

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
@@ -202,7 +202,7 @@
 
 <script setup lang="ts">
 import '@/ace-config';
-import { cloneDeep, isEmpty, orderBy } from 'lodash';
+import { cloneDeep, isEmpty, isEqual, orderBy } from 'lodash';
 import Accordion from 'primevue/accordion';
 import AccordionTab from 'primevue/accordiontab';
 import Button from 'primevue/button';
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -563,6 +565,12 @@ const initialize = async () => {
 };
 
 const onSelectConfiguration = (config: ModelConfiguration) => {
+	// Checks if there are unsaved changes to current model configuration
+	if (isEqual(selectedModelConfig.value, knobs.value.transientModelConfig)) {
+		applyConfigValues(config);
```
YohannParis, line 570: if they are equals, why applying them?

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -458,6 +458,7 @@ useClientEvent(ClientEventType.TaskGollmConfigureFromDataset, configModelEventHa
 
 const selectedOutputId = ref<string>('');
 const selectedConfigId = computed(() => props.node.outputs.find((o) => o.id === selectedOutputId.value)?.value?.[0]);
+const selectedModelConfig = ref<ModelConfiguration | null>(null);
```
shawnyama, line 461: I think this can be a regular variable instead of a `ref`. 

As an aside, after working in here myself I noticed that the `originalConfig` is retrieved in `resetConfiguration()`. Maybe you can just use the `selectedModelConfig` you made here there to avoid the pointless `find()`...maybe renaming it to `originalConfig` would make more sense too.

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
YohannParis: What is the answer to my question?

PR #4463, opened 2024-08-14
feat(fixes unit entry): specify characters you want to reject from tera-input

shawnyama:
# Description

* Reject specific characters we don't want with `charactersToReject` prop in `tera-input-text`
* This handles the rejection of spaces in the unit entry better

https://github.com/user-attachments/assets/1b90f2bc-0646-41c9-9fb2-6185483aae3e


YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-input-text.vue
```diff
@@ -83,8 +84,13 @@ function formatValue(value: string | number | undefined) {
 
 const updateValue = (event: Event) => {
 	const target = event.target as HTMLInputElement;
-	const value = target.value;
-	emit('update:model-value', value);
+	if (props.charactersToReject && !isEmpty(props.charactersToReject)) {
```
mwdchang, line 87: Probably sufficient with just the `isEmpty` check, don't think you need to check the variable is falsy

mwdchang (approved):

packages/client/hmi-client/src/components/widgets/tera-input-text.vue
shawnyama: Yeah I tried that but typescript isn't aware of lodash checking for that so it still underlines the variables thinking I'd be passing a potentially undefined value

PR #4466, opened 2024-08-15
Fix: Use correct Id in calibrate drilldown

Tom-Szendrey:
# Description
Correcting the ID used in the calibration drilldown for the charts.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -719,7 +719,8 @@ watch(
 	async () => {
 		// Update selected output
 		if (props.node.active) {
-			selectedOutputId.value = props.node.active;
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -719,7 +719,8 @@ watch(
 	async () => {
 		// Update selected output
 		if (props.node.active) {
-			selectedOutputId.value = props.node.active;
+			const active = props.node.active;
+			selectedOutputId.value = props.node.outputs.find((o) => o.id === active)?.value?.[0];
```
mwdchang, line 723: A bit confused - selectedOutputId doesn't seem to be used in this component, it is declared at the top and assigned here, so how is it causing an issue?

Tom-Szendrey: Apologizes this is not required here!

PR #4472, opened 2024-08-15
update client side workflow to avoid version mismatch

mwdchang:
### Summary
The version of each sub-part is updated when things change, however the client does not refresh its copy so at some point the client may fall behind, and subsequent update will be ignored (the red arrow). This PR updates the client-side workflow with the server response (adding in the blue arrows ).

Now we have
- WorkflowWrapper#load(...) which replace the workflow entirely, and
- WorkflowWrapper#update(...) which does piece-wise merges of database's version of wf into client's side version of the workflow.

image

### Note
While testing (locally) I did some offhand benchmarks and noticed a few things, the logic that handles the workflow merge is decently fast, we are talking sub 25 ms for a workflow of 6, 7 nodes. However the overall time is on the order of 900+ ms, some 35-45X the business logic. 

It seems like `TerariumAssetServiceWithSearch` is quite slow because of ElasticSearch reindexing, perhaps we are using the `_wait_for` option with 1.0s refresh cycles?? - that would match the observed behaviours.  

Testing with extending `TerariumAssetServiceWithoutSearch`  instead lowered the overall time to ~40 ms mark.
A 900 ms update round-trip will have a lot of UX and possibly unexpected issues.

### Testing
Run calibrate, after operation finishes, refresh the page, the states and outputs should not reset.


Tom-Szendrey (approved):
Tested out bug and was unable to reproduce
Solution adds up 

Awesome! Thank you

dvince2 (approved):

PR #4474, opened 2024-08-15
Fix model edit reset in wizard

shawnyama:
# Description

* If reset is clicked in wizard view, rerender the decomposed canvas. This needs to be rerendered since changes in the decomposed view are typically done by user actions (after the initial `flattenedToDecomposed` conversion). It worked from the Notebook view since the `onMounted` was retriggered on tab change.
* Removed unneeded class and prop passing

https://github.com/user-attachments/assets/0c40e064-4bc0-4444-a471-d59ae21345ee


YohannParis (approved):

mwdchang (approved):
Didn't test - code looks okay!

PR #4477, opened 2024-08-16
Add error plot

jryu01:
# Description

Screenshot 2024-08-26 at 9 58 01 AM

Screenshot 2024-08-16 at 10 34 48 AM

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts
```diff
@@ -58,22 +59,24 @@ export function getErrorData(groundTruth: DataArray, simulationData: DataArray,
 	);
 	if (relevantGroundTruthColumns.length === 0) return errors;
 
+	// Fix the time values by using the index of the value since the values are floats not integer e.g t: 0.904522613065327 instead of t: 1
+	// We assume that ground truth is sorted by time in ascending order
+	const groundTruthTimeValuesCorrected = groundTruth.map((ele, index) => ({ ...ele, [datasetTimeCol]: index }));
```
mwdchang, line 64: You can't make this assumption of using the index - what is the problem here?

.vscode/launch.json
```diff
@@ -48,6 +48,16 @@
...
+        {
+          "type": "java",
+          "name": "TerariumApplication (Staging Keycloak)",
+          "request": "launch",
+          "mainClass": "software.uncharted.terarium.hmiserver.TerariumApplication",
+          "projectName": "server",
+          "args": [
+              "--spring.profiles.active=default,secrets,local,staging-keycloak,ide"
+          ]
+        },
```
mwdchang, line 60: @kbirk do you care ?? ☝️

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts
mwdchang: So then this may be something that @Tom-Szendrey needs to resolve in the MAE function, since we are already doing matching resolutions on time-point there. Maybe check with @liunelson to see if this is an immediate concern that needs to be addressed or we can back-burn it.

The data you are showing looks like it came from Julia's simulation, which prioritizes on emulating the accurate shape of the curve rather than providing useful forecast output.

In the meantime @jryu01 use a different dataset for ground truth:
- Run a simulation
- Grab the dataset of the simulation from S3 and use that as the ground truth

jryu01: Hm yeah.. it looks like the dataset I was testing with have such weirdness for `t`, I'll grab different dataset and test with it  again.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -210,12 +210,30 @@
 							:chart-config="{ selectedRun: 'fixme', selectedVariable: selectedVariables }"
 							:multi-select="true"
 							:show-remove-button="false"
-							:variables="Object.keys(pyciemssMap).filter((c) => modelPartTypesMap[c] !== 'parameter')"
+							:variables="Object.keys(pyciemssMap).filter((c) => modelPartTypesMap[c] === 'state')"
```
Tom-Szendrey, line 213: Why are we adding this filter?
If I am understanding correctly as a user I will no longer be able to check observables in this chart?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -210,12 +210,30 @@
...
 						<template v-for="variable of node.state.selectedVariables" :key="variable">
 							<vega-chart expandable :are-embed-actions-visible="true" :visualization-spec="preparedCharts[variable]" />
 						</template>
+						<h5>Errors</h5>
+						<tera-chart-control
+							:chart-config="{ selectedRun: 'fixme', selectedVariable: selectedErrorVariables }"
+							:multi-select="true"
+							:show-remove-button="false"
+							:variables="
+								Object.keys(pyciemssMap).filter(
```
Tom-Szendrey, line 225: Should the `:variables` not be something like:
`Object.keys(errorData.value?.[0]) ? []`
This way out keys are coming from the same place as our data is and we know we have like with like?

mwdchang (approved):
Looks okay, we want want to add a temporary guard against `selectedErrorVariables` usage so older operators created before this change do not go boom!

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -420,50 +438,54 @@ const chartSize = computed(() => drilldownChartSize(outputPanel.value));
...
 	const { result, resultSummary } = mergeResults(
-		runResult.value,
 		runResultPre.value,
-		runResultSummary.value,
-		runResultSummaryPre.value
+		runResult.value,
+		runResultSummaryPre.value,
+		runResultSummary.value
 	);
+	const error = getErrorData(groundTruth, result, mapping.value);
```
Tom-Szendrey, line 464: Can use errorData.value here as its being defined on via `props.node.active watcher`

If we prefer utilizing this then we should consider removing errorData.value

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: didn't notice that was there. I fixed it thanks!

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: errorData also include `sample_id` which we don't want here. We still need to filter the list and map the variable to the correct target. I just simplified a logic a bit here though.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: We don't want to show all the unnecessary variables for selection for this dropdown list for the variable chart. We want to show the only state variables.

Tom-Szendrey: Although looking at a parameter like beta may be better seen on the chart above this one 
I think that seeing observables like hospitalized is worth allowing the user to do here
Previously we were filtering not Param, if you want to be explicit with what we are (instead of are not) allowing i think we should allow `state` as well as `observable`

https://github.com/DARPA-ASKEM/terarium/blob/d415c37efe59f1908c2e4289f94b45cd6c10f551/packages/client/hmi-client/src/services/model.ts#L195-L201

jryu01: Ok I just added observables to the list.

PR #4479, opened 2024-08-16
Do not mirror workflow

mwdchang:
### Summary
Remove ES-mirroring, this speeds up the total round-trip time for workflow update by about 30-35 tiems, since it was waiting for ES to refresh. 

We have moved to PG, ES is no longer the primary data storage and this mirroring is not needed.


YohannParis (approved):
Lovely

dvince2 (approved):

PR #4486, opened 2024-08-16
another attempt at fixing workflow bugs

mwdchang:
### Summary
We cannot just do `nodeA = Object.assign(nodeA, update)` because some of our operators are written to be very reactive - in their watcher they do not check if the new/old values are equivalent. So when we sync up with the database it triggers, among other things, unwanted append-output cycles, which itself invalidates downstream and marks the workflow dirty in order to trigger another update ... and so the cycle repeats.

So instead we need to update node piecemeal


YohannParis (approved):

Tom-Szendrey (approved):
At first glance this makes sense and LGTM
Passes a quick round of monkey testing

shawnyama (approved):

shawnyama (approved):

PR #4490, opened 2024-08-19
better workflow save/update timing

mwdchang:
### Summary
- Avoid save/update cycle when node being dragged, which can create weird jitter effects and disconnected edges
- Ensure save is called when changing between workflows

### Testing
Make some change to existing workflow, navigate away immediately, come back to save workflow and check the change persisted.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -63,7 +63,9 @@
 					top: `${node.y}px`,
 					left: `${node.x}px`
 				}"
+				@dragstart="nodeDragging = true"
 				@dragging="(event) => updatePosition(node, event)"
+				@dragend="nodeDragging = false"
 			>
 				<tera-operator
 					ref="teraOperatorRefs"
@@ -256,6 +258,7 @@ let currentPortPosition: Position = { x: 0, y: 0 };
 let isMouseOverPort: boolean = false;
 let saveTimer: any = null;
 
+const nodeDragging = false;
 let workflowDirty: boolean = false;
 let startTime: number = 0;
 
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -63,7 +63,9 @@
 					top: `${node.y}px`,
 					left: `${node.x}px`
 				}"
+				@dragstart="nodeDragging = true"
```
Tom-Szendrey, line 66: overwriting a const
`Type 'true' is not assignable to type 'false'.ts(2322)
`

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
mwdchang: weird ... I fixed that with my last commit...

Tom-Szendrey: Its declared as a constant and not with `let` or a `ref` so you are unable to reassign 
Your last commit may not have been pushed? I only see 2 commits from ~10 hours ago

mwdchang: Yes ... but okay weirdly enough the pre-commit hook thinks otherwise. I think I managed to trick it this time though.

```
Running secrets encryption check
✔ Preparing lint-staged...
✔ Hiding unstaged changes to partially staged files...
✔ Running tasks for staged files...
✖ Prevented an empty git commit!
↓
  ✖ lint-staged failed due to a git error.
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

  ⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration

husky - pre-commit hook exited with code 1 (error)
```

mwdchang: Oh ... I think prettier changed my `let` to a const` 😠

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -256,6 +258,7 @@ let currentPortPosition: Position = { x: 0, y: 0 };
 let isMouseOverPort: boolean = false;
 let saveTimer: any = null;
 
+const nodeDragging = ref<boolean>(false);
```
shawnyama, line 261: `isDraggingNode` would sound better

shawnyama (approved):

PR #4496, opened 2024-08-19
style: stretch toggleable input to fill space

shawnyama:
# Description

Using `flex-1` to fill in the space for the input field. Shifted around some html to accommodate this.

BEFORE: 
image

AFTER:

https://github.com/user-attachments/assets/131751bc-2bcc-4cb6-9d67-3be9cc88e3ca

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
@@ -48,6 +48,10 @@ header > div {
 	display: inline-flex;
 	gap: var(--gap-1);
 	align-items: center;
+
+	&:first-child {
+		flex: 1;
+	}
 }
 
 section {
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
```diff
@@ -48,6 +48,10 @@ header > div {
 	display: inline-flex;
 	gap: var(--gap-1);
 	align-items: center;
+
+	&:first-child {
+		flex: 1;
+	}
```
mwdchang, line 54: May be little heavy-handed? There are a lot of component running with tera-drilldown-section and are are changing the first-child without really knowing what the first child is?

shawnyama: This is the first child of the header which typically contains a title or instructions. The second child contains buttons. 

I don't think there would be a case where the first child would inhibit the second one. Granting the first child with a `flex: 1` just enables it's children to use `flex: 1`.

mwdchang: okay

mwdchang (approved):

blanchco (approved):

PR #4499, opened 2024-08-19
Dvince/clean up workflows

dvince2:
# Description

This is a tentitive proposal for how to clean up workflows.  Functionally this works however I'm not sure if its the "best" solution available and any feedback would be welcome. Basically...

* In the workflow table, we check the `nodes` and `edges` columns which are json arrays, and collect any workflows which have a node and an edge marked for deletion
* These workflows are returned in a set as there may be some overlap between the two collections.
* In Java we remove the edges and nodes from their lists inside of the workflow
* The workflows are all saved.

Is it possible to find both nodes and edges at once in a single call or will it always be two?


packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/WorkflowRepository.java
```diff
@@ -1,9 +1,33 @@
...
+	/**
+	 * Find all workflows who have at least one edge or node marked for deletion
+	 */
+	@Query(
+		value = "SELECT w.* FROM Workflow w " +
+		" cross join json_array_elements(edges) as edgesData" +
+		" where edgesData->>'isDeleted' = 'true'",
+		nativeQuery = true
+	)
+	List<Workflow> findEdgesToBeDeleted();
```
bigglesandginger, line 21: This would be more correctly called `find WorkflowsWithEdgesToBeDeleted`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/repository/data/WorkflowRepository.java
```diff
@@ -1,9 +1,33 @@
...
+	/**
+	 * Find all workflows who have at least one edge or node marked for deletion
+	 */
+	@Query(
+		value = "SELECT w.* FROM Workflow w " +
+		" cross join json_array_elements(nodes) as nodesData" +
+		" where nodesData->>'isDeleted' = 'true'",
+		nativeQuery = true
+	)
+	List<Workflow> findNodesToBeDeleted();
```
bigglesandginger, line 32: Similarly this would be more correctly called `findWorkflowsWithNodesToBeDeleted`

bigglesandginger (approved):
I think this is a "best of both worlds" approach.  It is clear, the SQL is easy to read, and the Java leverages the existing knowledge space.  Could it be faster all in PostGres... maybe but for not much of a gain.

mwdchang (approved):
Didn't test, but code seems okay

PR #4502, opened 2024-08-19
fix menuitem generation logic

mwdchang:
### Summary
The generator is faulty because it only looks at the first item from the outputMap, but there can be multiple items when an operator produces 2 or more different asset types (e.g. calibration). This PR reworks the logic to take this into account.


blanchco (approved):
nice

asylves1 (approved):

PR #4509, opened 2024-08-20
fix(compare-models): overview is saved in state, add markdown renderer

shawnyama:
# Description
- overview is generated once and saved in the node state
- I grant the model compare a default status once two models are attached to it
- added markdown-it package to render markdown

https://github.com/user-attachments/assets/9b038dd3-19f2-4dfe-843b-6c990a722db9


packages/client/hmi-client/src/components/workflow/ops/model-comparison/model-comparison-operation.ts
```diff
@@ -5,6 +5,7 @@ const DOCUMENTATION_URL =
 	'https://githubicom/gyorilab/mira/blob/7314765ab409ddc9647269ad2381055f1cd67706/notebooks/hackathon_2023.10/dkg_grounding_model_comparison.ipynb#L307';
 
 export interface ModelComparisonOperationState extends BaseState {
+	overview: string;
```
mwdchang, line 8: We should avoid storing giant strings, or large data structures in operator state. It slow things down and makes all other workflow related tasks more difficult to coordinate.

packages/client/hmi-client/package.json
```diff
@@ -73,6 +73,7 @@
 		"vue": "3.3.13",
 		"vue-feather": "2.0.0",
 		"vue-gtag": "2.0.1",
+		"vue-markdown-render": "^2.2.1",
```
mwdchang, line 76: This library looks fairly simple - have we done any evaluation here on what it adds? 
It may be better to just work with "markdown-it" directly instead of having to maintain dozens and dozens of depfendencies.

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison-node.vue
```diff
@@ -9,16 +9,22 @@
...
 	node: WorkflowNode<ModelComparisonOperationState>;
 }>();
 
-const hasAtLeastTwoValues = computed(() => props.node.inputs.filter((input) => input.value).length >= 2);
+const hasAtLeastTwoValues = computed(() => {
+	const flag = props.node.inputs.filter((input) => input.value).length >= 2;
+	// This is a custom way of granting a default status to the operator, since it has no output
+	if (flag) emit('update-status', OperatorStatus.DEFAULT);
+	else emit('update-status', OperatorStatus.INVALID);
+	return flag;
```
mwdchang, line 26: Generally speaking, I think we should avoid having computed properties having these types of "side effects" where it can modify external environments. It can cause unexpected behaviours since people do not tend to think of computed behaving in this manner.

shawnyama: sure. I think it'll make more sense to set it to `DEFAULT` once the overview is generated

packages/client/hmi-client/package.json
shawnyama: so i am just using `markdown-it` now. seems like the dependencies added to the `yarn.lock` are slightly less and more up to date actually

packages/client/hmi-client/src/components/workflow/ops/model-comparison/model-comparison-operation.ts
shawnyama: for this pass, I am actually not going to save anything into the state. I changed it to an `overviewId` but not gonna use it until I chat with @kbirk about how we should cache them

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison.vue
```diff
@@ -294,16 +290,32 @@ async function buildJupyterContext() {
...
+}
+
+async function generateOverview() {
+	// Generate if there is no overview and the comparison task has been completed
+	if (!compareModelsTaskOutput || props.node.state.overviewId) return;
+	// const newOverviewId = uuidv4(); // TODO: Save overview to S3
+	assignOverview(compareModelsTaskOutput);
+	// const state = cloneDeep(props.node.state);
+	// state.overviewId = newOverviewId;
+	// emit('update-state', state);
```
Tom-Szendrey, line 310: Is this meant to still be commented out?

Tom-Szendrey (approved):
LGTM as first pass (Still not saving to state as you mentioned)

mwdchang (approved):

PR #4516, opened 2024-08-20
Fix broken missing indices in tests and disable problematic async stuff in tests

kbirk:


dvince2 (approved):

bigglesandginger (approved):
I am sure there is some proper "mock" way to do this, but I understand this :)

PR #4518, opened 2024-08-20
4452 Added expand feature for the vega chart

### LINKED ISSUE #4452: [TASK]: Investigate showing (popping up) larger version of vegalite chart
Pascale was asking if it is possible to "pop up" a larger version of the Vegalite chart, similar to how model-comparison has this "preview" kind of things with PrimeVue images (https://primevue.org/image/)

```
          <li v-for="(image, index) in structuralComparisons" :key="index">
            <label>Comparison {{ index + 1 }}</label>
            <Image id="img" :src="image" :alt="`Structural comparison ${index + 1}`" preview />
          </li>
```

Not a priority I think, because Vegalite itself can already be detatched into Vegalite-editor in a different window/tab ... but might be good to spend maybe half a day looking at what options are there.

jryu01:
# Description

- Allow vega chart to be expanded and rendered in a larger modal once expand button is clicked.
- Added `expand` prop to the vegachart to show the expand button on the chart.
- When a function is provided to `expand`,  function returns the customized chart spec to be rendered in the expand mode. 

https://github.com/user-attachments/assets/a46440be-f05f-4816-8d10-c7c10ce9f6c3

Resolves #4452

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,7 @@
             "runtimeExecutable": "yarn",
             "cwd": "${workspaceFolder}/packages/client/hmi-client",
             "runtimeArgs": [
-                "dev"
+                "staging"
             ],
             // Run vite dev server using `yarn dev` and then luanch the chrome browser
             "serverReadyAction": {
--- a/packages/client/hmi-client/src/components/widgets/VegaChart.vue
+++ b/packages/client/hmi-client/src/components/widgets/VegaChart.vue
@@ -70,8 +70,8 @@ const isExpanded = ref(false);
 const onExpand = async () => {
 	if (vegaContainerLg.value) {
 		const defaultSize = {
-			width: window.innerWidth / 1.3,
-			height: window.innerHeight / 1.3
+			width: window.innerWidth / 1.5,
+			height: window.innerHeight / 1.5
 		};
 		let spec = deepToRaw(props.visualizationSpec) as any;
 		spec.width = defaultSize.width;
--- a/packages/client/hmi-client/src/components/widgets/VegaChart.vue
+++ b/packages/client/hmi-client/src/components/widgets/VegaChart.vue
@@ -49,7 +49,7 @@ const props = withDefaults(
 		areEmbedActionsVisible: true,
 		intervalSelectionSignalNames: () => [],
 		config: null,
-		expand: true
+		expand: false
 	}
 );
 const vegaContainer = ref<HTMLElement>();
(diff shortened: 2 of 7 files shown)


.vscode/launch.json
```diff
@@ -11,7 +11,7 @@
             "runtimeExecutable": "yarn",
             "cwd": "${workspaceFolder}/packages/client/hmi-client",
             "runtimeArgs": [
-                "dev"
+                "staging"
```
mwdchang, line 14: This doesn't impact me, but you should check with your VsCode friends to see if they are okay with this.

Tom-Szendrey: I am not running via my editor directly so this should not impact me 
I am utilizing the following in `iTerm`
```
./hmiServerDev.sh start staging run
yarn local
```

packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -32,18 +44,47 @@ const props = withDefaults(
 		 */
 		intervalSelectionSignalNames?: string[];
 		config?: Config | null;
+		/**
+		 * Whether to show the expand button.
+		 * If a function is provided, it will be called before expanding the chart, and the returned spec will be used for the expanded chart.
+		 */
+		expand?: boolean | ((spec: VisualizationSpec) => VisualizationSpec);
```
mwdchang, line 51: What is the use-case for passing-in `expand` as a function? Is there an example?

mwdchang: nevermind found it ... search was being weird

mwdchang (approved):
nice!

packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -32,18 +44,47 @@ const props = withDefaults(
...
 	}>(),
 	{
 		areEmbedActionsVisible: true,
 		intervalSelectionSignalNames: () => [],
-		config: null
+		config: null,
+		expand: false
```
Tom-Szendrey, line 57: Nit pick: Is it worth to rename to `expandable` ?

Tom-Szendrey (approved):
Thank you!!

packages/client/hmi-client/src/components/widgets/VegaChart.vue
jryu01: sure

PR #4519, opened 2024-08-21
show spinner for model-config extraction

asylves1:
# Description

Add spinner to model config node when user extracts inputs

Screenshot 2024-08-21 at 8 53 27 AM

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
@@ -10,8 +10,8 @@ export interface ModelConfigOperationState extends BaseState {
 	transientModelConfig: ModelConfiguration;
 	notebookHistory: NotebookHistory[];
 	hasCodeRun: boolean;
-	datasetModelConfigTaskId?: string;
-	documentModelConfigTaskId?: string;
+	datasetModelConfigTaskId: string;
+	documentModelConfigTaskId: string;
 }
 
 export const blankModelConfig: ModelConfiguration = {
@@ -43,7 +43,9 @@ export const ModelConfigOperation: Operation = {
 		const init: ModelConfigOperationState = {
 			transientModelConfig: blankModelConfig,
 			notebookHistory: [],
-			hasCodeRun: false
+			hasCodeRun: false,
+			datasetModelConfigTaskId: '',
+			documentModelConfigTaskId: ''
 		};
 		return init;
 	}
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
@@ -10,6 +10,8 @@ export interface ModelConfigOperationState extends BaseState {
 	transientModelConfig: ModelConfiguration;
 	notebookHistory: NotebookHistory[];
 	hasCodeRun: boolean;
+	datasetModelConfigTaskId?: string;
+	documentModelConfigTaskId?: string;
 }
 
 export const blankModelConfig: ModelConfiguration = {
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
```diff
@@ -10,6 +10,8 @@ export interface ModelConfigOperationState extends BaseState {
 	transientModelConfig: ModelConfiguration;
 	notebookHistory: NotebookHistory[];
 	hasCodeRun: boolean;
+	datasetModelConfigTaskId?: string;
+	documentModelConfigTaskId?: string;
```
mwdchang, line 14: Generally speaking, we prefer not to have optional fields as a part of operators' state. It is a lot easier to handle when we know precisely what is there and what is not.

Can we remove the optionals here?

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -13,19 +14,42 @@
...
+	if (event.data?.id !== taskIdRefs[event.type].value) return;
+	if ([TaskStatus.Success, TaskStatus.Cancelled, TaskStatus.Failed].includes(event.data.status)) {
+		taskIdRefs[event.type].value = '';
+	}
+};
+
+useClientEvent(ClientEventType.TaskGollmConfigureModel, configModelEventHandler);
+useClientEvent(ClientEventType.TaskGollmConfigureFromDataset, configModelEventHandler);
+
+const isLoading = computed(() => datasetModelConfigTaskId.value !== '' || datasetModelConfigTaskId.value !== '');
```
mwdchang, line 51: Typo or you are checking the wrong things ... you are saying `A !== '' || A !== ''`

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -633,6 +619,20 @@ const resetConfiguration = () => {
...
+		documentModelConfigTaskId.value = props.node.state.documentModelConfigTaskId ?? '';
+	}
+);
+
+watch(
+	() => props.node.state.datasetModelConfigTaskId,
+	() => {
+		datasetModelConfigTaskId.value = props.node.state.datasetModelConfigTaskId ?? '';
+	}
+);
```
mwdchang, line 634: I think e can clean up things a bit here with a string trick, instead of having so many watchers and trying to manage so many variables....we cao do something like

```
const isLoading = ref(false)
...

watch(
	() => props.node.state.datasetModelConfigTaskId + ": " + props.node.state.datasetModelConfigTaskId
	(watchStr) => {
    if (watchStr !== ':') isLoading.value = true;
    else isLoading.value = false;
	}
);
```

Then we can remove one watcher, and remove the extra refs documentModelConfigTaskId and datasetModelConfigTaskId

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
asylves1: yup typo

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -79,4 +103,30 @@ watch(
 	},
 	{ deep: true }
 );
+
+watch(
```
Tom-Szendrey, line 107: This section is confusing me at first glance 
We have 2 watchers on the state to update our id fields
In a separate location we then have a computed on these id fields
and in a 3rd location we have a watcher on the computed which sometimes updates the state

We may be able to clean this up but im not certain on whats happening here

Tom-Szendrey: Thanks for the explanation!

mwdchang (approved):

PR #4521, opened 2024-08-21
4476 task revisit elasticsearchservice semantic

dvince2:
# Description

Resolves #(issue)


packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchServiceTests.java
```diff
@@ -72,9 +73,12 @@ public void testItCanSearchPrivateProjectAsReader() throws Exception {
 		project.setDescription("my description");
 
 		projectSearchService.indexProject(project);
+		projectSearchService.forceESRefresh();
 		projectSearchService.addProjectPermission(project.getId(), searcherId);
+		projectSearchService.forceESRefresh();
```
mwdchang, line 78: You only need one at the end of the logic if you are doing forced-refresh index-wide.

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -136,6 +136,10 @@ public void indexProject(final Project project) throws IOException {
 		elasticService.indexWithRouting(getAlias(), project.getId().toString(), doc, routing);
 	}
 
+	public void forceESRefresh() throws IOException {
```
bigglesandginger, line 139: This is only used in tests - correct?

bigglesandginger (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
dvince2: correct!

PR #4524, opened 2024-08-21
Handle error state for the notification

jryu01:
# Description

There was an issue that error state notification appeared as normal and can't be cleared out from the notification panel. 
This fixes the issue.
Screenshot 2024-08-21 at 11 45 20 AM
E.g ^ notification errored out but looks fine and wasn't being cleared out when clear notification was clicked. 

Resolves #(issue)


Tom-Szendrey (commented):
What is the difference between error and failed?
Is there a reason to keep both around?

mwdchang (approved):
Seems okay, but we should consolidate the various status codes

jryu01:
> What is the difference between error and failed? Is there a reason to keep both around?

Both are from simulation progressState, and it looks like they both are used for failure status.

asylves1 (approved):

PR #4527, opened 2024-08-21
change update to a debounce model

mwdchang:
### Summary
This should make the updates a bit smoother, both with shortened intervals and pushing the update request instead of just marking the workflow as dirty and waiting for next update interval.

This is not a fool proof solution - various actions are happening too fast the UI and server can get out of sync.


Tom-Szendrey (approved):
Seems good 
Thanks!

shawnyama (approved):
works fine for me

PR #4531, opened 2024-08-22
Cleanup utils and experiments

mwdchang:
### Summary
Another round of clean up
- Remove unused theia experiment and loss function rendering
- Move workflow-localstorage functions to workflow-service
- Remove unused petrinet matrix service


Tom-Szendrey (approved):
Thanks Dan!! 
Without all of these cleanup PRs I am scared to think what our codebase would look like knowing what it does now

asylves1 (approved):

PR #4535, opened 2024-08-22
Generate error for calibration

Tom-Szendrey:
# Description
Getting error data for calibration utilizing the incoming dataset + post calibration simulation run's data
Create a util to get mean absolute errors.
Create a helper function to get calibration data into correct form


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts
```diff
@@ -29,3 +30,64 @@ export const mergeResults = (
...
+			The timestamp column is titled: timepoint_id
+	* Assume that the mapping is in the calibration form:
+			Ground truth will map to datasetVariable
+			Simulation data will map to modelVariable AND not include _State
+ * transform data, utilize mae, return mean aboslute error for charts.
+ * Note: This will only compare rows with the same timestep value.
+ */
+export async function getErrorData(groundTruth: DataArray, simulationData: DataArray, mapping: CalibrateMap[]) {
+	const errors: DataArray = [];
+	const pyciemssMap = await parsePyCiemssMap(simulationData[0]);
```
mwdchang, line 48: Pretty sure that `parsePyCiemssMap` does not need to be async, because it is just doing string twiddlings, and thus `getErrorData` does not nee to be async, and so forth.

packages/client/hmi-client/src/utils/stats.ts
```diff
@@ -11,3 +13,15 @@ export const stddev = (numberList: number[], usePopulation = false) => {
...
+// Note this will only compare values that share the same key.
+export const mae = (map1: Map<number, number>, map2: Map<number, number>) => {
+	const sharedKey = [...map1.keys()].filter((key) => map2.has(key));
+	const error = _.meanBy(sharedKey, (key) => {
+		const firstValue = map1.get(key) as number;
+		const secondValue = map2.get(key) as number;
+		return Math.abs(firstValue - secondValue);
+	});
+	return error;
+};
```
mwdchang, line 27: Maybe okay for now but slightly confusing, consider a more general usage scenario of a developing utility toolkit ...

People will generally have two or more time series that want to run the metric on, and not two maps. So from a design and API perspectives this is not particularly friendly - you are forcing people to transform their data before they can use this function, further more, you are transforming the Maps back into arrays for various operations here ... so maybe the question is why you chose to use Maps in the first place?

```
export const mae = (dataArray1: DataArray, dataArray2: DataArray) ..
```

Would be a better choice in my opinion, let me know if you want to have a little chat.

Tom-Szendrey: Utilizing `export const mae = (dataArray1: DataArray, dataArray2: DataArray, keyField: string, valueField: string)`  as discussed
I am exporting the DataArray type rather than importing it from our simulation service as this is in a util file and should not depend on simulation service. Let me know if i should change this

mwdchang (approved):

jryu01 (approved):

PR #4552, opened 2024-08-22
style: proper colors for model compare legend

shawnyama:
# Description

- updated model compare colors in notebook legend
- also removed the `overviewId` from the node state, the caching is actually done my gollm itself when you recall `compareModels` with the same params

Screenshot 2024-08-22 at 3 12 51 PM

Resolves #(issue)


Tom-Szendrey (approved):
handy caching is done for you!

jryu01 (approved):

PR #4558, opened 2024-08-22
Create beaker cell components

blanchco:
# Description

Resolves #(issue)


packages/client/hmi-client/src/components/llm/tera-beaker-code-cell-output.vue
```diff
@@ -0,0 +1,60 @@
...
+onMounted(() => {
+	updateOutputs();
+});
+
+watch(
+	() => props.jupyterMessage,
+	() => {
+		updateOutputs();
+	},
+	{ deep: true }
```
mwdchang, line 56: Not sure how complicated are these messages, but do we need deep? If it is a message wouldn't every message be unique?

blanchco: You're right its not needed, removed `deep`

mwdchang (approved):
Don't really know how to test the cache parts, the other things look okay

blanchco:
> Don't really know how to test the cache parts, the other things look okay

Beforehand you would see that the outputs would take time to show up since we would re-run all the code cells to show the outputs (you can see this currently on staging).  Now when you open the drilldown, you can see them right away

packages/client/hmi-client/src/components/llm/tera-beaker-code-cell-output.vue
```diff
@@ -0,0 +1,59 @@
...
+onMounted(() => {
+	updateOutputs();
+});
+
+watch(
+	() => props.jupyterMessage,
+	() => {
+		updateOutputs();
+	}
+);
```
shawnyama, line 56: add an immediate: true so you dont need onmounted

shawnyama (approved):

packages/client/hmi-client/src/components/llm/tera-beaker-code-cell-output.vue
blanchco: Doing this breaks it unfortunately...I think its because onMounted makes sure the DOM is fully rendered and the way the outputs are render are quite unique 😬

PR #4564, opened 2024-08-23
Optimize constraint - adding observables 

Tom-Szendrey:
# Description
-Add observables as an option for optimize constraints.
-appending `_state` and `_observable` in hmi-client as it readily has the model accessible while we are creating these dropdown options anyways. This seems like a better spot than pyciemss-service where we would need to do more

Related +  required: 
https://github.com/DARPA-ASKEM/pyciemss-service/pull/114

# Testing
While running pyciemss-service branch locally:
-running optimize drilldown with a state and ensure that it still runs + the output setting correctly gets defaulted
-running optimize drilldown with an observable, ensure it runs correctly + the output setting correctly gets defaulted.

image


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -656,13 +659,20 @@ const initialize = async () => {
...
+		value: `${state.id}_state`
+	}));
+
+	// Add obs:
+	if (model.value?.semantics?.ode.observables) {
+		modelStateAndObsOptions.value.push(
+			...model.value.semantics.ode.observables.map((observable: any) => ({
+				label: observable.id,
+				value: `${observable.id}_observable`
+			}))
```
mwdchang, line 673: Are these the field-names used in pyciemsss-service (label / value) ? Wondering if we can be a bit more explicit than just saying "value".

Would be a bit easier to read if you extract out
- model.value?.semantics?.ode, and
- model.value?.model

Tom-Szendrey: I will extract these out

Label = dropdown label for UI
value = the value being used for the pyciemss-service call

Maybe I can rename value to something more relating to constraint target variable

mwdchang (approved):

mwdchang:
merging cause we already merged the pyciemss-service couterpart

PR #4565, opened 2024-08-23
fix: disable wizard for complex models

shawnyama:
# Description

- model edit drilldown defaults to notebook
- wizard is disabled if there are more than 50 transitions (every transition is a template that triggers an `amr-to-mmt` call)

image


Tom-Szendrey (approved):
I thought the issue was that the state's notebook code is getting reset?
That being said im not up to date with this

shawnyama:
> I thought the issue was that the state's notebook code is getting reset? That being said im not up to date with this

this fixes a different problem. very stratified models end up overwhelming the server with too many `amr-to-mmt` requests

blanchco (approved):

PR #4582, opened 2024-08-26
dataset column information redesign

shawnyama:
# Description

* Enriching columns occurs reactively
* Column information can now be edited and saved similarly to the model parts
* I am only showing one concept now even though an array of them are generated after enrichment, designs should be updated
* I am not using that `enrichDataset()` in the dataset util - doesn't seem to do anything helpful now

https://github.com/user-attachments/assets/4f9eccfe-d8b3-4410-b0aa-1043c97711c9

### Testing
* Add a new dataset
* Enrich it
* Make some edits and click the save button
* See if its saved


Tom-Szendrey:
Hitting reset does not correctly clear the concept
1) start with a dataset
2) add a new concept 
3) hit reset

image

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -217,6 +283,22 @@ const card = computed(() => {
 const description = computed(() => dataset.value?.description?.concat('\n', card.value?.DESCRIPTION ?? '') ?? '');
 const author = computed(() => card.value?.AUTHOR_NAME ?? '');
 
+function updateColumn(index: number, key: string, value: any) {
+	if (!transientDataset.value?.columns?.[index]) return;
+	if (key === 'unit' || key === 'name') {
+		transientDataset.value.columns[index].metadata[key] = value;
```
Tom-Szendrey, line 289: metadata is optional
If I have an unenriched dataset and the metadata does not exist for my column then this will fail.

packages/client/hmi-client/src/components/dataset/tera-column-info.vue
```diff
@@ -0,0 +1,193 @@
...
+import type { FeatureConfig } from '@/types/common';
+
+type ColumnInfo = {
+	symbol: string;
+	dataType?: ColumnType;
+	description?: string;
+	grounding?: Grounding;
+	// Metadata
+	unit?: string;
+	name?: string;
```
Tom-Szendrey, line 83: Can we utilize `DatasetColumn` instead of this `ColumnInfo`? I may be misunderstanding but this looks to be a subsection of that?

Also what is the point in symbol and name? Are they not both just `DatasetColumn.name`?
If they are both derived from the same `DatasetColumn.name` how does updating one work?

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -149,12 +149,13 @@ const optionsMenuPt = {
 
 async function updateModelContent(updatedModel: Model) {
 	if (!useProjects().hasEditPermission()) {
+		logger.error('You do not have permission to edit this model.');
```
YohannParis, line 152: I would prefer for the button or actions allowing the users to edit the model to be disabled if they do not have the rights to do so.

packages/client/hmi-client/src/components/widgets/tera-asset-enrichment.vue
```diff
@@ -114,17 +114,17 @@ function closeDialog() {
 
 const confirm = async () => {
 	isLoading.value = true;
+	closeDialog();
 
 	if (dialogType.value === DialogType.ENRICH) {
-		sendForEnrichment();
+		await sendForEnrichment();
```
YohannParis, line 120: Why the need to add `await`?

packages/client/hmi-client/src/components/model/tera-model.vue
shawnyama: I'll write a comment about that

packages/client/hmi-client/src/components/widgets/tera-asset-enrichment.vue
shawnyama: I want to call the `emit('finished-job')` (which refetches the dataset) after I get a response. I need to wait for the dataset to update so then I can fetch the updated version. Added a comment explaining this.

packages/client/hmi-client/src/components/dataset/tera-column-info.vue
shawnyama: Yeah I tried to do this at first but I had to go for a more UI based approach so the object I mutate have the exact values that I want and are easy to access. Some of the values here are from metadata (see how I construct it in `columnInformation` in `tera-dataset`). I did a similar approach when editing model parts

The `name` is initialized to what the `symbol` is. However it is saved and then read from in `metadata.name` once any edits are applied to it.

YohannParis (approved):

packages/client/hmi-client/src/components/dataset/tera-column-info.vue
```diff
@@ -0,0 +1,208 @@
...
+		if (concept) {
+			emit('update-column', { key: 'concept', value: { curie: concept.curie, name: concept.name } });
+		}
+	}
+}
+
+watch(
+	() => props.column.grounding?.identifiers,
+	async (identifiers) => {
+		// console.log(identifiers); // FIXME: Multiple identifiers are held in here after enrichment! Designs have to be updated to handle more.
```
Tom-Szendrey, line 121: Not a big deal:
If there is a ticket assigned for these designs/fix it would be helpful to have it here so this doesnt get lost?

Tom-Szendrey (approved):

PR #4583, opened 2024-08-26
fix: matrix editing in model config

shawnyama:
# Description
- injected configuration values into mmt that matrix is based on

https://github.com/user-attachments/assets/2ce0e85d-2ee4-4720-a603-0191ec8583ae


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -467,6 +467,24 @@ const model = ref<Model | null>(null);
 const mmt = ref<MiraModel>(emptyMiraModel());
 const mmtParams = ref<MiraTemplateParams>({});
 
+const configuredMmt = computed(() => {
+	const mmtCopy = cloneDeep(mmt.value);
+	if (!mmtCopy) return mmtCopy;
```
mwdchang, line 472: Not exactly sure what you are trying to say here??? `if (!a) return a` ?

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -467,6 +467,24 @@ const model = ref<Model | null>(null);
...
+			mmtInitial.expression = initial.expression;
+		}
+	});
+	knobs.value.transientModelConfig.parameterSemanticList.forEach((parameter) => {
+		const mmtParameter = mmtCopy.parameters[parameter.referenceId];
+		if (mmtParameter) {
+			mmtParameter.value = parameter.distribution.parameters.value;
+		}
+	});
+	return mmtCopy;
```
mwdchang, line 485: You also hooked this up to observables in the template section, so should we also process observables?

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
shawnyama: oof my bad

mwdchang (approved):

PR #4595, opened 2024-08-27
extra workflow update guard

mwdchang:
### Summary
I can't think of a scenario of why this might happen, but just in case we will verify the ids are the same before allowing update to continue.


blanchco (approved):
okay 😄

shawnyama (approved):

PR #4597, opened 2024-08-27
save as model configuration

blanchco:
# Description

* Save as functionality for model config

Disable `Save` on value changes only

https://github.com/user-attachments/assets/a852e443-a768-4a76-80ce-77391e8c2509

`Save` when non value changes

https://github.com/user-attachments/assets/4b3964a1-ee3d-4954-a2fc-409947c4adc7

`Save as` when value changes

https://github.com/user-attachments/assets/69e27370-f469-4c26-a69f-fb8597f5d0e3


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -301,8 +294,56 @@ const calibratedConfigObservables = computed<Observable[]>(() =>
...
+const isSaveDisabled = computed(
+	() =>
+		knobs.value.transientModelConfig.name === '' ||
+		isEqual(originalConfig, knobs.value.transientModelConfig) ||
+		haveValuesChanged.value
+);
+
+// a property to check if the values have changed (not metadata such as name, description, sources...)
+const haveValuesChanged = computed(() => {
+	const modelConfig = cloneDeep(knobs.value.transientModelConfig);
```
mwdchang, line 307: Consider
- Moving this function out into a util, so we can unit test it
- Using cache maps so the lookup is more efficient and not `O(n^2)` , cause we have stratified models
- Observables

blanchco: Good points, all of the above has been added

packages/client/hmi-client/src/utils/model-configuration.ts
```diff
@@ -0,0 +1,59 @@
+import { ModelConfiguration } from '@/types/Types';
+import { cloneDeep, isEqual } from 'lodash';
+
+export const areModelConfigValuesEqual = (
+	originalConfig: ModelConfiguration | null,
+	newConfig: ModelConfiguration | null
+) => {
```
mwdchang, line 7: Thanks for creating a util function. I think maybe we can just move this to `ops/model-config/model-config-operation.ts`. It is a bit more thematic and easier to find.


Secondly, you are checking if "every-item-are-the-same" is false, another way to think about it is whether "some-items-are-different" is true. If we believe that people are going to tweak value fields more than they are going to tweak metadata fields for model-configuration, then the latter option would be a more efficient check as it would short circuit exit much earlier on average.

blanchco: Sure, I've moved the function and now I'm using `.some` instead.

I've also created a function to check if the model configurations are equal, the way we were doing it originally was not working as intended.  This has been fixed with that new function `isModelConfigsEqual`.  Let me know what you think 😄

packages/client/hmi-client/src/components/workflow/ops/model-config/model-config-operation.ts
```diff
@@ -50,3 +51,122 @@ export const ModelConfigOperation: Operation = {
...
+		'inferredParameterList'
+	];
+	const baseKeysAreEqual = isEqual(omit(originalConfig, keysToOmit), omit(newConfig, keysToOmit));
+
+	if (!baseKeysAreEqual) return false;
+
+	// secondly check if the lists are equals
+	const listKeysToOmit = ['createdOn', 'updatedOn', 'id'];
+	const sortedOriginalConfigLists = {
+		initialSemanticList: sortBy(
```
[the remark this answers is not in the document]
mwdchang: Let's have a quick sync up, I think we may be a bit into the weeds about the meaning of equality, but also lodash has some neat util functions that make some of these things more succinct with less boilder-plates.

```
const listX = [{a: 1, b: 2}, {a: 3, b: 4}, {a: 5, b: 6}]
const listY = [{a: 5, b: 6}, {a: 3, b: 4}, {a: 1, b: 2}]
const listZ = [{a: 1, b: 2}, {a: 3, b: 4}, {a: 5, b: 5}]


_.differenceWith(listX, listY, _.isEqual)
> []

_.differenceWith(listX, listZ, _.isEqual)
> [{…}]
```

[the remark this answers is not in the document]
blanchco: okay, I've added this.  I _think_ everything looking good now!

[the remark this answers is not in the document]
blanchco: also hiding the dropdown

mwdchang (approved):
looks fine thanks, there are a couple of things we should probably add as a later todo:
- Calling "initialize" when you save a config is kind of expensive, we can probably work around that by breaking initialize() into more basic parts
- As discussed, we will want to create a "set-outputport" rather than "append-outputport"

asylves1 (approved):

PR #4602, opened 2024-08-27
show model parts in model edit

shawnyama:
# Description

* added `model-parts` to model edit (output looks similar to stratify)
* threw in a spinner as well (shows when fetching the amr or while the edited model is coming)
* removed some old classes and divs
* removed pointless wrapper function and just merged the code running into one function

https://github.com/user-attachments/assets/e45e9492-c7ff-441b-b369-da7ab42b4a75


asylves1 (approved):

blanchco (approved):
looks fine

PR #4603, opened 2024-08-27
Model Edit append output on run

Tom-Szendrey:
# Description

- [x] Append new output on run and reset button.
- [x] remove outdated draft stuff - we never prompted the user for creating on close and do not intend to going fwd.
- [x] Utilize watcher instead of onUpdate as onUpdate was causing very weird interactions with the workflow (explained below)

# Testing:
Opening model edit and running some code to see a new output
Opening model edit and hitting reset to see a new output
Utilizing new outputs in downstream operators such as model configuration.

image

Weird interaction that was found with onUpdate was that when the drilldown appended an output the onUpdate would call updateModel. Im not certain why but this would then cause the workflow to become uninteractive until I refreshed the page


packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit-node.vue
```diff
@@ -28,12 +28,21 @@ const props = defineProps<{
...
 };
+
 onMounted(updateModel);
-onUpdated(updateModel);
+
+// Updates output selection
+watch(
+	() => props.node.active,
+	async () => {
+		await updateModel;
```
shawnyama, line 45: `await updateModel()`

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/model-edit/tera-model-edit-node.vue
shawnyama: try this, you can remove the onMounted too
```
// Updates output selection
watch(
	() => props.node.active,
	() => {
		updateModel();
	},
	{ immediate: true }
);
```

mwdchang: Let's rename this function please, update model sounds confusing when you are actually fetching the model

PR #4604, opened 2024-08-28
Added chart annotation endpoints

jryu01:
# Description

Implemented
  GET /chart-annotations/search (by nodeId)
  POST /chart-annotations
  DELETE /chart-annotations/{id}

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ChartAnnotationController.java
```diff
@@ -0,0 +1,175 @@
...
+		);
+
+		final List<ChartAnnotation> chartAnnotations = chartAnnotationService.getAnnotationsByNodeId(
+			body.nodeId,
+			permission
+		);
+
+		if (chartAnnotations.isEmpty()) {
+			return ResponseEntity.noContent().build();
+		}
```
mwdchang, line 91: This feels a little bit weird - wouldn't we want to return an empty array instead of non-content 204? 
Are we doing this across the app @dvince2 ?

That way the client-side doesn't need to have two different branches dealing with it.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ChartAnnotation.java
```diff
@@ -0,0 +1,54 @@
...
+	public enum ChartType {
+		@JsonAlias("variable")
+		VARIABLE,
+		@JsonAlias("parameter")
+		PARAMETER,
+		@JsonAlias("loss")
+		LOSS,
+		@JsonAlias("error")
+		ERROR
+	}
```
mwdchang, line 35: I am wondering if we want to hold off adding the enum and leave it as a TODO, as I am not sure we have the full listing of chart-types, and since enums are translated into constraints via hibernate it makes things messy if we want to add or remove types.

Otherwise we'd need to be as comprehensive as possible with all possible charts upfront.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ChartAnnotationController.java
jryu01: Hm.. yeah I think it's preferable to return an empty array instead of 204. I just followed code from other examples like summary endpoint. If there's no specific reason to return 204 I'll just updated it to return an empty array if it's empty.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ChartAnnotation.java
jryu01: Yeah we probably are okay without the  `chartType` at the moment considering we are creating an annotation for each individual chart, I can comment out for now if updating enum value might be messy. But I'm also curious on how the migration happens with enum values with hibernate. Does it automatically handles the migration when adding new enum values or we need a separate migration script? What about delete or update cases? @dvince2

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ChartAnnotationController.java
dvince2: We do return `no content` other places, but there isn't a set standard that the front end works with.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/ChartAnnotation.java
dvince2: For just _adding_ cases everything should be handled by `magic` and you're good.  However any time you want to remove, rename or reorder the enum then you'll be writing a migration for that

jryu01: I left tis as TODO

mwdchang (approved):

dvince2 (approved):

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -88,6 +88,15 @@ export interface Artifact extends TerariumAsset {
     metadata?: any;
 }
 
+export interface ChartAnnotation extends TerariumAsset {
+    nodeId: string;
+    outputId: string;
```
Tom-Szendrey, line 93: What is a chart annotation outputId used for?

Tom-Szendrey (approved):

packages/client/hmi-client/src/types/Types.ts
mwdchang: I think it is meant to map to the outputport-id

PR #4606, opened 2024-08-28
address sync problem of deletes within a short interval

mwdchang:
### Summary
Should alleviate the issue when trying to delete nodes in quick succession, and causing the DB/client to become de-synced with each other.

The problem is getNodes/getEdges more excludes deleted elements, and we should use the full list when sync'ing


shawnyama (approved):

Tom-Szendrey (approved):
thanks for the explanation

PR #4608, opened 2024-08-28
feat: added empty state image and moved welcome message into a closeable panel

jamiewaese-uncharted:
Apparently some users didn't realize they could edit the content in the Overview page editor.

These changes should make that more apparent.

https://github.com/user-attachments/assets/6b65d6d4-e038-47e6-9566-8eb53bbe2a32


dgauldie (approved):
LGTM

mwdchang (changes requested):
This doesn't work if I create a new project. 

Also we tend to use triple equals `===` in javascript for comparisons.

mwdchang (approved):

Issue #4488, opened 2024-08-16
[TASK]: Better way to deal with optional types that are not really optional


mwdchang:
### Preamble
The majority of our Typescript types are based off 'TerariumEntity' and 'TerariumAsset'.  Roughly speaking the types generated looks something like this combined.

```
{
  id? : string,
  name?: string,
  description?: string,
  ...
}
```

However from data-engineering point of view, `id`, `name` are not really optional per se, they are either null or some kind of string/text. And indeed they are model that way in Java which is the used to generated the Typescript typings.

The crux of the issue is the use of `@TSOptional` to satisfy a specific requirement from the user-interface, that is: **when we create an entity/asset - we do not usually have an id that we can assign**

However this decision created many other pain-points for the UI:
- The code is, generally speaking, more verbose with optional-chaining syntax sugar `a?.b".c?....`
- It is harder to write for and reason with .... does the field actually needed or not??
- It makes the interface messier because data structure is passed through the component properties

### Proposal
We should not generate typings for a specific use case, rather we should just produce the needed typings on a need-basis. For example, so we have 

```
interface Person {
  id: string;
  name: string;
  occupation: string
}
```

If we want to create a new Person, we don't have an id so it is a typing error, but there are a suite of Typescript utilities that can be used to address this. We can create:

```
interface NewPerson = Omit<Person, 'id'>

function createPerson(person: NewPerson) {
}
```

And that will give us a Person definition without the id field for the creation process. In this way, we can use Person everywhere with all the fields well-defined, while on creation we will use the NewPerson typing.

### Task
`id` field should not be optional. Other than that we can probably argue over the other fields as well. But we should do an assessment and get agreement over the basic Java classes and see exactly what we want to model.

Unfortunately because of the inheritance usage, it may not be possible to do this little by little, so actual implementation is best when there is a bit of a lull on the front-end side of things.

### References
- https://www.typescriptlang.org/docs/handbook/utility-types.html

dvince2:
Opting to close this issue.  given the time remainng for the project this won't be done, but its a good one

PR #4613, opened 2024-08-28
save as intervention policy

blanchco:
# Description

* save as for intervention policy

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
@@ -17,10 +17,12 @@ import { WorkflowNode, WorkflowPortStatus } from '@/types/workflow';
 import Button from 'primevue/button';
 import TeraOperatorPlaceholder from '@/components/operator/tera-operator-placeholder.vue';
 import { cloneDeep } from 'lodash';
-import { blankIntervention } from '@/services/intervention-policy';
-import { InterventionPolicyState } from './intervention-policy-operation';
+import { blankIntervention, createInterventionPolicy } from '@/services/intervention-policy';
+import { getInterventionPoliciesForModel } from '@/services/model';
+import { InterventionPolicy } from '@/types/Types';
+import { InterventionPolicyOperation, InterventionPolicyState } from './intervention-policy-operation';
 
-const emit = defineEmits(['open-drilldown', 'update-state']);
+const emit = defineEmits(['open-drilldown', 'update-state', 'append-output']);
 const props = defineProps<{
 	node: WorkflowNode<InterventionPolicyState>;
 }>();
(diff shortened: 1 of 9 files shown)


packages/client/hmi-client/src/services/save-asset.ts
```diff
@@ -47,24 +51,27 @@ export async function saveAs(
...
 	}
 
-	const projectId = useProjects().activeProject.value?.id;
-	if (!projectId) {
-		logger.error(`Asset can't be saved since target project doesn't exist.`);
-		return;
-	}
-	await useProjects().addAsset(assetType, response.id, projectId);
+	// TODO: Potentially add a flag in case we don't want to add the asset to the project
+	if (assetType !== AssetType.InterventionPolicy) {
```
mwdchang, line 55: Isn't this what the function parameter `openOnSave` is suppose to guard against?

packages/client/hmi-client/src/services/intervention-policy.ts
```diff
@@ -17,6 +17,8 @@ export const getInterventionPolicyById = async (policyId: string): Promise<Inter
 
 export const createInterventionPolicy = async (policy: InterventionPolicy): Promise<InterventionPolicy> => {
 	delete policy.id;
+	delete policy.createdOn;
+	delete policy.updatedOn;
```
mwdchang, line 21: How do we get a createdOn/updatedOn fields before we actually created the object?

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
```diff
@@ -30,17 +32,41 @@ const isModelInputConnected = computed(() => modelInput?.status === WorkflowPort
...
+				isSelected: false,
+				state
+			});
+		} else {
+			state.interventionPolicy = {
+				modelId,
+				interventions: [blankIntervention]
+			};
+			emit('update-state', state);
+		}
```
mwdchang, line 69: Why are we changing this watcher and creating a "default" intervention policy?

packages/client/hmi-client/src/services/save-asset.ts
mwdchang: Resolved in a chat with Cole. There are two things here
- whether we want to save into project-asset, and
- whether we wan to navigate to the asset

So these are handled in a single if-block ... may be clearer if they are independent.

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
blanchco: I thought a default policy would make sense, but a blank one doesnt really 😬.  I've removed this

packages/client/hmi-client/src/services/intervention-policy.ts
blanchco: We don't, but since they don't exist nothing will happen.  This is more for the case when we want to create a new intervention policy with "save as."  I pass in the whole intervention policy which does have those keys.

packages/client/hmi-client/src/services/save-asset.ts
blanchco: I've made saving and navigating into separate if-blocks

packages/client/hmi-client/src/services/intervention-policy.ts
mwdchang: Ideally the API should just ignore, or overwrite, any extra fields it doesn't use. 
I am guessing this is related to https://github.com/DARPA-ASKEM/terarium/issues/4488 so feel free to add pain-points on that ticket.

mwdchang (approved):
great, thanks!

PR #4616, opened 2024-08-29
force observable field to exist

mwdchang:
### Summary
Lack of observables in a model becomes a null-field. This creates problems downstream. To solve the issue we will convert them into empty array

### Testing
Upload the following AMR, when viewing it there shouldn't be any mira-task errors and the model-diagram should be visible

image

[dris.json](https://github.com/user-attachments/files/16802629/dris.json)


dvince2 (approved):
looks like a test may need to be updated

kbirk (approved):

PR #4626, opened 2024-08-29
Jamie transform dataset update page design

jamiewaese-uncharted:
Style updates to the Transform Dataset page.
image


blanchco:
@jamiewaese-uncharted Ill take this over from here - going to be changing the use of absolutes and there is some buggy behaviour

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
blanchco, line 1: took this opportunity to remove some stuff that wasn't being used here

blanchco (approved):

blanchco:
This should be better now @jamiewaese-uncharted , got rid of the absolutes and transformer should be working as expected

YohannParis (approved):

PR #4627, opened 2024-08-29
Added general purpose LLM request GoLLM endpoint

jryu01:
# Description

Added a gollm endpoint for general purpose llm request which supports json formatted response type. 
We will use this endpoint to ask to generate the chart annotation instead of using existing summary endpoint. 

Resolves #(issue)


packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/tasks/TaskServiceTest.java
```diff
@@ -161,6 +161,23 @@ public void testItCanSendGoLLMGenerateSummaryRequest() throws Exception {
...
+		final GenerateResponseHandler.Input input = new GenerateResponseHandler.Input();
+		input.setInstruction("Give me a simple random json object");
+		final JsonNode resFormat = new ObjectMapper().readTree("{\"type\": \"json_object\"}");
+		input.setResponseFormat(resFormat);
+		req.setInput(input);
+
+		final TaskResponse resp = taskService.runTaskSync(req);
+
+		log.info(new String(resp.getOutput()));
+	}
```
mwdchang, line 179: I am wondering if we really want to invoke a unit test where the result is inherently stochastic 🤔

mwdchang (approved):
Didn't test - code looks okay.

dvince2 (approved):

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/tasks/TaskServiceTest.java
jryu01: Just checking if things don't break :)

PR #4629, opened 2024-08-29
Update intervention-policy.md

ngraham76:
Added some new interventions so that we cover all our bases with possible intervention types.

# Description

Resolves #(issue)

### DIFF
--- a/testing/manual/intervention-policy.md
+++ b/testing/manual/intervention-policy.md
@@ -19,25 +19,39 @@ Report any issues into GitHub: [open an issue](https://github.com/DARPA-ASKEM/te
 2. Drag and drop _the Model_ onto the workflow
 3. Add the operator `Create intervention policy` and connect the model to it
 
-### 4. Create a static criteria
-1. Create a _static_ criteria named `Start masking`
-2. Set parameter `beta` to `0.5` starting at `50`
+### 4. Create a static parameter criteria
+1. Click `Open` on the `Create intervention policy` operator node to enter its drill down
+2. Edit the default intervention card, name it `Static Parameter` and leave it as _Static_.
+3. Set Parameter `β` to value `3` starting at timestep `20`.
+4. Click `+ Add`.
+5. In the new fields, set value to `4` starting at timestep 30`.
 
-### 5. Create a dynamic criteria
-1. Create a _dynamic_ criteria named `Vaccination`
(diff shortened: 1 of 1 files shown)


testing/manual/intervention-policy.md
```diff
@@ -19,25 +19,39 @@ Report any issues into GitHub: [open an issue](https://github.com/DARPA-ASKEM/te
...
 
-### 4. Create a static criteria
-1. Create a _static_ criteria named `Start masking`
-2. Set parameter `beta` to `0.5` starting at `50`
+### 4. Create a static parameter criteria
+1. Click `Open` on the `Create intervention policy` operator node to enter its drill down
+2. Edit the default intervention card, name it `Static Parameter` and leave it as _Static_.
+3. Set Parameter `β` to value `3` starting at timestep `20`.
+4. Click `+ Add`.
+5. In the new fields, set value to `4` starting at timestep 30`.
```
mwdchang, line 27: So kind of going back to what I was saying in the channel; whether we are testing if intervention-policy works?, or whether we are testing if intervention-policy's UI works? 

The values you have chosen here would likely make the simulation unstable (the parameters are usually between 0 and 1) and make the simulation result harder to verify. But again I am not sure if the goal is to test if intervention-policy works, or if the intervention-policy works.

ngraham76: I initially had tried to use parameters that were between 0 and 1, I was getting some errors. Chatted with Tom, he shared a policy he had created that worked without any errors, so I switched to his configuration / values. Maybe what would be best - if we want to test to see if it works with real world numbers - if you or someone could give me some examples of some numbers that should be input. I'll update this test write up with these correct values.

mwdchang: Well, if we are getting errors with values chosen from `[0, 1]` that would mean the UI works but the actual intervention-policy does not and we should probably raise the issue with the CIEMSS team. 

@Tom-Szendrey ?

Tom-Szendrey: I checking Neil's original test that failed before bringing it to the ciemss team. 
I would like to see what the error is as well as if it was caused by something such as asking for two conflicting things to happen

mwdchang: okay, let's take this up in a different thread

mwdchang (approved):

YohannParis (approved):

PR #4640, opened 2024-08-30
cache curries to not repeat calls for concepts

Tom-Szendrey:
# Description
cache curries to not repeat calls for concepts


packages/client/hmi-client/src/services/concept.ts
```diff
@@ -20,14 +21,29 @@ interface EntityMap {
...
+
+		// If a rename function is defined, loop over the first row
+		const response = await promise;
+		if (response?.status === 200 && response?.data) return response.data;
+		if (response?.status === 204) console.warn('No DKG entities found for curies:', curies);
+		return null;
+	} catch (err) {
+		logger.error(err);
+		return null;
+	}
```
mwdchang, line 46: I think this setup will still get a ton of errors if the call fails - which goes into the catch block with the splendour of 100 toasters?

mwdchang: Yeah we may need to tweak the axio interceptor to not spam toasters

mwdchang (approved):
The cache parts look fine, but the error handling will probably need to be rethought.

packages/client/hmi-client/src/services/concept.ts
YohannParis: Yes, spamming the user with 4XX and 5XX isn't great.

YohannParis (approved):
I would prefere to cache the `response.data` than the promise myself.

PR #4649, opened 2024-09-03
sort calibrate updates by progress

mwdchang:
### Summary
Force sort updates by progress, this was missing and/or returning in random order

image

### Testing
Run calibrate (with a sensible calibration dataset), after it has finished, refresh the page 
- verify that the loss-graph is still there, 
- verify that the loss-chart makes sense, it should be a curve going down


Tom-Szendrey (approved):
Thank you!

Issue #4559, opened 2024-08-23
[TASK]: Review save-as semantic for model-configuration and intervention policy


mwdchang:
We need to be able to create model-configurations and intervention policies, such that they are available in the project, but not automatically picked up the their parent models.

For example: Calibrate operation will create model-configurations, but these model configurations should not be visible in the model-configure operation's resource panel, until they are actively saved into the "project" by the modeller.

PR #4654, opened 2024-09-03
save for reuse optimize and calibrate

blanchco:
# Description

*Save for reuse on optimize and calibrate

When we create an intervention policy or model config we with run we set the `temporary` flag to `true`.  This makes sure it is not captured in the intervention policy or model configuration side panels.  When we actually click save for re-use this flag is then set to `false` so then it will be shown

DC Edit:
Closes https://github.com/DARPA-ASKEM/terarium/issues/4559


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -148,7 +148,19 @@
 
 		<!-- Output section -->
 		<template #preview>
-			<tera-drilldown-preview>
+			<tera-drilldown-section class="ml-4 mr-2 pt-3">
+				<template #header-controls-left>
+					<h5 v-if="configuredModelConfig?.name">{{ configuredModelConfig.name }}</h5>
+				</template>
```
YohannParis, line 154:
```suggestion
				<template #header-controls-left v-if="configuredModelConfig?.name">
					<h5>{{ configuredModelConfig.name }}</h5>
				</template>
```

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -746,8 +746,8 @@ export function createSuccessCriteriaChart(
 	options: BaseChartOptions
 ): any {
 	const targetState = `${targetVariable}_state`;
-	const data = riskResults[targetState]?.qoi || [];
-	const risk = riskResults[targetState]?.risk?.[0] || 0;
+	const data = riskResults?.[targetState]?.qoi || [];
+	const risk = riskResults?.[targetState]?.risk?.[0] || 0;
```
mwdchang, line 750: Maybe we can add a FIXME here, it is rather odd that at the time we are rendering a chart that we need to check if the result we just got are there or not. CC @Tom-Szendrey

blanchco: sure Ill add the comment, I just noticed it was breaking here at some points and this fixed it

Tom-Szendrey: I suspect this type of check is being used in the `createForecastChart` and maybe others as well 
Theyre all set up on computed values the same way if you create a ticket for the fixme you can throw @jryu01 or myself on it to check it out

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -330,6 +339,15 @@
...
 		/>
 	</Dialog>
+	<tera-save-asset-modal
+		:initial-name="optimizedInterventionPolicy?.name"
+		:is-visible="showSaveInterventionPolicy"
+		:asset="optimizedInterventionPolicy"
+		:asset-type="AssetType.InterventionPolicy"
+		@close-modal="showSaveInterventionPolicy = false"
+		@on-save="onSaveForReuse"
+		is-overwriting
```
asylves1, line 349: I made a change to the overwriting code you may want to check

blanchco: thanks, changed the prop and it should be good now!

YohannParis (approved):

PR #4655, opened 2024-09-03
Jamie model design polish

jamiewaese-uncharted:
Lots of small design tweaks... pretty pretty model page
Model page tweaks small


packages/client/hmi-client/src/assets/css/style.scss
```diff
@@ -93,6 +93,13 @@
 		box-shadow: none;
 	}
 
+
+	/* Correct secondary button styles */
+	.p-button.p-button-secondary.p-button-text, .p-buttonset.p-button-secondary > .p-button.p-button-text, .p-splitbutton.p-button-secondary > .p-button.p-button-text {
+		color: var(--text-color-secondary);
+	}
```
mwdchang, line 100: I am not sure how all these style changes will play into the primevue update - should we wait or bite the bullet and redo stuff again when primevue 4 rolls out @YohannParis ?

YohannParis: There is so much to be updated in terms of CSS for PrimeVue 4. that we might never update to it...
I have a [working branch](https://github.com/DARPA-ASKEM/terarium/tree/real-primevue-4-move) but so much to fix...
So we keep moving forwards nonetheless

YohannParis (approved):

mwdchang (approved):

PR #4657, opened 2024-09-03
Stratify default output

Tom-Szendrey:
# Description
- Create defaults for stratify model node.
- Rename stratifiedAmr -> outputAmr as the default is not stratified so it may cause confusion


packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-node-mira.vue
```diff
@@ -16,15 +16,41 @@ import TeraModelDiagram from '@/components/model/petrinet/model-diagrams/tera-mo
...
+	async () => {
+		const input = props.node.inputs[0];
+		if (!input || !props.node.outputs[0].value) return;
+		// Create a default if we dont have an output yet:
+		let modelId: string | null = null;
+		if (input.type === 'modelId') {
+			modelId = input.value?.[0];
+		} else if (input.type === 'modelConfigId') {
+			modelId = await getModelIdFromModelConfigurationId(input.value?.[0]);
+		}
```
mwdchang, line 40: Note: We will need clarification from pascale if we want to continue to support using model-configuration as a possible input to stratify. 

There is a gap in terms of the expected behaviour, that if you use the model-config the output model will have values as per the model-configuration's values....but we actually fall back to the model itself.

e.g. if my model has population constant = 1000, then in config-X I set the population to 5000, if I pop the configuration into stratify and run it, the output's population is 1000 instead of 5000.

Tom-Szendrey: I think this discussion can be had for another issue/PR let me know if you disagree

YohannParis (approved):

mwdchang (approved):

PR #4668, opened 2024-09-04
Clone project notification

asylves1:
# Description

First part of the change of the Fork Project feature
- change backend call to use `NotificationGroupInstance`
- Add notification to the Notification Panel

Screenshot 2024-09-04 at 11 28 56 AM

Resolves #(issue)


mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -502,23 +532,54 @@ public ResponseEntity<Project> updateProject(@PathVariable("id") final UUID id,
...
+				return cloneProject;
+			});
+			notificationInterface.sendFinalMessage("Cloning complete");
+			clonedProject = project.get();
+		} catch (ExecutionException e) {
+			log.error("Execution Exception exporting project", e);
+			throw new ResponseStatusException(HttpStatus.SERVICE_UNAVAILABLE, messages.get("postgres.service-unavailable"));
+		} catch (CancellationException e) {
+			log.error("Canceled exporting project", e);
+			throw new ResponseStatusException(HttpStatus.SERVICE_UNAVAILABLE, messages.get("postgres.service-unavailable"));
```
dvince2, line 572: Pedantic, but, is a `cancellationException` caused by the user cancelling something or that there was a problem trying to cancel? If its the former we probably don't want to throw an exception (or at least, a postgres error for the front end?

PR #4669, opened 2024-09-04
[FEAT]: Export Model Config and Model at once, zipped #4316

### LINKED ISSUE #4316: [FEAT]: Export Model Config and Model at once, zipped
Currently when a user exports a model config from the workflow node, what is exported is a configured AMR. If this is then imported back into terarium a default configuration is created for the mode with the correct values, however, its missing the "extra" information from that configuration like its name, description, sources etc

Rather than exporting a configured AMR (or in addition to?), we want to export both the model's JSON and the configuration's JSON inside of a zip archive. This zip archive will then be able to be uploaded back into terarium to import both the model *and* the configuration with all of its information intact

Another option is using the AMR's metadata, but, this will result in a lot of edge cases to consider.

**Additional context**

dvince2:
# Description

When a user downloads a model configuration now they will receive a zipped archive that includes the original model and the configuration.  This will preserve all the metadata and information that was included in the configuration and not present it as a 'default configuration'.  I've opted to not use the extension `.zip` because it would be slightly harder to verify whats coming in to us.

Resolves #4316


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ParameterSemantic.java
YohannParis, line 1: You can remove the override if you are just calling `super`

dvince2: This was just a change pulled in from `main`

Tom-Szendrey (approved):
Looks good didnt test

PR #4670, opened 2024-09-04
model config state not stored on outputs

blanchco:
# Description

* Do not store `transientModelConfig` in outputs, we just need a single state to store a draft, otherwise we can use the side panel resource .

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -590,9 +588,7 @@ const initialize = async () => {
 		applyConfigValues(suggestedConfigurationContext.value.tableData[0]);
 	} else {
 		originalConfig.value = await getModelConfigurationById(selectedConfigId.value);
-
-		// FIXME: Bug sept-03, state.transientModelConfig may not be saved properly before this date
-		if (state.transientModelConfig.id !== originalConfig.value.id) {
+		if (!overwriteWithState) {
 			knobs.value.transientModelConfig = cloneDeep(originalConfig.value);
 		} else {
 			knobs.value.transientModelConfig = cloneDeep(state.transientModelConfig);
```
mwdchang, line 594: If we are not keeping transientModelConfig around, that means we wouldn't want to copy from it either because it wouldn't be reliable. I am guessing it is either:

- applyConfigValues(suggestedConfigurationContext.value.tableData[0]) , or
- knobs.value.transientModelConfig = cloneDeep(originalConfig.value);

and we don't need this "overwriteWithState" flag.

YohannParis (approved):

mwdchang (approved):

PR #4674, opened 2024-09-04
Cache extraction service

kbirk:
# Description

Resolves #(issue)


bigglesandginger (commented):
As a first pass I think this looks great (I will review a second time with a eye for nit-picky details).  My one thought is - should this be cached in Redis (with a TTL etc) or simply written to the database to live longer/forever.   The DB live forever method would require (or I think would desire) a way for a user to force a new extraction "for reasons".  Though I get, how does one judge when to get a new extraction.

Additionally, should an extraction expire (again the TTL) unless it errored?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
YohannParis, line 1: Why we do not need to handle deadlock

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskService.java
kbirk: That is from the old caching approach they would scale across instances of hmi-server. I dropped that a while ago for a simpler model. This is just a leftover param that isn't used anymore.

bigglesandginger (approved):

PR #4694, opened 2024-09-05
Jamie dataset design tweaks

jamiewaese-uncharted:
Misc styling tweaks to the dataset page
- also corrected the mis-counted number of rows in the dataset datatable
image

### DIFF
--- a/packages/client/hmi-client/src/components/dataset/tera-dataset-datatable.vue
+++ b/packages/client/hmi-client/src/components/dataset/tera-dataset-datatable.vue
@@ -3,7 +3,7 @@
 		<!-- Toggle histograms & column summary charts -->
 		<div class="datatable-toolbar">
 			<span class="datatable-toolbar-item">
-				{{ rawContent.headers.length || 'No' }} columns | {{ rawContent.csv.length || 'No' }} rows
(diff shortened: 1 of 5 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/dataset/tera-dataset-datatable.vue
```diff
@@ -3,7 +3,7 @@
 		<!-- Toggle histograms & column summary charts -->
 		<div class="datatable-toolbar">
 			<span class="datatable-toolbar-item">
-				{{ rawContent.headers.length || 'No' }} columns | {{ rawContent.csv.length || 'No' }} rows
+				{{ rawContent.headers.length || 'No' }} columns | {{ rawContent.csv.length - 1 || 'No' }} rows
```
mwdchang, line 6: This doesn't seem right? Why do we have `length - 1`?

YohannParis: Because the first row are the headers?

mwdchang: Looks like the backend include headers in csv.

mwdchang (approved):

PR #4697, opened 2024-09-05
UX revisions to drill down header

ngraham76:
# Description

* Removed the kebab menu icon button next to the output dropdown as we no longer need the Save function here - it's now been added to operator's output header as a `Save for re-use` button
* Added a 1px bottom border (surface-border-alt) to the operator drilldown headers - defined in our Figma designs, was missing

image


packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<header>
+	<header class="border-b" :style="{ borderColor: 'var(--surface-border-alt)' }">
```
YohannParis: You can just change line 46 to have a different border-color

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
```diff
@@ -59,10 +59,6 @@
 							:output="selectedOutputId"
 							@update:selection="(e) => emit('update:selection', e)"
 						/>
-						<section v-if="!isEmpty(menuItems)" class="mx-2">
-							<Button icon="pi pi-ellipsis-v" rounded text @click.stop="toggleEllipsisMenu" />
-							<Menu ref="ellipsisMenu" :model="menuItems" popup />
```
YohannParis, line 64: You can remove the prop menuItems now that is not used anymore.

YohannParis (approved):

packages/client/hmi-client/src/components/drilldown/tera-drilldown-header.vue
ngraham76: Oh. Right. :) I'll do that now.

packages/client/hmi-client/src/components/drilldown/tera-drilldown.vue
ngraham76: image

mwdchang (changes requested):
We need to clean up the callers so they no longer generate and send in "menu-items" unnecessarily.

mwdchang:
TA3 operators likely needs a bit of design tweaks for removing these menu-item paradigms, as they are saving different types of asset or assets. See follow up in:

https://github.com/DARPA-ASKEM/terarium/issues/4713

mwdchang (approved):

PR #4699, opened 2024-09-05
guard inputs changes for model configuration and intervention policy

blanchco:
# Description

* Adding an extra guard here for the watcher on the model config and intervention policy nodes.  The extra guard will not change the `state` if the incoming input model id is the same as the model config's/ intervention policy's `state` `modelID`


YohannParis (approved):
LGTM

mwdchang (approved):

PR #4700, opened 2024-09-06
Added output settings panel for the calibrate drilldown

jryu01:
# Description

https://github.com/user-attachments/assets/97090b2a-9193-4a1c-a105-dbdbfdac8ff1

- Added output settings panel to the calibrate operator drilldown
- Moved chart selection controls to the settings panel

**Note:** I'm not sure why `trasnform-dataset.md` file is keep adding to my commit. But it looks like there are two of them with same name but one with capitalized name. e.g. `trasnform-dataset.md` and `Trasnform-dataset.md` and git is trying to fix the duplication and the file from this PR is the latest one it looks like. 

Resolves #(issue)


packages/client/hmi-client/src/components/widgets/tera-chart-settings-item.vue
```diff
@@ -0,0 +1,34 @@
...
+	}
+});
+defineEmits(['open', 'remove']);
+</script>
+
+<style scoped>
+.chart-settings-item {
+	border-left: 4px solid #667085;
+	padding: 0.75rem;
+	padding-left: 1rem;
```
YohannParis, line 27:
```suggestion
	padding: var(--gap-3);
	padding-left: var(--gap-4);
```

packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -0,0 +1,75 @@
...
+		backdrop-filter: blur(3px);
+		&.shadow {
+			box-shadow: 0 1px 4px 0 rgba(0, 0, 0, 0.1);
+		}
+		button {
+			height: 2.5rem;
+		}
+	}
+	.content {
+		padding: 1rem;
```
YohannParis, line 72:
```suggestion
		padding: var(--gap-4);
```

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -966,4 +1058,19 @@ img {
 		text-align: center;
 	}
 }
+.chart-settings-item-container {
+	gap: 0.5rem;
```
YohannParis, line 1062:
```suggestion
	gap: var(--gap-2);
```

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -966,4 +1058,19 @@ img {
...
+}
+
+.output-settings-panel {
+	padding: 1rem;
```
YohannParis, line 1066:
```suggestion
	padding: var(--gap-4);
```

YohannParis (approved):

testing/manual/transform-dataset.md
```diff
@@ -1,23 +1,33 @@
-## Transform Dataset Operator
-Please go through __every__ step of the test scenario.\
-Report any issues into GitHub: [open an issue](https://github.com/DARPA-ASKEM/terarium/issues/new?assignees=&labels=bug%2C+Q%26A&template=qa-issue.md&title=%5BBUG%5D%3A+).
+Please go through **every** step of the test scenario.  
+Report any issues into GitHub: [open an issue](https://github.com/DARPA-ASKEM/terarium/issues/new?assignees=&labels=bug%2C+Q%26A&template=qa-issue.md&title=%5BBUG%5D%3A+).
```
mwdchang, line 2: Why is `transform-dataset.md` here? It is weirdly out of place - if you are making testing notes changes let's put that into a different PR to make things easier to track.

packages/client/hmi-client/src/types/workflow.ts
```diff
@@ -178,3 +178,16 @@ export interface AssetBlock<T> {
 	isCollapsed?: false;
 	asset: T;
 }
+
+export enum ChartSettingType {
+	VARIABLE = 'variable',
+	PARAMETER = 'parameter',
+	ERROR = 'error'
+}
```
mwdchang, line 186: I sorta feel like we should introduce better naming now then deal with it later down the road.  Maybe:
- variable (eg simulate infected + recovered)
- variable-comparison (eg calibrate infected before and after)
- distribution-comparison (eg calibrate beta)
- error-distribution
- intervention

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -870,3 +872,47 @@ export function setLocalStorageTransform(id: string, canvasTransform: { x: numbe
...
+/**
+ * Deletes the chart settings with the given id and returns the new settings.
+ *
+ * @param {ChartSetting[]} settings - The current array of chart settings.
+ * @param {string} id - The id of the chart setting to delete.
+ * @returns {ChartSetting[]} The updated array of chart settings.
+ */
+export function removeChartSettingById(settings: ChartSetting[], id: string) {
+	return settings.filter((setting) => setting.id !== id);
+}
```
mwdchang, line 918: The two functions `updateChartSettingsBySelectedVariables` and `removeChartSettingById` are exactly workflow related - perhaps we can move them to chart.js of a new service?

testing/manual/transform-dataset.md
jryu01: It looks like this stated showing up in multiple people's PR lol

packages/client/hmi-client/src/services/workflow.ts
jryu01: Yeap I think that's better, I wasn't quite sure where to put these function and put it under workflow for now. I'll just make the change.

packages/client/hmi-client/src/types/workflow.ts
jryu01: Done

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -286,6 +261,82 @@
...
+							v-for="settings of chartSettings.filter(
+								(setting) => setting.type === ChartSettingType.DISTRIBUTION_COMPARISON
+							)"
+							:key="settings.id"
+							:settings="settings"
+							@open="activeChartSettings = settings"
+							@remove="removeChartSetting"
+						/>
+						<tera-chart-control
+							:chart-config="{ selectedRun: 'fixme', selectedVariable: selectedParameters }"
```
Tom-Szendrey, line 292: I dont see `tera-chart-control` using the `selectedRun` field.
I understand we are not sending this field, but is what else is this fixme for?

Tom-Szendrey (approved):
LGTM thanks for this!

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: Hm.. not exactly sure. It looks like `fixeme` was used for temporary to make it type safe. `tera-chart-control` emits back the `selectedRun` value to the parent and I'm not exactly sure what other places use it. If this isn't used in any other places in our app. I think we should modify the tera-chart-control component itself to remove that field.

mwdchang: This is from back the the days when we tried to support multiple runs per chart instead of singular runs, and the typing requires a selectedRun id. We labelled it as a fixme so readers know something is off.

@jryu01 as the notion of chart-config/chart-settings become more mature we should look to replace this.

mwdchang (approved):

PR #4710, opened 2024-09-06
4514 feat refine the fork a project workflow

asylves1:
# Description

Add a temporary card when cloning a project
image

Resolves #(issue)


dvince2 (approved):

packages/client/hmi-client/src/page/Home.vue
```diff
@@ -163,14 +168,28 @@ import Dropdown from 'primevue/dropdown';
...
+const cloningProjects = computed(() => {
+	const items: any = [];
+	notificationItems.value.forEach((item) => {
+		if (item.type === ClientEventType.CloneProject && item.status === ProgressState.Running) {
+			const project = myFilteredSortedProjects.value.find((p) => p.id === item.assetId);
+			items.push(project);
+		}
+	});
+	return items;
+});
```
mwdchang, line 191: How does `items.length` become 0 in order to trigger the refresh? Is there somewhere we would remove the "ClientEventType.CloneProject" item from notificationItems?

asylves1: Is there an edge case that I'm missing?

I have the computed triggering on the status change only because I'm using it to check if the process is still running inside of the computed.

mwdchang: I am not sure, but it seems like you are depending on the notificationItems to change from zero => not_zero => zero. There doesn't seem to be anything to automatically "remove" stuff from notificationItems??

packages/client/hmi-client/src/components/home/tera-project-card.vue
```diff
@@ -44,8 +44,12 @@
 			</div>
 			<section>
 				<div class="title" ref="titleRef">
+					<span v-if="isCopying">Copy of</span>
```
YohannParis, line 47:
```suggestion
					<template v-if="isCopying">Copy of</template>
```

packages/client/hmi-client/src/components/home/tera-project-card.vue
```diff
@@ -44,8 +44,12 @@
...
 					{{ project.name }}
 				</div>
+				<div v-if="isCopying">
```
YohannParis, line 50: Is the div necessary

YohannParis (approved):

packages/client/hmi-client/src/components/home/tera-project-card.vue
asylves1: removed

packages/client/hmi-client/src/page/Home.vue
mwdchang: I found what I am looking for - we update existing notifications as oppose to always appending new notifications.

PR #4724, opened 2024-09-09
4673 task create server side notification for workflow changes

### LINKED ISSUE #4673: [Task]: Create server side notification for workflow changes

dvince2:
# Description

* Server side updates when workflow changes and delete
* If these fail, no front end exception is thrown
* Public projects blast all users, non public blast only users who need to know.

Resolves #4673


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
```diff
@@ -223,13 +269,37 @@ public ResponseEntity<ResponseDeleted> deleteWorkflow(
...
+					.getReaders(rebacProject)
+					.stream()
+					.map(Contributor::getUserId)
+					.toList();
+				clientEventService.sendToUsers(event, userIds);
+			}
+		} catch (final Exception e) {
+			log.error("Unable to notify users of deleted  workflow", e);
+			// No response status exception here because the workflow was deleted successfully, and it's just the update that's failed.
+		}
```
YohannParis, line 301: Why not create a method at the top of this controller that accepts an event and a projectId as arguments? This way, any new endpoints can easily access this logic. Alternatively, I would suggest placing this method in the ClientEvent class, as it simply determines who should read the event based on the projectId.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
dvince2: I'm going to tuck that first idea away as the next step if we end up needing more collaboration updates :)

PR #4725, opened 2024-09-09
Split codeQL

YohannParis:
# Description

* _Java_ code QL would run on front-end PR slowing down PR reviews
* _Javascript_ code QL would run on back-end PR slowing down PR reviews
* _Python_ code QL never ran


mwdchang (approved):
Seems alright

PR #4728, opened 2024-09-09
4529 feat layout improvements to the notebook views

### LINKED ISSUE #4529: [FEAT]: Layout improvements to the notebook views
New Designs:
Screenshot 2024-08-21 at 4 25 39 PM
Screenshot 2024-08-21 at 4 26 01 PM

brittanykondo:
# Description

* Adjusted the layout and styles to match designs, new layout looks like this for all notebook tabs:
Screenshot 2024-09-09 at 3 02 50 PM

* Changed the llm input into an Autocomplete component so we can display the suggestions before and as the user types. Once the user's query gets longer (exceeds a max amount of chars), switch to a text area to allow for multi-line input:
Screenshot 2024-09-09 at 3 04 20 PM

Resolves #4529


YohannParis (approved):
Great work, looks good to me.

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison.vue
```diff
@@ -407,13 +405,17 @@ ul {
 }
 
 /* TODO: Improve this pattern later same in (tera-model-input) */
+
+.notebook-section {
+	background-color: var(--surface-disabled);
+	border-right: 1px solid var(--surface-border-dark);
+	padding: var(--gap);
+}
```
shawnyama, line 413: This rule is being repeated. See about moving the rule to `tera-drilldown-section` and I believe the class would still be read from here

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/model-comparison/tera-model-comparison.vue
brittanykondo: Done!

PR #4734, opened 2024-09-09
save intervention policy when one is not selected

blanchco:
# Description

* Ability to `save` an intervention policy when one is not selected (when you first open the intervention policy operator).


YohannParis (approved):
LGTM

PR #4735, opened 2024-09-09
fix(model concept editing): accept empty value and input value

shawnyama:
# Description

- When adding the concept input to `tera-dataset`'s `tera-column-info`  I found that going back to an empty concept or accepting a valid input (without choosing an autocomplete option) wasn't accepted as a valid edit
- Here is the fix done for the model page
- Also cleaned up a conditional button

https://github.com/user-attachments/assets/2b3d6410-3371-443c-bf39-aaa09554e46c


mwdchang (approved):
Looks fine

PR #4744, opened 2024-09-10
fix: other values now show source

blanchco:
# Description

* Show `source` instead of `target`/`id` for source column in Model configuration "other values" dialog


shawnyama (approved):

mwdchang (approved):
Didn't test, but code looks fine.

PR #4745, opened 2024-09-10
Optimize spinner progress

Tom-Szendrey:
# Description

- Allow for optimize's intermediate results to be captured in `CiemssStatusUpdate` by adding optional fields
- Adding `currentProgress` to optimize state
- Adding % message to optimize node and drilldown based off of the optimize iteration step.
- Adding a static message when forecast runs are running.

# DEPENDS ON
https://github.com/DARPA-ASKEM/pyciemss-service/pull/116

# Testing:
-Running the above pyciemss-service branch we can run optimize which will now have intermediate results sent to rabbitmq
-We can also confirm the type change has no impact on calibration's intermediate results

# Pictures
image
image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -254,7 +254,10 @@
 					/>
 				</template>
 				<tera-progress-spinner v-if="showSpinner" :font-size="2" is-centered style="height: 100%">
-					<div>{{ props.node.state.currentProgress }}% of maximum iterations complete</div>
+					<div v-if="node.state.inProgressOptimizeId !== ''">
+						{{ props.node.state.currentProgress }}% of maximum iterations complete
+					</div>
+					<div v-else>Optimize complete. Running simulations</div>
 				</tera-progress-spinner>
 				<tera-operator-output-summary v-if="node.state.summaryId && !showSpinner" :summary-id="node.state.summaryId" />
 				<!-- Optimize result.json display: -->
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
@@ -239,7 +239,6 @@
 		<template #preview>
 			<tera-drilldown-section
 				class="ml-3 mr-3"
-				:is-loading="showSpinner"
 				:class="{ 'failed-run': optimizationResult.success === 'False' ?? 'successful-run' }"
 			>
 				<template #header-controls-left v-if="optimizedInterventionPolicy?.name">
@@ -254,10 +253,16 @@
 						@click="showSaveInterventionPolicy = true"
 					/>
 				</template>
-
+				<section v-if="showSpinner" class="spinner-message">
+					<tera-progress-spinner :font-size="2" is-centered style="height: 12rem" />
+					<p>{{ node.state.currentProgress }}% of maximum iterations complete</p>
+				</section>
 				<tera-operator-output-summary v-if="node.state.summaryId && !showSpinner" :summary-id="node.state.summaryId" />
 				<!-- Optimize result.json display: -->
-				<div v-if="optimizationResult && displayOptimizationResultMessage" class="result-message-grid mt-2 mb-2">
+				<div
+					v-if="optimizationResult && displayOptimizationResultMessage && !showSpinner"
+					class="result-message-grid mt-2 mb-2"
+				>
 					<span class="flex flex-row">
 						<p class="mt-2">For debugging</p>
 						<Button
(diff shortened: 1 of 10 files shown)


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -74,7 +78,19 @@ const pollResult = async (runId: string) => {
...
+			if (runId === props.node.state.inProgressOptimizeId && data.updates.length > 0) {
+				const checkpoint = _.first(data.updates);
+				if (checkpoint) {
+					const state = _.cloneDeep(props.node.state);
+					state.currentProgress = +((100 * checkpoint.data.progress) / checkpoint.data.totalPossibleIterations).toFixed(
+						2
+					);
+					emit('update-state', state);
+				}
+			}
```
mwdchang, line 92: So I don't have a good notion of the frequency of updates in optimize, now that we have some sense of how fast the numbers changes - you may want to take the opportunity to tune the poller parameters:

```
		.setInterval(5000)
		.setThreshold(100)
```

E.g. if it is the case that we see changes every 10 seconds, maybe we don't need to poll as often.

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -4,7 +4,11 @@
 			<template v-if="!node.inputs[0].value"> Attach a model configuration </template>
 		</tera-operator-placeholder>
 		<template v-if="node.inputs[0].value">
-			<tera-progress-spinner v-if="showSpinner" :font-size="2" is-centered style="height: 100%" />
+			<div v-if="showSpinner">
+				<p>{{ node.state.currentProgress }}% of maximum iterations complete</p>
+				<tera-progress-spinner :font-size="2" is-centered style="height: 100%" />
+			</div>
```
mwdchang, line 10: Your usage of `tera-progress-spinner` is a bit odd since it is slotted for a message but you are putting the message outside I think?

Can we do something like this so the usage is more idiomatic across the app?
```
   <tera-progress-spinner v-if="inProgressCalibrationId" :font-size="2" is-centered style="height: 100%">
     <div>{{ props.node.state.currentProgress }}%</div>
   </tera-progress-spinner>
```

packages/client/hmi-client/src/types/Types.ts
```diff
@@ -554,9 +554,11 @@ export interface CalibrationRequestCiemss {
 }
 
 export interface CiemssStatusUpdate {
-    loss: number;
-    progress: number;
     jobId: string;
+    progress: number;
+    loss?: number;
+    currentResults?: number[];
+    totalPossibleIterations?: number;
```
mwdchang, line 561: Any particular reason we have the same update-object servicing different operations?  
Is it too much problem to have them separate? It is a bit awkward because as a new reader you don't really know which fields should go with which other fields.

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -254,10 +253,16 @@
 						@click="showSaveInterventionPolicy = true"
 					/>
 				</template>
-
+				<section v-if="showSpinner" class="spinner-message">
+					<tera-progress-spinner :font-size="2" is-centered style="height: 12rem" />
+					<p>{{ node.state.currentProgress }}% of maximum iterations complete</p>
+				</section>
```
shawnyama, line 259: utilize slot here as well

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
Tom-Szendrey: Good catch thank you, will do

packages/client/hmi-client/src/types/Types.ts
Tom-Szendrey: I will look to change this into two classes

Tom-Szendrey: Ive split this into two classes, this involved updating how we parse these in `SimulationEventService`.
After speaking with Kevin this seems like a reasonable approach for this

kbirk (approved):

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/SimulationEventService.java
```diff
@@ -151,11 +155,30 @@ private void onScimlAllInstanceReceive(final Message message) {
 
 	@RabbitListener(queues = "${terarium.simulation-status}", concurrency = "1")
 	private void onPyciemssOneInstanceReceives(final Message message, final Channel channel) throws IOException {
-		final CiemssStatusUpdate update = ClientEventService.decodeMessage(message, CiemssStatusUpdate.class);
-		if (update == null) {
+		// Parse message to get message type:
+		final String updateType = (ClientEventService.decodeMessage(message, JsonNode.class)).get("type").asText();
+		if (updateType.equals(CiemssStatusType.OPTIMIZE.toString())) {
```
mwdchang, line 160: FYI this `CiemssStatusType` enum isn't very useful, because you are converting everything to a string at the end anyway. It is functionally equivalent to a class with static string fields but needs more steps.

mwdchang (approved):

PR #4748, opened 2024-09-10
Added chart annotation control UI

jryu01:
# Description

https://github.com/user-attachments/assets/606d3d82-ff66-476b-8b73-9e135f5c49aa

Created and added UIs for adding and removing chart annotation with fake annotation data. 
Hooking up the chart annotations and services are not done yet and further needed to be implemented. 

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
@@ -43,7 +43,7 @@ import TeraInputText from '@/components/widgets/tera-input-text.vue';
 const props = defineProps<{
 	activeSettings: ChartSetting | null;
 	annotations?: ChartAnnotation[];
-	generateAnnotation?: (setting: ChartSetting) => Promise<ChartAnnotation>;
+	generateAnnotation?: (setting: ChartSetting, query: string) => Promise<ChartAnnotation>;
 }>();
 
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -6,26 +6,82 @@
...
 import Button from 'primevue/button';
 import { ChartSetting } from '@/types/common';
 import { ChartAnnotation } from '@/types/Types';
+import TeraInputText from '@/components/widgets/tera-input-text.vue';
 
-defineProps<{
+const props = defineProps<{
 	activeSettings: ChartSetting | null;
 	annotations?: ChartAnnotation[];
+	generateAnnotation?: (setting: ChartSetting, query: string) => Promise<ChartAnnotation>;
```
mwdchang, line 52: Can you comment on why we are passing in a functor here as a prop? As oppose to emit the query as an event and let the parent-component handle the generation? Is there a specific advantage you see doing things this way?

mwdchang: Hmm, maybe we can talk about this some more - but should be okay for now I think.

I am just a bit wary of passing down a function closure that can hold onto other references, may need to keep an eye on upstream changes, memory leaks ... that type of things.

mwdchang (approved):

Tom-Szendrey (approved):
sweet

PR #4761, opened 2024-09-11
clean(backend): remove searches and getAlls that were used by explorer or not even used at all

shawnyama:
# Description

* Deleting anything search-like that was used in the data explorer
* Deleting unused "getAll" endpoints


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -378,65 +339,6 @@ ResponseEntity<Model> getModelFromConfigId(
...
-				)
-			),
-			@ApiResponse(
-				responseCode = "500",
-				description = "There was an issue retrieving models from the data store",
-				content = @Content
-			)
-		}
-	)
-	public ResponseEntity<List<Model>> searchModels(
```
dvince2, line 400: @kbirk do we want to strike this or will it still be useful for the new project-page searching?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/climatedata/ClimateDataProxy.java
```diff
@@ -12,9 +12,6 @@ public interface ClimateDataProxy {
 	@GetMapping("/status/{uuid}")
 	ResponseEntity<JsonNode> status(@PathVariable("uuid") String uuid);
 
-	@GetMapping("/search/esgf")
```
dvince2, line 15: Removing the searching of climate data effectively drops support for climate data as users won't have a way to get this data into terrarium. This is more a problem for @dgauldie to be aware of!

dvince2 (approved):
Looks good! Just wait on @kbirk to confirm the model search code before merging?

mwdchang (approved):

YohannParis (approved):
Have you made sure that no services uses those endpoints by checking @dgauldie dashboard?

shawnyama:
> Have you made sure that no services uses those endpoints by checking @dgauldie dashboard?

I shall do this next week

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
kbirk: Nah that's good to nuke. The guts of it is in the service.

kbirk (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/search/SearchByAssetTypeController.java
kbirk, line 1: Actually, can you keep this file? We will piggy back on this code for the project search

packages/client/hmi-client/src/services/search.ts
```diff
@@ -0,0 +1,16 @@
+import API from '@/api/api';
+import { AssetType, TerariumAssetThatSupportsAdditionalProperties } from '@/types/Types';
+import { logger } from '@/utils/logger';
+
+export async function searchByAssetType(
+	text: string,
+	assetType: AssetType
+): Promise<TerariumAssetThatSupportsAdditionalProperties[]> {
```
YohannParis, line 8: This should return the same type than used for `AssetType`.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/proxies/climatedata/ClimateDataProxy.java
shawnyama: Yeah, you know what I'll add this search back in and we can worry about this if we want to clean up climate data. I can see that happening.

packages/client/hmi-client/src/services/search.ts
shawnyama: I tried some ways of doing that using generics but it doesn't really work. AssetType is an enum.

PR #4763, opened 2024-09-12
fix compound type propagation for selectOutput

mwdchang:
### Summary
`WorkflowWrapper#selectOutput(...)` is used to propagate output changes, this was copying the values directly which does not work for compounded outputs such as calibrate. 

### Testing
The following sequence of actions to build a workflow should work and not crash
- add calibrate
- add simulate
- connect calibrate => simulate
- run calibarate
- run simulate

### DIFF
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -473,8 +473,29 @@ export class WorkflowWrapper {
 				const targetPort = targetNode.inputs.find((port) => port.id === edge.targetPortId);
 				if (!targetPort) return;
 				edge.sourcePortId = selected.id; // Sync edge source port to selected output
+
+				const selectedTypes = selected.type.split('|').map((d) => d.trim());
+				const allowedInputTypes = targetPort.type.split('|').map((d) => d.trim());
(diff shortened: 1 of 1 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -473,8 +473,29 @@ export class WorkflowWrapper {
...
+
+				// Not supported if there are more than one match
+				if (intersectionTypes.length > 1) {
+					console.error(`Ambiguous matching types [${selectedTypes}] to [${allowedInputTypes}]`);
+					return;
+				}
+				if (intersectionTypes.length === 0) {
+					return;
+				}
+				if (selectedTypes.length > 1) {
```
Tom-Szendrey, line 489: nitpick: may want to add comment with #intersectionTypes is implicitly length 1

mwdchang: added a bit more comments

Tom-Szendrey: Thanks!

blanchco (approved):

PR #4771, opened 2024-09-12
add intervention policy to simulate

blanchco:
# Description

* Added intervention policy to simulate node.
* I reworked some logic around chart configs to determine whether or not the chart should show interventions or not
* I cleaned up some logic around displaying intervention annotations to make them easier for devs to display them
NOTE: this is a breaking change... all current simulate, simulate-ensemble, and calibrate-ensemble nodes will need to be removed and added again.

Cole Edit:
This is no longer a breaking change, and am not touching the chart configs any longer

https://github.com/user-attachments/assets/9b95dda8-c66d-44a4-834f-632c5e7ab5bf


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
```diff
@@ -6,7 +7,7 @@ const DOCUMENTATION_URL = 'https://github.com/ciemss/pyciemss/blob/main/pyciemss
 
 export interface SimulateCiemssOperationState extends BaseState {
 	// state shared across all runs
-	chartConfigs: string[][];
+	chartConfigs: Partial<ChartConfig>[];
```
mwdchang, line 10: Please note the notion of chart-config is getting overhauled as per annotations/output-setting changes, I would avoid changing too much stuff here right now because they will likely collide with what @jryu01 is doing.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
```diff
@@ -11,7 +12,7 @@ export interface EnsembleCalibrateExtraCiemss {
 }
 
 export interface CalibrateEnsembleCiemssOperationState extends BaseState {
-	chartConfigs: string[][];
+	chartConfigs: Partial<ChartConfig>[];
```
mwdchang, line 15: You may want to leave ensemble-related operators alone for the moment - I think Nelson is catching up with TA3 folks on what we want to do in this space, so not particularly worthwhile to get them into the latest when we don't precisely know what we want.

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
blanchco: Im not changing the config anymore as of our discussion!

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
blanchco: Not longer touching these at the moment

mwdchang (approved):

YohannParis (approved):

PR #4781, opened 2024-09-12
Model Config only append to output when we dont yet have one.

Tom-Szendrey:
# Context:
here is a watcher on model config that will append a new output on every input change.
Issue is:
user plugs in a new model (1)
user plugs in a new document/dataset to enrich with (2)
user plugs in a new document/dataset to enrich with (3)

We already have 3 essentially duplicate outputs appended to our node.
This is causing unnecessarily large nodes

Here we check for the value as our default output will be assigned and its value is null.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
@@ -82,18 +82,21 @@ watch(
 			}
 			configs = await getModelConfigurationsForModel(modelId);
 		}
-		// Auto append output if and only if there isnt already an output
-		if (!props.node.outputs.at(0)?.value && configs[0]?.id) {
+
+		if (configs[0]?.id) {
 			const config = configs[0];
 			state.transientModelConfig = config;
 			emit('update-state', state);
-			emit('append-output', {
-				type: ModelConfigOperation.outputs[0].type,
-				label: config.name,
-				value: config.id,
-				isSelected: false,
-				state: omit(state, ['transientModelConfig'])
-			});
+			// Auto append output if and only if there isnt already an output
+			if (!props.node.outputs.at(0)?.value) {
+				emit('append-output', {
+					type: ModelConfigOperation.outputs[0].type,
+					label: config.name,
+					value: config.id,
+					isSelected: false,
+					state: omit(state, ['transientModelConfig'])
+				});
+			}
 		}
 	},
 	{ deep: true }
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -82,8 +82,8 @@ watch(
 			}
 			configs = await getModelConfigurationsForModel(modelId);
 		}
-		// Auto append output
-		if (configs[0]?.id) {
+		// Auto append output if and only if there isnt already an output
+		if (!props.node.outputs.at(0)?.value && configs[0]?.id) {
```
blanchco, line 86: I _think_ this may cause issues with this flow:
1. We have 2 models `Model_A`, `Model_B`
2. Attach `Model_A` to a model config node
3. Unlink `Model_A` from model config node
4. Attach `Model_B` to the same model config node

Will opening the model config here show the default configuration and output from `Model_B`? It does that currently

mwdchang: Good question, we should clarify what is the expected behaviour here!

Tom-Szendrey: I have now updated this to be more in line with how other operators work:
```
We have 2 models Model_A, Model_B
Attach Model_A to a model config node
Unlink Model_A from model config node
Attach Model_B to the same model config node
Currently we still have model_a's configuration.
Now make any changes and click save for a new output 
Now we will have model_b's configuration.
```

This is similar to if we 
```
plug in model_config_a to simulate run
back out
unplug model_config_a
plug in new model config_b
now simulate run's output does not line up with the model config thats technically connected

```

Tom-Szendrey: I think I raised this awhile ago and this was Dan's solution but I will reach out to design channel

Tom-Szendrey: Based off of the chat in the design channel it seems like having this working like the other operators is okay as we may want to overhaul how they work as a whole.
Please let me know if you disagree

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
blanchco: Sure, as long as everyone's on the same page seems good to me

blanchco (approved):

PR #4799, opened 2024-09-13
style(clean): apply a class to `tera-slider-panel` if you want it grey

shawnyama:
# Description

* No need to write the same css rules for every input config slider
* Just apply the `input-config` class to the slider to make it grey

image


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -1079,23 +1084,6 @@ watch(
 </script>
 
 <style scoped>
-/* Left sidebar styles */
-:deep(.slider-content) {
-	background-color: var(--surface-100);
-	border-right: 1px solid var(--surface-border-light);
-	height: auto;
-	padding-bottom: 7rem;
```
brittanykondo, line 1087: This bottom padding isn't being applied anymore

brittanykondo: Actually, you could probably add this bottom padding to the slider component, it's good to have on all of them so the content doesn't sit too close to the bottom of the panel

brittanykondo (approved):

packages/client/hmi-client/src/components/widgets/tera-slider-panel.vue
```diff
@@ -139,6 +156,10 @@ h5 {
 	writing-mode: vertical-lr;
 }
 
+.p-button.p-button-icon-only.p-button-rounded {
+	height: 2.5rem;
```
asylves1, line 160: this can be `var(--gap-10)`

blanchco (approved):

asylves1 (approved):

Issue #4780, opened 2024-09-12
[BUG]: Post stratification error


mwdchang:
Stratification seemed to work, but we don't seem to handle the result properly?

The console is showing several of these errors

```
TypeError: Cannot read properties of null (reading 'emitsOptions')
    at shouldUpdateComponent (main-AwGyjmr-.js:20:50562)
    at hu (main-AwGyjmr-.js:20:36548)
    at du (main-AwGyjmr-.js:20:36184)
    at Us (main-AwGyjmr-.js:20:32110)
    at uu (main-AwGyjmr-.js:20:35264)
    at au (main-AwGyjmr-.js:20:34436)
    at tu (main-AwGyjmr-.js:20:32852)
    at Us (main-AwGyjmr-.js:20:32074)
    at ReactiveEffect.jp [as fn] (main-AwGyjmr-.js:20:37223)
    at ReactiveEffect.run (main-AwGyjmr-.js:9:1934)
```

While the "save-for-reuse" button became available and then greyed out without any user intervention, also it seemed like the operation run did not result in a new "output" for the operator.

mwdchang:
This is a workflow sync issue. An attempt to resolve is in this PR https://github.com/DARPA-ASKEM/terarium/pull/4801

PR #4801, opened 2024-09-13
work around workflow de-sync problem in stratify

mwdchang:
### Summary
There are sync'ing issues when an output-creation is split into multiple stages, where each stage end up invoking workflow-updates. For example here when stratify finishes:
- update the state with the code written
- save a new model and append output

because the two not grouped together it means at the time we are ready to update the new output, it could have been overwritten by the response from the update-state. This PR adds logic to try to guard against these cases.

### In this PR:
- Tweak the model creation ordering, so we don't update the preview unless we have successfully create a model
- Group code-state update and appending output together 
- Add a few FIXMES for workflow-logic, and deprecation in tera-drilldown
- Better default output port labels (model-names)

### TODO: 
Ideally I think update-state and update-output would be their own thing so they do not cross contaminate, this will rerouting, additional API endpoints to do partial updates, and moving parts of the workflow management server-side.  

### Testing
Run stratify on a non-trivial model (not SIR) with different configurations, shouldn't see errors, and the number of outputs produced should be the number of times we have run stratification.

See: https://github.com/DARPA-ASKEM/terarium/issues/4780

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
@@ -263,16 +263,6 @@ const stratifyModel = () => {
 	});
 };
 
-// const handleStratifyResponse = (data: any) => {
-// 	const executedCode = data.content.executed_code;
-// 	if (executedCode) {
-// 		codeText.value = executedCode;
-//
-// 		// If stratify is run from the wizard, save the code but set `hasCodeBeenRun` to false
-// 		saveCodeToState(executedCode, false);
-// 	}
-// };
-
 const handleModelPreview = async (data: any) => {
 	const amrResponse = data.content['application/json'] as Model;
 	isStratifyInProgress.value = false;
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
```diff
@@ -254,36 +253,50 @@ const stratifyModel = () => {
...
-const handleStratifyResponse = (data: any) => {
-	const executedCode = data.content.executed_code;
-	if (executedCode) {
-		codeText.value = executedCode;
-
-		// If stratify is run from the wizard, save the code but set `hasCodeBeenRun` to false
-		saveCodeToState(executedCode, false);
-	}
-};
+// const handleStratifyResponse = (data: any) => {
```
Tom-Szendrey, line 266: Unsure why this is left as commented our rather than just removed
If we do not want to use `stratify_response` to `saveCodeToState` and thus the one liner is sufficient/no helper is needed can we just remove?

Are we intending to bring this back in the future? -> If we are can we leave a comment about when/why?

Tom-Szendrey (approved):
Thank you for this! Interesting bug you found and fixed!

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-mira.vue
mwdchang: removed

PR #4804, opened 2024-09-13
Apply chart annotations to the forecast charts on calibrate operator 

### LINKED ISSUE #4722: Hook up the `chart-annotation` and annotation generation endpoints to the UI to support adding and persisting chart annotations in calibrate drilldown settings panel

jryu01:
# Description

Apply chart annotations to the forecast charts and persist the annotation changes.
Annotation changes are synced across multiple sessions/users. 

https://github.com/user-attachments/assets/95d8985a-020c-4600-b9e1-4319e7d79317

Resolves #4722


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ChartAnnotationController.java
```diff
@@ -117,15 +130,39 @@ public ResponseEntity<ChartAnnotation> createChartAnnotation(
...
 			throw new ResponseStatusException(org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR, error);
 		}
+
+		final ClientEvent<ChartAnnotation> event = ClientEvent.<ChartAnnotation>builder()
+			.type(ClientEventType.CHART_ANNOTATION_CREATE)
+			.data(chartAnnotation)
+			.build();
+
+		try {
+			final RebacProject rebacProject = new RebacProject(projectId, reBACService);
```
dvince2, line 149: This looks like the same code as in the workflow sync messages?
If we're already reusing it we should move ahead with @YohannParis 's suggestion and move this into some sort of shared class.  Maybe not for this task, but, would you mind opening a new ticket to extract this and the code in `WorkflowController`?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ChartAnnotationController.java
jryu01: Yeah I think we should. I created a ticket for that: https://github.com/DARPA-ASKEM/terarium/issues/4806

packages/client/hmi-client/src/services/chart-settings.ts
```diff
@@ -44,3 +54,213 @@ export function updateChartSettingsBySelectedVariables(
...
+      "layer": [
+        {
+          "mark": {
+            "type": "rule",
+            "strokeDash": [4, 4]
+          },
+					"encoding": {
+						"x": { "datum": 200, "axis": { "title": ""} }
+					}
+        },
```
mwdchang, line 143: FYI: there seem to be some quirky tab formatting going on here.

mwdchang (approved):
Looks alright!

packages/client/hmi-client/src/services/chart-settings.ts
jryu01: Fixed it

PR #4818, opened 2024-09-17
feat: funman input update

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/funman/funman-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/funman/funman-operation.ts
@@ -6,12 +6,7 @@ const DOCUMENTATION_URL = 'https://github.com/siftech/funman';
 export interface ConstraintGroup {
 	borderColour: string;
 	name: string;
-
-	// One of
-	// - monotonicityConstraint
-	// - stateConstraint
 	constraintType: string;
-
 	variables: string[]; // If len = 1, need to rename to "variable" for request formatting
 	weights?: number[]; // 1 to 1 mapping with variables
 	timepoints?: FunmanInterval;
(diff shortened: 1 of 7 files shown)


packages/client/hmi-client/src/components/workflow/ops/funman/funman-operation.ts
```diff
@@ -3,21 +3,35 @@ import type { FunmanInterval, TimeSpan } from '@/types/Types';
...
-	// - stateConstraint
-	constraintType: string;
+export enum DerivativeType {
+	LessThan = 'less than',
+	GreaterThan = 'greater than',
+	Increasing = 'increasing',
+	Decreasing = 'decreasing',
+	LinearlyConstrained = 'linearly constrained',
+	Following = 'following'
+}
```
mwdchang, line 19: Your terminologies are confusing here, as only 2 out of 6 items here have something to do with derivatives. 

I would recommend these changes to better reflect the vocabulary/semantic at hand cc @liunelson 
- ConstraintType => Constraint
- DerivativeType => ConstraintType

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-constraint-group-form.vue
```diff
@@ -1,295 +1,191 @@
...
+						}
+						emit('update-self', { key: 'constraintType', value: $event });
+					}
+				"
+			/>
+			<tera-input-number
+				v-if="config.constraintType === ConstraintType.LessThan"
+				auto-width
+				:model-value="config.interval?.ub ?? 0"
+				@update:model-value="emit('update-self', { key: 'interval', value: { lb: config.interval?.lb, ub: $event } })"
```
mwdchang, line 57: If `config.intevel.ub` is not defined, you would set the model-value as a literal `0`, then for update you would still emit `config.intervel.ub` which is still undefined no?

The setup seems weird here.

shawnyama: It changes to `0` as soon as you click out of the input. This is not right. Once I activate this input `config.interval.ub` should be `0`

Tom-Szendrey (approved):

PR #4820, opened 2024-09-17
Disable "increases to above" dropdown in the Create Intervention Poli…

### LINKED ISSUE #4769: [TASK]: Disable "increases to above" dropdown in the `Create Intervention Policy` operator
**Describe the task**
For dynamic interventions in the `Create Intervention Policy` operator, we need to remove the `increases to above` dropdown and replace it with plain text that say `crosses the threshold` since PyCIEMSS does not distinguish between threshold crossing under different directions and will not support this for a while.

**Screenshots**
Screenshot 2024-09-12 at 10 56 48 AM

mloppie:
# Description

* replaces the threshold direction drop down with plain text since PyCIEMSS does not distinguish between threshold crossing under different directions and will not support this for a while

Resolves #4769


packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
```diff
@@ -105,13 +105,14 @@
 						option-value="value"
 						placeholder="Select a trigger"
 					/>
-					<Dropdown
+					crosses the threshold
+					<!-- <Dropdown
```
Tom-Szendrey, line 109: I dont think we plan to introduce this any time soon. safe to remove rather than comment out

Tom-Szendrey:
We should also update this:
https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-intervention-summary-card.vue#L14

This is utilized in our simulate and calibrate drilldowns for eg here:
image

Tom-Szendrey:
@liunelson As far as I am aware pyciemss is not intending to support this unless theyre explicitly told they have to.
Should we remove this isGreaterThan from our dynamic typing? Or mark it as depreciated? 
https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/DynamicIntervention.java#L13

Cc: @mwdchang

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
YohannParis: 💯 we have a policy of not keeping code commented out.
We have git if people want to re-use old code.

YohannParis (approved):

liunelson:
> @liunelson As far as I am aware pyciemss is not intending to support this unless theyre explicitly told they have to.
> Should we remove this isGreaterThan from our dynamic type? Or mark it as depreciated? 
> https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/DynamicIntervention.java#L13
> 
> Cc: @mwdchang 
> 

You're correct. I don't know what impact either choice would make, so I'd leave that up to you.

Tom-Szendrey (approved):
Thanks!

PR #4824, opened 2024-09-17
spinner on enrich model

blanchco:
# Description

* We need to listen to the events source for model enrichment to determine whether it is loading or not (enrichment for dataset doesnt need this)


packages/client/hmi-client/src/components/widgets/tera-asset-enrichment.vue
```diff
@@ -65,6 +66,18 @@ const props = defineProps<{
...
+useClientEvent(ClientEventType.TaskGollmModelCard, (event: ClientEvent<TaskResponse>) => {
+	const { modelId } = event.data?.additionalProperties || {};
+	const { status } = event.data || {};
+
+	if (props.assetType !== AssetType.Model || modelId !== props.assetId) {
+		return;
+	}
+
+	isLoading.value = status !== TaskStatus.Success;
+});
```
mwdchang, line 79: Something seems a little weird here, because in the `confirm` function we also set "isLoading" to true/false, and in addition we emit a "finish-job" event. Are they like different things or are we mixing up our messaging events?

```
const confirm = async () => {
	isLoading.value = true;
	closeDialog();

	// Await for enrichment/extraction so once we call finished-job the refetched dataset will have the new data
	if (dialogType.value === DialogType.ENRICH) {
		await sendForEnrichment();
	} else if (dialogType.value === DialogType.EXTRACT) {
		await sendForExtractions();
	}

	isLoading.value = false;
	emit('finished-job');
	getRelatedDocuments();
};
```

YohannParis (approved):
All good, just resolve Daniel's comment.

shawnyama (approved):

packages/client/hmi-client/src/components/widgets/tera-asset-enrichment.vue
blanchco: Looks like somethings are sort of mixed up but it looks like the flow here for enrichment:

For dataset:
1. call `/api/knowledge/profile-dataset/`
2. wait for response from `/api/knowledge/profile-dataset/`
3.  send `finished-job` event

For models
1. call `api/gollm/model-card`
2. get response that gollm task has been kicked off
3. send `finished-job` event (job isnt actually finished we just got the notification that gollm has started)
**NEW**
4. listen to client event for the gollm task to finish

YohannParis: For models, if a cache version exist it is returned directly, and the client-event is redundant.

mwdchang: But `finish-job` event invokes refetching of dataset and models - if finish-job isn't actually job finished we have a possibility of fetching a model that is yet to be enriched?

blanchco: I'm merging this and made an [issue](https://github.com/DARPA-ASKEM/terarium/issues/4836) regarding sending the cached response through the client event as per our discussion this morning

Issue #4562, opened 2024-08-23
[TASK]: Workflow branching is not exactly working for diamond shaped formations


mwdchang:
If I have a workflow of a diamond formation like below

``` 
     B
  /     \
A        C
  \.    /
     X
```

And we branch on B

PR #4825, opened 2024-09-17
fix branching workflow in diamond arrangement

mwdchang:
### Summary
There is a bug with branching when the subgraph forms a diamond, the upstream edges are not calculated correctly.

Before: note one of the simulate is disconnected from its upstream model-config
image

After:
image

### Testing
In a workflow
- add model, model-config, intervention-policy, simulate
- connect model=>model-config, model=>intervention-policy, {intervention-policy, model-config} => simulate
- branch on either the intervention-policy or model-config node

The resulting new subgraph should still form a diamond formation with the existing graph

Fixes: https://github.com/DARPA-ASKEM/terarium/issues/4562

### DIFF
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -369,9 +369,12 @@ export class WorkflowWrapper {
 			});
 		}
 
-		// 3. Collect the upstream edges of the anchor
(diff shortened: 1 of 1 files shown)


YohannParis (approved):

blanchco (approved):
nice!

PR #4830, opened 2024-09-17
4790 bug app fails to extract configurations from timeseries datasets

dgauldie:
# Description

Resolves #(issue)


kbirk (approved):

packages/mira/Dockerfile
```diff
@@ -11,10 +11,24 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	software-properties-common
 RUN add-apt-repository ppa:deadsnakes/ppa
 RUN apt-get update && apt-get install -y --no-install-recommends \
-	python3.10 \
-	wget \
-	python3-pip && \
-	rm -rf /var/lib/apt/lists/*
+ python3.10 \
+ python3.10-venv \
```
YohannParis, line 15: Why funman is running on Python 3.8 but mira 3.10?

YohannParis (approved):

packages/mira/Dockerfile
dgauldie: no idea. but some of the sub images they use 3.8

mwdchang: One of Funman's staggered docker mages happens to use ubuntu-20 which defaults to python 3.8. we can probably upgrade if it is desired so long as the C-bindings the numeric libraries still work.

PR #4831, opened 2024-09-17
Model to latex with mira task 

Tom-Szendrey:
# Description
Replacing sciml's `model-to-latex` with a mira task runner job.

- This means we will have `\operatorname` rather than `\mathrm` I dont think this is an issue but please let me know
- These equations are now able to have observables. I am confirming if we want to add these or not
- The test added here is following the same idea of `amr-to-mmt`'s test. AKA can the task be created.
If we want more in depth testing please let me know either here or in the `askem-server` channel 

# Follow up:
Next PR will include the remaining removal of sciml. Just trying to limit the scope of this PR for readability.

# Testing:
- Go to a model page via project side panel, or model's drilldown in workflow and see equations:
image

- Go to a broken model page and see that it correctly errors out on the task runner rather than waiting for a timeout (no pic of logs thats kindve silly?)
Here is a sample broken model where I have changed the expression to be a bit nonsense for `t0` This will fail having a diagram as well as equations
[SIR with failed equations.json](https://github.com/user-attachments/files/17044884/SIR.with.failed.equations.json)


packages/mira/tasks/generate_model_latex.py
```diff
@@ -0,0 +1,76 @@
...
+                odeterms[var] += t.rate_law.args[0]
+
+        # Time
+        symb = lambda x: sympy.Symbol(x)
+        try:
+            time = model.time.name
+        except:
+            time = "t"
+        finally:
+            t = symb(time)
```
mwdchang, line 40: This try/except block feels a bit lazy - can we do proper checks to see if time is valid rather than letting it blow up and fall into except/final blocks?

liunelson: Lines 35-39 are probably not necessary. The sympy equations just need a time symbol (which is defined in Line 40).

YohannParis (approved):

packages/mira/tasks/generate_model_latex.py
```diff
@@ -0,0 +1,76 @@
...
+		    #Reformat:
+        odesys = "\\begin{align} \n    " + " \\\\ \n    ".join([eq for eq in odesys]) + "\n\\end{align}"
+
+        taskrunner.write_output_dict_with_timeout({"response": odesys})
+        print("Generate latex succeeded")
+
+    except Exception as e:
+        sys.stderr.write(f"Error: {str(e)}\n")
+        sys.stderr.write(traceback.format_exc())
+        sys.stderr.flush()
```
mwdchang, line 66: Can you provide what your failure case looks like? 

Also the subscript looks. a bit weird here in your screenshot, note the "c" can cases.

image

packages/mira/tasks/generate_model_latex.py
Tom-Szendrey: Yep I will fix this

packages/mira/tasks/generate_model_latex.py
mwdchang: caused by underscore in variable name

PR #4837, opened 2024-09-18
Provide localhost adobe key on staging

YohannParis:
# Description

* Adobe service can only provide one allowed URL domain per API keu
* Using `yarn [dev|staging]` will create issues were the client uses `localhost` and the api key is for `app.[dev|staging].terarium.ai`
* The solution check if the origin of the `/api/adobe` call to get the key as `Origin` containing `localhost` and will return the appropriate key.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/AdobeController.java
```diff
@@ -15,16 +16,19 @@
 public class AdobeController {
 
 	@Value("${adobe.api-key}")
-	String key;
+	String adobeKey;
```
dvince2, line 19: Can't `adobeKey` just be defined with the local value in some local application.properties, so that we don't need to have this sort of logic in the controller class? `application-local.properties`?

YohannParis: The point is to be able to use the local `hmi-client` with the _staging_ or _dev_ `hmi-server`.
So adding it to the local properties will not solve the problem.

dvince2 (approved):

Issue #4828, opened 2024-09-17
[BUG]: workflow nodes get misconnected for no obvious reason


pascaleproulx:
**Describe the bug**
I am working on a workflow. While in it, nodes start to appear misconnected (one end of a link missing the round ball). No idea why it is happening. 

https://app.staging.terarium.ai/projects/59d1af73-e373-42da-9e88-ac020a39cc1f/workflow/8dfacd1a-d997-4d22-86fc-c6ea8c3f5a6a

**Screenshots**
Only thing I can see as error in the console is this:
![image](https://github.com/user-attachments/assets/b6b27a02-30ba-40cf-b733-133af1180ab0)
![image](https://github.com/user-attachments/assets/1d633fd0-a7bd-4eb3-991f-52e05cf8e6c7)
![image](https://github.com/user-attachments/assets/d877879f-faa6-449d-a9f6-ea1a16a4ef93)

mwdchang:
I am not sure how this bug is happening - I have yet to recreate.

One possible conjecture is that interaction are happening really fast, and a previous update partially overwrite the action to link the edge.

- Add model-operation
- Add model-config operation
- Move node around
- Add edge

The update from moving node interfere with add edge before the update is sent server side...

Issue #4406, opened 2024-08-08
[BUG]: Workflow node input ports persist after connections are unlinked


ngraham76:
**Describe the bug**
Workflow operator node input ports persist after a connection is unlinked, resulting in empty input ports stacked up with no way to remove them.

**To Reproduce**
See attached video capture below.

**Expected behavior**
When inputs unlinked, I would expect the port to disappear. When all inputs are removed, I would expect the node to return to its default state. If there is an optional port for that input (say for a Dataset) and all Datasets have been unlinked, only one Dataset (optional) port should remain.

Additionally, when a connection is made to an "optional" port, we don't need to display the word "(optional)" next to the connected input port label.

**Screen captures**

https://github.com/user-attachments/assets/2134d9dc-7703-4fb3-a8b2-a35f27527878

https://github.com/user-attachments/assets/5502ae89-4d35-4654-b595-3acdddef866e

PR #4842, opened 2024-09-19
Do not update if there is a save pending

mwdchang:
### Summary
This attempts to solve another round synchronization issue - the suspected culprits are the debouncers we are using for both saving-to-db and update-from-db going out of sync with each other. The primary reason for debouncers here is because we cannot send in a stream of updates across REST API without jamming-up every one else, so to resolve around that we introduce a couple of heuristics below.

To illustrate the potential sync problem: 
- Action 1: if we move operator-A, then
- Action 2: link operator-A to operator-B
- Action 1 is send to DB, and Action 2 is queued
  - As Action 2 is queued up, the response from Action 1 comes back and partially overwrite what we would've written with Action 2, at this point Action 2 is flushed tot he DB.

Heuristic 1: Do not update if there is a save action pending

Heuristic 2: Do not update if we are dragging

In both cases, we acknowledge there is an update by incrementing the version number, but skip over the fields. The idea is that, for
- heuristic 1: there is a save in progress that will result in a update-from-db, we will use that one 
- heuristic 2: drag-end results in a save, which is heuristic 1

The assumption is that only a single user works on an operator at a time.

Relates to
- https://github.com/DARPA-ASKEM/terarium/issues/4828
- https://github.com/DARPA-ASKEM/terarium/issues/4406

### Testing
This should remedy strange jerkiness/flickers, as well as weird connection states between operator ports.


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -300,28 +305,34 @@ const teraOperatorRefs = ref();
...
 // eslint-disable-next-line
 const _updateWorkflow = (event: any) => {
-	if (event.data.id !== wf.value.getId()) return;
-
-	// console.log('update workflow', event.data);
-	wf.value.update(event.data as Workflow);
+	if (event.data.id !== wf.value.getId()) {
+		return;
+	}
+	wf.value.update(event.data as Workflow, pendingSave, isDragging);
```
YohannParis, line 325: If you do not like one line returns, why not reverse the conditional test?
```suggestion
	if (event.data.id === wf.value.getId()) {
		wf.value.update(event.data as Workflow, pendingSave, isDragging);
	}
```

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -65,6 +65,28 @@ export class WorkflowWrapper {
 		const updatedNodeMap = new Map<string, WorkflowNode<any>>(updatedWF.nodes.map((n) => [n.id, n]));
 		const updatedEdgeMap = new Map<string, WorkflowEdge>(updatedWF.edges.map((e) => [e.id, e]));
 
+		if (pendingSave || isDragging) {
```
YohannParis, line 68: Why not creating a single boolean test with a more appropriate names, or a comment explaining why this test is happening?

YohannParis (approved):

shawnyama (approved):

packages/client/hmi-client/src/services/workflow.ts
Tom-Szendrey, line 68: I was not understanding why we would want to update the local version here 
Thank you for your help
May help to have a quick comment explaining this is to allow the user's outdated version to overwrite things prior to pulling

Tom-Szendrey (approved):
Awesome!

packages/client/hmi-client/src/services/workflow.ts
mwdchang: will do

PR #4859, opened 2024-09-19
Update Configured SIR.json

pascaleproulx:
# Description

Initial values need to be float.
Parameter values need to be much smaller 1/1000 of non-normalized initial values

Resolves #(issue)


YohannParis (approved):
Why do they need to be float? PyCIEMSS does not normalize it's input?

pascaleproulx:
Does not look like it. It fails if they are not.

PR #4881, opened 2024-09-20
Select only relavant variables for the chart

### LINKED ISSUE #4731: Chart performance and memory investigation
Every time we select a variable in Vegalite to display in a chart, we effectively replicate the dataset with our grammar.

eg
```
{ a: 123, b: 123, time: 1 }
{ a: 222, b: 222, time: 2 }
```

becomes this if we select both a & b
```
{ time::1, a: 123, b: 123, field: "a", value: 123 }
{ time: 1, a: 123, b: 123, field: "b", value: 123 }
{ time::2, a: 123, b: 222, field: "a", value: 222 }
{ time: 12 a: 123, b: 222, field: "b", value: 222 }
```

Let maybe spend a few hours to see if there's anything in Vegalite grammar that we can do where the dataset is independent of the variables selected.

If not - then I think we will just chop off the unused variables from the data set before sending it into Vegalite.

jryu01:
# Description

It looks like there's no way to filter or select data columns to reduce the size of the data using only vegalite chart spec. So , I just chopped off the unused variables from the datasets before passing it to the chart. 

Before we had an issue of browser crashing when more than 3 variables are selected. Now more variables can be selected. 

Screenshot 2024-09-20 at 2 15 58 PM

Note: We still have issues when too many variables are selected with large dataset. For this particular example (https://app.staging.terarium.ai/projects/f6a5af08-8d19-4462-bc5b-c12a9d6f6616/workflow/d7bf01c5-7c6e-4503-800c-f2d03dea9d1e?operator=bff38ac4-ca7a-44c4-8eca-e82f2a69dccb) , it will still crash the browser when all variables are selected. We may want to introduce the limit on the chart to prevent user from selecting too many variables. 

Resolves #4731

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
@@ -354,7 +354,6 @@ const preparedCharts = computed(() => {
 
 	const result = runResults.value[selectedRunId.value];
 	const resultSummary = runResultsSummary.value[selectedRunId.value];
-
 	const reverseMap: Record<string, string> = {};
 	Object.keys(pyciemssMap).forEach((key) => {
 		reverseMap[`${pyciemssMap[key]}_mean`] = key;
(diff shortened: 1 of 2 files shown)


mwdchang (approved):

YohannParis (approved):
LGTM, if this become an issue, we could just do the filtering in the back-end and only fetch what is needed for big dataset.

PR #4891, opened 2024-09-23
update funman, new feature to accept observables

shawnyama:
# Description

* Tested sir model with `noninf` observable
* I get true boxes when upper bound is greater than 1000
* Less than 1000 gives me false boxes

[obs-input.json](https://github.com/user-attachments/files/17100968/obs-input.json)


mwdchang:
@shawnyama your test cases are vacuously true so I don't think they exercised the Funman logic that we are looking for

You have starting conditions of S=1000, I=1, and R=0. 

Since this is a closed system, with  `noninf=S+R` it will never exceed 1001, so your upper bound > 1000 is always true.
And since S=1000 to start, your upper bound < 1000 will fail by default.

Can we check if the behaviour is as expected with more realistic scenarios? We can
- Take the model and run it via pyciemss to see what the result looks like
- Test it against funman with some time-period in the middle, e.g. what are the expected values of noninf between time=[30, 70]

mwdchang (approved):

PR #4895, opened 2024-09-23
4864 histogram chart fix

### LINKED ISSUE #4864: [BUG]: issues with parameter charts in calibrate
- Scenario: https://github.com/DARPA-ASKEM/terarium/blob/main/testing/manual/calibrate.md
- Step: 4
- Link to the project: https://app.staging.terarium.ai/projects/388d3bde-461d-47f7-b59e-21e231e6a8e0/workflow/f87dd1ea-5640-439d-a20a-3d45e7bb0b42?operator=30a5941b-b9bb-4819-88b7-d28577fa0cb4

**Describe the issue**
1. Count is too big:
Explanation from @liunelson :
Parameter plots show a count in the range of 300 when there are only 10 samples. Each sample has 100 timepoints. So it looks like 10*1000 count. It should use only the 10 sample points for the first timepoint, not all of them.

2. before/after histogram bins should be the same

image

jryu01:
# Description

**Before:**
Screenshot 2024-09-23 at 4 20 33 PM

**After:**
Screenshot 2024-09-23 at 4 20 59 PM

Resolves #4864

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -219,6 +219,7 @@
 						<br />
 						<template v-for="setting of selectedParameterSettings" :key="setting.id">
 							<vega-chart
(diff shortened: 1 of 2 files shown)


mwdchang (approved):
thx!

blanchco (approved):
nice

PR #4900, opened 2024-09-24
Fix dockerfiles.

kbirk:


packages/equation_extraction/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN ./gradlew bootJar
 WORKDIR /
 
 # Copy the jar to the root directory
-RUN mv /taskrunner/build/libs/*.jar .
+RUN cp /taskrunner/build/libs/*.jar .
```
kbirk, line 46: this is the fix, copy the jar files, dont move them

packages/text_extraction/Dockerfile
```diff
@@ -26,11 +26,7 @@ RUN ./gradlew bootJar
 
 WORKDIR /
 
-# Copy the jar to the root directory
```
kbirk, line 29: We dont need to move these for this container.

dgauldie:
there are a couple more fixes I want to put in this PR

packages/equation_extraction/Dockerfile
dgauldie: how come we need to copy in this build but not in the others?

dgauldie (approved):
ok

PR #4903, opened 2024-09-24
4854 extract equations from image

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -716,6 +724,109 @@ public ResponseEntity<TaskResponse> createEnrichAMRTask(
...
+		@RequestParam(name = "mode", required = false, defaultValue = "ASYNC") final TaskMode mode,
+		@RequestBody final String image
+	) {
+		final Schema.Permission permission = projectService.checkPermissionCanRead(
+			currentUserService.get().getId(),
+			projectId
+		);
+
+		// validate the image
+		if (image == null || image.isEmpty()) {
```
dvince2, line 756: I think that this can be done in the parameters area with a "required" flag?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
YohannParis: I do not think you can make a body required?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dgauldie: fixed

dgauldie: The body is required by default. It won't let you have an empty body in the post call. I removed this check

PR #4902, opened 2024-09-24
4751 feat create intervention policy   misc adjustments

asylves1:
# Description

image

Changes:
- static interventions in a intervention will have the same value for `timestep` 
- each static intervention will  have `type` which can be 'state/parameter'

# Note Required:
https://github.com/DARPA-ASKEM/pyciemss-service/pull/120

PR #4906, opened 2024-09-24
save for reuse for simulation assets

blanchco:
# Description

* Ability to save for reuse for simulation objects, depending on the simulation type (optimize, simulate, calibrate) assets will be saved/updated.
* Created a backend endpoint to handle saving/updating the assets
* Created a new modal to save assets when saving for re-use on a simulation object, I figured a new component needed to made here since the logic is pretty different to how we are using the `tera-save-asset-model`.
 Simulation -> save dataset
 Calibrate -> save dataset, update model configuration
 Optimization -> save dataset, update intervention policy

EDIT: 
As discussed, output labels will now change if there are multiple outputs.  It will arbitrarily choose the first asset to use as the name


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
```diff
@@ -81,6 +88,55 @@ public SimulationUpdate appendUpdateToSimulation(
 		return created;
 	}
 
+	public Dataset createDatasetFromSimulation(
```
blanchco, line 91: moved this function from the controller to the service

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SimulationController.java
```diff
@@ -349,41 +347,89 @@ public ResponseEntity<Dataset> createFromSimulationResult(
...
-				projectAssetService.createProjectAsset(project.get(), AssetType.DATASET, dataset, permission);
-				// underlying asset does not exist
-				return ResponseEntity.status(HttpStatus.CREATED).body(dataset);
-			} else {
-				log.error("Failed to add the dataset from simulation {} result", id);
-				return ResponseEntity.internalServerError().build();
+			List<TerariumAsset> assets = new ArrayList<>();
+			// Update the intervention policy name if the simulation is an optimization
+			if (simulationType == SimulationType.OPTIMIZATION) {
+				final Optional<InterventionPolicy> interventionPolicy = interventionPolicyService.getAsset(assetId, permission);
```
dvince2, line 409: There is an incoming breaking change to Interventions in [this PR](https://github.com/DARPA-ASKEM/terarium/pull/4902). We probably want to wait on merging PR #4902 first before merging this or else it will propagate the breakage

dvince2: I don't think that this PR will need any code changes though for these changes, more so, we just don't want to be saving references to assets that are immediately going to be blown away

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
```diff
@@ -81,6 +88,55 @@ public SimulationUpdate appendUpdateToSimulation(
...
+			this.copySimulationResultToDataset(sim.get(), dataset);
+			datasetService.updateAsset(dataset, projectId, permission);
+
+			if (!addToProject) {
+				return dataset;
+			}
+
+			// Add the dataset to the project
+			final Optional<Project> project = projectService.getProject(projectId);
+			if (project.isPresent()) {
```
dvince2, line 127: The above code isn't too expensive, but, it _will_ create a dataset in the database. If this project check fails then this dataset would be orphaned. If that's not expected then we may want to check for the presence of the project and simulation right at the start to know that we can proceed

dvince2 (changes requested):
See comments ^

packages/client/hmi-client/src/services/models/simulation-service.ts
```diff
@@ -245,6 +248,25 @@ export async function makeEnsembleCiemssCalibration(params: EnsembleCalibrationC
...
+export async function createSimulationAssets(
+	projectId: string,
+	simulationId: string,
+	simulationType: SimulationType,
+	newName: string | null,
+	assetId?: string
+) {
+	try {
+		const response: AxiosResponse<TerariumAsset[]> = await API.post(
+			`/simulations/${simulationId}/create-assets-from-simulation/${projectId}?name=${newName}&simulation-type=${simulationType}${assetId ? `&asset-id=${assetId}` : ''}`
```
mwdchang, line 260: Your POST request is rather unconventional that it is like a GET request, it may be better to have a proper POST-body instead of this URL string logic which can have encoding issues.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
blanchco: makes sense - Ill make those changes

packages/client/hmi-client/src/services/models/simulation-service.ts
blanchco: fair, will make those changes

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
blanchco: I've added the project check at the start of the function

packages/client/hmi-client/src/services/models/simulation-service.ts
blanchco: changes made

blanchco:
Im putting this back in draft as per our discussion, I will need to change the output port names

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SimulationController.java
YohannParis, line 1: Why moving createDatasetFromSimulation from the service to the controller?

blanchco: I was trying to make the PR a little more readable as if those changes were never, but Ill revert it back since its probably better in the service anyways

PR #4916, opened 2024-09-25
remove runType

Tom-Szendrey:
# Description
Removing run type as we no longer have julia


shawnyama (approved):

YohannParis (approved):
I love a good cleaning!

Issue #4848, opened 2024-09-19
[BUG]: Simulate in `Normal` preset not working


YohannParis:
- Scenario: [Simulate](https://github.com/DARPA-ASKEM/terarium/blob/main/testing/manual/simulate.md
- Step: 4.1
- Link to the project: https://app.staging.terarium.ai/projects/8583471a-9f09-49c2-8501-cf239870ba58/workflow/ab7f641f-99c3-4489-8672-d834c7bde053?operator=e40abfe4-5131-4764-95ae-f0a9de682d4b

**Describe the issue**
The following error was returned instead: `7a052ce4-6fe0-4c23-85a3-f6c95f6fa22a`

```python
[...]

PR #4921, opened 2024-09-25
fix simulate preset selection not triggering update

mwdchang:
### Summary
Change the order of variable updates for presets to take effect

### Testing
- Select a model with some parameter value variability (with lower/upper bound not equal). 
- Hook up model => model-config => simulate
- Run "normal" simulation, then run "fast" simulation, it should be obvious the two runs are different as the former has significantly more samples than the latter.

Fixes: https://github.com/DARPA-ASKEM/terarium/issues/4848


Tom-Szendrey (approved):
Nice catch

YohannParis (approved):
LGTM

PR #4923, opened 2024-09-25
Funman output

shawnyama:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-slider-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-slider-panel.vue
@@ -151,6 +151,15 @@ header:not(.tab) {
 		background-color: var(--surface-100);
 		border-right: 1px solid var(--surface-border-light);
 	}
+
+	/** Override default accordion styles */
+	&:deep(.p-accordion-header-link) {
+		background-color: var(--surface-100);
+	}
+
+	&:deep(.p-accordion-content) {
+		background-color: var(--surface-100);
+	}
 }
 
 .tab {
--- a/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
@@ -643,15 +643,6 @@ ul {
 	padding: 0 var(--gap-2);
 }
 
-/** Override default accordion styles */
-:deep(.p-accordion-header-link) {
-	background-color: var(--surface-100);
-}
-
-:deep(.p-accordion-content) {
-	background-color: var(--surface-100);
-}
-
 /* Override grid template so output expands when sidebar is closed */
 .overlay-container:deep(section.scale main) {
 	grid-template-columns: auto 1fr;
--- a/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
@@ -145,8 +145,8 @@
 					<!--c205809b-3236-4bf3-92ad-6c3355b32698-->
 					<!--e44bd2c5-5c93-406e-b848-5ec4b3192d17-->
 					<tera-funman-output
-						v-if="true"
-						:fun-model-id="'c205809b-3236-4bf3-92ad-6c3355b32698'"
+						v-if="activeOutput"
+						:fun-model-id="activeOutput?.id"
 						:trajectoryState="node.state.trajectoryState"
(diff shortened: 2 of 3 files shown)


YohannParis (commented):
You should remove those `gollm/gollm_task.egg-info` folder and files.
Update from `main` to have your `.gitignore` up to snuff.

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -804,3 +805,179 @@ export function createInterventionChart(
...
+/// /////////////////////////////////////////////////////////////////////////////
+// Funman charts
+/// /////////////////////////////////////////////////////////////////////////////
+
+enum BoxType {
+	Satisfactory = 'Satisfactory',
+	Unsatisfactory = 'Unsatisfactory',
+	Ambiguous = 'Ambiguous',
+	ModelChecks = 'Model checks'
+}
```
mwdchang, line 818: This feels like you are jamming two different ideas into the same enum and might be a point of confusion later on, or that it is not well-named.

Because funman has only Satisfactory/Unsatisfactory "boxTypes"  it is a bit confusing to introduce new things and only use parts of it and apply to things that are not really boxes.

If we are using this to encode colour then we should just say so, and rather than `modelChecks[i].boxType` it may be clearer just to say `modelChecks[i].colour` or something.

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -804,3 +805,179 @@ export function createInterventionChart(
...
+	// Show checks for the selected state
+	const stateIdConstraints = constraints.filter((c) => c.variables.includes(stateId));
+	const modelChecks = stateIdConstraints.map((c) => ({
+		boxType: BoxType.ModelChecks,
+		startX: c.timepoints.lb,
+		endX: c.timepoints.ub,
+		// If the interval bounds are within the min/max values of the line plot use them, otherwise use the min/max values
+		startY: Math.max(c.additive_bounds.lb, minY),
+		endY: Math.min(c.additive_bounds.ub, maxY)
+	}));
```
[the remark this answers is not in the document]
shawnyama: Yeah I decided to just focus on the line rather than the model checks. I set it up like this because if we choose greater than the upper bound is super high and with less than the lower bound is super low. In those cases you can't have a good idea of the trajectory of line as it either goes in a straight horizontal line or in some cases the graph doesn't render properly (just empty space).

packages/client/hmi-client/src/services/charts.ts
shawnyama: that makes sense thanks

mwdchang (changes requested):
1) The results returned from Funman doesn't seem to make sense, for example you are showing the model variable "R"
- It goes up by a factional amount, which is very suspicious given global population is like 1001
- It stays constant after going up a little bit, this seemingly violates the math

2) It would be helpful to have visual markers on the chart, where the points are taken. For example your screenshot has the time going by increments of 10, however your chart draws a grid line every 5 steps. This gives a wrong illusion of what Funman is returning. Having the markers are also very helpful for debugging, and we can turn them off later when things are stable.

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -804,3 +805,198 @@ export function createInterventionChart(
...
+	const modelChecks = stateIdConstraints.map((c) => ({
+		legendItem: FunmanChartLegend.ModelChecks,
+		startX: c.timepoints.lb,
+		endX: c.timepoints.ub,
+		// If the interval bounds are within the min/max values of the line plot use them, otherwise use the min/max values
+		startY: Math.max(c.additive_bounds.lb ?? minY, minY),
+		endY: Math.min(c.additive_bounds.ub ?? maxY, maxY)
+	}));
+
+	console.log(boxLines);
```
jryu01, line 849:
```suggestion
```

mwdchang (approved):

YohannParis (approved):
good job!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ValidateModelConfigHandler.java
```diff
@@ -91,6 +91,8 @@ public TaskResponse onSuccess(final TaskResponse resp) {
 			final HttpEntity fileEntity = new ByteArrayEntity(bytes, ContentType.TEXT_PLAIN);
 			simulationService.uploadFile(simulationId, resultFilename, fileEntity);
 
+			System.out.println(result.get("response"));
```
dvince2: debug comment

Issue #4764, opened 2024-09-12
[BUG]: Model page shows incorrect/missing metadata when models has parameters named `*_*`


liunelson:
**Describe the bug**

I have a model (see below) for which the model page is showing incorrect parameter information.
1. the parameter `NPI_mult` is shown as `NPI` only and is missing all its metadata
2. ditto for the parameter `vacc_mult`
3. all parameters `r_*` are forced into a single matrix, even though this model is not a product of stratification

This is probably due to the (incorrect) assumptions about IDs with `*_*`.

The model:
[SEIRHD vacc model for LA County t0 = 10_28_2021 (1).json](https://github.com/user-attachments/files/16972411/SEIRHD.vacc.model.for.LA.County.t0.10_28_2021.1.json)

**To Reproduce**
1. Upload this model
2. Inspect its model page
3. Create a Configure Model operator and plug the model to it
4. Inspect the parameter table to see the lack of values of `NPI_mult, vacc_mult` and the incorrect matrix for `r_*`

**Screenshots**
![Screenshot 2024-09-11 at 9 55 07 PM](https://github.com/user-attachments/assets/dd4bcb89-7243-47b7-8e23-28ae75bcba59)
![Screenshot 2024-09-11 at 9 54 57 PM](https://github.com/user-attachments/assets/a3cdef9b-c559-4a11-b691-e729c60b7bcd)
![Screenshot 2024-09-11 at 9 54 47 PM](https://github.com/user-attachments/assets/bb94c35d-bcbb-4638-8d77-9f671681114d)
![Screenshot 2024-09-11 at 9 37 08 PM](https://github.com/user-attachments/assets/371fc64b-7b8d-4eba-b723-c62aa6e28425)

mwdchang:
I think we can add additional heuristics to detect cases like this - but at some point these become educated guesses.

PR #4930, opened 2024-09-25
improve parsing for stratified parameters

mwdchang:
### Summary
Improve parser to reduce the number of false positives where we incorrectly identify something as stratified but it is not. Use the context (grounding info) data, if applicable to determine whether a parameter is the result of an earlier stratification process.

### Testing
Load and view the attached AMR model 

[SEIRHD.vacc.model.for.LA.County.t0.10_28_2021.1.json](https://github.com/user-attachments/files/17153595/SEIRHD.vacc.model.for.LA.County.t0.10_28_2021.1.json)

parameters like `NPI_mult` and `vacc_mult` will denoted as stand-alone parameters and not grouped under a stratified category. Ditto for the`r_*` variables.

Closes: https://github.com/DARPA-ASKEM/terarium/issues/4764


packages/client/hmi-client/src/model-representation/mira/mira.ts
```diff
@@ -65,20 +67,22 @@ export const getContextKeys = (miraModel: MiraModel) => {
...
 				Object.entries(miraConcept.context).forEach(([k, v]) => {
 					addWithGuard(k, v as string);
 				});
 			});
 		}
 	});
 
-	const modifiers = [...modifierKeys];
-	return modifiers;
+	// const modifiers = [...modifierKeys];
```
YohannParis, line 77:
```suggestion
```

packages/client/hmi-client/src/model-representation/mira/mira.ts
```diff
@@ -107,14 +112,27 @@ export const collapseParameters = (miraModel: MiraModel, miraTemplateParams: Mir
 			continue;
 		}
 
+		const nameContainsContext = _.intersection(tokens, contextValues).length > 0;
+		const hasNumericSuffixes =
+			tokens.length > 1 &&
+			tokens.every((d, idx) => {
```
YohannParis, line 118: Why the `tokens.length > 1` test, `every` works on empty array as well. Or just to be sure?

packages/client/hmi-client/src/model-representation/mira/mira.ts
```diff
@@ -33,8 +33,9 @@ export const emptyMiraModel = () => {
 /**
  * Collection of MMT related functions
  * */
-export const getContextKeys = (miraModel: MiraModel) => {
```
YohannParis, line 36: You could keep this function with the following, so you do not have to modifed the service. Could still be useful
```javascript
export const getContextKeys = (miraModel: MiraModel) => getContext(miralModel).keys;
```

YohannParis (approved):
LGTM

packages/client/hmi-client/src/model-representation/mira/mira.ts
mwdchang: We want `x_123` to be true, but `x_abc` to be false and `x` to be false.

blanchco (approved):
tested and looks good

Issue #4863, opened 2024-09-19
Default charts for calibrate


pascaleproulx:
Currently, calibrate plots by default the variables that were selected in the mapping. This is a good start but we should also display:

- [ ]  the error plots for those same variables and 
- [ ] the parameter charts (only those that were not constant)
- [ ] each section (Loss, Parameters, Variables, Errors) should be collapsable

Other enhancements:

- [ ] The output settings panel should be collapsed by default
- [ ] The charts should use the space available

PR #4931, opened 2024-09-25
Show default charts for variables, errors and parameters on calibrate run

jryu01:
# Description

On initial calibrate run, following charts will be selected by default

- variables that were selected in the mapping
- the error plots for those same variables and
- the parameter charts (only those that were not constant)

see: https://github.com/DARPA-ASKEM/terarium/issues/4863#issue-2537328926

https://github.com/user-attachments/assets/5dd7cd5d-f1fa-4fa0-9e38-35c67effbf9f

Resolves #(issue)


Tom-Szendrey (approved):
Thanks

YohannParis (approved):

packages/client/hmi-client/src/services/model.ts
```diff
@@ -208,6 +208,10 @@ export const getTypesFromModelParts = (model: Model) => {
 	return typeMapping;
 };
 
+export function getModelParameters(model: Model): ModelParameter[] {
+	return model?.semantics?.ode?.parameters ?? [];
+}
```
mwdchang, line 213: There are some overlapping here, to get parameters there is also
- model-representation/services#getParameters `https://github.com/DARPA-ASKEM/terarium/blob/921e3ee9e74950a5bc989dd6b2ced5b981315f55/packages/client/hmi-client/src/model-representation/service.ts#L104-L114`

CC @blanchco

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -1007,16 +1008,12 @@ async function getAutoMapping() {
...
-		modelPartTypes,
-		modelParameters: modelParams
-	} = await setupModelInput(modelConfigId.value);
+	const { model, modelConfiguration, modelOptions, modelPartUnits, modelPartTypes } = await setupModelInput(
+		modelConfigId.value
+	);
 	modelConfig.value = modelConfiguration;
 	modelStateOptions.value = modelOptions;
-	modelParameters.value = modelParams ?? [];
+	modelParameters.value = !model ? [] : getParameters(model);
```
YohannParis: Could we instead improve `getParameters` to return `[]` if `model` is null or undefined.

mwdchang: I would argue no.

I think at some level, we need to have functions that act on concrete objects, if we constantly have to check if the parameters we pass in are null or undefined it becomes quite difficult to to write coherent code, or to debug problems once we have nested several levels of these type of logic.

packages/client/hmi-client/src/services/model.ts
jryu01: I switched to use the existing function. Thanks!

PR #4936, opened 2024-09-26
Calibration Configuration inferred params fix

Tom-Szendrey:
# Description
Fix inferred params list so that it is not empty when we save model configs from calibration
The issue was that we were not providing the reference to the model config for the mapping, so we would create them and our `POST` would return a correct result but any `GET` after would not be able to find this field

# To Test:
Run a calibration,
Save it as a model config for reuse
Open it in the model config drilldown and see that this section is not all 0
image


dvince2 (approved):
lgtm

blanchco (approved):
Nice!

PR #4957, opened 2024-09-26
multiple input field open when selecting one

asylves1:
# Description

Issue:
When a user clicks on an equation, all of the equations input field appear. This is due to the equations all sharing the same name:

Change:
Add the array index of the equation to the name parameter.

Testing:
Add 2 equations using the add equation field and select one of the equations. Only one input field should appear.

image
image

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
@@ -113,6 +113,19 @@ const docText = ref<string | null>();
 const isFetchingPDF = ref(false);
 const clonedState = ref(cloneDeep(props.node.state));
 
+function updateOutputPort() {
+	const outputPort = cloneDeep(props.node.outputs?.find((port) => port.type === 'documentId')) || null;
+	if (!outputPort) return;
+	outputPort.value = [
+		{
+			...outputPort.value?.[0],
+			equations: clonedState.value.equations
+		}
+	];
+
+	emit('append-output', outputPort);
+}
+
 onMounted(async () => {
 	if (props.node.state.documentId) {
 		isFetchingPDF.value = true;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
```diff
@@ -113,6 +113,19 @@ const docText = ref<string | null>();
...
+	if (!outputPort) return;
+	outputPort.value = [
+		{
+			...outputPort.value?.[0],
+			equations: clonedState.value.equations
+		}
+	];
+
+	emit('append-output', outputPort);
+}
```
mwdchang, line 127: - We have different meanings for "appending-output" and "update-output", please rename the function so to not cause confusion down the road.

- I do not know how operator state is managed in this operator, as usually we do not clone the state from existing output, but rather clone from the current operator's state. The idea is that if we need to switch output we would be able to restore the state that was used to derive the output. If this isn't clear, let me know and I can explain in better detail, if it is irrelevant in this operator feel free to ignore this comment.

mwdchang (changes requested):
Can you please provide better PR description and testing instructions? 

Your description "stops all input fields from opening when only the selected one should" seems like it has nothing to do with the source code change, which is about appending output-ports from what I can gather. Not sure if we are missing some commits, but better clarity would help the reviewers as well as tracking purposes.

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
```diff
@@ -164,6 +176,7 @@ function onUpdateInclude(asset: AssetBlock<DocumentExtraction>) {
 
 	outputPort = cloneDeep(props.node.outputs?.find((port) => port.type === 'documentId')) || null;
 	if (!outputPort) return;
+	console.log('clonedState.value[portType]', clonedState.value[portType]);
```
YohannParis, line 179:
```suggestion
```

asylves1:
Removing code change from `tera-document-operation-drilldown.vue`
Will open a new PR because the design has changed

mwdchang (approved):
I am not sure what the changes relate to or how to test it.
Code looks okay.

YohannParis (approved):

PR #4962, opened 2024-09-27
on-save plumbing for model operator

mwdchang:
### Summary
This PR changes the "save as" semantic for model-operator, such that on top of
- creating a new model
- adding the model into the project

It will also update the model-operator to take in the newly create model, any downstream operators will be invalidated as if a new output has been chosen.

### Note
Because model-operator ceeds all control to tera-model component, we don't have a graceful way of isolating the logic to  the model-operator itself. What we end up is tera-model will detect whether it is on the workflow-route or not, and emit an event back to the invoker. We will then catch that event in model-operator and do the usual update-state, append-output to propagate new changes.

### Testing
- Add a model onto workflow canvas
- Link the model to a model-config operator
- Drilldown into the model operator, save as a new model with a different name

Check new model is created and available within the project
Check the new model replaces the existing one in the model-operator
Check the new model is the output of model-operator and downstream operators are invalidated because there is a new instance


packages/client/hmi-client/src/components/workflow/ops/model/tera-model-node.vue
```diff
@@ -93,6 +94,16 @@ onMounted(async () => {
 		await getModelById(state.modelId);
 	}
 });
+
+watch(
+	() => props.node.active,
+	async () => {
+		const newModelId = props.node.outputs.find((d) => d.id === props.node.active)?.value?.[0];
+		if (newModelId) {
```
YohannParis, line 102: Should check that the `newModelId` is not equal to the current `model.value.id`.

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -136,8 +141,11 @@ function onSaveAs() {
 }
 
 // Save modal
-function onModalSave() {
+function onModalSave(event: any) {
 	showSaveModal.value = false;
+	if (isWorkflow.value) {
+		emit('on-save', event);
+	}
```
YohannParis, line 148: Why the test if we are in a workflow or not? What does this prevent to emit that event in the model asset page?

YohannParis (approved):
LGTM

blanchco (approved):

packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -124,6 +125,10 @@ const isNaming = computed(() => isEmpty(props.assetId) || isRenaming.value);
 const hasChanged = computed(() => !isEqual(model.value, temporaryModel.value));
 const hasEditPermission = useProjects().hasEditPermission();
 
+const route = useRoute();
+
+const isWorkflow = computed(() => route.params.pageType === 'workflow');
```
shawnyama, line 130: Don't think this has to be computed. Maybe passing an `isInWorkflow` prop would be better

packages/client/hmi-client/src/components/workflow/ops/model/tera-model-drilldown.vue
```diff
@@ -5,21 +5,35 @@
...
+	state.modelId = event.id;
+	emit('update-state', state);
+	emit('append-output', {
+		type: ModelOperation.outputs[0].type,
+		label: event.header.name,
+		value: [event.id],
+		state,
+		isSelected: false
+	});
+};
```
shawnyama, line 38: do we want to append an output or do we want to replace it?

YohannParis: We replacing it, will fix this.

mwdchang: You want to append, we don't want to replace.

YohannParis: I'm editing this PR to do that. We need to be able to edit current output. Then the goal is to also propagate changes if the ouput has been updated.

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -159,11 +159,11 @@
...
 			:spawn-animation="drilldownSpawnAnimation"
 			@append-output="(event: any) => appendOutput(currentActiveNode, event)"
+			@close="addOperatorToRoute(null)"
+			@select-output="(event: any) => selectOutput(currentActiveNode, event)"
+			@update-output="(event: any) => updateOutput(currentActiveNode, event)"
 			@update-state="(event: any) => updateWorkflowNodeState(currentActiveNode, event)"
 			@update-status="(status: OperatorStatus) => updateWorkflowNodeStatus(currentActiveNode, status)"
-			@select-output="(event: any) => selectOutput(currentActiveNode, event)"
-			@close="addOperatorToRoute(null)"
-			@update-output-port="(event: any) => updateOutputPort(currentActiveNode, event)"
```
mwdchang, line 166: @YohannParis what is the rationale behind these name changes? 

There are operator components that will emit `update-output-port` , see https://github.com/search?q=repo%3ADARPA-ASKEM%2Fterarium%20update-output-port&type=code so they are going to break.


I think we should revert here and focus on the save-as feature, and not try to to do too many different things all at once.

YohannParis: As you wish, I will make a small PR just for that renaming.

YohannParis (approved):

PR #4983, opened 2024-09-30
update SerializeWithoutTerariumFields to allow customizing what field…

dgauldie:
…s to keep/delete

# Description

Resolves #(issue)

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
@@ -11,7 +11,10 @@ import jakarta.persistence.Column;
 import jakarta.persistence.MappedSuperclass;
 import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/TerariumAsset.java
```diff
@@ -77,44 +80,53 @@ protected TerariumAsset cloneSuperFields(final TerariumAsset asset) {
...
+				}
+			}
 		}
-		objectNode.set("metadata", metadata);
-
-		return objectNode.toString();
 	}
 
-	public String serializeWithoutTerariumFieldsKeepId() {
+	public String serializeWithoutTerariumFields(
```
YohannParis, line 98: Could you add some JavaDoc so it's easier to read the refactored code done in the controllers?

dgauldie: added

kbirk (approved):

PR #4985, opened 2024-09-30
Add equation scroll bar

asylves1:
# Description

Issue:
- when the equations longer then the asset block the full panel including all of the different equation becomes scrollable. 

Testing:
Add an equation longer then the asset block. The scrollbar show appear under the one longer equation. The full panel should no longer be scrollable.

image

Resolves #(issue)


YohannParis (approved):
Great job!

dgauldie (approved):

PR #4990, opened 2024-09-30
fix with equations not appearing

asylves1:
# Description

Issue
- The equations after starting a new project and uploading a new doc would not appear. This was caused by assigning the equations to a unused ref

Testing:
- Open a new project and upload a new PDF.
- Add doc to workflow and connect model from equation node
- open node and the extracted equations should appear

image

Resolves #(issue)


shawnyama (approved):
lgtm

blanchco (approved):

Tom-Szendrey (approved):
looks like we want to use `clonedState.value.equations` but I am not overly familiar with this component

PR #4992, opened 2024-09-30
Calibrate timestamp field UI tweak

Tom-Szendrey:
# Description

1) Timestamp column has its own state field.
2) Timestamp column has its own section in the mapping (see screenshot below)
3) On open node try to fill in the timestamp column if its not yet filled in. We will look for the first thing that starts with "time"

# Warning:
This updates the state object for calibration so older calibration nodes will run into issues with their mappings.

# Pics

image

With designs here: https://www.figma.com/design/SLO56PGmGq7kFQRFCeaPsY/Terarium---September-2024?node-id=235-584&node-type=canvas&t=3E0tQ1xlWHmsXuxp-0


YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -26,13 +26,59 @@ t d
...
+										:options="datasetColumns?.map((ele) => ele.name)"
+									/>
+								</template>
+							</Column>
+							<Column field="deleteRow">
+								<template #header>
+									<span class="column-header"></span>
+								</template>
+								<template #body="{ data }">
+									<Button class="p-button-sm p-button-text" icon="pi pi-trash" @click="deleteMapRow(data.id)" />
```
shawnyama, line 72: I don't think time should have a delete button, designs don't have it

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
Tom-Szendrey: As this table is just all of the fields that have been selected as `timestamp` on the model side it seems helpful that a user can delete these.
If a user ends up selecting `timestamp` then they may end up in a state which they want to correct it but can only do so via recreating the node
image

If I do not allow the user to select the `timestamp` in this dropdown to avoid this exact scenario then we may run into edge cases that either do or do not exist yet so I think it is worth keeping to give the user more control

mwdchang:
Per discussion, we should figure out if we want to
- allow potential invalid scenarios, and have mechanism to fix them after the fact, for example - having timestamp mapped multiple times doesn't really make sense. Or
- prevent invalid scenarios from happening in the first place

mwdchang (approved):
Thanks

PR #5012, opened 2024-10-01
Resize charts based on available space

### LINKED ISSUE #4863: Default charts for calibrate

jryu01:
# Description

https://github.com/user-attachments/assets/b9dd0584-2497-40cb-a73f-7af583ef77c3

Charts should resize on simulate and optimize drilldowns as well. 

Resolves #4863


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -875,7 +876,7 @@ const LOSS_CHART_DATA_SOURCE = 'lossData'; // Name of the streaming data source
 const lossChartRef = ref<InstanceType<typeof VegaChart>>();
 const lossChartSpec = ref();
 const lossValues = ref<{ [key: string]: number }[]>([]);
-const updateLossChartSpec = (data: string | Record<string, any>[]) => {
+const updateLossChartSpec = (data: string | Record<string, any>[], size: { width: number; height: number }) => {
```
Tom-Szendrey, line 879: It looks to me we are passing `lossChartSize` as a parameter `size` 
Why not just use `lossChartSize.value` for this similar to how you are using `chartSize` sorry if i am missing something obvious

jryu01: Just a preference, I just wanted to make it more explicit with the dependent variables and just wanted it to be decoupled from parent scope although in this case, its simple enough that it doesn't make much difference.

jryu01: It helps in some cases for example, when you look at the watcher

```
watch(
	[() => props.node.state.inProgressCalibrationId, lossChartSize],
	([id, size]) => {
		if (id === '') {
			showSpinner.value = false;
			updateLossChartSpec(lossValues.value, size);
```

By passing the size param (which is watched variable), you can easily tell that `updateLossChartSpec` depends on that variable. If `lossChartSize.value` were used inside the function and size wasn't passed as param here, it wouldn't be clear why `lossChartSize` is being watched here unless you go into and look at the content of the function.

Tom-Szendrey: This makes total sense!
Is the simulation chart in the same situation as this? If it is are we able to make these two consistent with eachother?

jryu01: No simulation drilldown doesn't have this piece of code and the loss chart.

Tom-Szendrey (approved):

shawnyama (approved):
neat!

PR #5018, opened 2024-10-02
Link equations to page

asylves1:
# Description

Design change
- when the user clicks on an equation the pdf's page will change to the page containing the equation.

Testing
- Should see the current selected equation in the pdf viewer

image

Resolves #(issue)


YohannParis (approved):
LGTM

blanchco (approved):
cool!

PR #5019, opened 2024-10-02
[FEAT]: Add AMR enrichment to 'Enrich metadata with AI assistant'

### LINKED ISSUE #4856: [FEAT]: Add AMR enrichment to 'Enrich metadata with AI assistant'
image

dvince2:
# Description 

Adding a new CompoundTask which can run successive secondary tasks after the first finishes. Using this to run AMR enrichment and model card generation.

Resolves #4856


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -1069,4 +1110,71 @@ public ResponseEntity<Void> cancelTask(@PathVariable("task-id") final UUID taskI
 		taskService.cancelTask(taskId);
 		return ResponseEntity.ok().build();
 	}
+
+	private TaskRequest getModelCardTask(DocumentAsset document, Model model, UUID projectId) {
```
YohannParis, line 1114: Should `getModelCardTask` and `getEnrichAMRTaskRequest` methods be in a service file instead of this controller?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -744,6 +720,101 @@ public ResponseEntity<TaskResponse> createGenerateResponseTask(
...
+	 * @param modelId UUID of the model to enrich
+	 * @param documentId UUID of the document to use for enrichment
+	 * @param mode TaskMode to run the task in (is this ASYNC?)
+	 * @param projectId UUID of the project to associate the task with for permissions
+	 * @param overwrite boolean to determine if the model should be overwritten
+	 * @return TaskResponse with the task ID
+	 */
+	@GetMapping("/enrich-model-metadata")
+	@Secured(Roles.USER)
+	@Operation(summary = "Dispatch a multiple GoLLM tasks to enrich model metadata")
```
YohannParis, line 735:
```suggestion
	@Operation(summary = "Dispatch multiple GoLLM tasks to enrich model metadata")
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -744,6 +720,101 @@ public ResponseEntity<TaskResponse> createGenerateResponseTask(
...
+		// Grab the model
+		final Optional<Model> model = modelService.getAsset(modelId, permission);
+		if (model.isEmpty()) {
+			log.warn(String.format("Model %s not found", modelId));
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("model.not-found"));
+		}
+
+		final TaskRequest req;
+
+		if (document.isPresent()) {
```
YohannParis, line 784: Do we need this condition with the test on [Line 770](https://github.com/DARPA-ASKEM/terarium/pull/5019/files#diff-45634443c497336d9e2e95ad796b7a09e0d7f9548c020d74a9497ef52759e636R770)?

YohannParis (approved):
LGTM

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dvince2: We do - the check on line 770 is checking to make sure that if there is a document it has content (which is a safety check that was there before so it feels important). However it is a valid case to have *no* document passed in, in which case we still need this.

packages/client/hmi-client/src/services/goLLM.ts
```diff
@@ -51,6 +51,33 @@ export async function interventionPolicyFromDocument(
...
+			ondata(data, closeConnection) {
+				if (data?.status === TaskStatus.Failed) {
+					closeConnection();
+					throw new FatalError('Task failed');
+				}
+				if (data.status === TaskStatus.Success) {
+					closeConnection();
+				}
+			}
+		});
```
jryu01, line 75: I think we can remove this code. Since, instead of creating connection for each task, we can subscribe/unsubscribe on global sse event channel whenever needed using `subscribe` in `ClientEventService.ts` or `useClientEvent.ts`

jryu01: And I believe the endpoint for subscribing for each individual task event got removed in the server side at some point and this code won't work.

PR #5028, opened 2024-10-03
Add dataset transformer output to node

blanchco:
# Description

* Add a single output to the datatransformer node, this can be extended to show multiple outputs in the future if needed.
* `on-selected` to response component to select the node
* `update-selected-output` to chat component to update which outputs are selected which are stored in the state.  These state outputs can then be accessed to be shown on the node

https://github.com/user-attachments/assets/62177ce7-b328-477c-b149-6481de1fc220


packages/client/hmi-client/src/components/llm/tera-beaker-code-cell-output.vue
YohannParis, line 1: Nice CSS job here!

packages/client/hmi-client/src/components/workflow/ops/dataset-transformer/tera-dataset-transformer-node.vue
```diff
@@ -3,7 +3,17 @@
...
-		<Button v-else label="Edit" @click="emit('open-drilldown')" severity="secondary" outlined />
+		<template v-if="node.inputs[0].value">
+			<template v-if="!isEmpty(node.state.selectedOutputs)">
+				<tera-beaker-code-cell-output
+					v-for="m in node.state.selectedOutputs?.[0]?.messages"
+					:key="m.header.msg_id"
+					:jupyter-message="m"
+					:feature-config="{ isPreview: true }"
+				/>
+			</template>
```
YohannParis, line 14:
```suggestion
				<tera-beaker-code-cell-output
				  v-if="!isEmpty(node.state.selectedOutputs)"
					v-for="m in node.state.selectedOutputs?.[0]?.messages"
					:key="m.header.msg_id"
					:jupyter-message="m"
					:feature-config="{ isPreview: true }"
				/>
```

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/workflow/ops/dataset-transformer/tera-dataset-transformer-node.vue
blanchco: I think using `v-if` and `v-for` on the same component is not recommended by Vue which is why I had the `template` tag there

YohannParis: Why not do `v-for="m in (node.state.selectedOutputs?.[0]?.messages ?? [])" then and avoid the v-if?

blanchco: that works, I changed this up in my latest commit

PR #5060, opened 2024-10-04
Update the channel where we are grabbing the llm thoughts.

Tom-Szendrey:
# Description
Beaker has been updated to have the thoughts section more based off of the llm thoughts and added a new section for the responses. This means when we ask questions like the screenshot below we were no longer getting the response, we were only getting the thoughts.

Before:
image

After:
image


YohannParis (approved):
Good job

shawnyama (approved):

PR #5067, opened 2024-10-04
Fixed the issue that error chart not showing up

jryu01:
# Description

Fixed the issue that error chart wasn't showing up in the calibrate operator drilldown.
The issue was caused because the dataset time column was moved from `mapping` (`CalibrateMap`) to `knobs.value.timestampColName` but `getErrorData` function wasn't updated accordingly. This change fixed the issue.

Resolves #(issue)


YohannParis (approved):

Tom-Szendrey (approved):
Thank you for catching this for me!

PR #5075, opened 2024-10-07
Fix sorting of extraction pages

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -832,7 +833,15 @@ public Future<EquationExtraction> extractEquationsFromPDF(
...
-			Collections.sort(keys);
+			Collections.sort(
+				keys,
+				new Comparator<String>() {
+					@Override
+					public int compare(final String s1, final String s2) {
+						return Integer.compare(Integer.parseInt(s1), Integer.parseInt(s2));
+					}
+				}
+			);
```
YohannParis, line 844: What is the difference with a lambda function? I'm a front-end, so I'm curious about the pattern here?
```java
Collections.sort(keys, (s1, s2) -> Integer.compare(Integer.parseInt(s1), Integer.parseInt(s2)));
```

YohannParis: CoPilot is writting even worse code than I would think:
```java
Collections.sort(keys, Comparator.comparingInt(Integer::parseInt));
```

kbirk: Because copilot told me so. I updated it to lambdas :P

YohannParis: I see we are both Java via CoPilot engineers!

dgauldie (approved):

PR #5077, opened 2024-10-07
Fix messaging around GoLLM tasks #5076

### LINKED ISSUE #5076: Fix messaging around GoLLM tasks

dvince2:
# Description

So it turns out that the events were already set up but we forgot to …remove the old task handler code. Removing that fixes the exceptions.

Resolves #5076


YohannParis (approved):
Nice clean up!

Tom-Szendrey (approved):
Did not test but do love it when a FIXME is fixed

PR #5079, opened 2024-10-07
5038 bug intervention policies are not described well in simulate

YohannParis:
* [`tera-intervention-summary-card.vue`](diffhunk://#diff-d43cc643432e06747638bedd33b38419a2e2bc847266871ce29b0b6c8712e3c1R1-R53): Moving it to its own folder
* [`tera-intervention-policy-card.vue`](diffhunk://#diff-ac3d676055fff5e0086667851c655ecb871d1440565272c366d86fa4170fbd40L2-L20): Improved the structure and styles, including the removal of unnecessary elements and the addition of a header section.
* [`tera-intervention-policy-drilldown.vue`](diffhunk://#diff-54c31526a779022f9231558c720e27ce9e02ed6da5df3ff490e94bd8bc891248L48-R48): Enhanced styles for better visual consistency and added padding to the `intervention-settings-section` class.

Screenshot 2024-10-07 at 15 23 08


packages/client/hmi-client/src/components/intervention-policy/tera-intervention-summary-card.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+	<div>
+		<aside>{{ !isEmpty(intervention.dynamicInterventions) ? 'Dynamic' : 'Static' }}</aside>
+		<h5>{{ intervention.name }}</h5>
+		<ul>
+			<li v-for="(staticIntervention, index) in intervention.staticInterventions" :key="`static-${index}`">
```
dgauldie, line 6: see comment

YohannParis:
> @liunelson , doesn't it make more sense for the static strings to read "set parameter beta to 0.6 starting at day 15" instead of "set parameter beta to 0.6 at start time 15 day"
> 
> also... is "day" reasonable? I guess once we move to dates and timestamps, we can change...

We use the words days, because that is what the user sees when creating those interventions, once we do the unit better, we can change that, but that is out of scope from this PR

YohannParis:
> ok, I still think we should change the format of the string... keeping day

Yes, that has been done already!

liunelson:
> @liunelson , doesn't it make more sense for the static strings to read "set parameter beta to 0.6 starting at day 15" instead of "set parameter beta to 0.6 at start time 15 day"
> 
> also... is "day" reasonable? I guess once we move to dates and timestamps, we can change...
> 

"day" is really just a placeholder or default value until the UI gets the time unit from the model AMR. With datetime, then the time unit is still relevant since it seems always helpful to see how much time has elapsed since the starting datetime.

PR #5087, opened 2024-10-08
Adding Output Console to drilldowns with notebooks

Tom-Szendrey:
# Description:
-Add new component 
-Utilize this new component in drilldowns that contain a notebook.
-Update tera-resize component to allow for resizing from top rather than just the bottom

# Testing:
write some code that includes print statements or uses the logger
Here is an sample:
```
import logging

def sample(aString):
     logging.basicConfig()
     logger = logging.getLogger()
     try:
        print(aString[3])
     except Exception as e:
       logger.warning(f"break here")

sample("test") # -> This will pass fine
sample("t") # -> This will have execute_reply as "ok" but will also have the stderr from the logger 
```
You can also use a more realistic example of running the new rate law code with an invalid rate law:
```
import sympy

def replace_rate_law_sympy(model, template_name, new_rate_law):
    """Replace the rate law of transition. The new rate law passed in will be a sympy.Expr object

    Parameters
    ----------
    template_name :
        The ID of the transition whose rate law is to be replaced, this is
        typically the name of the transition
    new_rate_law :
        The new rate law to replace the existing rate law with

    Returns
    -------
    :
        The updated model as an AMR JSON
    """
    assert isinstance(model, TemplateModel)
    tm = model
    local_dict = { key:sympy.Symbol(key) for key, _value in tm.parameters.items() }
    for k, v in tm.get_concepts_name_map().items():
        local_dict[k] = sympy.Symbol(k)

    for template in tm.templates:
        if template.name == template_name:
            template.set_rate_law(new_rate_law, local_dict=local_dict)
    return tm

# Transmission by direct (fluid) contact
model = replace_rate_law_sympy(
    model, 
    "inf", 
    "i * (S))"
)
```

# Screenshots:
image

image

image


YohannParis (approved):
🎉 Great work!

asylves1 (approved):

PR #5097, opened 2024-10-08
Fix model 500s on main page

kbirk:


packages/client/hmi-client/src/services/model.ts
```diff
@@ -38,8 +38,10 @@ export async function createModelAndModelConfig(file: File, progress?: Ref<numbe
  * Get Model from the data service
  * @return Model|null - the model, or null if none returned by API
  */
-export async function getModel(modelId: string): Promise<Model | null> {
-	const response = await API.get(`/models/${modelId}`);
+export async function getModel(modelId: string, projectId?: string): Promise<Model | null> {
+	const response = await API.get(`/models/${modelId}`, {
+		params: { 'project-id': projectId }
```
YohannParis, line 43: This should be passed on by api.ts, why adding it here?

YohannParis (approved):
Those registerHandler need to be redone, why we need to fetch the whole asset just for a name that can be given by the back-end on the event.data.

packages/client/hmi-client/src/services/model.ts
kbirk: Because there is no project open when this gets called from notifications in the main page.

PR #5104, opened 2024-10-09
Added custom number formatting to better format the numbers on the chart

### LINKED ISSUE #4947: [BUG]: simulation chart y axis does not handle small numbers very well
- Scenario: https://github.com/DARPA-ASKEM/terarium/blob/main/testing/manual/intervention-policy.md
- Step: 9
- Link to the project: https://app.staging.terarium.ai/projects/59d1af73-e373-42da-9e88-ac020a39cc1f/workflow/8f52229b-276a-474a-a534-f30d3c409f54?operator=aa816360-051e-41cf-9089-7c251b5524dd

**Describe the issue**
Chart y axis can't handle small numbers very well. Here beta starts at 0.0006 and it changed to 0.0001 at timestep 20...
image

jryu01:
# Description

- Added Vega chart custom formatting function to better format small numbers. 
- Added and leverage an utility function to fix javascript floating point precision errors when applying formatting or showing display number. 
- All Vega charts will inherit this formatting by default but it can be also overriden by individual chart spec. 
- Added tests

**Before: (lookt at the numbers on the y axis)**
Screenshot 2024-10-09 at 12 33 54 PM

**After**
Screenshot 2024-10-09 at 11 44 17 AM

**Other Examples**
Y Axis with long numbers with many 0s
Screenshot 2024-10-09 at 12 04 38 PM

Formatting with numbers > 1 (this is same as existing behaviour)
Screenshot 2024-10-09 at 11 44 29 AM

Resolves #4947


packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -21,15 +21,39 @@
...
+
+// Define the custom expression functions that can be registered and used in the Vega charts
+const expressionFunctions = {
+	// chartNumberFormatter is a custom number format that will display numbers in a more readable format
+	chartNumberFormatter: (value: number) => {
+		const correctedValue = fixPrecisionError(value);
+		if (value > -1 && value < 1) {
+			return countDigits(correctedValue) > 6 ? correctedValue.toExponential(3) : correctedValue.toString();
+		}
+		return format(NUMBER_FORMAT)(correctedValue);
```
YohannParis, line 41: Why not using the number util? You spent all that time improving it, and not using it at the end.

YohannParis (approved):

packages/client/hmi-client/src/components/widgets/VegaChart.vue
jryu01: I think existing `displayMumber` util isn't the best fit for display number for the charts especially for the axis since it either display long numbers as exponential or numbers with spaces. 

For numbers > 1 or < -1, I think it's better to stick with current d3 formatter for the charts (e.g we want 100K instead of 1 00 000 or 1M instead of 1.000e+6). 

For number like `0.00001`, I think it's better to display as it is instead of `0.000 01`  that you would get with the existing function. 

Also existing function didn't handle the precession error, and you would get axiss like 0.1 0.2 '3.000e-1' 0.4, without correcting precision error. 

So I implemented some enhancements to the existing chart and also used custom function for the charts for better numbers.

YohannParis: Thank you for the detailled answer!

Tom-Szendrey (approved):
sweet nice fix

PR #5105, opened 2024-10-09
feat: funman graph improvements

shawnyama:
# Description

- Prevent constants from being picked in parameters of interest
- Only show parameter graphs of interest
- Add ambiguous boxes
- The latest timepoint should be shown not the furthest one specified
   - By default I am showing all boxes and there is a checkbox to just see the latest one
- Add option to prioritize the view of the model check

image


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
```diff
@@ -9,12 +9,20 @@
...
 	</header>
+
 	<Accordion multiple :active-index="[0, 1, 2, 3]">
 		<AccordionTab header="Summary"> Summary text </AccordionTab>
 		<AccordionTab>
 			<template #header> State variables<i class="pi pi-info-circle" /> </template>
+			<!--TODO: Will put these checkbox options in output settings later-->
+			<div class="flex align-items-center gap-2 ml-4 mb-3">
+				<Checkbox v-model="onlyShowLatestBox" binary @change="renderCharts" /><label>Only show latest box</label>
+			</div>
```
mwdchang, line 20: I think the language here needs a bit of tweaking? I don't see anywhere where we explained what `latest box` means. 

Also this slotted in the "state variables" section, but the effect of showing all boxes also modifies the parameter's section.

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -900,11 +912,12 @@ export function createFunmanStateChart(
...
 }
 
-export function createFunmanParameterChart(
+export function createFunmanParameterCharts(
 	parametersOfInterest: { label: 'all'; name: string; interval: FunmanInterval }[],
 	boxes: FunmanBox[]
 ) {
-	const parameterRanges: { parameterId: string; boundType: string; lb?: number; ub?: number }[] = [];
+	const parameterRanges: { parameterId: string; boundType: string; lb?: number; ub?: number; tick?: number }[] = [];
+	const parameterIdsOfInterest: string[] = [];
```
mwdchang, line 920: What is the motivation for this? we could run Funman without any parameter of interests (eg just a bare-bone non-negativity check) and it should still return some parameter ranges, wouldn't we want to show those?

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
shawnyama: Look at `getBoxesEndingAtLatestTimestep()` in `funman-service.ts`. This just shows the lines that make it to the furthest timestep.

It's slotted in the state variables section for now, I was thinking I would add these checkboxes to an Output Settings slider in another pass

packages/client/hmi-client/src/services/charts.ts
shawnyama: I just thought the point of the parameters of interest dropdown was to show the ones we want to see. It makes sense to show them all if none are selected though, so I'll do that. I am wondering if we should just show them all now, regardless of what was selected.

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
mwdchang: Maybe my first point wasn't clear, I meant since we never introduced the term "box" anywhere (that I know of), having a checkbox "only show latest box" is at best mysterious and at worst confusing. 

It is also conceivably possible to have multiple boxes at furtherest timestamp. For example at t=150 there are 2 true boxes where beta > 0.4 and beta > 0.2, and there is a false box for beta < 0.2.

So I dunno if your parsing logic needs to change, but "show only the last box" reads weird and can use a bit of rephrasing. How about "show only results with furtherest progress" or something akin to that?

shawnyama: I changed the label. Parsing logic accounted for multiple so that's fine.

packages/client/hmi-client/src/services/charts.ts
shawnyama: Now I am just filtering out the constants, since Pascale and Nelson just want to see the distributions

mwdchang (approved):
let's go!

PR #5110, opened 2024-10-09
5059 bug unable to see intervention in output chart for calibrate

### LINKED ISSUE #5059: [BUG]: Unable to see Intervention in Output Chart for Calibrate
Calibrated with an intervention and my intervention for beta at time 10 is not displayed in the output chart. 

* It's not displayed in the output panel by default.
* It's impossible to try to display, because the input panel is scrolling up to the top when clicking on the button to do so.
* If selected to be displayed on the node, it is not.

https://app.staging.terarium.ai/projects/9440f1c2-0c17-4692-8b71-5781ecadb665/workflow/21d9786f-1251-47dc-8499-8aa5cfe3cb2f?operator=3ff458c2-ba0d-4863-857f-59aaff07849d

mloppie:
# Fixes missing intervention marker in the calibrate node and calibrate drilldown

Screenshot 2024-10-09 153541
Screenshot 2024-10-09 153532

Resolves #5059


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -749,7 +750,7 @@ export function createInterventionChartMarkers(
 			type: 'text',
 			align: 'left',
 			angle: 90,
-			dx: 5,
+			dx: labelXOffset,
 			dy: -10
```
YohannParis, line 754: Why only `dx` deserves a constant?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -210,8 +210,11 @@ const preparedCharts = computed(() => {
 			}
 		);
 		applyForecastChartAnnotations(chart, annotations);
-		chart.layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
-
+		Object.entries(groupedInterventionOutputs.value).forEach((item) => {
+			item[1].forEach((intervention) => {
+				chart.layer.push(...createInterventionChartMarkers([intervention], false, -80));
```
YohannParis, line 215: Why does the node chart uses `-80` and not `-180` like the drilldown?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -792,7 +792,11 @@ const preparedCharts = computed(() => {
 			annotations
 		);
 
-		charts[variable].layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
+		Object.entries(groupedInterventionOutputs.value).forEach((item) => {
+			item[1].forEach((intervention) => {
+				charts[variable].layer.push(...createInterventionChartMarkers([intervention], false, -180));
+			});
+		});
 	});
```
jryu01, line 800: It looks like we are applying all the interventions to the chart. We are going to end up plotting plotting interventions for variables like I, S, and R all on the same chart. It will create a issue such that chart for variable `I`  also have intervention for other variables, for example `S` or `R` along with all interventions for parameters as well. I'm not sure if that's what we want to do.  If we want still draw all interventions for parameters for the chart, we should at least filter out the intervention for state variables (e.g. I, S, R etc) I guess.

packages/client/hmi-client/src/services/charts.ts
mloppie: It's a param, there's nowhere in the code that changes `dy`

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
YohannParis: I think this is something @pascaleproulx needs to figure out with the design team. So until they do this, that PR can be a step forward, and we should merge it as a first version.

jryu01: Then I think we should include the interventions for the the same variable if it's for state variable, and all for parameters and filter out the intervention for other state variables. For example chart for `I` should include interventions for `I` only and plus the all interventions for the parameters (e.g `beta`).  @mloppie you can probably use `modelPartTypesMap` in line 590 to check the type to see if a variable is state variable or a parameter.

jryu01: But I still think that without proper design how we want to show parameter intervention on the state chart, just plotting all parameter interventions markers as vertical lines on the state chart won't be that useful. Parameter intervention were intentionally left out I believe. And this is going to be reworked depending on how we want to do it properly. I feel it's wise to pause until we get better idea how we should draw them on the chart from Pascale.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -782,9 +796,42 @@ const preparedCharts = computed(() => {
 			),
 			annotations
 		);
-
-		charts[variable].layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
 	});
+
```
Tom-Szendrey, line 800:
```suggestion
// Intervention Charts:
```

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -782,9 +796,42 @@ const preparedCharts = computed(() => {
 			),
 			annotations
 		);
-
-		charts[variable].layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
```
jryu01, line 786: Why do we delete this line? Do we not show the variable intervention on the variable chart anymore?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -210,8 +210,11 @@ const preparedCharts = computed(() => {
 			}
 		);
 		applyForecastChartAnnotations(chart, annotations);
-		chart.layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
-
+		Object.entries(groupedInterventionOutputs.value).forEach((item) => {
+			item[1].forEach((intervention) => {
+				chart.layer.push(...createInterventionChartMarkers([intervention], false));
+			});
+		});
```
jryu01, line 217: This is going to end up drawing all the interventions (variables + parameters) for each chart. 
```suggestion
chart.layer.push(...createInterventionChartMarkers(groupedInterventionOutputs.value[variable]));
```
Don't we need to revert back to the previous code like ^?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
mloppie: I suppose we could leave it even though it's duplicated ion the intervention chart.

mloppie:
After discussion this PR now adds an intervention charts section to the drilldown
Screenshot 2024-10-21 at 12 16 10 PM

jryu01:
Thanks! LGTM

jryu01 (approved):

PR #5145, opened 2024-10-11
5070 create new gollm endpoint that takes in a list of latex equations a returns cleaned up ones

### LINKED ISSUE #5070: create new gollm endpoint that takes in a list of latex equations a returns cleaned up ones

dgauldie:
# Description

Resolves #5070 

image


packages/gollm/gollm_openai/tool_utils.py
```diff
@@ -59,6 +60,47 @@ def get_image_format_string(image_format: str) -> str:
     return format_strings.get(image_format.lower())
```
YohannParis, line 60: This `return dict.get()` look like it will return a string based on `format_strings` format.
Why change the return type to be a `dict`?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -236,9 +246,47 @@ public ResponseEntity<UUID> equationsToModel(
...
+			throw new ResponseStatusException(HttpStatus.SERVICE_UNAVAILABLE, messages.get("task.gollm.timeout"));
+		} catch (final InterruptedException e) {
+			log.warn("Interrupted while waiting for task response", e);
+			throw new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, messages.get("task.gollm.interrupted"));
+		} catch (final ExecutionException e) {
+			log.error("Error while waiting for task response", e);
+			throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, messages.get("task.gollm.execution-failure"));
+		}
+
+		// create a clone of the request and set the cleaned up equations
```
YohannParis, line 274: Why do we need a clone here?

YohannParis (approved):
LGTM

packages/gollm/gollm_openai/tool_utils.py
dgauldie: good catch. I didn't want to change this one

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
dgauldie: so that I don't modify the original request.

PR #5157, opened 2024-10-15
additional heuristic for template parsing handling multiple outputs

mwdchang:
### Summary
This address an incorrect assumption we made earlier that a transition's signature/key can be calculated from its input and output component, for the purpose of calculating transition groups. This is not the case, e.g we can have this scenario:

```
transition1_xx: [ input: [a], output: [b] }
transition2_yy: [ input: [a], output: [b] }
```

Where we would want to keep these transitions as different groups. To do this we add the base transition name as part of the key.

### Testing
The model 
[pascale-strata-problem-templates.json](https://github.com/user-attachments/files/17383266/pascale-strata-problem-templates.json), when loaded into Terarium should look like

image

Instead of:

image

Note the correct version should have 4 transition matrices instead of 3.

### DIFF
--- a/packages/client/hmi-client/src/model-representation/mira/mira.ts
+++ b/packages/client/hmi-client/src/model-representation/mira/mira.ts
@@ -230,6 +230,12 @@ export const rawTemplatesSummary = (miraModel: MiraModel) => {
 	return allTemplates;
 };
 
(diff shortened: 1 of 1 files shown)


asylves1 (approved):

YohannParis (approved):
LGTM

PR #5159, opened 2024-10-16
Intervention validations

Tom-Szendrey:
# Purpose:
- Prevent users from creating invalid intervention policies and provide simple toast to them for their understanding what is disallowed
- An invalid intervention policy is one that has a static intervention with a negative timestamp, or any duplicate appliedTo + time pairs in static interventions

# Description
- add a validation check for interventions. This will throw an error when:
     - There is a negative timestamp found in a static intervention.
     - There are two static interventions that are duplicated within a single intervention
     - There are two static interventions that are duplicated within different interventions within a single policy
- added some simple test cases for the validations
- Allow for user to delete interventions via the sidebar
- Add the confirm back into app.vue so we can utilize primevue's useConfirm -> Yohann can clean this later
   
# Follow up:
(I want to reduce how big the PR is for readability)
Will remove the menu in the side panel and add a trash symbol.
When user clicks on a card it will select it
When user clicks on the trash it will delete

# Pictures:
image

image

image

image


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
```diff
@@ -32,4 +34,36 @@ public Intervention clone() {
 		}
 		return intervention;
 	}
+
+	// Check for no negative timesteps
+	// Check for no duplicate time + appliedTo pairs in static interventions
+	public Boolean validateIntervention() throws Exception {
```
dvince2, line 40: I think that these `validate` methods on `Intervention` and `InterventionPolicy` should possibly exist in `InterventionController` or `InterventionService` since it doesn't really have anything to do with the model class.

Also, this method is only returning `true` or throwing an exception... it can never be `false` not valid.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
```diff
@@ -32,4 +34,36 @@ public Intervention clone() {
...
+			final Number time = this.staticInterventions.get(i).getTimestep();
+			if (time.doubleValue() < 0) {
+				final String errorMessage = String.format(
+					"The intervention %s has a timestep %s which is less than 0.",
+					this.getName(),
+					time.toString()
+				);
+				throw new Exception(errorMessage);
+			}
+			for (int j = 0; j < this.staticInterventions.size(); j++) {
```
dvince2, line 51: Right now this validation is O(n^4).  I think you can simplify it a bit doing something like...

```
if(staticInterventions.size() > 1){
   for(int i = 1; i < staticInterventions.size(); i++){
      ...
      	final String appliedToOne = this.staticInterventions.get(i).getAppliedTo();
				final Number timeOne = this.staticInterventions.get(i).getTimestep();
				final String appliedToTwo = this.staticInterventions.get(i-1).getAppliedTo();
				final Number timeTwo = this.staticInterventions.get(i-1).getTimestep();
   }
}
```

This would let you get rid of your `if` statement as well because `appliedToTwo` would never be `appliedToOne`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionPolicy.java
```diff
@@ -40,4 +42,45 @@ public InterventionPolicy clone() {
 		}
 		return clone;
 	}
+
+	// Check each intervention this policy contains
+	// If any of them are are invalid the entire policy is invalid.
+	public Boolean validateInterventionPolicy() throws Exception {
+		for (int i = 0; i < this.interventions.size(); i++) {
```
dvince2, line 49: Similarly here, if you start at 1 for `i` you can drop the entire second for loop and use index `i-1`

dvince2 (dismissed):

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionPolicy.java
Tom-Szendrey: Every intervention needs to be checked with each other. For example the following should fail as the first and last both contain "A"
Intervention One: [A, B, C]
Intervention Two: [D, E, F]
Intervention Three[A, G, H]

Here is a picture of what this looks like with the current UI
image

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
```diff
@@ -32,4 +35,43 @@ public Intervention clone() {
...
+
+	// Check for no negative timesteps
+	// Check for no duplicate time + appliedTo pairs in static interventions
+	public Boolean validateIntervention() throws Exception {
+		//Sort static list
+		Collections.sort(this.staticInterventions, (staticOne, staticTwo) -> {
+			final String stringOne = staticOne.getAppliedTo() + staticOne.getTimestep().toString();
+			final String stringTwo = staticTwo.getAppliedTo() + staticTwo.getTimestep().toString();
+			return stringOne.compareTo(stringTwo);
+		});
```
mwdchang, line 47: What is the desired sorting order here? Cause you are sorting strings with a numerical component and it rarely does what you want.

Tom-Szendrey: Did not mean to push these changes yet! still playing with this idea

The idea being that if we sort first we can compare `i` and `i - 1` rather than `i` and `j`
Sorting here would mean that we put any duplicate pair beside eachother 
Duplicate pair means appliedTo and timestep are identical

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
```diff
@@ -32,4 +35,43 @@ public Intervention clone() {
...
+					this.getName(),
+					time.toString()
+				);
+				throw new Exception(errorMessage);
+			}
+			if (i > 0) {
+				final String appliedToOne = this.staticInterventions.get(i).getAppliedTo();
+				final Number timeOne = this.staticInterventions.get(i).getTimestep();
+				final String appliedToTwo = this.staticInterventions.get(i - 1).getAppliedTo();
+				final Number timeTwo = this.staticInterventions.get(i - 1).getTimestep();
```
mwdchang, line 63: Generally speaking, it is more common to see people do 

```
for (int i = 0; i < this.staticInterventions.size() - 1; i++) {
```

Rather than looping the entire list and have a special if-clause to check consecutive elements.

Also generally speaking if you want to check for duplicates, people tend to just use Map or Set instead of this kind of iteration.

```
final Set<String> blah = new HashSet();
for (int i = 0; i < this.staticInterventions.size(); i++) {
  final String key = ....
  if (blah.contains(key)) {
    // duplicate
  } else {
    blah.add(key)
  } 
```

Tom-Szendrey: Good idea.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionPolicy.java
```diff
@@ -40,4 +42,45 @@ public InterventionPolicy clone() {
...
+					validateInterventionPair(this.interventions.get(i), this.interventions.get(j)); //validate pair with eachother
+				}
+			}
+		}
+		return true;
+	}
+
+	// Takes a list of two interventions and will check their static intervention lists to ensure
+	// there are no duplicate time + appliedTo pairs in their indiviudal static interventions
+	private Boolean validateInterventionPair(Intervention interOne, Intervention interTwo) throws Exception {
```
mwdchang, line 62: This seem to suggest you want either
- encode uniqueness through overriding toString or hasCode functions, or
- implement the Comparable interface and use compareTo(...)

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
mwdchang: I would just use Set or Map for duplicate checks, then it is order-invariant and you save yourself some computation-cycles as well. You are saving a bit of resources in terms of space-complexity doing it your way, but it is largely negligible in the grand scheme of things and it increases computational complexity.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/Intervention.java
Tom-Szendrey: Thanks again for this!

YohannParis: You do not even need to check for duplicate, just add into the Set, it will take care of being only one of each.

Tom-Szendrey: I am checking here so that if we do find a duplicate we can propagate that error up to the user
If youre suggesting to change the datastructure of intervention to be a set instead of a list i think that will lead to confusion when the user tries to save the following:

image

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionPolicy.java
```diff
@@ -40,4 +42,44 @@ public InterventionPolicy clone() {
...
+					final String errorMessage = String.format(
+						"The intervention %s has a timestep %s which is less than 0.",
+						this.getName(),
+						time.toString()
+					);
+					throw new Exception(errorMessage);
+				}
+				// Check for duplicate appliedTo timestep pairs:
+				final String key = staticIntervention.getAppliedTo() + staticIntervention.getTimestep().toString();
+				if (foo.contains(key)) {
```
mwdchang, line 71: You may want a delimiter here, in the event that someone comes along and create a model and interventions like
- beta1 should be 0.5 at time 5 => beta15
- beta should be 0.2 at time 15 => beta15
Which would be invalid in your scheme I thihnk.

Also please pick a better name than "foo" :)

Tom-Szendrey: foo to `duplicateCheckSet` 
key to include `atTheTime:` in the middle of appliedTo and timestep

mwdchang (approved):

PR #5162, opened 2024-10-16
4754 notebook re run notifications

### LINKED ISSUE #4754: Notebook re-run notifications
### Issue 1: Add 'Re-run all cells' button and notification
**CURRENT**
When you open a previously generated **Transform dataset** operator, it loads the input and output text of the previous run, and it indicates that the kernel is "Ready", but none of the transformations are actually available to use. 

Image

**PROPOSED**
There should be an indicator to tell users that the cells need to be re-run if they want to make any changes or use anything the operator generated downstream.

There should also be a "Re-run all cells" button on the toolbar.

It would be nice if you can make the page look more like this:
Image

mloppie:
# Description

* Adds a "Rerun all cells" button to the transform dataset drilldown. 
* Adds a message explaining that cells need to be rerun to restore the context.
* Disables "Save for reuse" when the status is invalid or error

Screenshot 2024-10-16 at 11 44 12 AM

Resolves #4754


packages/client/hmi-client/src/components/llm/tera-jupyter-chat.vue
```diff
@@ -8,6 +8,10 @@
 			tabindex="0"
 			class="message-container"
 		>
+			<div v-if="showRerunMessage" class="rerun-message">
+				Re-run all the cells to restore the context if you need to make any changes or use them downstream.
+				<Button class="close-mask" icon="pi pi-times" text rounded aria-label="Close" @click="hideRerunMessage" />
+			</div>
```
YohannParis, line 14: It will be fun if clicking on the banner did the same thing.
```suggestion
			<div v-if="showRerunMessage" class="rerun-message"  @click="window.dispatchEvent(new Event(EventType.RunAllCells))">
				Re-run all the cells to restore the context if you need to make any changes or use them downstream.
				<Button class="close-mask" icon="pi pi-times" text rounded aria-label="Close" @click.stop="hideRerunMessage" />
			</div>
```

packages/client/hmi-client/src/components/llm/tera-jupyter-chat.vue
```diff
@@ -489,4 +501,11 @@ section {
 	height: calc(100% - 3.5rem);
 	overflow-y: auto;
 }
+.rerun-message {
+	display: flex;
+	background-color: var(--surface-warning);
+	justify-content: space-between;
+	align-items: center;
+	padding: 0.5rem;
```
YohannParis, line 509: Here a list of all our CSS variables: https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/assets/css/theme/_variables.scss#L981
```suggestion
	padding: var(--gap-2);
```

YohannParis (approved):

PR #5163, opened 2024-10-16
feat: funman parameter graph interactions

shawnyama:
# Description

* Clicking on a parameter tick has you view its bounds and status. Corresponding ticks in other parameter charts activate as well
* ~I set up the logic for highlighting a specific line in the state graph that would use a `boxId`. However getting the `boxId` I want from the parameter graph is what I am missing. So that needs to be handled later. I am stuck on emitting the `boxId` out of the parameter graph, I posted a stackoverflow question to a vega-lite slack channel (since that's how they wanted it): https://stackoverflow.com/questions/79094839/trying-to-get-scoped-param-value-out-of-vega-lite-spec~
* Using the `chart-click` emit to pass the parameter chart's `boxId` to highlight the state chart's corresponding line.

https://github.com/user-attachments/assets/6f07c525-1077-410b-b068-18f38406f69a


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-output.vue
```diff
@@ -30,7 +30,11 @@
 		</AccordionTab>
 		<AccordionTab>
 			<template #header>Parameters<i class="pi pi-info-circle" /></template>
-			<vega-chart :visualization-spec="parameterCharts" :are-embed-actions-visible="false" />
+			<vega-chart
+				:visualization-spec="parameterCharts"
+				:are-embed-actions-visible="false"
+				@chart-click="onParameterChartClick"
```
mwdchang, line 36: Can you explain a bit how the visual encodings work with respect to click on the parameters' rug plot? I played around with it and it is confusing what had been selected.

mwdchang (changes requested):
Probably not directly related to this PR, but since you are working in the area: There is something strange with how you are collecting the trajectories for funman results. Notice below you have several markers on the same x-axis - this means it is no longer a function of time and is therefore incorrect.

image

See: https://app.staging.terarium.ai/projects/33f364c1-1da2-4cf7-a176-9870ef1a3ab6/workflow/25bde950-a3c3-41b1-bf32-bfee8add72ae?operator=648f4cb2-ff15-4c01-907a-4b1b4e241dbc

shawnyama:
Added white borders to the ticks since they could be hard to see in these cases 
image

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -847,6 +847,8 @@ export function createFunmanStateChart(
 		endY: focusOnModelChecks ? c.additive_bounds.ub : Math.min(c.additive_bounds.ub ?? maxY, maxY)
 	}));
 
+	console.log(trajectories);
```
YohannParis:
```suggestion
```

mwdchang (approved):
Thanks, as discussed we probably want to revisit the parameter chart, making it larger and a bit more obvious what are the interactions and their respective responses. This is a good start.

PR #5173, opened 2024-10-16
Model config pdf viewer

asylves1:
# Description

Design Change:
- Add Pdf viewer to the model config

image

Resolves #(issue)


packages/client/hmi-client/src/components/widgets/tera-pdf-panel.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+	<div class="container">
```
YohannParis, line 2: Is the div.container needed here?

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -7,6 +7,18 @@
...
 		<template #sidebar>
+			<tera-slider-panel
+				v-if="pdfData"
+				v-model:is-open="isDocViewerOpen"
+				header="Document Viewer"
+				content-width="700px"
+			>
+				<template #content>
+					<tera-drilldown-section :is-loading="isFetchingPDF">
+						<tera-pdf-panel :pdfs="pdfData" />
```
blanchco, line 18: should we hide the pdf panel when there is no attached document to the model config node?
Screenshot 2024-10-17 at 10 48 18 AM

PR #5174, opened 2024-10-16
prevent rendering if data dependencies are not ready

mwdchang:
### Summary
It looks like with the resize-observer getting kicked off right away, it is possible to render a graph without the correct dependencies and this eventually result in JS errors that partially break the UI. This is a quick-and-dirty fix to add a sanity check to exit the rendering function if things are not in place.

### Testing
Adding a model onto the cavas should not result in any vue-warnings or js-errors


Tom-Szendrey (approved):

YohannParis (approved):
LGTM

Issue #5161, opened 2024-10-16
[TASK]: Sympy equation parsing issues


mwdchang:
Sympy has issues evaluating and/or parsing certain tokens because they are reserved keywords and/or create conflicts.

The known tokens are (case sensitive):
- lambda
- S 

And follow suit on some of the MIRA special parsing logic for dealing with AMRs:
- https://github.com/gyorilab/mira/blob/bfb0975a59aa05f84bb1982f8c9ce71ef68d7057/mira/metamodel/utils.py

PR #5178, opened 2024-10-17
Sympy parsing workaround for specific tokens

mwdchang:
### Summary
When rate-laws contain specific tokens sympy parsing tend to break as they are a part of reserved symbols. To parse them properly we encode them to something else, then revert them back when we need the parsing result.

So far, we have
- lambda
- S
- Ci

For example:

```
>>> parse_expr("A")
A
>>> parse_expr("A + 1")
A + 1
>>> parse_expr("S + 1")
ValueError: Error from parse_expr with transformed code: 'S +Integer (1 )'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dchang/venv/3.11/lib/python3.11/site-packages/sympy/parsing/sympy_parser.py", line 1087, in parse_expr
    raise e from ValueError(f"Error from parse_expr with transformed code: {code!r}")
  File "/Users/dchang/venv/3.11/lib/python3.11/site-packages/sympy/parsing/sympy_parser.py", line 1078, in parse_expr
    rv = eval_expr(code, local_dict, global_dict)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dchang/venv/3.11/lib/python3.11/site-packages/sympy/parsing/sympy_parser.py", line 906, in eval_expr
    expr = eval(
           ^^^^^
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'SingletonRegistry' and 'One'
```

### Testing
Import this model 
[pascale-strata-problem-templates.json](https://github.com/user-attachments/files/17437446/pascale-strata-problem-templates.json)

You should be able to view it, open up parameter matrices, open up template matrices without causing JS errors.

Alternatively you can mess around in the engineering interface (http://localhost:8080/pyodide-test) ... note there are JS errors here because it does partial evaluations as you type, but the output (eval, symbols...etc) should make sense.

Closes [#5161](https://github.com/DARPA-ASKEM/terarium/issues/5161)
Closes https://github.com/DARPA-ASKEM/terarium/issues/5185


YohannParis (approved):
LGTM

Tom-Szendrey (approved):

blanchco (approved):
seems to work well

PR #5181, opened 2024-10-17
Apply metadata from the pre-MIRA model to the post-MIRA model if the name of variables matches

kbirk:
#5094

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ModelController.java
```diff
@@ -190,7 +191,8 @@ ResponseEntity<Model> getModel(
 								new TypeReference<>() {}
 							);
 
-							// Append the Document extractions to the Model extractions, just for the front-end.
+							// Append the Document extractions to the Model extractions, just for the
+							// front-end.
```
YohannParis, line 195: @dgauldie Do we still do this? Maybe in another PR when we remove SKEMA-TR

YohannParis (approved):
LGTM

mwdchang (approved):
Didn't test, code looks okay.

PR #5188, opened 2024-10-17
Add better handling for partial failures. Remove cosmos

kbirk:


YohannParis (approved):
Thanks for the cleanup

PR #5203, opened 2024-10-18
fix(model-edit): only mutate input model, not adding on to output model

shawnyama:
# Description

* Notebook code only edits the original model, instead of editing the last output model

https://github.com/user-attachments/assets/7b57a62c-8eab-460d-9499-cd73c42b7f7b


YohannParis (approved):

Tom-Szendrey (approved):
Thanks for updating these var names!
wasnt able to test atm sorry

PR #5207, opened 2024-10-18
adding initial dates to model config pt1

blanchco:
# Description
Adding initial dates to model configuration to be used with timepoints, part 1/4

https://github.com/user-attachments/assets/71196d8b-3ce8-448c-a398-12e8be768b21

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -18,6 +18,19 @@
 			<!--amr_to_mmt doesn't like unit expressions with spaces, removing them here before they are saved to the amr-->
 			<template v-else-if="showUnit">
 				<template v-if="featureConfig.isPreview"><label>Unit</label>{{ item.unitExpression }}</template>
+				<!-- we use a dropdown for units with time semantic-->
+				<Dropdown
+					v-else-if="isTimePart"
+					:model-value="item.unitExpression"
+					option-label="label"
+					option-value="value"
+					:options="[
+						{ label: 'Days', value: 'days' },
+						{ label: 'Months', value: 'months' },
+						{ label: 'Years', value: 'years' }
+					]"
+					@change="$emit('update-item', { key: 'unitExpression', value: $event.value })"
+				/>
 				<tera-input-text
 					v-else
 					label="Unit"
@@ -80,7 +93,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, computed, watch } from 'vue';
+import { ref, computed, watch, inject } from 'vue';
 import TeraInputText from '@/components/widgets/tera-input-text.vue';
 import AutoComplete from 'primevue/autocomplete';
 import Button from 'primevue/button';
@@ -89,6 +102,7 @@ import { stringToLatexExpression } from '@/services/model';
 import type { DKG } from '@/types/Types';
 import { getCurieFromGroundingIdentifier, getNameOfCurieCached, searchCuriesEntities } from '@/services/concept';
 import type { FeatureConfig } from '@/types/common';
+import Dropdown from 'primevue/dropdown';
 
 const props = defineProps<{
 	item: ModelPartItem;
@@ -96,6 +110,7 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits(['update-item']);
+const isTimePart = inject('isTimePart');
 
 const query = ref('');
 const results = ref<DKG[]>([]);
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -93,7 +93,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, computed, watch, inject } from 'vue';
+import { ref, computed, watch } from 'vue';
 import TeraInputText from '@/components/widgets/tera-input-text.vue';
 import AutoComplete from 'primevue/autocomplete';
 import Button from 'primevue/button';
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -89,13 +102,15 @@ import { stringToLatexExpression } from '@/services/model';
...
 import type { FeatureConfig } from '@/types/common';
+import Dropdown from 'primevue/dropdown';
 
 const props = defineProps<{
 	item: ModelPartItem;
 	featureConfig: FeatureConfig;
 }>();
 
 const emit = defineEmits(['update-item']);
+const isTimePart = inject('isTimePart');
```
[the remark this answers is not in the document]
blanchco: Good to know, I can use props instead. I just didnt want to pass it all the way down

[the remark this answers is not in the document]
mwdchang: Well ... I think you can tell whether it is a time field by parsing the "ModelPartItem" since it is a pre-agreed upon kind of thing.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
```diff
@@ -28,6 +29,9 @@ public class ModelConfiguration extends TerariumAsset {
 	@TSOptional
 	private UUID simulationId;
 
+	@TSOptional
+	private Date temporalContextDate;
```
mwdchang, line 33: Note java.util.date has a lot of things deprecated and considered somewhat obsolete ... what are you planning to do with it?

blanchco: Im storing the date the user wants to set timestep 0 at on the ModelConfiguration.  If they want April 10th, 2024 at timestep 0 they select that and it is stored on the model configuration.  Maybe I should use java.sql.Timestamp instead? We seem to use that a lot

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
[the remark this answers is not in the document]
blanchco: The `ModelPartItem` itself doesn't have anything uniquely identifying on it to determine what semantic type (parameter, observable, time) it is from what I can see.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
mwdchang: yeah I think probably "java.sql.Timestamp"

dvince2: For what its worth `java.sql.Timestamp` is just a convenience wrapper around `java.util.date` and a few versions back Java basically incorperated JodaTime's library into a new `java.time` package.... but thats just a fun fact for our next pub trivia ;)

`Timestamp` is used everywhere else, yeah may as well use it

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
```diff
@@ -28,6 +29,9 @@ public class ModelConfiguration extends TerariumAsset {
 	@TSOptional
 	private UUID simulationId;
 
+	@TSOptional
+	private Timestamp temporalContextDate;
```
YohannParis, line 33:
```suggestion
	private Timestamp temporalContext;
```

Date means a full date, but with a timestamp you could uses seconds as well. If you see what I mean?

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -503,6 +523,30 @@ const mmtParams = ref<MiraTemplateParams>({});
 
 const configuredMmt = ref(makeConfiguredMMT(mmt.value, knobs.value.transientModelConfig));
 
+const calendarSettings = computed(() => {
```
YohannParis, line 526: I think this should be on its own component tera-calendar, as it can be used somewhere else.
But that can be done by the next person re-using this code.

YohannParis (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -19,6 +19,19 @@
 				<!--amr_to_mmt doesn't like unit expressions with spaces, removing them here before they are saved to the amr-->
 				<template v-else-if="showUnit">
 					<template v-if="featureConfig.isPreview"><label>Unit</label>{{ item.unitExpression }}</template>
+					<!-- we use a dropdown for units with time semantic-->
+					<Dropdown
+						v-else-if="isTimePart"
+						:model-value="item.unitExpression"
+						option-label="label"
+						option-value="value"
+						:options="[
```
Tom-Szendrey, line 28: Likely will want these as an enum in future sections of this
For this section this is not required

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
blanchco: In my part 2 ticket I've added this as a util instead!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/model/configurations/ModelConfiguration.java
blanchco: Makes sense, done!

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
blanchco: Agreed - I think I might do this in part 4 as some final cleanups

PR #5215, opened 2024-10-19
funman: add onFailure hook, fix update-hook python error

mwdchang:
### Summary
Tidy up Funman task runner to be more complete, and update the python task as it seemed like the interface we are hitting against had changed a little bit.
- Add a onFailure to mark the funman-simulation as failed
- Fix python update datetime error in validation task's update hook, there is a date comparison that needs start_time to be initialized

@shawnyama The progression may not be "smooth" but you should have some progress number coming back after this PR, you just need to hook it up, it is part of the simulation object being polled. `simulation.progress` field. 

image

### Testing
Run funman process to validate a model-configuration. Note you may need to exec into the funman-taskrunner container and do:

```
cd /funman_task/
pip install  -e .
```
To refresh the installed tasks.

If funman runs successfully, the funman log should be free of errors such as 
```
2024-10-18 21:15:04 [WARN] [1f6971c9-15c4-4293-ac8d-c9173ed91840] stderr:     time_delta = now - self.start_time [software.uncharted.terarium.taskrunner.service.Task:415]
```

If funman fails, the UI should stop and not be doing infinite polling

### DIFF
--- a/packages/funman/tasks/validate_modelconfig.py
+++ b/packages/funman/tasks/validate_modelconfig.py
@@ -35,13 +35,14 @@ def run_validate(model: PetrinetModel, request, taskrunner):
         request=request,
         parameter_space=ParameterSpace(),
     )
+    current_results.start()
 
     # Update callback
     def update_current_results(scenario: AnalysisScenario, results: ParameterSpace) -> FunmanProgress:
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
LGTM

dvince2 (approved):

Issue #5154, opened 2024-10-15
[BUG]: PDF extraction fails but still returns success


dgauldie:
**Describe the bug**
I ran a document extraction locally. The task returns a success

![Image](https://github.com/user-attachments/assets/ab588d19-ad65-4a0a-905f-62510f17572b)

However, there are no equations extracted and there was an error logged in the equations task-runner.

```
2024-10-15 14:11:47 [ERROR] Unexpected failure for task 749bd71f-7327-4c9f-b322-42ae02cefe3b [software.uncharted.terarium.taskrunner.service.TaskRunnerService:176]
java.util.concurrent.TimeoutException: Reading from pipe took too long for task 749bd71f-7327-4c9f-b322-42ae02cefe3b
	at software.uncharted.terarium.taskrunner.service.Task.readProgressWithTimeout(Task.java:283)
	at software.uncharted.terarium.taskrunner.service.TaskRunnerService.dispatchSingleInputSingleOutputTask(TaskRunnerService.java:150)
	at software.uncharted.terarium.taskrunner.service.TaskRunnerService.onTaskRequest(TaskRunnerService.java:94)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.springframework.messaging.handler.invocation.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:169)
	at org.springframework.amqp.rabbit.listener.adapter.KotlinAwareInvocableHandlerMethod.doInvoke(KotlinAwareInvocableHandlerMethod.java:45)
	at org.springframework.messaging.handler.invocation.InvocableHandlerMethod.invoke(InvocableHandlerMethod.java:119)
	at org.springframework.amqp.rabbit.listener.adapter.HandlerAdapter.invoke(HandlerAdapter.java:75)
	at org.springframework.amqp.rabbit.listener.adapter.MessagingMessageListenerAdapter.invokeHandler(MessagingMessageListenerAdapter.java:277)
[...]

Issue #5123, opened 2024-10-10
[BUG]: Upload of a document not finishing


YohannParis:
**Describe the bug**
 
* Sometimes, when sending a document, the status can be stuck to 90% at the end of the `textExtractionFuture.get()`.
But we might never get the last 10% that allow us to close the notification and clear/clean it.

![Image](https://github.com/user-attachments/assets/e2130832-2b21-4c50-a128-9edae3c073f7)

* The percentages give to the document extraction are not accurate, here you can see that the table extraction has started but the percentage is still 0%

![Image](https://github.com/user-attachments/assets/b1ea2a48-3e39-425d-97fd-3e0880e6ea37)

dgauldie:
we should clean up the back-end to ensure we have better notifications and messages

PR #5223, opened 2024-10-21
[BUG]: Better extraction failure notifications

dvince2:
# Description

There are two fixes here. Firstly, for #5154 we now send up an error message for the front end with a comma seperated list of what extraction services failed. Secondly for #5123 we forgot to send a *Final* message in the failure case so the extraction just sat forever at its last reported %

Resolves ##5123 and #5154


YohannParis (approved):
Thank you!

kbirk (approved):

PR #5224, opened 2024-10-21
add date/timestep interchangeability to calibrate and optimize

blanchco:
# Description

* Created a component to handle the timestep/calendar interchangeability 
* Added functionality to calibrate and optimize

Optimize
Screenshot 2024-10-21 at 3 43 07 PM

Calibrate
Screenshot 2024-10-21 at 3 43 22 PM

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-timestep-calendar.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-timestep-calendar.vue
@@ -7,7 +7,7 @@
 			@update:model-value="emit('update:model-value', $event)"
 		/>
 		<Calendar
-			v-if="modelConfiguration.temporalContext"
+			v-if="startDate"
 			:model-value="date"
 			showIcon
 			iconDisplay="input"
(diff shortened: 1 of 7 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/widgets/tera-timestep-calendar.vue
```diff
@@ -0,0 +1,62 @@
...
+import { computed, ref } from 'vue';
+import teraInputNumber from './tera-input-number.vue';
+
+const props = defineProps<{
+	label: string;
+	modelValue: number;
+	model: Model;
+	modelConfiguration: ModelConfiguration;
+	disabled?: boolean;
+}>();
```
mwdchang, line 40: This component feels awkwardly constructed - if your intention is to create a general timestep/calendar component it shouldn't know anything about model and model-configuration.

Widgets, like the utils folders, are originally meant to house domain agnostic items. So I would suggest we either make this generic, or to move this out of widget-folder and give it a more specific name.

blanchco: Makes sense. I've made the component more generic such that it doesn't rely on `model` or `model-configuration` anymore.  Let me know what you think!

mwdchang (approved):
Thanks for the update

PR #5240, opened 2024-10-22
Design change document node

asylves1:
# Description

Design Change

- Add notification progress bar to the document Node
- Add a thumbnail of the Pdf document

image

Testing:

- Start up local build
- Create new workflow and upload a new doc.
- The thumbnail should appear
- The progress bar should also appear after a moment when the extraction starts


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DocumentController.java
```diff
@@ -362,16 +369,57 @@ private ResponseEntity<Void> uploadDocumentHelper(
...
+					g2d.dispose();
+
+					ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+					ImageIO.write(resizedBufferedImage, "png", outputStream);
+
+					byte[] thumbnailBytes = outputStream.toByteArray();
+					outputStream.close();
+
+					document.get().setThumbnail(thumbnailBytes);
+					pdfDocument.close();
```
dvince2, line 407: There are a lot of calls to `close` various things in here that won't get triggered if there is an exception at some point in creating the thumbnail... consider using the `finally` block for these? Somethings will have to be re-scoped for this to work

```
try{
...
} catch (Exception e) {
...
} finally {
g2d.dispose()
outputStream.close()
...
}
```

dvince2 (approved):
This is really cool!

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-node.vue
```diff
@@ -5,7 +5,30 @@
 			<h6>
 				<span class="truncate-after-three-lines">{{ documentName }}</span>
 			</h6>
-			<tera-operator-placeholder :node="node" />
+			<tera-operator-placeholder v-if="!thumbnail" :node="node" />
+			<section v-else>
+				<img class="pdf-thumbnail" :src="thumbnail" alt="Pdf's first page" />
+			</section>
```
shawnyama, line 11: could this just be an img tag?

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-node.vue
```diff
@@ -128,6 +161,24 @@ watch(
...
 );
+
+watch(
+	() => document.value?.thumbnail,
+	() => {
+		if (document.value?.thumbnail?.length) {
+			thumbnail.value = `data:image/png;base64, ${document.value.thumbnail}`;
+		}
+	}
+);
```
shawnyama, line 172: i think you can just add this if statement to the above watcher and things will be the same

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-node.vue
```diff
@@ -138,4 +189,44 @@ watch(
...
+}
+
+.pdf-thumbnail {
+	padding-bottom: var(--gap-small);
+	border: 1px solid var(--surface-border-light);
+	border-radius: var(--border-radius-medium);
+}
+
+.done-container {
+	margin-top: 0.5rem;
```
shawnyama, line 218: use gap variables

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-node.vue
```diff
@@ -138,4 +189,44 @@ watch(
...
+}
+
+.action {
+	font-size: var(--font-caption);
+	color: var(--text-color-secondary);
+	text-align: center;
+}
+
+.pdf-thumbnail {
+	padding-bottom: var(--gap-small);
```
shawnyama, line 212: I think we want to use the gap variables with the numbers instead

shawnyama (approved):

PR #5253, opened 2024-10-23
Updated model comparison schema and prompt

liunelson:
# Description

* Expanded the descriptions in the model comparison response schema for more longer and detailed answers
* Added some helper instructions to the prompt
* The goal is to make the comparison less of a repeat of the model metadata table (e.g. "This model has N states") and focus more on relevant questions (e.g. "what are the intended use cases and rank the models by their fit for purpose").


YohannParis (approved):

packages/gollm/gollm_openai/prompts/model_meta_compare.py
```diff
@@ -1,15 +1,33 @@
...
- - Fill in the semanticComparison fields for all models.
+Think step by step. 
+
+Carefully consider the descriptions in the provided JSON schema.
+
+Think and respond to the best ability of an academic or professional with subject matter expertise in epidemiology and mathematical modelling of diseases.
+
+Adhere to relevant best practices and domain-specific language.
+
+If any of the AMR models have gollmCard information, only use those models to fill out metadataComparison fields.
```
dgauldie, line 12: I wonder if this is why we only get 2 models instead of 3. maybe we just get rid of this section. and add something like if a model does not have information for a section then state that it can't be compared, or something like that

dgauldie:
Loving the improved descriptions!!! great job @liunelson

packages/gollm/gollm_openai/prompts/model_meta_compare.py
liunelson: I followed your suggestion and replaced those instructions with one that does not tell it to ignore models.

dgauldie (approved):

PR #5255, opened 2024-10-23
add dates to forecast charts

blanchco:
# Description

* If we have a time unit in the model and a defined start date within the model configuration, then show forecast charts as dates

Screenshot 2024-10-23 at 3 43 14 PM
Screenshot 2024-10-23 at 3 43 00 PM
Screenshot 2024-10-23 at 3 42 45 PM


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -42,6 +48,28 @@ export interface ErrorChartOptions extends Omit<BaseChartOptions, 'height' | 'yA
...
+export interface InterventionMarkerOptions {
+	hideLabels?: boolean;
+	labelXOffset?: number;
+	dateOptions?: DateOptions;
+}
+
+function transformDateFormatFn(
+	date: Date,
+	datum: string,
+	type: 'year' | 'yearmonth' | 'yearmonthdate' = 'yearmonthdate'
```
mwdchang, line 60: Any particular reason why we have two sets of time-measurement? Possible to use a single set of controlled vocabulary?
- CalendarDateType.DATE, CalendarDateType.MONTH, CalendarDateType.YEAR
- yearmonthdate, yearmonth, year

blanchco: I wasn't able to find a good way to use the same enum across the different 3rd party components we use.

These `CalendarDateType` enum is being used for the Primevue Calendar view https://v3.primevue.org/calendar/#api.calendar.props.view while the 'yearmonthdate', 'yearmonth', 'year' are for the vega chart https://vega.github.io/vega-lite/docs/timeunit.html definitions.

jryu01:
I think AI annotation feature probably won't work for the charts with date type. We will need to update the prompt accordingly to reflect this change. We should handle that n a separate issue I think.

packages/client/hmi-client/src/services/model.ts
```diff
@@ -247,3 +247,7 @@ export function stringToLatexExpression(expression: string): string {
 	latexExpression = latexExpression.replace(/([a-zA-Z0-9]+)\/([a-zA-Z0-9]+)/g, '\\frac{$1}{$2}');
 	return latexExpression;
 }
+
+export function getTimeUnits(model: Model): CalendarDateType {
+	return model?.semantics?.ode?.time?.units?.expression || '';
+}
```
mwdchang, line 253: I feel like we should mark down a canonical place to house functions to set/get model attributes/properties. Right now they are kind of splattered over
- services/models, and
- model-representation/service

Do you have any preference since you created a large subset of these functions? We don't need to do the refactor now but would be good to get consensus and mark it down as a TODO.

packages/client/hmi-client/src/utils/date.ts
```diff
@@ -1,4 +1,7 @@
-import { Model } from '@/types/Types';
+import { DateOptions } from '@/services/charts';
+import { getTimeUnits } from '@/services/model';
+import { CalendarDateType } from '@/types/common';
+import { Model, ModelConfiguration } from '@/types/Types';
```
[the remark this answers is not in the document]
blanchco: will move whats needed to services!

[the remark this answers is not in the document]
blanchco: This no longer depends on any services, those functions that needed them have been moved out

packages/client/hmi-client/src/services/charts.ts
blanchco: This is no longer needed since the underlying chart is now using all number as discussed in our discussion late yesterday

packages/client/hmi-client/src/services/model.ts
blanchco: Ill mark a TODO...would be best to have these in one place.  I think services/models might be best

packages/client/hmi-client/src/services/model.ts
```diff
@@ -247,3 +248,42 @@ export function stringToLatexExpression(expression: string): string {
...
+
+	return { view, format };
+}
+
+export function getVegaDateOptions(
+	model: Model | null,
+	modelConfiguration: ModelConfiguration | null
+): DateOptions | undefined {
+	let dateOptions;
+	if (model && modelConfiguration && modelConfiguration.temporalContext) {
```
shawnyama, line 282:
```suggestion
	if (model && modelConfiguration?.temporalContext) {
```

shawnyama (commented):
On your second screenshot (the month graph) there are still some years, how do I know what month that year is on?

shawnyama (approved):

mwdchang (approved):

blanchco:
> On your second screenshot (the month graph) there are still some years, how do I know what month that year is on?

That was using vegalite's system for dates, I've changed the labels to use our own format so it should show like Month, Year

Tom-Szendrey (approved):

PR #5275, opened 2024-10-24
date in intervention descriptions

blanchco:
# Description

* Include dates in intervention descriptions

Screenshot 2024-10-24 at 10 41 02 AM


shawnyama (approved):

packages/client/hmi-client/src/utils/date.ts
```diff
@@ -162,3 +162,20 @@ export function dateTypeToVegaDate(type: CalendarDateType) {
 			return 'yearmonthdate';
 	}
 }
+
+export function getTimePointString(timePoint: number, startDate?: Date, calendarSettings?: CalendarSettings): string {
```
mwdchang, line 166: I would avoid creating an interface that has two consecutive optionals, it makes it difficult to understand the intent and to reason with. The functional signature seem to suggest that you can specify calendarSettings but ignore startDate, but that would be incorrect ... and even if you can, calendarSettings is ignored because you short-circuit on startDate.

Preferable might be creating an Options object.

Also as noted before - different people may see different dates because they are in different time-zones, we don't need to deal with it here in this PR, but this is something we will want to resolve.

blanchco: Sure Ill make an options object

blanchco: added!

mwdchang (approved):
okay!

PR #5279, opened 2024-10-24
show negative timesteps as invalid in the intervention policy drilldown

mloppie:
# Description

* Expands the element validation to some negative timesteps as invalid
* Also fixes some formatting in the test description

Screenshot 2024-10-24 at 1 32 26 PM

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
@@ -28,8 +28,8 @@
 				Starting at day
 				<tera-input-number
 					auto-width
+					invalidate-negative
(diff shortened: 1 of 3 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-card.vue
```diff
@@ -29,6 +29,7 @@
 				<tera-input-number
 					auto-width
 					:model-value="intervention.staticInterventions[0].timestep"
+					:invalidate-negative="true"
```
blanchco, line 32: we can just do `invalidate-negative` if the prop is always true (see auto-width)

blanchco (approved):

PR #5285, opened 2024-10-24
Intervention pdf goto page

asylves1:
# Description

Design Change

- For interventions with page number, clicking on them will jump to the matching the page in the PDF

Testing:
- connect a document to the intervention's drill down
- extract from inputs\
- select a policy and click on an intervention

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-pdf-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-pdf-panel.vue
@@ -1,14 +1,14 @@
 <template>
 	<TabView v-if="pdfs?.length" class="container">
 		<TabPanel :header="pdf.name" v-for="pdf in pdfs" :key="pdf.name">
-			<tera-pdf-embed v-if="pdf.isPdf" :pdf-link="pdf.data" :title="pdf.name || ''" />
+			<tera-pdf-embed v-if="pdf.isPdf" ref="pdfRef" :pdf-link="pdf.data" :title="pdf.name || ''" />
 			<tera-text-editor v-else :initial-text="pdf.data" />
 		</TabPanel>
 	</TabView>
 </template>
 
 <script setup lang="ts">
-import { ref } from 'vue';
+import { onMounted, ref } from 'vue';
 
 import TabView from 'primevue/tabview';
 import TabPanel from 'primevue/tabpanel';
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/widgets/tera-pdf-panel.vue
```diff
@@ -24,11 +24,16 @@ interface PdfData {
...
+const emit = defineEmits(['pdf-ref']);
+
 const props = defineProps<{
 	pdfs: PdfData[];
 }>();
 
 const pdfs = ref<PdfData[]>(props.pdfs);
+const pdfRef = ref(null);
+
+onMounted(() => emit('pdf-ref', pdfRef));
```
mwdchang, line 36: Probably @shawnyama or @blanchco can comment on this, I think we are using "defineExpose" generally, for exposing children-components to that of their parents, but I am not 100% sure, or if there are any trade offs between emit vs expose.

shawnyama: ya looking at this `defineExpose` would be a cleaner solution

shawnyama, line 36:
```suggestion
defineExpose({pdfRef})
```

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
```diff
@@ -256,6 +266,18 @@ const newBlankInterventionPolicy = ref({
...
+interface SelectedIntervention {
+	index: number;
+	name: string;
+	extractionDocumentId?: string;
+	extractionPage?: number;
+	staticInterventions: StaticIntervention[];
+	dynamicInterventions: DynamicIntervention[];
+}
+
+const pdfViewer = ref();
```
shawnyama, line 278: You can remove the `getPdfRef()`
```suggestion
const pdfViewer = computed(() => pdfPanelRef.value?.pdfRef[0])
```

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
```diff
@@ -15,7 +15,7 @@
 			>
 				<template #content>
 					<tera-drilldown-section :is-loading="isFetchingPDF">
-						<tera-pdf-panel :pdfs="pdfData" />
+						<tera-pdf-panel :pdfs="pdfData" @pdf-ref="getPdfRef" />
```
shawnyama, line 18:
```suggestion
						<tera-pdf-panel :pdfs="pdfData" ref="pdfPanelRef" />
```
create a `pdfPanelRef` in the js too

shawnyama (approved):

PR #5289, opened 2024-10-24
Model config pdf goto page

asylves1:
# Description

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -301,6 +301,8 @@ const isFetchingPDF = ref(false);
 const isDocViewerOpen = ref(true);
 
 const pdfData = ref<{ document: any; data: string; isPdf: boolean; name: string }[]>([]);
+const pdfPanelRef = ref();
+const pdfViewer = computed(() => pdfPanelRef.value?.pdfRef[0]);
```
Tom-Szendrey, line 305: Why does this need to be compute?
If we only use it in `onSelectConfiguration` can we not just define it in there?

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
asylves1: There is a delay when the pdf ref is loaded. I was using `watch` before but this is cleaner to read

PR #5301, opened 2024-10-25
Layout improvement for model diagram

mwdchang:
### Summary
This PR attempts to improve edge-routing in model-diagrams after one or more nodes have been dragged on the canvas. 

Currently no re-routing is done.

image

Note: I made the AStar algorithm a bit more flexible, but it didn't work nicely at the end - the unconstrained greedy paths makes it harder to create "eye-pleasing" edges. I left the changes in here because it doesn't hurt, but reviewers can largely ignore this portion.

I restarted with a new orthogonal edge router from [this medium article](https://medium.com/swlh/routing-orthogonal-diagram-connectors-in-javascript-191dc2c5ff70), it is somewhat similar in spirit to that of AStar (uses Dijkstra per the article) but the search space is constrained by pre-creating a graph/grid. The demo source code has some limitations which we modified:
- Add an extra `obstacles` array that gets injected in the grid/graph creation process, this allows use to route edges for more than 2 nodes

The Renderer interface has changed to make  underlying logic a bit simpler, instead of providing a bare-bone edge router, now caller must specify one via `options.options.edgeReroutingFn`. We now provide a routing function in `services/graph#rerouteEdges`, which also houses the logic to "split apart" incidental edges so they are not overlayed on top of each other.

### Things to look at
Where the important changes are located
- model-representation/services.ts
- services/graph.ts
- utils/ortho-router.ts

### Results: initial layout
image

image

### Results after node drag
image

image

### TODO
- I hardcoded the configurations such that source is the "right" and target is the "left". We can probably figure out better alignment based on relative position of the source/target nodes.
- Expose more aspect of the ortho-router interface like margins/paddings

### Testing
Check model diagram of various AMRs, simple ones, stratified ones .. etc


packages/client/graph-scaffolder/src/core/traverse.ts
```diff
@@ -38,13 +38,15 @@ export interface IGrid {
...
-): IPoint[] => {
+export interface AStarOptions {
+	collider: ColliderFn;
+	gridCell?: IGrid;
+	searchLimit?: number;
+	distanceType?: 'Euclidean' | 'Manhattan';
+}
+
+export const getAStarPath = (start: IPoint, goal: IPoint, options: AStarOptions): IPoint[] => {
+	console.log('debug');
```
mloppie, line 49: Is this still needed?

packages/client/hmi-client/src/services/graph.ts
```diff
@@ -26,6 +27,56 @@ function interpolatePointsForCurve(a: Position, b: Position): Position[] {
...
+	edges.forEach((edge: IEdge<any>) => {
+		if (edge.data.isObservable) return;
+		if (!sourceNodeCounter.has(edge.source)) {
+			sourceNodeCounter.set(edge.source, 0);
+		}
+		let counter = sourceNodeCounter.get(edge.source) as number;
+		const denominator = edgeSourceMap.get(edge.source) as number;
+		counter++;
+
+		console.log('>', edge.source, edge.target, counter, denominator);
```
mloppie, line 63: Do we need to log this?

mloppie (approved):

shawnyama (approved):

PR #5305, opened 2024-10-25
[Small]: Limit preview on dataset node

Tom-Szendrey:
# Description
There is no limit on the preview in the dataset node.
This leads to HILARIOUS previews that make the node hard to work with and insane.

# Problem:

image

# Solution

https://github.com/user-attachments/assets/556ba80c-52a7-4668-bbe3-68e8fcfb7e7d


mloppie (approved):

shawnyama:
does the show more option appear if you are to hover over the text. If you just want to show less you can just apply this css to the div the text is in 

```
        white-space: pre-line;
	display: -webkit-box;
        -webkit-line-clamp: 5; /*Amount of lines you want here*/
	-webkit-box-orient: vertical;
	overflow: hidden;
	text-overflow: ellipsis;
```

shawnyama (approved):

PR #5308, opened 2024-10-25
fix(funman): limit y range and round timepoints to third decimal place in preview

shawnyama:
# Description

- timepoints are rounded to third decimal place instead of rounded down to 0
- limit y axis on graph to not go so far that it goes to NaN

https://github.com/user-attachments/assets/15bfa00e-c3dc-4574-9965-38c7b2f9c34e

### DIFF
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -909,8 +909,6 @@ export function createFunmanStateChart(
 		endY: Math.max(upperBounds[index] ?? maxY, maxY)
 	}));
 
-	console.log('modelChecks', modelChecks);
-
 	return {
 		$schema: VEGALITE_SCHEMA,
 		id: stateId,
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -915,16 +915,6 @@ export function createFunmanStateChart(
 		};
 	});
 
-	console.log(lowerBounds, upperBounds);
-	console.log(
-		'startY',
-		modelChecks.map((m) => m.startY)
-	);
-	console.log(
-		'endY',
-		modelChecks.map((m) => m.endY)
-	);
-
 	return {
 		$schema: VEGALITE_SCHEMA,
 		id: stateId,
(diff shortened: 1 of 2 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -875,22 +875,44 @@ export function createFunmanStateChart(
...
+	const upperBounds = stateIdConstraints.map((c) => c.additive_bounds.ub);
+	const yPoints = trajectories.map((t) => t.values[stateId]);
+
+	const potentialMinYs = focusOnModelChecks && !isEmpty(lowerBounds) ? lowerBounds : yPoints;
+	const potentialMaxYs = focusOnModelChecks && !isEmpty(upperBounds) ? upperBounds : yPoints;
+	let minY = Math.floor(Math.min(...potentialMinYs));
+	let maxY = Math.ceil(Math.max(...potentialMaxYs));
+	// Limit to MAX as very large numbers cause the chart to have NaN in the y domain
+	minY = minY < -MAX ? -MAX : minY;
+	maxY = maxY > MAX ? MAX : maxY;
```
mwdchang, line 894: I think this needs some more investigation here, if it is the case that Funman's result goes to infinity, then that may indicate that something had gone wrong and we either need to communicate it to the Funman team or perhaps change the process logic on our side.

Do we have any additional data on this? We should only mask problems as a last resort.

shawnyama: I don't get an `Infinity` but I do get a `1.7e308`. It renders fine if I have a y range of `[0, 1.7e308]` but it blows up when I have `[-1.7e308, 1.7e308]` giving me this:
image

mwdchang: Do you still have the request and response payloads? They would be useful to have, to determine what you are asking for, and what exactly you are getting back.

If the request looks sensible and the response does not, we should bump this over the Funman team to take a look, check in with @liunelson if you need a hand determining the sanity of either one.

In the meantime, we should probably add a FIXME here indicating that this is likely an abnormal event, and let's "console.error" some information here as well so it is more apparent.

shawnyama: This happens when we check off the "Focus on model checks" box. Since the boxes are going to be drawn to infinity this makes sense. 

https://app.staging.terarium.ai/projects/34e7b955-f408-4e8b-be27-9429480ad6aa/workflow/ae644619-7fe5-40d7-bb90-880c447ba9a4?operator=330e2760-d3e1-4204-b8ed-08be949af5ed

https://github.com/user-attachments/assets/4cde8689-ce44-4ff3-9617-56f98bb0283a

mwdchang (approved):
Thanks, let's roll with the heuristic threshold for a few days and see how things look and reevaluate.

PR #5310, opened 2024-10-28
Intervention service minor update

Tom-Szendrey:
# Description
As the hmiserver can fail to create an intervention it is important that our client's service can as well.
This update is propagated to optimize as well where technically the result can be failed to create
This will now better handle this fail case

### DIFF
--- a/packages/client/hmi-client/src/services/intervention-policy.ts
+++ b/packages/client/hmi-client/src/services/intervention-policy.ts
@@ -27,7 +27,6 @@ export const createInterventionPolicy = async (policy: InterventionPolicy): Prom
 		delete policy.createdOn;
 		delete policy.updatedOn;
 		const response = await API.post<InterventionPolicy>(`/interventions`, policy);
-		console.log(response);
 		if (response.status !== 201) {
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/services/intervention-policy.ts
```diff
@@ -20,12 +21,21 @@ export const getInterventionPolicyById = async (policyId: string): Promise<Inter
...
-	delete policy.updatedOn;
-	const response = await API.post<InterventionPolicy>(`/interventions`, policy);
-	return response?.data ?? null;
+export const createInterventionPolicy = async (policy: InterventionPolicy): Promise<InterventionPolicy | null> => {
+	try {
+		delete policy.id;
+		delete policy.createdOn;
+		delete policy.updatedOn;
+		const response = await API.post<InterventionPolicy>(`/interventions`, policy);
+		console.log(response);
```
dvince2, line 30: is this log kept in on purpose?

dvince2 (approved):

packages/client/hmi-client/src/services/intervention-policy.ts
Tom-Szendrey: Is it not thank you for catching this

shawnyama (approved):

PR #5312, opened 2024-10-28
Automap string comparison should not care for casing

Tom-Szendrey:
# Description
Allow for the following to successfully automap (note no concepts were used in this automapping)
image

### DIFF
--- a/packages/client/hmi-client/src/services/concept.ts
+++ b/packages/client/hmi-client/src/services/concept.ts
@@ -213,7 +213,7 @@ const autoEntityMapping = async (sourceEntities: Entity[], targetEntities: Entit
 		targetEntities.forEach((target) => {
 			if (
 				!distinctSources.includes(source.id) &&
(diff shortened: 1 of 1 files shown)


jryu01 (approved):

packages/client/hmi-client/src/services/concept.ts
```diff
@@ -213,7 +213,7 @@ const autoEntityMapping = async (sourceEntities: Entity[], targetEntities: Entit
 		targetEntities.forEach((target) => {
 			if (
 				!distinctSources.includes(source.id) &&
-				(target.id.startsWith(source.id) || source.id.startsWith(target.id))
+				(target.id.startsWith(source.id.toLocaleLowerCase()) || source.id.startsWith(target.id.toLocaleLowerCase()))
```
jryu01, line 216: Is `target.id` always in lower case?

blanchco (approved):

PR #5316, opened 2024-10-28
5278 bug gollm comparison of model results in an overview of object object

dgauldie:
# Description

Resolves #(issue)

### DIFF
--- a/containers/scripts/docker-compose-data-services.yml
+++ b/containers/scripts/docker-compose-data-services.yml
@@ -10,7 +10,7 @@ services:
     environment:
       HMI_SERVER_URL: "http://${local_host_name}:3000"
       JUPYTER_SERVER: "http://${local_host_name}:3050"
-      MIRA_REST_URL: "http://mira-epi-dkg-lb-c7b58edea41524e6.elb.us-east-1.amazonaws.com/"
+      MIRA_REST_URL: "http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com/"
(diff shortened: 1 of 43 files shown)


blanchco:
It looks like there's a case with unenriched models its will spin forever.
```
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at hasNonEmptyValue (tera-model-comparison.vue?t=1730209061178:168:21)
    at generateOverview (tera-model-comparison.vue?t=1730209061178:240:13)
    at processCompareModels (tera-model-comparison.vue?t=1730209061178:304:9)
    at async tera-model-comparison.vue?t=1730209061178:326:7
```

I think you can just put a check if the object is null/undefined in the `hasNonEmptyValue` function and it should fix it

blanchco (approved):

liunelson (commented):
Pascale has told me that we prefer the model comparison sections and subsections to be shown as part of a larger body of text, as opposed to a sequence of short text snippets. For example, the subsections `common, unique, conclusion` should look like a single paragraph; the sections themselves `states, parameters, transitions, observables` may be better as paragraphs under a single header.

dgauldie:
> Pascale has told me that we prefer the model comparison sections and subsections to be shown as part of a larger body of text, as opposed to a sequence of short text snippets. For example, the subsections `common, unique, conclusion` should look like a single paragraph; the sections themselves `states, parameters, transitions, observables` may be better as paragraphs under a single header.

They are concatenated together in the front end so that they look like a paragraph. All paragraphs are structured as common - unique - conclusion.

PR #5323, opened 2024-10-29
fix(funman): saving model config for reuse instead of auto save

shawnyama:
# Description

- I am not automatically saving the model configuration once I get the funman output, the config is now just a part of the funman output
- If the user wants to save the configuration then they have to click the "Save for reuse" button
- This allows them to name the config 
- Once saved it'll show up in their Model configuration operator
- Note: The invalid date in the video shows up as a date now

https://github.com/user-attachments/assets/9dfb5ee3-905e-410a-814a-9d572fe5f7d0


Tom-Szendrey (approved):
Did not test, looks good!

PR #5325, opened 2024-10-29
updating mira endpoints

Tom-Szendrey:
# Description
They have updated epi dkg endpoint

https://github.com/DARPA-ASKEM/orchestration/pull/440

### DIFF
--- a/packages/client/hmi-client/src/services/concept.ts
+++ b/packages/client/hmi-client/src/services/concept.ts
@@ -62,7 +62,7 @@ async function searchCuriesEntities(query: string): Promise<Array<DKG>> {
 }
 
 function getCurieUrl(curie: string): string {
-	return `http://mira-epi-dkg-lb-c7b58edea41524e6.elb.us-east-1.amazonaws.com/entity/${curie}`;
+	return `http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com//entity/${curie}`;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/services/concept.ts
```diff
@@ -62,7 +62,7 @@ async function searchCuriesEntities(query: string): Promise<Array<DKG>> {
 }
 
 function getCurieUrl(curie: string): string {
-	return `http://mira-epi-dkg-lb-c7b58edea41524e6.elb.us-east-1.amazonaws.com/entity/${curie}`;
+	return `http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com//entity/${curie}`;
```
dvince2, line 65: Double `//`

packages/server/src/main/resources/application.properties
```diff
@@ -148,7 +148,7 @@ terarium.amazon.s3.default.credentials-id=default
 ########################################################################################################################
 github.url=https://api.github.com
 jsdelivr.url=https://cdn.jsdelivr.net
-mira-api.url=http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com
+mira-api.url=http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com/
```
dgauldie, line 151: you need the ending slash?

dvince2, line 151: Also new `/` here

dgauldie (approved):

packages/client/hmi-client/src/services/concept.ts
Tom-Szendrey: Thank you so much

dvince2:
The function in `concept.ts` is never called anywhere, and application.proeprties is unchanged. Do we actually just want to remove that function in `concept.ts`?

dvince2 (approved):

PR #5328, opened 2024-10-29
dismissable warning banner on workflow

blanchco:
# Description
* A dismissible warning banner has been added to the workflow when invalid nodes are on the screen
* Storing a variable in local storage so that the user has the option not to see this again.

Screenshot 2024-10-29 at 11 47 12 AM

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -51,6 +51,11 @@
 					<ContextMenu ref="addComponentMenu" :model="contextMenuItems" style="white-space: nowrap; width: auto" />
 				</div>
 			</div>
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -264,6 +269,8 @@ const drilldownSpawnAnimation = ref<'left' | 'right' | 'scale'>('scale');
 const route = useRoute();
 const router = useRouter();
 
+const warningBannerDismissed = ref(false);
+
```
mwdchang, line 273: Can we just call this variable `warningBanner`?

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
blanchco: done

shawnyama (approved):

PR #5330, opened 2024-10-29
Calibrate knobs to update with output selection

Tom-Szendrey:
# Description
The calibration knobs will now update when we change our output selection 

https://github.com/user-attachments/assets/f43c54bc-d72a-4159-9715-106d26128cba

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -1094,6 +1094,17 @@ async function getAutoMapping() {
 }
 
 const initialize = async () => {
+	// Update Wizard form fields with current selected output state
+	knobs.value = {
+		numIterations: props.node.state.numIterations ?? 1000,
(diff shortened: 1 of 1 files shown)


asylves1 (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -1094,6 +1094,17 @@ async function getAutoMapping() {
...
 const initialize = async () => {
+	// Update Wizard form fields with current selected output state
+	knobs.value = {
+		numIterations: props.node.state.numIterations ?? 1000,
+		numSamples: props.node.state.numSamples ?? 100,
+		endTime: props.node.state.endTime ?? 100,
+		stepSize: props.node.state.stepSize ?? 1,
+		learningRate: props.node.state.learningRate ?? 0.1,
+		method: props.node.state.method ?? CiemssMethodOptions.dopri5,
+		timestampColName: props.node.state.timestampColName ?? ''
```
mwdchang, line 1105: nitpick: would be better if we factor out `const state = props.node.state` so we are not repeat the same thing 7,8 times.

mwdchang (approved):

PR #5332, opened 2024-10-30
5142 bug y range of the loss chart in the calibrate operator is too wide   should be relative to the data not start from zero always 1

### LINKED ISSUE #5142: [BUG]: Y range of the loss chart in the `Calibrate` operator is too wide - should be relative to the data, not start from zero always
**Describe the bug**

In the `Calibrate` operator output panel, there is a "Loss" line chart. The y-axis range (min, max) should be based on the min/max range of the loss dataset, instead of always having the min = 0.

The reason is that the absolute value of the loss has no inherent meaning - the user only needs to know whether the loss as a function of iterations is "well-behaved". It is well-behaved when it decreases monotonically. The current chart with a fixed Y min = 0 prevents this observation when `loss >> max(loss) - min(loss)`.

mloppie:
# Fits the scale of the loss chart in the calibrate node and drill down
Screenshot 2024-10-29 at 9 32 15 PM

Resolves #5142


mwdchang (approved):
didn't test, but code looks fine!

jryu01 (approved):
Looks fine to me!

PR #5335, opened 2024-10-30
fix(funman): output should be a temp config and save for reuse makes it a non temp config

shawnyama:
# Description
- fix save for reuse
- temporary config in funman drilldown must be saved for reuse to be accessed in the config operator

https://github.com/user-attachments/assets/b1d8a142-6d7d-41d9-9770-c5e3c73cf251

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
@@ -153,8 +153,8 @@
 				<template v-else-if="!isEmpty(node.state.runId)">
 					<header class="flex align-items-start">
 						<div>
-							<h4>{{ activeOutput?.label }}</h4>
-							<span class="secondary-text">{{ formatShort(activeOutput?.timestamp) }}</span>
+							<h4>{{ validatedModelConfiguration?.name }}</h4>
+							<span class="secondary-text">{{ formatShort(validatedModelConfiguration?.createdOn) }}</span>
 						</div>
 						<div class="btn-group">
 							<Button label="Add to report" outlined severity="secondary" disabled />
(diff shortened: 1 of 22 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ValidateModelConfigHandler.java
```diff
@@ -128,12 +128,19 @@ public TaskResponse onSuccess(final TaskResponse resp) {
 				props.newModelConfigName,
 				contractedModel.getDescription()
 			);
+			contractedModelConfiguration.setTemporary(true); // Config should only be accessible within the operator for now
```
mwdchang, line 131: I would reword or just remove this comment, as the notions of workflow operators are decouples from that of the taskrunner, you could in theory use the task runners without any workflows and the comment will make no sense in that context.

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
```diff
@@ -819,10 +821,12 @@ async function renderCharts() {
 	updateStateCharts();
 	updateParameterCharts();
 
+	if (!validatedModelConfiguration.value) return;
+
```
mwdchang, line 825: This function feels a bit odd and maybe some reorganization would be good? Are you actually wanting to return or do we want to throw an error? 

But moreover, the underlying model and MMT do not change across validation runs, so `getModel`/`getMMT` are likely better if they are in onMounted hook since they need to be loaded once and once-only. What is the reason we only get results and in "renderCharts"?

packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -122,8 +122,13 @@ export const processFunman = (result: any, onlyShowLatestResults: boolean) => {
 	const boxes: FunmanBox[] = [];
 	const trajectories: any[] = [];
 
+	console.log('trueBoxes', trueBoxes);
+	console.log('falseBoxes', falseBoxes);
+	console.log('ambiguousBoxes', ambiguousBoxes);
```
jryu01, line 127: We should not leave console.log on the committed code.

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
shawnyama: I did the clean ups and I figured out a way to just process the funman results once. I was filtering the data depending on the "Show furthest results checkbox", but I decided to just mark which boxes are the furthest to use them later.

dvince2 (approved):

mwdchang (approved):
thanks!

PR #5342, opened 2024-10-31
SymPy latex_to_sympy task

kbirk:


packages/mira/Dockerfile
```diff
@@ -63,5 +66,8 @@ COPY ./packages/mira /mira_task
 WORKDIR /mira_task
 RUN pip install --no-cache-dir -e .
 
+# install sympy
+RUN pip install sympy antlr4-python3-runtime==4.11
+
```
mwdchang, line 71: I think installing MIRA will come with sympy, so not entirely sure if there may be version problems if we overwrite the installation.

Also why is everything here twice in this Dockerfile??

dvince2 (approved):

mwdchang (approved):

PR #5358, opened 2024-11-01
bug: stop image from being dragged off of node

asylves1:
# Description

https://github.com/user-attachments/assets/1ca90e53-10f9-40bc-8349-59b6664099ba

Issues: Images in the Dataset Transformation node could be dragged, and would create a new node.

Testing: 
- https://github.com/DARPA-ASKEM/terarium/blob/main/testing/manual/transform-dataset.md

- Create a new dataset transformation and show a graph image on the node
- Try dragging image

Resolves #(issue)


shawnyama (approved):

mwdchang (approved):
thanks

PR #5362, opened 2024-11-01
Hook up new ES based DKG lookups #5309

### LINKED ISSUE #5309: Hook up new ES based DKG lookups

dvince2:
# Description
Hooking up to the new Elasticseach hosted DKG. Right now this is just doing a basic text search over the title and description, with title taking priority. There is an issue logged to use embeddings to improve this search.

Resolves #5309


packages/server/src/main/resources/application-devES.properties
```diff
@@ -0,0 +1,8 @@
+
+########################################################################################################################
```
dgauldie, line 2: should this be in the base application.properties? And how are you going to handle the local dev version where project search will be local ES and dkg will be pointing to dev?

dvince2: I'd put it here so the default application.properties wouldn't be affected for local dev, but, you're right that it won't handle local project searching at all like this...

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/DKGService.java
```diff
@@ -0,0 +1,78 @@
...
+		final String index,
+		final Integer page,
+		final Integer pageSize,
+		final String q,
+		final SourceConfig source
+	) throws IOException {
+		final SearchRequest.Builder builder = new SearchRequest.Builder().index(index).from(page).size(pageSize);
+
+		if (q != null && !q.isEmpty()) {
+			Query query = QueryBuilders.multiMatch().fields(DKG.NAME + EMPHASIS, DKG.DESCRIPTION).query(q).build()._toQuery();
```
mwdchang, line 50: Maybe better if "emphasis" is a function/method to make things a little more future proof?

```
private static String emphasis(String s, int boost) {
  return s + "^" + String.valueOf(boost);
}
```

dvince2: oh true!

mwdchang (approved):

PR #5367, opened 2024-11-01
if we can't find a grounding, we set as a default empty grounding

dgauldie:
# Description

Resolves #(issue)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -858,32 +858,57 @@ public ResponseEntity<TaskResponse> createEnrichModelMetadataTask(
 		// Update State Grounding
 		if (!model.isRegnet()) {
 			final List<State> states = model.getStates();
-			states.forEach(state -> TaskUtilities.performDKGSearchAndSetGrounding(miraProxy, state));
+			states.forEach(state -> {
```
dvince2, line 861: I'm not sure if these are going to work as expected? if the element in these lists are null, will `state = new State();` correctly init it in the list of objects? I think that is effectively the same as doing a `continue` because its not added back into the lists?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -230,34 +230,59 @@ private void enrichModel(
 		}
 
 		// Update State Grounding
-		if (!model.get().isRegnet()) {
+		if (model.get().isPetrinet()) {
```
dvince2, line 233: `!model.get().isRegnet()` was purposful as we also have roughed in support for stock and flow, which isn't regnet and looks to go down this path?

dvince2: oh i see `isPetrinet()`is new. is stock and flow still relevant? our front end seems to think it is

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
YohannParis: why not just filtering first?
```java
states.stream()
	.filter(Objects::nonNull)   // filters out null values
	.forEach(state -> TaskUtilities.performDKGSearchAndSetGrounding(miraProxy, state));
```

YohannParis (approved):

dvince2 (approved):

packages/client/hmi-client/src/services/goLLM.ts
YohannParis, line 1: The async/await and Promise<void> can be removed if we do not need any result from it.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
dgauldie: I fixed this up. But I noticed that skema also return "generic math models" on top of petrinets, and regnets.

packages/client/hmi-client/src/services/goLLM.ts
dgauldie: well... I got rid of events and we wait on this to finish up... not sure what the best behaviour was. But I didn't like that the model card event happens and the model stays updating because it's working on other parts of the enrichment.

Feel free to make another issue to improve UX on this.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
dgauldie: we need to make sure that there is a base grounding since the dkg search can return a null value. This null value causes Mira to fail.

PR #5371, opened 2024-11-04
Erronious user in spicedb

bigglesandginger:
# Description

Should deal with bad user ids accidentially inserted into SpiceDB

Resolves #(issue)

On Staging one cannot view the users in the Public group (in the admin panel)


YohannParis (approved):
LGTM

dvince2 (approved):

PR #5375, opened 2024-11-04
Allow editor/writer to change project user permissions and accessibility

YohannParis:
### Permission Adjustments:
* `ProjectController.java`: Changed permission checks from `canAdministrate` to `canWrite` for making a project public and setting project user permissions.


packages/client/hmi-client/src/types/Types.ts
```diff
@@ -801,9 +801,9 @@ export interface ModelUnit {
 }
 
 export interface GroundedSemantic {
-    grounding?: ModelGrounding;
     id: string;
     name?: string;
+    grounding?: ModelGrounding;
```
dvince2, line 806: This worryingly seems to change the order of items in this interface on every PR now?

dvince2 (approved):

Tom-Szendrey (approved):

PR #5395, opened 2024-11-05
Support partial import failures

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TerariumAssetCloneService.java
```diff
@@ -241,36 +241,40 @@ public Project importProject(final String userId, final String userName, final P
...
 
-			TerariumAsset asset = assetExport.getAsset();
+				TerariumAsset asset = assetExport.getAsset();
 
-			// upload the files (do this first as the asset creation my use the files)
-			for (final Map.Entry<String, FileExport> entry : assetExport.getFiles().entrySet()) {
-				final String filename = entry.getKey();
-				final FileExport fileExport = entry.getValue();
+				// upload the files (do this first as the asset creation my use the files)
+				for (final Map.Entry<String, FileExport> entry : assetExport.getFiles().entrySet()) {
```
dvince2, line 252: Maybe outside the scope of this issue, but, looking at the code here it looks like we'll upload the files of an asset before creating the asset. If we're potentially expecting the creation to fail, do we want to do the file upload after or delete the files on failure so that we don't have orphaned files on s3?

dvince2 (approved):

YohannParis (approved):
Thanks!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TerariumAssetCloneService.java
kbirk: For some assets the file needs to exist before we create the asset.

PR #5396, opened 2024-11-05
show add project button when there is more than one project

shawnyama:
# Description

* Show add project option when there is more than one project
_* Note: When we do the `tera-asset` clean ups we should move these options in there, also a searchbar above the dropdown would be helpful_

https://github.com/user-attachments/assets/579574f8-34ab-48d7-b1fb-4d853dbf10bc


YohannParis (approved):
Thank you!

Tom-Szendrey (approved):

Issue #5350, opened 2024-10-31
[BUG]: Can't save a dataframe as a dataset for reuse, Add cell button not working, yellow header always there


pascaleproulx:
**Describe the bug**
Can't save a dataframe as a dataset for reuse

**To Reproduce**
1. In a new workflow, add two or more datasets and plug them into a transform dataset

- Issue 1: Since I have not done anything yet, the yellow header should not be present at this point.
![Image](https://github.com/user-attachments/assets/351fa4f4-6aa8-44dc-8c22-90364e0ed682)

2. Click the 'Add a cell ' button in the header (not the link at the bottom)
- Issue 2: Nothing happens. That button should create a new cell but it does not. Only the bottom green link works.

**3. [biggest issue]** Ask the AI to filter the dataset in some way. At the top, you can select the new dataframe (e.g. filtered_df). The button 'Save for reuse' never gets enabled which makes it impossible to save the result of this operator for another use. Maybe it is because the yellow header is always there (i.e. it always act as if the session was not live)...
![Image](https://github.com/user-attachments/assets/fec9c1d8-ed4c-425b-9b68-de74af85b0f9)

PR #5412, opened 2024-11-06
5350 [part 2] fixes permanently disabled "save for reuse" button

mloppie:
# There are really 3 bugs in https://github.com/DARPA-ASKEM/terarium/issues/5350, I am submitting the fixes for them separately for clarity and ease of review

* Fixes the permanent invalid state of the node to allow saving
* Preserves the ability to disable saving when actually invalid

Screenshot 2024-11-06 at 3 03 52 PM

### DIFF
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -244,7 +244,7 @@ export class WorkflowWrapper {
 				state: {}
 			})),
 
-			status: OperatorStatus.INVALID,
+			status: OperatorStatus.DEFAULT,
 			width: nodeSize.width,
 			height: nodeSize.height
 		};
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -244,7 +244,7 @@ export class WorkflowWrapper {
 				state: {}
 			})),
 
-			status: OperatorStatus.INVALID,
+			status: OperatorStatus.DEFAULT,
 			width: nodeSize.width,
```
mwdchang, line 248: It might be better to change the banner logic such that
- the operator state is invalid, and
- there is a non-empty output on the operator

Tom-Szendrey (approved):
Did not test this but cant see what would be wrong with it

YohannParis (approved):
LGTM

PR #5413, opened 2024-11-06
ensemble mapping row type

blanchco:
# Description

* This will be shared in calibrate and simulate ensemble which @Tom-Szendrey is working on as well, so just creating this type for us to coordinate using the same types
* Creating a type to be used in calibrate and simulate ensemble new UI design
* May be used down the line in the backend, so created it in the backend as well

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/parts/EnsembleMappingRow.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/parts/EnsembleMappingRow.java
@@ -1,0 +1,21 @@
+package software.uncharted.terarium.hmiserver.models.simulationservice.parts;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+import java.util.Map;
+import lombok.Data;
+import lombok.experimental.Accessors;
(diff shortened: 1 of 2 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/parts/EnsembleMappingRow.java
```diff
@@ -0,0 +1,21 @@
...
+import com.fasterxml.jackson.annotation.JsonAlias;
+import java.util.Map;
+import lombok.Data;
+import lombok.experimental.Accessors;
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+
+@Data
+@Accessors(chain = true)
+@TSModel
+public class EnsembleMappingRow {
```
Tom-Szendrey, line 12: Can you update this to include the name Calibrate?
The simulate version of this mapping will not include a dataset_mapping so it will be its own unique type.

blanchco: will do!

Tom-Szendrey (approved):

blanchco:
closing 😛, not necessary at this point

PR #5416, opened 2024-11-07
clearer unsaved changes message in model config

shawnyama:
# Description

- User will know if they will lose changes to name and description now. Also I decided to make the wording more concise and clear
- As a side cleanup I changed `originalModelConfig` to a regular variable instead of a `ref`

https://github.com/user-attachments/assets/ba14ca6c-f4b0-4c29-afe1-422f6f4647a0

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
@@ -63,7 +63,7 @@
 				</template>
 			</tera-slider-panel>
 		</template>
-		<tera-drilldown-section :is-loading="initializing" :tabName="ConfigTabs.Wizard" class="px-3 mb-10">
+		<tera-drilldown-section :is-loading="initializing" :tabName="DrilldownTabs.Wizard" class="px-3 mb-10">
 			<template #header-controls-left>
 				<tera-toggleable-input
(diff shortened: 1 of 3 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -639,15 +634,28 @@ const onSelectConfiguration = async (config: ModelConfiguration) => {
...
+	if (!isModelConfigValuesEqual(knobs.value.transientModelConfig, originalConfig)) lostItems.push('values');
+	if (knobs.value.transientModelConfig.name !== originalConfig?.name) lostItems.push('name');
+	if (knobs.value.transientModelConfig.description !== originalConfig?.description) lostItems.push('description');
+
+	let lostItemsStr = '';
+	if (lostItems.length > 1) {
+		lostItemsStr = `${lostItems.slice(0, -1).join(', ')} and ${lostItems.slice(-1)}`;
+	} else {
+		lostItemsStr = lostItems[0];
+	}
```
mwdchang, line 654: - There are a handful of `knobs.value.transientModelConfig` repeated here, can you you stick that into a variable so the code is easier to read?
- The 'and-string concatenation`, let's put that into util/text.ts, also you seem to be missing the case to handle an empty array.

mwdchang (approved):

PR #5422, opened 2024-11-07
send user id for workflow update client-events

mwdchang:
### Summary
This is an attempt to get around workflow collab/sync problems, where slowness in server response can cause a modeler's prior action to overwrite their current action.

We can kind of get around this problem, by selectively updating the UI with only events not from the same modeler.


Tom-Szendrey (approved):
sweet

blanchco (approved):

PR #5424, opened 2024-11-07
UI change - Input field error background

asylves1:
# Description

Design Change
- Add an background highlight for blank inputs on Model Config

https://github.com/user-attachments/assets/67d27820-7bd9-49eb-9842-0757f6d90359

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
@@ -1,5 +1,5 @@
 <template>
-	<div class="parameter-entry flex flex-column flex-1">
+	<div class="parameter-entry flex flex-column flex-1" :class="{ 'empty-input': isInputEmpty }">
 		<header>
 			<div class="flex">
 				<strong>{{ parameterId }}</strong>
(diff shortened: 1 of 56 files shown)


asylves1:
Will open a different PR for  empty input count and the left border highlighting

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -148,6 +149,7 @@
 						:model-configurations="filteredModelConfigurations"
 						:mmt="configuredMmt"
 						:mmt-params="mmtParams"
+						:error-empty="true"
```
mwdchang, line 152: Probably double check with @shawnyama here, it seems to me like we are unnecessarily bubbling up `errorEmpty`, which is to say start lower and thus not needed for
- tera-parameter-table
- tera-initial-table
- tera-model-config-drilldown

Arguably we don't need the props for
- tera-initial-entry
- tera-parameter-entry 
Either, but I am a bit out of the look what exactly what is the change request at hand ...

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -42,12 +42,12 @@
 						:options="distributionTypeOptions()"
 						class="mr-3"
 					/>
-
 					<!-- Constant -->
 					<tera-input-number
 						v-if="getParameterDistribution(modelConfiguration, parameterId).type === DistributionType.Constant"
 						label="Constant"
 						:model-value="getParameterDistribution(modelConfiguration, parameterId)?.parameters.value"
+						:error-empty="errorEmpty"
```
shawnyama, line 50: ya I'd just assign them to true here it'll always be like that. 

Side tip: you can put `error-empty` as a short form for`:error-empty="true"`

YohannParis (approved):

packages/client/hmi-client/src/components/model/petrinet/tera-parameter-entry.vue
```diff
@@ -148,7 +150,7 @@ const props = defineProps<{
 	featureConfig?: FeatureConfig;
 }>();
 
-const emit = defineEmits(['update-parameter', 'update-source']);
+const emit = defineEmits(['update-parameter', 'update-source', 'has-empty-field']);
```
mwdchang, line 153: - It seems like `has-empty-field` is not been used?

mwdchang (approved):

PR #5425, opened 2024-11-07
Calibrate ensemble inputs

blanchco:
# Description

* Ensemble calibrate input UI update

https://github.com/user-attachments/assets/08fa3493-d325-4d84-b05f-810f7ed9f632

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
@@ -28,20 +28,25 @@ export interface EnsembleCalibrateExtraCiemss {
 	learningRate: number;
 }
 
+export interface CalibrateEnsembleMapEntry {
+	modelConfigId: string;
+	compartmentName: string; // State or Obs that is being mapped to newName
+}
+
+export interface CalibrateEnsembleWeight {
+	modelConfigurationId: string;
+	value: number;
+}
+
 export interface CalibrateEnsembleMappingRow {
 	name: string;
 	datasetMapping: string;
-	modelConfigurationMappings: { [key: string]: string };
+	modelConfigurationMappings: CalibrateEnsembleMapEntry[];
 }
-
-export interface CalibrateEnsembleWeights {
-	[key: string]: number;
-}
-
 export interface CalibrateEnsembleCiemssOperationState extends BaseState {
 	chartConfigs: string[][];
 	ensembleMapping: CalibrateEnsembleMappingRow[];
-	configurationWeights: CalibrateEnsembleWeights;
+	configurationWeights: CalibrateEnsembleWeight[];
 	timestampColName: string;
 	extra: EnsembleCalibrateExtraCiemss;
 	inProgressCalibrationId: string;
@@ -68,7 +73,7 @@ export const CalibrateEnsembleCiemssOperation: Operation = {
 		const init: CalibrateEnsembleCiemssOperationState = {
 			chartConfigs: [],
 			ensembleMapping: [],
-			configurationWeights: {},
+			configurationWeights: [],
 			timestampColName: '',
 			extra: {
 				solverMethod: speedPreset.method,
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
@@ -29,7 +29,7 @@ export interface EnsembleCalibrateExtraCiemss {
 }
 
 export interface CalibrateEnsembleMappingRow {
-	name: string;
+	newName: string;
 	datasetMapping: string;
 	modelConfigurationMappings: { [key: string]: string };
 }
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
@@ -1,18 +1,34 @@
+import { CiemssMethodOptions } from '@/services/models/simulation-service';
+import { CiemssPresetTypes } from '@/types/common';
 import { Operation, WorkflowOperationTypes, BaseState } from '@/types/workflow';
-import type { EnsembleModelConfigs } from '@/types/Types';
 import calibrateEnsembleCiemss from '@assets/svg/operator-images/calibrate-ensemble-probabilistic.svg';
 
 const DOCUMENTATION_URL = 'https://github.com/ciemss/pyciemss/blob/main/pyciemss/interfaces.py#L156';
 
 export interface EnsembleCalibrateExtraCiemss {
-	solverMethod: string;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -5,137 +5,178 @@
 		@on-close-clicked="emit('close')"
 		@update-state="(state: any) => emit('update-state', state)"
 	>
-		<section :tabName="CalibrateEnsembleTabs.Wizard">
-			<tera-drilldown-section class="ml-3 mr-2 pt-3">
-				<template #header-controls-right>
-					<Button :disabled="isRunDisabled" label="Run" icon="pi pi-play" @click="runEnsemble" />
-					<tera-pyciemss-cancel-button class="mr-auto" :simulation-run-id="cancelRunId" />
+		<section tabName="Wizard" class="wizard">
```
Tom-Szendrey, line 8: Opportunity to use `import { DrilldownTabs } from '@/types/common';`

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -5,137 +5,178 @@
...
+										<tera-input-text disabled model-value="ADAM" />
+									</div>
+								</div>
+							</fieldset>
+						</AccordionTab>
+					</Accordion>
+				</template>
+			</tera-slider-panel>
 		</section>
 		<section :tabName="CalibrateEnsembleTabs.Notebook">
```
Tom-Szendrey, line 175: If you are going to change over to `DrilldownTabs` you can use it here as well

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -15,7 +15,7 @@
 					selectedRun: props.node.state.forecastRunId,
 					selectedVariable: config
 				}"
-				:mapping="props.node.state.ensembleConfigs as any"
+				:mapping="props.node.state.ensembleMapping as any"
```
Tom-Szendrey, line 18: Does this break the chart now that you have updated the typing here?
This chart is sure to be updated but in the meantime you may need to use the util function you created?

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
blanchco: Done!

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
blanchco: added that transform fn here

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
```diff
@@ -1,18 +1,47 @@
...
 	numIterations: number;
+	endTime: number;
+	stepSize: number;
+	learningRate: number;
+}
+
+export interface CalibrateEnsembleMappingRow {
+	name: string;
+	datasetMapping: string;
+	modelConfigurationMappings: { [key: string]: string };
```
Tom-Szendrey, line 34: Already discussed this but just so no one merges your PR while youre making a change
If a user plugs in the same model config they will have the same key
I doubt that its a huge deal but if you change this to a _type_[] then you can use the index instead

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
```diff
@@ -1,18 +1,47 @@
...
+	presetType: CiemssPresetTypes;
+	solverMethod: CiemssMethodOptions;
 	numIterations: number;
+	endTime: number;
+	stepSize: number;
+	learningRate: number;
+}
+
+export interface CalibrateEnsembleMappingRow {
+	name: string;
```
shawnyama, line 32: If you look at @Tom-Szendrey `SimulateEnsembleMappingRow` he has it as `newName`. See about settling for similar naming between the operators for consistency

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
mwdchang: What are the expected key-value of of `modelConfigurationMappings`? If the keys are variable name/id that may create issues.

Using indices as Tom suggest may be okay, but can be temperamental considering that the additional input may be added and existing ones unplugged, so the index positions will shift ...

blanchco: Originally I was using maps for this and the key was the `ModelConfiguration.id` which was mapped to its respective value.  I've switched it over to using lists.

The issue was that this object represents each row in that "table" on the front end.  If 2 of the same `model configurations` (same id) were plugged in both model configurations would share the same value.



In the example below if we had 2 model Antwerps plugged in the model config if we selected something from one dropdown it would reflect in the other.  When I switch to using the list this fixes it.  I hadn't thought of the using the same model config twice on the same node 😬 .
Screenshot 2024-11-07 at 5 17 35 PM

mwdchang: Gotcha.

I think I would rather have
- Prevent nonsensical things from happening in the first place and have a simpler data structure, rather than
- Leveraging a complex data structure to allow weird events to happen that goes nowhere

It makes maintenance easier in the long run and less errors.

As I mentioned above, indices are very annoying to work with when the elements in the list change their absolute or relative positions 😬

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
blanchco: @Tom-Szendrey I've switch to using newName for consistency sake

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-ciemss-operation.ts
blanchco: https://github.com/DARPA-ASKEM/terarium/issues/5443 new ticket for plugging in same model configs

PR #5431, opened 2024-11-08
5415 reduce dev deployment and fix local dev

dgauldie:
# Description

Resolves #(issue)


containers/docker-compose-local.yml
```diff
@@ -1,6 +1,9 @@
 ---
 include:
   - scripts/docker-compose-common.yml
-  - scripts/docker-compose-taskrunner.yml
   - scripts/docker-compose-data-services.yml
   - scripts/docker-compose-sim-services.yml
+################################################
+# Uncomment this to run any taskrunner locally #
+################################################
+#  - scripts/docker-compose-taskrunner.yml
```
YohannParis, line 9: Please add this to the README as well.

containers/scripts/docker-compose-hmi.yml
YohannParis, line 1: Should we just remove those comments instead of updating them?

packages/server/src/main/resources/application.properties
```diff
@@ -133,7 +133,7 @@ server.error.include-stacktrace=never
 ########################################################################################################################
 # AWS credentials configuration
 ########################################################################################################################
-aws-url=http://s3.t1.uncharted.software
+aws-url=
```
YohannParis, line 136: Can we remove it instead of empty, if it is defined in all other profiles?

YohannParis (approved):

packages/server/src/main/resources/application.properties
dgauldie: it needs to be empty for aws to work

containers/scripts/docker-compose-hmi.yml
dgauldie: removed

dvince2 (approved):

PR #5432, opened 2024-11-08
5350 bug cant save a dataframe as a dataset for reuse add cell button not working yellow header always there part 3

### LINKED ISSUE #5350: [BUG]: Can't save a dataframe as a dataset for reuse, Add cell button not working, yellow header always there

mloppie:
# There are really 3 bugs in https://github.com/DARPA-ASKEM/terarium/issues/5350, I am submitting the fixes for them separately for clarity and ease of review

* Bubbles a dirty flag when the code in the editor changes to show the rerun banner at the appropriate time

Resolves #5350

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
 		"husky": "8.0.3",
 		"lint-staged": "15.2.10",
 		"prettier": "3.3.3",
-		"prettier-plugin-java": "2.6.4",
-		"typescript": "5.4.5"
+		"prettier-plugin-java": "2.6.4"
 	}
 }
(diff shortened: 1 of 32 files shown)


YohannParis (approved):
LGTM

mwdchang (approved):
thx

PR #5447, opened 2024-11-08
ability allow only 1 unique asset id in an input

blanchco:
# Description

* Adding a `unique` flag to the inputs so that we can check if we can add multiple of that value to the target port.
* Allows us to only connect unique model configs (or any other asset ids) into target ports

Edit
* Now a user can define whether or not the inputs are unique at an operator level by using the `uniqueInputs` key


YohannParis (approved):
LGTM, but should the uniqueness of an input be managed byt the node directly, and we should not have to precise that on the emit add input port.

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -322,6 +323,11 @@ export class WorkflowWrapper {
 			return;
 		}
 
+		// check if the port value is unique, if so, we should not connect the incoming edge
+		if (targetNode.inputs.some((input) => input.unique && input.value?.[0] === sourceOutputPort.value?.[0])) {
+			return;
+		}
```
mwdchang, line 329: A little weird as Yohann had pointed out, as the control would be better modelled at the operator level - this might also have problems with placeholder `nulls` from a general sense.

blanchco: Sure - I'll look into a way to do this at the operator level

blanchco: @YohannParis @mwdchang 
Okay, this should be at the operator level now. I've also added a check to see if the input value is `null` so we should be fine on that front as well.

YohannParis (approved):

mwdchang (approved):

PR #5449, opened 2024-11-11
Project search assets

kbirk:


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/elasticsearch/ElasticsearchService.java
```diff
@@ -471,6 +471,34 @@ public <T, Partial> void update(final String index, final String id, final Parti
 		}
 	}
 
+	/**
+	 * Update a document from an index.
+	 *
+	 * @param index The index to remove the document from
+	 * @param id    The id of the document to remove
```
YohannParis, line 478: Your JavaDoc is missing `partial` and `routing`.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
```diff
@@ -205,10 +327,203 @@ public List<ProjectDocument> searchProjectsForUser(
...
+									index = i;
+									break;
+								}
+							}
+							if (index != -1) {
+								projectDoc.getAssetEmbeddings().set(index, projectAssetEmbedding);
+							} else {
+								projectDoc.getAssetEmbeddings().add(projectAssetEmbedding);
+							}
+						}
```
YohannParis, line 386: Why not deleting the embeddings and adding the new generated ones?
I'm not sure I understand, maybe it's an ES logic to do it this way.

YohannParis (approved):
LGTM

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetServiceWithSearch.java
```diff
@@ -364,4 +380,39 @@ public void updateIndexInPlace() throws IOException {
 		// refresh the index
 		elasticService.refreshIndex(getAssetAlias());
 	}
+
+	protected boolean isRunningTestProfile() {
```
dvince2, line 384: I'm starting to see this method copied and pasted around more and more (I just had to use it for checking test with the DKG import). I wonder if we should centralize this....

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/elasticsearch/ElasticsearchService.java
kbirk: Done

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/ProjectSearchService.java
kbirk: The embeddings are part of a larger document. So we update them in place and overwrite the entire document.

PR #5451, opened 2024-11-11
5414 hook up local dev to static es index on s3

### LINKED ISSUE #5414: Hook up local dev to static ES index on S3
https://github.com/DARPA-ASKEM/terarium/pull/5362

This PR was errantly merged. Point local dev to a static ES index on S3 to grab the DKG info

dvince2:
# Description

* Checking for the existence of an index.
* If one is not found [this would only be in the case of local] its made then populated.
* The mapping and data come from a static dump on s3 (actual s3 not a local minio)
* Based off of pantera's implementation

Resolves #5414


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/DownloadStaticESIndexService.java
```diff
@@ -0,0 +1,134 @@
...
+@RequiredArgsConstructor
+public class DownloadStaticESIndexService {
+
+	private final S3ClientService s3ClientService;
+	private final ElasticsearchService elasticsearchService;
+
+	@Value("${terarium.static-index-path}")
+	private String staticIndexPath;
+
+	private static final String EPI_ROOT = "epi_dkg_20241030";
```
YohannParis, line 40: Should this be in the application properties for ease of update? And have potential different version based on environment?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/DownloadStaticESIndexService.java
```diff
@@ -0,0 +1,134 @@
...
+	private static final String AWS_ID = "static-index";
```
YohannParis, line 41: Should use `terarium.file-storage-s3-client-name` from application properties

YohannParis (approved):
LGTM, but this is still above my head. I understand what is happening, but I'm not sure about some naming convention that makes it hard to read.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/DownloadStaticESIndexService.java
dvince2: Yes it should!  However, this is happening during the `@PostConstruct` phase which apparently is before the configuration runs and grabs values from application.properties, so these all come in as `null`.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/s3/DownloadStaticESIndexService.java
dvince2: This is happening during the `@PostConstruct` phase which apparently is before the configuration runs and grabs values from application.properties, so these all come in as `null`.

kbirk (approved):

PR #5458, opened 2024-11-11
Ensemble Simulate - same typing as ensemble calibrate

Tom-Szendrey:
# Description
- Ensemble simulate now has the same mapping type as the calibration does.
- Now requires unique inputs
- provide a type for `modelConfigIdToNameMap`
- correct state updater via knobs as `watcher` should be `deep` not `immediate`

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss-drilldown.vue
@@ -40,11 +40,11 @@
 											<td>
 												<tera-input-text v-model="ele.newName" auto-focus class="w-full" placeholder="Add a name" />
 											</td>
-											<td v-for="(row, indx) in ele.modelConfigurationMappings" :key="indx">
+											<td v-for="key in Object.keys(ele.modelConfigurationMappings)" :key="key">
 												<Dropdown
 													class="w-full"
-													:options="allModelOptions[row.modelConfigId]"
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-ciemss-drilldown.vue
```diff
@@ -40,11 +40,11 @@
 											<td>
 												<tera-input-text v-model="ele.newName" auto-focus class="w-full" placeholder="Add a name" />
 											</td>
-											<td v-for="(row, indx) in ele.modelConfigurationMappings" :key="indx">
+											<td v-for="key in Object.keys(ele.modelConfigurationMappings)" :key="key">
```
mwdchang, line 43: `Object.keys(...)` does not guarantee order, does it matter here or do we need things to be consistent? 

It might mean that if current current order is `[a, b, c]` and you added `d`, it might become "[a, d, b, c]" instead of "[a, b, c, d]"

Tom-Szendrey: Great catch! Thank you
This may have caused user confusion
I will update this

Tom-Szendrey: I am not using `modelConfigurationIds` for this which is a list of the model configs from the inputs

mwdchang (approved):

PR #5460, opened 2024-11-11
Model config accordion empty input

asylves1:
# Description

Design Change:

- the grey bar on the left of the input will change to the error color style if there is an issue with the input
- also extracted empty inputs will be `NaN` by default in the model config parameter table

image

Resolves #(issue)


packages/client/hmi-client/src/components/model/petrinet/tera-initial-entry.vue
```diff
@@ -87,13 +92,19 @@ const emit = defineEmits(['update-expression', 'update-source']);
...
 const concept = ref('');
 const sourceOpen = ref(false);
 const showOtherConfigValueModal = ref(false);
 
 const getOtherValuesLabel = computed(() => `Other Values(${otherValueList.value?.length})`);
 
+function onExpressionChange(value) {
+	isExpressionEmpty.value = isNumberInputEmpty(value);
+	emit('update-expression', { id: props.initialId, value });
+}
```
mwdchang, line 106: If the expression is empty, which I assume to be invalid ... do we still want to emit an update with the invalid value?

likewise similar logic in `onParameterChange`

asylves1: change made

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -537,6 +522,32 @@ const initializing = ref(false);
 const isFetching = ref(false);
 const isLoading = ref(false);
 
+const amrInitials = ref();
+const amrParameters = ref();
+
+const getMissingInputsMessage = (amount, total) => {
+	const precent = (amount / total) * 100;
```
mwdchang, line 529: nitpic: typo "percent"?

mwdchang (approved):

shawnyama (approved):

PR #5461, opened 2024-11-12
fix: paginate model entries for better performance

shawnyama:
# Description

- Adding pagination to the model parts is what I settled for saving memory. UI wise I think it's good too since the page can get pretty long
- Added filters to the Accordion tabs to search within the model parts
- I tried using `tera-toggleable-input` to see if not always having an active input box will help out but it didn't make any noticeable difference in memory usage

https://github.com/user-attachments/assets/fb6b6bd5-f237-445c-8c4e-4da68eff0daa

Small cleanups:
- While adjusting `tera-model-part.vue` to support showing the filtered items I realized that there was no point in the watcher we have in there
- Added a scroll to the model equations since that can get huge

### Benchmark comparison
* Models that have 10-20 states typically uses 400 - 500mb in both versions
* A large model typically uses 1.4 - 1.8GB in this branch, currently in main it uses 2.3 - 3.3GB

Side note:
Adding pagination saved some memory but the model diagram and equations are taking up a lot too, the page will crash if you open the expanded view of [this model ](https://app.staging.terarium.ai/projects/04c55de2-c216-45c2-96dc-a603889a8a9d/model/336db118-bf95-4164-aa60-0f1d5fb2c485)


packages/client/hmi-client/src/components/model/petrinet/tera-initial-table.vue
```diff
@@ -122,7 +122,7 @@ ul {
 	flex-grow: 1;
 	padding-left: var(--gap-1);
 
-	li {
+	& li {
```
YohannParis, line 125: That's interesting that you need to add the ampersand 🤔

YohannParis (approved):
Great job, and neatily done!

packages/client/hmi-client/src/components/model/petrinet/tera-petrinet-parts.vue
```diff
@@ -3,60 +3,74 @@
...
+				:filter="transitionsFilter"
 				@update-transition="emit('update-transition', $event)"
 			/>
 		</AccordionTab>
 		<AccordionTab>
 			<template #header>
-				Time
-				<span class="artifact-amount">({{ time.length }})</span>
+				Time <span class="artifact-amount">({{ time.length }})</span>
+				<tera-input-text class="ml-auto" placeholder="Filter" v-model="timeFilter" />
```
mwdchang, line 65: I don't think we need a filter here - a situation where we have many time parameters is more like an error-case. Let's remove this.

packages/client/hmi-client/src/components/model/petrinet/tera-initial-table.vue
shawnyama: I think it still may work but the colouring is off in my editor. From what I remember having the & is the "right way" to do it

packages/client/hmi-client/src/components/model/model-parts/tera-model-part.vue
```diff
@@ -1,25 +1,29 @@
 <template>
 	<ul>
-		<li v-for="({ base, children, isParent }, index) in items" :key="index" class="model-part">
-			<template v-if="isParent && !isEmpty(parentEditingState)">
+		<li
+			v-for="({ base, children, isParent }, index) in filteredItems.slice(firstRow, firstRow + MAX_NUMBER_OF_ROWS)"
+			:key="index"
+			class="model-part"
+		>
```
mwdchang, line 7: I am apprehensive when we have a dynamic list and we use `:key=index` - we have had run into reactive issues where things doesn't work because while the underlying element changed, the index did not.

Is it possible to key by something else? for example the parameter/state ids should be unique.

We can probably still keep `getEditingState(index)` stuff, but vue binding/cache is somewhat out of our control so we should be careful around that.

packages/client/hmi-client/src/components/model/model-parts/tera-model-part.vue
```diff
@@ -101,19 +105,37 @@
...
-				</ul>
+				<div class="stratified" v-if="getEditingState(index).showChildren">
+					<ul>
+						<li
+							v-for="(child, j) in children.slice(
+								getEditingState(index).firstRow,
+								getEditingState(index).firstRow + MAX_NUMBER_OF_ROWS
+							)"
+							:key="j"
+						>
```
mwdchang, line 120: Ditto regarding key=index

https://github.com/DARPA-ASKEM/terarium/pull/5461/files#r1841079508

mwdchang (approved):

PR #5469, opened 2024-11-12
Rework Model, Dataset and Document models to use the same underlying dkg

### LINKED ISSUE #5468: Rework Model, Dataset and Document models to use the same underlying DKG representation

dvince2:
# Description

Using the `DKG` representation for groundings instead of `Identifier`.  Note I'm not going to go through the effort of renaming the fields from `identifier` as it will require a migration and substantial front end renaming which feels like a lot of effort for 0 gain.

Resolves #5468


Tom-Szendrey (approved):

mwdchang (approved):
Seems fine

PR #5476, opened 2024-11-12
Refactor charts and settings across calibrate, optimize and simulate nodes and drilldowns

jryu01:
# Description

Refactored charts and settings across the calibrate, optimize, and simulate nodes and drilldowns to reduce duplicate code.

- Extracted logic for managing chartSettings state into a reusable composable, `useChartSettings.ts`.
- Extracted chart specification preparation logic for each chart setting type into `useCharts.ts`.
- Moved chart data preparation logic for each operator into separate utility files, allowing shared use between nodes and drilldowns.


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts
```diff
@@ -76,3 +79,49 @@ export function getErrorData(
...
 	return errors;
 }
+
+export const mapModelVarToDatasetVar = (state: CalibrationOperationStateCiemss, modelVariable: string) => {
+	if (modelVariable === 'timepoint_id') return state.timestampColName ?? '';
+	return state.mapping.find((d) => d.modelVariable === modelVariable)?.datasetVariable || '';
+};
+
+export function usePreparedChartInputs(
+	props: {
```
Tom-Szendrey, line 89: It looks like this does not need this entire
```
 props: {
		node: WorkflowNode<CalibrationOperationStateCiemss>;
	}
```, but instead just utilizes the `calibrationId` and the `mapping` 

It might be nice to reduce the scope of what we are passing to this

Comment (YohannParis) [reply packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts:89]: I do not understand the issue, what is wrong to change to two props instead of one? they will be reactive as well.

File: packages/client/hmi-client/src/composables/useChartSettings.ts
Comment (YohannParis) [inline packages/client/hmi-client/src/composables/useChartSettings.ts:1]: This needs a lot of comments, as it is not self explanatory on what is happening.

File: packages/client/hmi-client/src/composables/useCharts.ts
Comment (YohannParis) [inline packages/client/hmi-client/src/composables/useCharts.ts:1]: Same here, some comment to be used as documentation would be great. Feel free to use an LLM for it.

File: packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts (hunk 1 above)
Comment (jryu01) [reply packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts:89]: It's either passing a single props as an argument vs multiple ractive/computed properties. 
```
usePreparedChartInputs(mapping: Ref<TypeofMapping>, state: Ref<typeofState>) { ... }

# In parent component
usePreparedChartInputs(computed(() => props.node.mapping), computed(() => props.node.state))
```
vs
```
usePreparedChartInputs(props: TypeofProps ) { ... }

# In parent component
usePreparedChartInputs(props)
```

File: packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue [hunk 2]
```diff
@@ -735,285 +715,58 @@ const lossChartSize = useDrilldownChartSize(lossChartContainer);
...
+	useParameterDistributionCharts
+} = useCharts(
+	props.node.id,
+	model,
+	modelConfig,
+	preparedChartInputs,
+	chartSize,
+	computed(() => interventionPolicy.value?.interventions ?? [])
+);
+const toDatasetVar = (modelVar: string) => mapModelVarToDatasetVar(props.node.state, modelVar);
```
Tom-Szendrey, line 764: I am not 100% sure if this is exclusively used for output?

If this is then I suspect we want to use the mapping from our output similar to my other comment

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -150,118 +142,26 @@ async function updateLossChartWithSimulation() {
...
+const { selectedVariableSettings, selectedInterventionSettings } = useChartSettings(props, emit);
+const { useInterventionCharts, useVariableCharts } = useCharts(
+	props.node.id,
+	model,
+	modelConfiguration,
+	preparedChartInputs,
+	toRef(chartSize),
+	computed(() => interventionPolicy.value?.interventions ?? [])
+);
+const toDatasetVar = (modelVar: string) => mapModelVarToDatasetVar(props.node.state, modelVar);
```
Tom-Szendrey, line 162: Same as my previous comment

If this is only being used for output I suspect we want to use the output's state's mapping here

Tom-Szendrey (approved):
Thank you for this!!

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
jryu01: Done. I've change the codes so that we grab the mapping from output settings

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
jryu01: Done

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-utils.ts
```diff
@@ -0,0 +1,30 @@
+import _ from 'lodash';
+import { computed, Ref } from 'vue';
+import { WorkflowNode } from '@/types/workflow';
+import { DataArray } from '@/services/models/simulation-service';
+import { SimulateCiemssOperationState } from './simulate-ciemss-operation';
+
+export function usePreparedChartInputs(
```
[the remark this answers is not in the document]
jryu01: For simulate, how we save the async results are a bit different than other two operator so that it was easier to pass pyciemssMap as a props. In order to make it similar to other two operator, more refactoring is needed on the other parts as well. Not sure if it's worth doing it but we can revisit this later.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-utils.ts
```diff
@@ -76,3 +79,49 @@ export function getErrorData(
...
+	},
+	runResult: Ref<DataArray>,
+	runResultSummary: Ref<DataArray>,
+	runResultPre: Ref<DataArray>,
+	runResultSummaryPre: Ref<DataArray>
+) {
+	const pyciemssMap = computed(() => (!runResult.value.length ? {} : parsePyCiemssMap(runResult.value[0])));
+
+	return computed(() => {
+		const state = props.node.state;
```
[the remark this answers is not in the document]
jryu01: I think this is fine since the output state is loaded as active state in this case and we are fetching output data with active state properties. 

We have bigger problem though. It seems once output states are saved and appended as new output, they are not updated afterward. Any state changes made on the active state don't get saved back to the state for the corresponding output. Which is a problem because once chart settings are saved after the run, you can no longer change it. When the output is on active state, you can modify it temporarily as long as it's active but when you select to different output and back to the original output, the new changes are gone and you are back to the original state that's created when you ran it.   

I think we should address the issue in a separate task.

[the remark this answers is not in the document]
Tom-Szendrey: It was my understanding that the output state was intended to be more static as it is though I may be mistaken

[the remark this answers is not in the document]
jryu01: Yes, it seems fine as it is for now, but it might lead to unexpected behaviour when updating chart settings while switching between outputs. I think it should be okay for now unless someone complains... 😅  It's been like this since we introduced the output settings panel.

mwdchang (approved):

PR #5479, opened 2024-11-12
Add a basic SIR model for stratification tests

liunelson:
# Description

New model AMR json for stratification manual tests.


YohannParis:
There is one already in the Google Drive referenced in the different tests.
We should not save random assets in the repo.

mwdchang:
@liunelson I uploaded the file onto the shared google drive

```
/testing/manual tests/stratify model/data
```

PR #5489, opened 2024-11-13
New intervention not working

Tom-Szendrey:
# Description
Previously the workflow for creating a new intervention is that we would POST a blank intervention and any updates would then allow the user to update this newly created intervention.
This is not a valid workflow as the blank intervention is not a valid intervention (It has no timestep and no value) thus it fails HMIServer's validation.

The new workflow is that we will reset the interface on the frontend and wait for the user to hit `Save` or `Save as...` before trying to POST this intervention 

https://github.com/user-attachments/assets/b273d284-f482-4bca-8e6c-d3749facb795

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
@@ -99,7 +99,7 @@ const onBlur = () => {
 watch(
 	() => props.modelValue,
 	(newValue) => {
-		if (isNaN(newValue)) return;
+		if (isNaN(newValue)) maskedValue.value = '';
 		maskedValue.value = newValue?.toString() ?? '';
 	},
 	{ immediate: true }
--- a/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
@@ -100,7 +100,7 @@ watch(
 	() => props.modelValue,
 	(newValue) => {
 		if (isNaN(newValue)) maskedValue.value = '';
-		maskedValue.value = newValue?.toString() ?? '';
+		else maskedValue.value = newValue?.toString() ?? '';
 	},
 	{ immediate: true }
 );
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
```diff
@@ -556,11 +548,16 @@ const onSaveInterventionPolicy = async () => {
...
 };
 
-const createNewInterventionPolicy = () => {
-	if (!model.value?.id) return;
-	showCreatePolicyModal.value = true;
-	newBlankInterventionPolicy.value.modelId = model.value.id;
-	showSaveModal.value = true;
+const resetInterventionPolicy = () => {
+	knobs.value.transientInterventionPolicy = {
+		modelId: '',
```
blanchco, line 553: You will need to define the model Id here, otherwise when you save the policy, it will not be associated with the model

blanchco (commented):
Looking at the preview, it looks like there are still some values that are persisting after you press the "Create New" button, I think these _should_ be blank.

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
Tom-Szendrey: Thanks

packages/client/hmi-client/src/components/widgets/tera-input-number.vue
```diff
@@ -99,7 +99,7 @@ const onBlur = () => {
 watch(
 	() => props.modelValue,
 	(newValue) => {
-		if (isNaN(newValue)) return;
+		if (isNaN(newValue)) maskedValue.value = '';
```
Tom-Szendrey, line 102: @blanchco can you confirm that this is okay before this gets merged?

mwdchang (approved):

shawnyama (approved):

blanchco (approved):

PR #5490, opened 2024-11-13
scenario templates

blanchco:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/project/tera-resource-sidebar.vue
+++ b/packages/client/hmi-client/src/components/project/tera-resource-sidebar.vue
@@ -55,13 +55,22 @@
 						</header>
 						<!-- New asset buttons for some types -->
 						<Button
-							v-if="type === AssetType.Model || type === AssetType.Workflow"
+							v-if="type === AssetType.Model"
 							class="new-button"
 							icon="pi pi-plus"
 							label="New"
 							size="small"
 							text
 							@click.stop="emit('open-new-asset', type)"
+						/>
+						<Button
+							v-else-if="type === AssetType.Workflow"
+							class="new-button"
+							icon="pi pi-plus"
+							label="New"
+							size="small"
+							text
+							@click.stop="emit('open-new-workflow', type)"
 						/>
 					</div>
 				</template>
@@ -163,7 +172,7 @@ defineProps<{
 	assetId: string;
 }>();
 
-const emit = defineEmits(['open-asset', 'remove-asset', 'open-new-asset']);
+const emit = defineEmits(['open-asset', 'remove-asset', 'open-new-asset', 'open-new-workflow']);
 
 const overview = { assetId: '', pageType: ProjectPages.OVERVIEW };
 
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
@@ -1,0 +1,20 @@
+import { Workflow } from '@/types/workflow';
+
+export abstract class BaseScenarioTemplate {
+	templateId: string;
+
+	templateName: string;
+
+	description: string;
+
+	workflowName: string;
+
+	constructor(nameId: string, name: string, description: string) {
+		this.templateId = nameId;
+		this.templateName = name;
+		this.description = description;
+		this.workflowName = '';
+	}
+
+	abstract createWorkflow(): Workflow;
+}
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-template.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-template.ts
@@ -1,0 +1,80 @@
+import { BaseScenarioTemplate } from '@/components/workflow/scenario-templates/scenario-template';
+import * as workflowService from '@/services/workflow';
+import { operation as ModelOp } from '@/components/workflow/ops/model/mod';
+import { operation as ModelConfigOp } from '@/components/workflow/ops/model-config/mod';
+import { operation as CalibrateCiemssOp } from '@/components/workflow/ops/calibrate-ciemss/mod';
+import { operation as DatasetOp } from '@/components/workflow/ops/dataset/mod';
+import { OperatorNodeSize } from '@/services/workflow';
+
+export class SituationalAwarenessScenarioTemplate extends BaseScenarioTemplate {
+	modelSpec: { id: string };
+
+	datasetSpec: { id: string };
+
+	modelConfigSpec: { id: string };
+
+	calibrateSpec: { ids: string[] };
+
+	constructor() {
+		super('situational-awareness', 'Situational Awareness', 'Template for situational awareness');
+		this.modelSpec = {
+			id: ''
+		};
+		this.datasetSpec = {
+			id: ''
+		};
+		this.modelConfigSpec = {
+			id: ''
+		};
+		this.calibrateSpec = {
+			ids: []
+		};
+	}
+
+	createWorkflow() {
+		const wf = new workflowService.WorkflowWrapper();
+
+		// Model
+		wf.addNode(
+			ModelOp,
+			{ x: 0, y: 0 },
+			{
+				size: OperatorNodeSize.medium
+			}
+		);
+
+		// Dataset
+		wf.addNode(
+			DatasetOp,
+			{ x: 0, y: 0 },
+			{
+				size: OperatorNodeSize.medium
+			}
+		);
+
+		// Model Configuration
+		wf.addNode(
+			ModelConfigOp,
+			{ x: 0, y: 0 },
+			{
+				size: OperatorNodeSize.medium
+			}
+		);
+
+		// Calibrate
+		wf.addNode(
+			CalibrateCiemssOp,
+			{ x: 0, y: 0 },
+			{
+				size: OperatorNodeSize.medium
+			}
+		);
+
+		const workflow = wf.dump();
+		workflow.name = this.workflowName;
+
+		return workflow;
+	}
+}
+
+export const SituationalAwarenessTemplate = new SituationalAwarenessScenarioTemplate();
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
@@ -1,0 +1,152 @@
+<template>
+	<div>
+		<h6>What's like to happen next?</h6>
+		<p>
+			Calibrates the model to historical data to obtain the best estimate of parameters for the present, then forecasts
+			into the near future.
+		</p>
+	</div>
+	<div>
+		<h6>Examples</h6>
+		<ul>
+			<li>Some example</li>
+		</ul>
+	</div>
+
+	<div>
+		<label>What would you like to call this workflow?</label>
+		<tera-input-text
+			:model-value="state.workflowName"
+			@update:model-value="
+				emit('update-state', {
+					workflowName: $event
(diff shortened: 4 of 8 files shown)


packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
```diff
@@ -0,0 +1,23 @@
...
+
+	workflowName: string;
+
+	constructor(nameId: string, name: string, description: string) {
+		this.templateId = nameId;
+		this.templateName = name;
+		this.description = description;
+		const workflows = useProjects().getActiveProjectAssets(AssetType.Workflow);
+		this.workflowName = `workflow ${workflows.length + 1}`;
+	}
```
mwdchang, line 20: This feels a bit awkward
- The workflowName looks like it would be provided from the design sketch, so I don't think we need to generate a name here
- It doesn't really make sense to have both templateId and templateName, it should probably be one or the other, but not both

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
```diff
@@ -0,0 +1,109 @@
...
+
+const onUpdateState = (state: any) => {
+	const scenario = getScenario();
+	if (!scenario) {
+		return;
+	}
+	Object.assign(scenario.template, state);
+};
+
+const getScenario = () => scenarios.value.find((s) => s.template.templateId === selectedTemplateId.value);
```
mwdchang, line 108: If we know it can never be null or undefined, let's just cast it so we don't get silly typescript warnings.

```
const getScenario = () =>  scenarios.find(...) as XYZ
```

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
```diff
@@ -0,0 +1,109 @@
...
+const scenarios = ref<ScenarioItem[]>([
+	{
+		template: new BlankCanvasScenarioTemplate(),
+		component: TeraBlankCanvasTemplate
+	},
+	{
+		template: new SituationalAwarenessScenarioTemplate(),
+		component: TeraSituationalAwarenessTemplate
+	}
+]);
```
mwdchang, line 68: Doesn't seem like there is a need to make scenarios reactive

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
```diff
@@ -0,0 +1,109 @@
...
+						/>
+						<label class="pl-2" :for="scenario.template.templateId">{{ scenario.template.templateName }}</label>
+					</div>
+				</aside>
+				<main class="col-10 flex flex-column gap-3">
+					<component
+						v-if="getScenario()"
+						:is="getScenario()?.component"
+						:state="getScenario()?.template"
+						@update-state="onUpdateState"
```
mwdchang, line 25: I think we should have a chat about a better way to do the plumbing...right now you are only updating the workflowName, but later on we are going to have intervention-policy id/name, config id/name ... etc etc and I suspect things will get quite messy.

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
blanchco: I've updated `templateId` and `templateName` to be static variables on each class as they are used for the radio button list on the left side

blanchco: For workflow name, I can change that...I figure we can just have the same behaviour we have now where the default name is the workflows.length + 1

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
blanchco: I do have this working with the situational awareness template on more than just the workflow name.  It is updating the model, model config, dataset, output parms etc...  We can chat about it tomorrow morning

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
```diff
@@ -0,0 +1,109 @@
...
+
+interface ScenarioItem {
+	template: BaseScenarioTemplate;
+	component: Component;
+}
+
+const scenarios = ref<ScenarioItem[]>([
+	{
+		template: new BlankCanvasScenarioTemplate(),
+		component: TeraBlankCanvasTemplate
```
mwdchang, line 66: There is something a bit odd here, you are doing this in the class file

```
export const BlankCanvasTemplate = new BlankCanvasScenarioTemplate();
```

But you are doing this here:
```
new BlankCanvasScenarioTemplate(),
```

So probably need to get the exports straightened out, I am presuming we do not want to export an instantiation of the scenario.


Also maybe we should drop the "Template" suffix ... not sure if it is really necessary and the names are getting quite long 🤔

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
mwdchang: Yeah I noticed, maybe it might be better to just use `:model` instead of emitting stuff. 
Since we are not going to use these templates beyond the modal it should be safe to just modify them in-place across parent/children.

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
blanchco: Ya was was using that exported instance for some experimental while making this and forgot to delete them, they're deleted now.  And I've removed the Template suffix 😄

packages/client/hmi-client/src/components/project/tera-resource-sidebar.vue
```diff
@@ -55,14 +55,23 @@
...
+							v-if="type === AssetType.Model"
 							class="new-button"
 							icon="pi pi-plus"
 							label="New"
 							size="small"
 							text
 							@click.stop="emit('open-new-asset', type)"
 						/>
+						<Button
+							v-else-if="type === AssetType.Workflow"
```
YohannParis, line 67:
```suggestion
							v-if="type === AssetType.Workflow"
```

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
YohannParis: I agree with @mwdchang, the name should be the one provided by the template.
If I use the template named `Extract model from Document`, I expect my new template to be named the same.

YohannParis: Which make the name not part of the `BaseScenario` class

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-template.ts
```diff
@@ -0,0 +1,82 @@
...
+
+	public static templateName = 'Situational Awareness';
+
+	modelSpec: { id: string };
+
+	datasetSpec: { id: string };
+
+	modelConfigSpec: { id: string };
+
+	calibrateSpec: { ids: string[] };
```
YohannParis, line 20: Why all the spacing, this ain't Java 😝 
```suggestion
	public static templateId = 'situational-awareness';
	public static templateName = 'Situational Awareness';
	modelSpec: { id: string };
	datasetSpec: { id: string };
	modelConfigSpec: { id: string };
	calibrateSpec: { ids: string[] };
```

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
YohannParis, line 1: Why all those `<div>` tags?

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+	<div>
+		<h6>What's like to happen next?</h6>
```
YohannParis, line 3: Is this an English idiome? I do not know this turn of phrase.

YohannParis (commented):
This looks like a really convulated and strange way to create templates.
I'm not part of that feature planning, so I will restrain to approve on this PR.
I will not comment on how the forms layout has been made, but it will need some cleaning up once this is done and fully functional.

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template.ts
blanchco: Sure, I've changed the name to be unique to each template

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
blanchco: Changed to "What's _likely_ to happen next?" 😄

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-template.ts
blanchco: The linter is doing it, I swear!

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
blanchco: Since the instance are being passed into the component, they need to be reactive

packages/client/hmi-client/src/components/workflow/tera-create-workflow-modal.vue
blanchco: I created a number of setters and no warnings.  Still works as expected

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-template.ts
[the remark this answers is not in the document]
mwdchang: Not entirely sure if we are talking about the same thing...

I think the idea is that some scenarios are more complicated to construct, such that a form-based input is more intuitive than creating a workflow in the style of the infinite-canvas.

For simples ones like these, it is about siz-half-a-dozen or slightly more cumbersome.

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
blanchco: This was mostly just to create a quick layout and this will probably change in future PRs.  I would say this PR is mostly focused on the data structures/flow for the foundation of scenario templates

mwdchang (approved):
Thanks, this is a great start

Tom-Szendrey (approved):
sweet

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/tera-situational-awareness-template.vue
```diff
@@ -0,0 +1,121 @@
...
+		if (!model) return;
+		modelConfigurations.value = await getModelConfigurationsForModel(modelId);
+		interventionPolicies.value = await getInterventionPoliciesForModel(modelId);
+
+		const modelOptions: any[] = model.model.states;
+
+		model.semantics?.ode.observables?.forEach((o) => {
+			modelOptions.push(o);
+		});
+		modelStateOptions.value = modelOptions;
```
shawnyama, line 116:
```suggestion
		modelStateOptions.value = [...model.model.states, ...model.semantics?.ode.observables];
```

PR #5492, opened 2024-11-13
5450 create a sample project

YohannParis:
# Description

* Add a boolean `sampleProject` to the Project.java
* Update the controller to allow Admins to make a project a sample project. This will make the project public, then will remove all previous permissions relationship to it. Only Admin can edit it.
* Display sample project in the right tab on the front-end.
* Allow admin to make a project a sample project


packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
```diff
@@ -491,11 +491,15 @@ public void removeRelationship(
...
 		final Schema.Relationship relationship
-	) throws Exception, RelationshipAlreadyExistsException {
+	) throws Exception {
 		userCache.invalidate(who.id);
 		invalidatePermissionCache(who, what);
 		final ReBACFunctions rebac = new ReBACFunctions(channel, spiceDbBearerToken);
-		CURRENT_ZED_TOKEN = rebac.removeRelationship(who, relationship, what);
+		try {
+			CURRENT_ZED_TOKEN = rebac.removeRelationship(who, relationship, what);
+		} catch (RelationshipAlreadyExistsException ignore) {
```
bigglesandginger, line 500: Seriously - RelationshipAlreadyExistsException is the exception for when a relationship does not exist when removing - lol - oh well silliness

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1125,6 +1125,80 @@ public ResponseEntity<JsonNode> makeProjectPublic(
...
+
+			/* Permissions */
+
+			// Getting the project
+			final RebacProject rebacProject = new RebacProject(id, reBACService);
+
+			// Delete all previous relationships attached to the project,
+			final List<Contributor> contributors = projectPermissionsService.getContributors(rebacProject);
+			for (final Contributor contributor : contributors) {
+				final RebacUser rebacUser = new RebacUser(contributor.getUserId(), reBACService);
```
bigglesandginger, line 1174: Need to check if getUserId() returns `null` - that is the signal that the contributor is a group (I guess - been many a passage of time since I wrote that code)

bigglesandginger (dismissed):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
YohannParis: I mean, that's why I ignored it, I want to remove it but it doesn't exist, good for me regardless. I am not going to check if it exist before deleting it?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1125,6 +1125,80 @@ public ResponseEntity<JsonNode> makeProjectPublic(
...
+				rebacUser.removeAllRelationships(rebacProject);
+			}
+
+			// Admin group to write the project
+			final RebacGroup adminGroup = new RebacGroup(ReBACService.ASKEM_ADMIN_GROUP_ID, reBACService);
+			adminGroup.canAdministrate(rebacProject);
```
bigglesandginger, line 1180: Not sure why we would check if the group can administrate - just force the group to be an ADMIN

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/ReBACService.java
bigglesandginger: Exactly - I was laughing at the curious name of the Exception (I created) but it is because the underlying exception is about a failure to create a relationship, even though the request is to delete the relationship

bigglesandginger:
Also, the projects that I set to be `Sample projects` simply appeared in the `Public projects` tab and not `Sample projects`

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
YohannParis: That's the problem with the word `can` is that a test or an affirmation? I will update this.

bigglesandginger: It is meant to be read in an if statement (where it is used elsewhere :) - but the I have always hated the `is` word tack on for boolean checking...

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
YohannParis: I added a test to contributor to check if it's a user or a group and changed wich method is called.
I wish we could simply do this from a RebacObject directly, but that's too much refactoring in one PR.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1125,6 +1125,85 @@ public ResponseEntity<JsonNode> makeProjectPublic(
...
+				} else if (contributor.isGroup()) {
+					final RebacGroup rebacGroup = new RebacGroup(contributor.getName(), reBACService);
+					rebacGroup.removeAllRelationships(rebacProject);
+				}
+			}
+
+			// Admin group to write the project
+			final RebacGroup adminGroup = new RebacGroup(ReBACService.ASKEM_ADMIN_GROUP_ID, reBACService);
+			adminGroup.createAdminRelationship(rebacProject);
+			adminGroup.createWriterRelationship(rebacProject);
```
bigglesandginger, line 1186: In theory WRITER is unnecessary, but not a problem

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1125,6 +1125,85 @@ public ResponseEntity<JsonNode> makeProjectPublic(
...
+			final RebacProject rebacProject = new RebacProject(id, reBACService);
+
+			// Delete all previous relationships attached to the project,
+			final List<Contributor> contributors = projectPermissionsService.getContributors(rebacProject);
+			for (final Contributor contributor : contributors) {
+				if (contributor.isUser()) {
+					final RebacUser rebacUser = new RebacUser(contributor.getUserId(), reBACService);
+					rebacUser.removeAllRelationships(rebacProject);
+				} else if (contributor.isGroup()) {
+					final RebacGroup rebacGroup = new RebacGroup(contributor.getName(), reBACService);
```
bigglesandginger, line 1178: Cannot remove based on name - must be an ID - as there are only two groups in the system might as well do the individually... a hassle but...

YohannParis: I update my removeAllRelationship with a removeAllRelationshipExceptOne

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/rebac/askem/RebacGroup.java
```diff
@@ -42,11 +42,32 @@ public void createAdminRelationship(RebacObject rebacObject) throws Exception, R
...
+		for (Schema.Relationship relationship : relationships) {
+			reBACService.removeRelationship(getSchemaObject(), rebacObject.getSchemaObject(), relationship);
+		}
+	}
+
+	/**
+	 * Remove all relationships between this group and the given object.
+	 * Except the provided relationship, if it doesn't exist, it will be added
+	 */
+	public void removeAllRelationsExceptOne(
```
bigglesandginger, line 57: 5% confusing, 100% get the job done with minimal faffing about - nice!

bigglesandginger:
Added change to Project.java so that sampleProject boolean is actually saved anywhere.

The UI does not show Sample Projects in the `sample project` tab

dvince2 (approved):
I think this looks ok but I will defer to @bigglesandginger on rebac magic

bigglesandginger (approved):

PR #5493, opened 2024-11-13
Minimize `getMmt()` calls

shawnyama:
# Description

- `getMmt` used to be called twice when opening `tera-model`, once for `tera-model-parts` and another time for `tera-model-diagram`
- This sets it up so if `tera-model-diagram` is a child of `tera-model` it will use the `mmtData` from the parent without the need to call for it again
- If the diagram is used by itself it will do the `getMmt` call as usual

Other cleanup:
- I also decided to delete `tera-model-parts` in favour of directly using `tera-petrinet-parts` since we aren't using regnets or stockflows anymore
- Some update logic from `tera-model-parts` was moved to `tera-model`


packages/client/hmi-client/src/components/model/tera-model.vue
```diff
@@ -204,13 +220,56 @@ async function updateModelName() {
 	onSave();
 }
 
+async function updateMMT() {
```
mwdchang, line 223: Since we are cleaning things, maybe a better name is `refreshMMT`

mwdchang (approved):

YohannParis (approved):
LGTM, thanks for the cleaning up

PR #5498, opened 2024-11-13
Ensemble calibrate output type

Tom-Szendrey:
# Description
- Ensemble Calibrate should output a dataset ID
- While we are at it we should also run a pre and post forecast (rather than just post)
- minor simulate ensemble fix


mwdchang (approved):
Didn't test, looks okay

YohannParis (approved):
LGTM

PR #5507, opened 2024-11-14
fix for config saving issue

asylves1:
# Description

Issue: 
- Extracting an inputs from a document in the model config then changing an input value and saving does not save the new value

Testing:
- upload a document and model to a workflow
- create a model config and add the model & document
-  extract the inputs from the document
- make changes to a few of the inputs and save
- check that the inputs have been updated in the new config

Resolves #(issue)


packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -235,3 +237,47 @@ export function getMissingInputAmount(modelConfiguration: ModelConfiguration) {
 	});
 	return missingInputs;
 }
+
+export function getMmtParameters(modelConfiguration, mmt, mmtParams) {
+	const collapsedParameters = collapseParameters(mmt, mmtParams);
+	const parameters = getParameters(modelConfiguration);
+	return Array.from(collapsedParameters.keys()).map((id) => {
```
mwdchang, line 244: Can you elaborate on what we are trying to do here? `collapseParameters` and likewise, `collapseInitials` are used for building matrices - they are heuristically based so do not always give you the right answer, and I think probably not what you are looking for in the context of extracting model-config from a document.

Moreover, if you want to extract information from MMT and the functions do not exist, they would be better placed somewhere in `src/model-representation/mira`

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -235,3 +237,47 @@ export function getMissingInputAmount(modelConfiguration: ModelConfiguration) {
...
+			distribution: {
+				type: DistributionType.Constant,
+				parameters: { value: 0 }
+			},
+			referenceId: id,
+			type: SemanticType.Parameter,
+			source: mmt.annotations.name
+		};
+
+		return childParameters.length ? { ...childParameters[0] } : parameter;
```
mwdchang, line 261: The function name is a bit confusing ... you are not "getting MMT parameters" per se, but looks like you are trying to merge two parameter lists? If so then following my other comment - don't use collapseXYZ functions, it will give you the wrong number of parameters, and also the wrong results. The question is then what are the two lists you are looking for?

Also, for operations dealing with two lists, it may be preferable to create a map instead of using "find()". The map will be a lot more performant.

For example:

```
const master = [ {k:1, v:1}, {k:2, v:2}, {k:3, v:3}, {k:4, v:4}]
const override = [ {k:1, v:9}, {k:3, v:99}, {k:9, v:3}]

const map = new Map();
override.forEach(d => {
	map.set(d.k, d);
});

const mergedList = master.map(d => {
	if (map.has(d.k)) {
		return map.get(d.k);
	}
	return d;
});

>> [ { k: 1, v: 9 }, { k: 2, v: 2 }, { k: 3, v: 99 }, { k: 4, v: 4 } ]
```

packages/client/hmi-client/src/services/model-configurations.ts
asylves1: removed collapseParameter() since it was not the right method to use in this situation.  Now using `service.ts`getParameters() & getInitials()`

packages/client/hmi-client/src/services/model-configurations.ts
asylves1: switch the `find()` to a `map()`

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-drilldown.vue
```diff
@@ -656,6 +681,24 @@ const onSelectConfiguration = async (config: ModelConfiguration) => {
...
+		}
+		const parameters = getModelParameters(
+			knobs.value.transientModelConfig,
+			mmt.value.annotations.name,
+			getAmrParameters(model.value)
+		);
+		if (parameters.length) {
+			knobs.value.transientModelConfig.parameterSemanticList = parameters;
+		}
+	}
```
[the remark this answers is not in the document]
shawnyama: i have a pr that moves to an onMounted but it's not ready, might be outdated too

blanchco (approved):

YohannParis (approved):

PR #5509, opened 2024-11-14
improve interactions when Funman times out

mwdchang:
### Summary
Increase Funman's time out to give it a better chance to finish, and better handling of situations when timeout happens

image


Tom-Szendrey (approved):
holy long running!

YohannParis (approved):

PR #5514, opened 2024-11-14
Switch between pdfs

asylves1:
# Description

https://github.com/user-attachments/assets/1ab6315e-b41e-4747-ba25-808428983cdc

Issue: 
- If there are multiple pdfs when the user selects a model config, the pdf viewer will always stay on the current selected pdf

Testing:
- Add a document & model to a workflow
- create a model config and extract the inputs
-  selecting a different config from different doc to open the correct doc

Resolves #(issue)


Tom-Szendrey (approved):

shawnyama (approved):

YohannParis (approved):
LGTM

PR #5519, opened 2024-11-15
Extract geolocation information when enriching model AMR

### LINKED ISSUE #4928: Extract geolocation information when enriching model AMR
The goal is to location-based stratification of epi models (e.g. inclusion of travel/flight networks) and support richer, possibly map-based visualization of the model initial conditions and parameters, beyond the current matrix-based view.

First step is to create an example input-output:
* input: a compartmental model stratified by some number of real locations (56 US states and territories)
https://github.com/liunelson/mira/blob/nliu/experiment/notebooks/nliu/data/example_models/model_seirhd_travel_stateName.json
* output: same model but all the state variables and parameters are grounded the relevant `geonames`
https://github.com/liunelson/mira/blob/nliu/experiment/notebooks/nliu/data/example_models/model_seirhd_travel_stateCurie.json

The difference between the two AMRs above is that the `model.states[].grounding.modifiers.location` and `semantics.ode.parameters[].modifiers.location_*` are populated with `geonames` "curies".

An idea on how to achieve this is to do `state_var_name.split('_')` and pass the relevant geographical text to MIRA DKG API. Alternatively, this could be done with a GoLLM task.
[http://mira-epi-dkg-lb-c7b58edea41524e6.elb.us-east-1.amazonaws.com/api/ground/alabama](http://mira-epi-dkg-lb-c7b58edea41524e6.elb.us-east-1.amazonaws.com/api/ground/alabama)

dvince2:
# Description

This is a first pass at getting geoname information. Right now we need to fall back to doing the DKG search on MIRA but in the future we could probably also host this information locally? I envision the search for geonames potentially happening seperately from regular DKG searches, with the locations being merged into the standard DKG search if needed.

Resolves #4928


YohannParis (approved):
LGTM

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/mira/DKG.java
```diff
@@ -19,6 +23,12 @@ public class DKG {
 	public static final String NAME = "name:string";
 	public static final String DESCRIPTION = "description:string";
 	public static final String EMBEDDINGS = "description:dense_vector";
+	public static final String GEONAMES = "geonames";
+
+	public DKG(String curie) {
+		this.curie = curie;
+		locations = new ArrayList<>();
+	}
```
mwdchang, line 31: I am not sure how this is modelled in MIRA, but the last time I worked with it on the WM project, it isn't such that "a curie has a list of geonames", but rather "geoname is a type of curie with id/description/label ...etc"

dvince2: The proposal from @liunelson has the geonames as a list of locations on the grounding - I'm under the impression that this is just for our own internal representation?

PR #5532, opened 2024-11-15
scenario template storage and layout

blanchco:
# Description

* Storing the scenario template object on the workflow
* add edges
* using dagre layout algorithm to layout workflow nodes

https://github.com/user-attachments/assets/c5a24625-2790-498d-92d6-ec31e87f80d7

Up next:
Adding input/state to populate the created nodes


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -584,6 +585,41 @@ export class WorkflowWrapper {
...
+	runDagreLayout() {
+		const g = new dagre.graphlib.Graph({ compound: true });
+		g.setGraph({});
+		g.setDefaultEdgeLabel(() => ({}));
+		g.graph().rankDir = 'LR';
+		this.getNodes().forEach((node) => {
+			g.setNode(node.id, {
+				label: node.displayName,
+				width: node.width ?? 50,
+				height: node.height ?? 50
```
mwdchang, line 598: A couple of minor things
- I would remove the `width/height ?? 50` thing cause we should always have the width/height
- Looking at the video, you may want to play it bit with `g.graph().nodesep` or `g.graph().ranksep` to space things out a bit more, another way is to artificially increase the width/height by some padding amount

mwdchang (approved):

packages/client/hmi-client/src/services/workflow.ts
blanchco: Thanks, I've added the suggestions

shawnyama (approved):

PR #5535, opened 2024-11-15
workflow clean up

mwdchang:
### Summary
Fix a potential problem where a delete node/edge may get undeleted.

Remove outdated version shim

Add createdBy auditing to better track when nodes/edges are created to better debug synchronization issues

### DIFF
--- a/packages/client/hmi-client/src/services/workflow.ts
+++ b/packages/client/hmi-client/src/services/workflow.ts
@@ -25,8 +25,6 @@ import {
 import { useProjects } from '@/composables/project';
 import useAuthStore from '@/stores/auth';
 
-const currentUserId = useAuthStore().user?.id;
-
 /**
  * A wrapper class around the workflow data struture to make it easier
  * to deal with CURD operations
@@ -213,6 +211,7 @@ export class WorkflowWrapper {
 	}
 
 	addNode(op: Operation, pos: Position, options: { size?: OperatorNodeSize; state?: any }) {
+		const currentUserName = useAuthStore().user?.username;
 		const nodeSize: Size = getOperatorNodeSize(options.size ?? OperatorNodeSize.medium);
 
 		const node: WorkflowNode<any> = {
@@ -225,7 +224,7 @@ export class WorkflowWrapper {
 			x: pos.x,
 			y: pos.y,
 
-			createdBy: currentUserId,
+			createdBy: currentUserName,
 			createdAt: Date.now(),
 
 			active: null,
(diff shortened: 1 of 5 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
```diff
@@ -130,19 +130,13 @@ public Optional<Workflow> updateAsset(
...
 
-				// FIXME: backwards compatibility for older workflows, remove in a few month.
-				// Aug 2024
-				if (dbNode.getVersion() == null) {
-					dbNode.setVersion(1L);
-					continue;
-				}
-
-				if (dbNode.getVersion().equals(node.getVersion())) {
+				if (dbNode.getIsDeleted() == false && dbNode.getVersion().equals(node.getVersion())) {
```
Tom-Szendrey, line 139: nitpick: Could use a comment here

Tom-Szendrey (approved):

blanchco (approved):

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -211,6 +212,13 @@ export class WorkflowWrapper {
 	}
 
 	addNode(op: Operation, pos: Position, options: { size?: OperatorNodeSize; state?: any }) {
+		let currentUserName: string | undefined = '';
+		try {
+			currentUserName = useAuthStore().user?.username;
```
YohannParis, line 217: Using the initials like we do for the user menu would make a lot more sense.
```suggestion
			currentUserName = useAuthStore().userInitials;
```

YohannParis (approved):
LGTM

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
mwdchang: done

packages/client/hmi-client/src/services/workflow.ts
YohannParis: If the goal is to keep track of which user and not for displaying, you should use the id.

mwdchang: You are not wrong, but between workflowId, nodeId, source, target, active-output, source-port, target-port and a few others, that is a whole lot of uuids to track and having one less helps.

PR #5537, opened 2024-11-18
Change project sample to be reversible

YohannParis:
### Sample Project Handling:

* Added a new method `setSample` to the `useProjects` composable to handle marking projects as sample projects.
* Updated the `ProjectController` to handle setting and unsetting sample projects via a new `PUT` endpoint.

### UI Updates:

* Modified the `tera-share-project` component to include a checkbox for marking projects as sample projects, visible only to admins.
* Removed the `makeSampleMenuItem` from the project menu and adjusted the project menu items accordingly.
* Updated the `tera-common-modal-dialogs` component to improve the UI for the delete project dialog.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
@@ -1196,9 +1196,12 @@ public class ProjectController {
 			@ApiResponse(responseCode = "500", description = "An error occurred verifying permissions", content = @Content)
 		}
 	)
-	@PostMapping("/set-sample/{id}")
+	@PutMapping("/set-sample/{id}/{sample}")
 	@Secured(Roles.USER)
-	public ResponseEntity<JsonNode> makeProjectSample(@PathVariable("id") final UUID id) {
+	public ResponseEntity<JsonNode> makeProjectSample(
+		@PathVariable("id") final UUID id,
+		@PathVariable("sample") final boolean isSample
+	) {
 		try {
 			// Only an admin can set a project as a sample project
 			projectService.checkPermissionCanAdministrate(currentUserService.get().getId(), id);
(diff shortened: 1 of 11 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
```diff
@@ -1208,32 +1211,80 @@ public ResponseEntity<JsonNode> makeProjectSample(@PathVariable("id") final UUID
...
+						projectPermissionsService.setProjectPermissions(
+							rebacProject,
+							rebacUser,
+							Schema.Relationship.READER.toString()
+						);
+					}
+
+					// When we revert a project to a non-sample project, and
+					// the user is the same as the project author it will become the creator of the project once more
+					if (!isSample && contributor.getUserId().equals(project.get().getUserId())) {
```
bigglesandginger, line 1268: This code block should never execute (to my knowledge) as the only contributor after a project is made as a `sample project` would be the ASKEM Admin group, therefore this shouldn't occur.  Instead I would suggest simple setting the project's userId to be the creator (assuming that is the original owner).

bigglesandginger (changes requested):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
bigglesandginger: FYI - I tested this with "ursula" and "adam" and once the project was de-sampled ursula was not made the creator.

YohannParis: Good catch!
I changed the condition test to be clearer.

bigglesandginger (approved):
As it sits currently, it allows making a project a sample, then setting back to not a sample and restores the creator.

Unknown side effects:
1) A sample project made not a sample becomes public even if it was restricted prior - easy to fix and not sure if this matters at all but good to recognise.
2) I added an additional user with Edit permissions prior to the project becoming a sample projects.  Once the project was un-sampled, the additional user only had Read permissions.  Again not sure if this matters but noted... :)

YohannParis:
> As it sits currently, it allows making a project a sample, then setting back to not a sample and restores the creator.
> 
> Unknown side effects:
> 
>     1. A sample project made not a sample becomes public even if it was restricted prior - easy to fix and not sure if this matters at all but good to recognise.
> 
>     2. I added an additional user with Edit permissions prior to the project becoming a sample projects.  Once the project was un-sampled, the additional user only had Read permissions.  Again not sure if this matters but noted... :)

Thank you @bigglesandginger !

Those are the expected behaviors I chose.
1. This is something that an Admin can change easily if they are not happy about that default. But wanting to make a public project a sample meant it was public before. I decided to allow even private project to be become sample to avoid any gotcha.
2. Keeping track of previous user permissions is too complicated for the purpose of this feature. Only Admin can make a project a Sample.

blanchco (approved):

PR #5540, opened 2024-11-18
replace input equations with cleaned equations (model from equations)

shawnyama:
# Description
* I am returning the `cleanedEquations` along with the `modelId` when running `equationsToModel`
* Replacing equations passed to request with the cleaned equations
* The cleaned equations coming back are formatted a bit differently from what's seen in the model page. 
* The model page uses a `getEquations()` function that uses mira to get the equations while the ones returned from `equationsToModel` are from golmm

https://github.com/user-attachments/assets/39f2dc0b-f1c7-454d-868c-33e5e55c2d33


shawnyama:
I changed the `equationsToAMR` in the backend to at least return the `cleanedEquations` if the model doesn't generate (requested by Nelson). Let me know if I am doing this right, not too confident with how I did it.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -238,21 +254,33 @@ public ResponseEntity<UUID> equationsToModel(
...
 				log.warn("Interrupted while waiting for task response", e);
+				if (!cleanedEquations.isEmpty()) {
+					return ResponseEntity.ok(new EquationsToModelResponse(null, cleanedEquations));
+				}
 				throw new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, messages.get("task.gollm.interrupted"));
 			} catch (TimeoutException e) {
 				log.warn("Timeout while waiting for task response", e);
+				if (!cleanedEquations.isEmpty()) {
+					return ResponseEntity.ok(new EquationsToModelResponse(null, cleanedEquations));
+				}
```
mwdchang, line 283: This pattern of returning `cleanedEquations` seems very strange.

If we are always going to return it then we should ignore the other Java exception and just return this, instead of returning it for every single type of exception? 

I don't exactly know what logic we want to have here, but this feels like an anti-pattern.

mwdchang: We checked in with gauldie - it is probably better to split this into two steps than try to around both returning results and exception handling.

- cleaned equation
- the skema stuff

asylves1 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -122,6 +122,48 @@ void init() {
...
+		try {
+			cleanupResp = taskService.runTask(TaskMode.SYNC, cleanupReq);
+		} catch (final JsonProcessingException e) {
+			log.warn("Unable to clean-up equations due to a JsonProcessingException. Reverting to original equations.", e);
+		} catch (final TimeoutException e) {
+			log.warn("Unable to clean-up equations due to a TimeoutException. Reverting to original equations.", e);
+		} catch (final InterruptedException e) {
+			log.warn("Unable to clean-up equations due to a InterruptedException. Reverting to original equations.", e);
+		} catch (final ExecutionException e) {
+			log.warn("Unable to clean-up equations due to a ExecutionException. Reverting to original equations.", e);
```
mwdchang, line 147: These `Reverting to original equations` logging doesn't make sense, I don't see any reverting process and you are just returning empty array if some exception happens. Are we actually trying to do something?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -122,6 +122,48 @@ void init() {
 		taskService.addResponseHandler(equationsCleanupResponseHandler);
 	}
 
+	/**
+	 * Clean up a list of equations
+	 *
+	 * @return List of cleaned-up equations
+	 */
+	@PostMapping("/get-cleaned-equations")
+	@Secured(Roles.USER)
```
mwdchang, line 131: Would prefer better wording here, you are POST'ing to a "GET" which might be a little confusing. Maybe better just to have `POST /clean-equations`.

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
```diff
@@ -363,21 +365,38 @@ function onCheckBoxChange(equation) {
 async function onRun() {
 	isOutputOpen.value = true;
 	isModelLoading.value = true;
-	const equations = clonedState.value.equations
+	const equationsText = clonedState.value.equations
 		.filter((e) => e.includeInProcess && !e.asset.extractionError)
 		.map((e) => e.asset.text);
+	const cleanedEquations = await getCleanedEquations(equationsText);
+	if (!cleanedEquations) return;
```
mwdchang, line 372: You created for yourself two exception cases:
- null
- empty array

This seem like it will handle null but not the empty-array case, are we okay with sending an empty-array down the road or do we need to stop?

YohannParis: It depends if an empty array is a valid response.
If there is an error within `getCleanedEquations()` it should return `null`.

packages/client/hmi-client/src/services/knowledge.ts
```diff
@@ -19,25 +19,35 @@ export interface EquationsToAMRRequest {
...
 
+/**
+ * Clean up a list of equations
+ * @param equations string[] - list of LaTeX or mathml strings representing a model
+ * @return {Promise<string[]>}
+ */
+export const getCleanedEquations = async (equations: string[]): Promise<string[] | null> => {
+	try {
+		const response = await API.post<string[]>(`/knowledge/get-cleaned-equations`, equations);
+		return response.data;
```
YohannParis, line 30:
```suggestion
		return response.data ?? null;
```

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
mwdchang: I always disliked how `null` is being leveraged to indicate errors in the front-end, it makes more sense if the callers handle exceptions as they see fit, so they can choose to halt or try to continue. But this is a different discussion.

Empty array is a valid response for "get-clean-equations" API call, but my guess is that it is probably not a valid input for "equation-to-model". since it doesn't make sense to create a model out of `[]`.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
shawnyama: Just added a proper revert to original equations by adding a fallback

PR #5543, opened 2024-11-18
Jamie style tweaks nov 18

jamiewaese-uncharted:
More style tweaks a visual bugs fixed

### Validate configuration style updates

BEFORE
Validate configuration - before

AFTER
Validate configuration - after

---------------------------------

### Tooltip - make text more readable

BEFORE
Tooltip - before

AFTER
Tooltip - after

---------------------------------

### Math equations should line wrap, not burst through their container

BEFORE
Math should linewrap - before

AFTER
Math should linewrap - after

---------------------------------

### Update home page banner text, make less generic
** I worked with Michael Crouch on this

BEFORE
Home page banner text - before

AFTER
Home page banner text - after

---------------------------------

### Make evaluation scenario modal look nicer

BEFORE
Evaluation scenario - before

AFTER
Evaluation scenario - after

---------------------------------

### Error message margins looks bad

BEFORE
Error message margins - before

AFTER
Error message margins - after

---------------------------------

### Dataset column information - make them look like other operators

BEFORE
Dataset column information - before

AFTER
Dataset column information - after

---------------------------------

### Calibrate output - empty headings when no output available feels weird

BEFORE
Calibrate output empty heading - before

AFTER
Calibrate output empty heading - after

---------------------------------

### Calibrate mapping labels -- too much bold

BEFORE
Calibrate - mapping labels - before

AFTER
Calibrate - mapping labels - after

---------------------------------

### Configure model notebook page - small adjustments

BEFORE
Configure model notebook page - before

AFTER
Configure model notebook page - after

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-input-number.vue
@@ -16,6 +16,7 @@
 				@focusout="$emit('focusout', $event)"
 				type="text"
 				:placeholder="placeholder"
+				@keydown="handleArrowKeys"
 			/>
 		</main>
 		<aside v-if="getErrorMessage"><i class="ml-2 pi pi-exclamation-circle" /> {{ getErrorMessage }}</aside>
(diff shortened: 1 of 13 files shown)


packages/client/hmi-client/src/components/widgets/tera-input-number.vue
```diff
@@ -16,6 +16,7 @@
 				@focusout="$emit('focusout', $event)"
 				type="text"
 				:placeholder="placeholder"
+				@keydown="handleArrowKeys"
```
mwdchang, line 19: We need to double-check if we can support this (eg where tera-input-number is being used), `handleArrowKeys` eventually calls `updateNumber` which emits an update event. We have had issues where we updated too often that we run into synchronization problems or performance problems.

You may want to add a debouncer here just in case

YohannParis: Done

YohannParis (approved):

asylves1 (approved):

PR #5551, opened 2024-11-19
situational awareness state population

blanchco:
# Description

* populating state information for the situational awareness template
* setting some foundation for state population for other nodes

https://github.com/user-attachments/assets/3d9eb447-d05c-4600-b219-84f3a98eb9ca


blanchco:
Will need https://github.com/DARPA-ASKEM/terarium/pull/5532 to be merged before this

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-operation.ts
```diff
@@ -26,6 +26,9 @@ export interface CalibrationOperationStateCiemss extends BaseState {
 	endTime: number;
 	stepSize: number;
 	learningRate: number;
+
+	// This is a state variable to indicate which model variables are selected for calibration, this is filled in from the workflow template generator
+	templateSelectedModelVariables: string[];
```
mwdchang, line 31: Not a fan of this because it couples two sort of orthogonal ideas together (operators and scenarios), would prefer if we can build scenarios on top of operators rather than beside (or within) operators.

Maybe we can have a chat and brainstorm about this

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-operation.ts
blanchco: Ya that sounds good

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -250,10 +250,19 @@ export class WorkflowWrapper {
...
 		}
 		this.wf.nodes.push(node);
+
+		if (options.outputValue) {
+			node.status = OperatorStatus.SUCCESS;
+			const outputPort = node.outputs[0];
+			outputPort.value = isArray(options.outputValue) ? options.outputValue : [options.outputValue];
+			node.active = outputPort.id;
+			this.selectOutput(node, outputPort.id);
+		}
```
mwdchang, line 265: It might be better to move this `outputValue` out of addNode(...), as this mechanism would only work for a couple of operators, the remaining operators need to more setup like output-state.

eg in situational-awareness createWorkflow, this sequence should create a workflow with the correct data
```
const model  = wf.addNode(...)
const modelConfig = wf.addNode(...)
const simulate = wf.addNode(...)

// .. add edges

Object.assign(model.outputs[0], ...) // state, value, status, operatorStatus
wf.selectOutput(model, model.outputs[0].id)

Object.assign(modelConfig.outputs[0], ...) // state, value, status, operatorStatus
wf.selectOutput(modelConfig, modelConfig.outputs[0].id)

Object.assign(simulate.outputs[0], ...)
wf.selectOutput(simulate, simulate.outputs[0].id) // state, value, status, operatorStatus

```

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-operation.ts
mwdchang: We checked in with gauldie - it is probably better to split this into two steps than try to around both returning results and exception handling.
- cleaned equation
- the skema stuff

blanchco: I think this should all be addressed now after our discussion, I am now setting the chartSettings state and removed the `templateSelectedModelVariables` entirely

packages/client/hmi-client/src/components/workflow/scenario-templates/situational-awareness/situational-awareness-scenario.ts
```diff
@@ -121,6 +149,52 @@ export class SituationalAwarenessScenario extends BaseScenario {
...
+		calibrateChartSettings = updateChartSettingsBySelectedVariables(
+			calibrateChartSettings,
+			ChartSettingType.ERROR_DISTRIBUTION,
+			this.calibrateSpec.ids
+		);
+		calibrateChartSettings = updateChartSettingsBySelectedVariables(
+			calibrateChartSettings,
+			ChartSettingType.INTERVENTION,
+			this.calibrateSpec.ids
+		);
```
jryu01, line 189:
```suggestion
```
You probably don't need to add intervention chart settings. If the template doesn't create interventions.

jryu01 (approved):

mwdchang (approved):

PR #5556, opened 2024-11-19
make all accordian incidices refs

blanchco:
# Description

* make all accordion indicies refs


mwdchang:
@blanchco we know this "fixes" the accordion expand/collapse problem, but what are the side effects, if any that are introduced because of this? 

Is it that we are manually forcing a collapsed accordion to open again? Are we rendering or umount/mount the same things/components multiple times?

mwdchang (approved):

YohannParis (approved):
Thank you very much!!!!

PR #5559, opened 2024-11-19
Project Search Upgrades

kbirk:
Search will prioritize direct matches to the projects name.

Added a re-index endpoint to migrate old projects.

Split embeddings by the field they are generated from.

Fixed the permission and asset type filtering logic to be applied correctly.


YohannParis (approved):
Works for me locally.

dgauldie (approved):

PR #5567, opened 2024-11-20
empty mathML workaround

mwdchang:
### Summary
A workaround for amr-to-mmt task failing because unit-expressions are blank

### Testing
You need to  run/rebuild mira-taskrunner locally

Goto http://localhost:8080/amr-petri-test and paste in the amr bellow, there should be no errors and the model-diagram should render.

[bad-amr-to-mmt.json](https://github.com/user-attachments/files/17830889/bad-amr-to-mmt.json)

### DIFF
--- a/packages/mira/tasks/amr_to_mmt.py
+++ b/packages/mira/tasks/amr_to_mmt.py
@@ -1,6 +1,7 @@
 import sys
 import json
 import traceback
+import time
 
 from taskrunner import TaskRunnerInterface
 from mira.sources.amr import model_from_json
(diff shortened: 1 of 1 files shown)


packages/mira/tasks/amr_to_mmt.py
```diff
@@ -12,23 +13,44 @@ def main():
...
+        taskrunner.log(f"number of templates {len(mmt.templates)}");
+
         for tm in mmt.templates:
             # Sanitize
             count = count + 1
             if tm.name == None or tm.name == "":
                 tm.name = "generated-" + str(count)
 
-            params = tm.get_parameter_names()
+            # params = tm.get_parameter_names()
```
Tom-Szendrey, line 52: I am not following why we are no longer grabbing param names here?
It looks like our `entry` below will always have an empty list?

Tom-Szendrey (approved):

packages/mira/tasks/amr_to_mmt.py
mwdchang: was from perf test a while ago, will revert, thx

YohannParis (approved):

PR #5568, opened 2024-11-20
Add ensemble variables charts on calibrate ensemble operator

jryu01:
# Description

First pass of ensemble variables chart on calibrate ensemble operator. 

Screenshot 2024-11-20 at 10 28 31 AM

Main changes are in `useCharts.ts`, `tera-calibrate-ensemble-ciemss-drilldown.vue` and other util files. Other files are changed because I modified one of the reusable functions. 

TODOs:
 - Support small multiples charts for each model 
 - Add chart setting options UI on the output setting panel to toggle small multiples 
 - Add support for multi lines on single chart to support simulate ensemble operator. 
 
Resolves #(issue)


packages/client/hmi-client/src/types/common.ts
```diff
@@ -188,7 +188,16 @@ export enum ChartSettingType {
 	INTERVENTION = 'intervention'
 }
 
-export interface ChartSetting {
+export type ChartSetting = ChartSettingBase | ChartSettingEnsembleVariable;
+
+export interface ChartSettingEnsembleVariable extends ChartSettingBase {
+	type: ChartSettingType.VARIABLE_ENSEMBLE;
```
Tom-Szendrey, line 194: I dont think we should force this type for `ChartSettingEnsembleVariable`?
We may want other valid types for ensemble such as weights?

jryu01: I think you are free to add new type extending the base setting when needed. I just added this type specific for `ensemble variables over time` charts.

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -40,30 +55,61 @@ interface ChartData {
...
 	// Setup annotations
 	const { getChartAnnotationsByChartId, generateAndSaveForecastChartAnnotation } = useChartAnnotations(nodeId);
 
 	const getUnit = (paramId: string) => {
-		if (!model.value) return '';
+		if (!model?.value) return '';
 		return getUnitsFromModelParts(model.value)[paramId] || '';
 	};
 
+	const createEnsembleVariableChartOptions = (setting: ChartSettingEnsembleVariable) => {
```
Tom-Szendrey, line 73: I see this has pre and post run labels
Is this specific to calibration? It may help to rename

jryu01: We use `:pre` label for the base data for all non ensemble calibrate, optimize (before and after optimize), and simulation (with or without intervention). I found it's a bit less confusing to just use same label here as well.

Tom-Szendrey: I am not understanding why a normal simulate has a baseline but this is irrelevant to this PR specifically
image

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -190,50 +190,75 @@
...
-						class="mb-2"
-					/>
-					<Button
-						class="p-button-sm p-button-text"
-						@click="chartProxy.addChart()"
-						label="Add chart"
-						icon="pi pi-plus"
-					/>
+				<section class="pb-3" ref="outputPanel">
+					<Accordion multiple :active-index="[0, 1]" class="px-2">
```
Tom-Szendrey, line 194: We will want to store the active index in a ref `currentActiveIndexes`

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -190,50 +190,75 @@
...
+						@close="activeChartSettings = null"
+					/>
+				</template>
+				<template #content>
+					<div class="output-settings-panel">
+						<tera-chart-settings
+							:title="'Ensemble variables over time'"
+							:settings="chartSettings"
+							:type="ChartSettingType.VARIABLE_ENSEMBLE"
+							:select-options="ensembleVariables"
```
Tom-Szendrey, line 251: I suspect we do not want to disallow the user from selecting regular model variables
image

For example I do not see a reason to limit the user to `Sus` and disallow `Susceptible_state`, `S`, and 'Susceptible'. 
If this is a more complex change than I am aware it may be worth doing in a follow up PR I guess

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -520,10 +570,15 @@ watch(
...
-			const output = await getRunResultCiemss(state.postForecastId, 'result.csv');
-			runResults.value = output.runResults;
 			lossValues.value = await getLossValuesFromSimulation(props.node.state.calibrationId);
-			lossChartSpec.value = updateLossChartSpec(lossValues.value, lossChartSize.value);
+			lossChartSpec.value = updateLossChartSpec(lossValues.value, chartSize.value);
+
+			// Fetch output data and prepare chart data
+			const data = await fetchOutputData(state.preForecastId, state.postForecastId);
+			chartData.value = data ? { ...data, translationMap: {} } : null;
+			console.log(data);
```
Tom-Szendrey, line 579: Logs to remove

packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -134,3 +139,20 @@ export const getGraphDataFromDatasetCSV = (
 
 	return graphData;
 };
+
+export function getSelectedOutput<S>(node: WorkflowNode<S>) {
```
Tom-Szendrey, line 143: This looks to be a repeat of `getActiveOutput` in `/services/operator.ts`
Worth double checking and keeping one or the other

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
jryu01: Let's address this in a separate issue.

packages/client/hmi-client/src/components/workflow/util.ts
jryu01: Nice catch, we had total 3 functions doing exactly same thing lol. I removed other two.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
jryu01: Done

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -5,27 +5,7 @@
...
-						props.node.state.configurationWeights
-					) as any
-				"
-				:initial-data="csvAsset"
-				:size="{ width: 190, height: 120 }"
-				has-mean-line
-				@configuration-change="chartProxy.configurationChange(index, $event)"
-			/>
-		</template>
+		<template v-if="!inProgressCalibrationId && !inProgressForecastId && runResults && csvAsset"> </template>
```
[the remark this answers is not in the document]
jryu01: Removed it

packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -134,3 +139,18 @@ export const getGraphDataFromDatasetCSV = (
...
+export type VariableMappings = CalibrateMap[] | CalibrateEnsembleMappingRow[] | SimulateEnsembleMappingRow[];
+
+export const modelVarToDatasetVar = (mapping: VariableMappings, modelVariable: string) => {
+	if (_.isEmpty(mapping) || isSimulateEnsembleMappingRow(mapping[0])) return '';
+	const found = (mapping as CalibrateMap[] | CalibrateEnsembleMappingRow[]).find((d) =>
+		isCalibrateMap(d) ? d.modelVariable === modelVariable : d.newName === modelVariable
+	);
+	if (!found) return '';
+	return isCalibrateMap(found) ? found.datasetVariable : found.datasetMapping;
+};
```
mwdchang, line 156: What is this function `modelVarToDatasetVar` suppose to do? It seems to touch upon all types of operators?

Would it be better if the signature is like?
```
modelVarToDatasetVar = (mapping: VariableMappings, modelVariable: string, operatorType: string) => {
  if (operatorType === "calibrate") {
  } else if (operatorType === "simulate") {
   ...
  }
}
```

jryu01: This function grabs a datasets variable for a given model variable from the provided mapping. I think we can just branch based on the mapping type since each operator (that uses mapping) uses different mapping object. It only touches the operator that has mapping object (calibrate, calibrate & simulate ensemble).  I'll just update the code so that it's more clear to read.

jryu01: Done

mwdchang (approved):

PR #5569, opened 2024-11-20
feat: preset configs for funman

shawnyama:
# Description

* Fast/precise presets to quickly configure funman
* Made `requestParameters` part of the `BasicKnobs`
* parameters that should be part of the presets are determined in `setRequestParameters()`

https://github.com/user-attachments/assets/b5b18865-f6e2-414f-9a82-38d6a8f8c1a6


YohannParis (approved):
LGTM

PR #5571, opened 2024-11-20
Validate config katex layout

asylves1:
# Description
image
image

Issue: 
- The list of parameters & initial states were not following the layout and just printing in a single line.
- Same would happen to the dropdown when selecting multiple options

image

Testing:
- Create a model-config with a large amount of parameters
- Create a new Validate configuration
- They should list the items and stay within the layout

Resolves #(issue)


YohannParis (approved):
Looks great to me 🎉

shawnyama (approved):

PR #5574, opened 2024-11-20
Ensemble minor util fix

Tom-Szendrey:
# Description
Correction to mapping. 
We do not want to redefine the entire mapping for every row we add. We want to add

Without this we will run into breaks while running with multiple mapping.


YohannParis (approved):
Nice catch

blanchco (approved):

PR #5583, opened 2024-11-20
Improve dkg search after model from equations

YohannParis:
### Shell Script Changes:
* Removed the `start_staging` function and updated references to use `start_local` instead. [[1]](diffhunk://#diff-e0e22af95c3c4a3a65f8a76cafe5636c7c61ab808c458de0d1beff2a65e55b65L82-L85) [[2]](diffhunk://#diff-e0e22af95c3c4a3a65f8a76cafe5636c7c61ab808c458de0d1beff2a65e55b65L170-R167)
* Updated valid environments list to remove `staging` from the options.

### Client-Side Changes:
* Removed the `enrichModelMetadata` import and the `isGeneratingCard` ref from `tera-model-from-equations-drilldown.vue`. [[1]](diffhunk://#diff-2f796b404e2219b0045f6ee5141b6caa46be4372ebd0720307afca92e0b412f2L185-R185) [[2]](diffhunk://#diff-2f796b404e2219b0045f6ee5141b6caa46be4372ebd0720307afca92e0b412f2L245)
* Added `workflowId` and `nodeId` to the `EquationsToAMRRequest` type and updated the `equationsToAMR` function to include these properties. [[1]](diffhunk://#diff-1cadbdc206294108b809908967930771b445649801686db0e4302c037ad81ca1R29-R50) [[2]](diffhunk://#diff-2f796b404e2219b0045f6ee5141b6caa46be4372ebd0720307afca92e0b412f2L381-L387)
* Added a new handler for `KnowledgeEnrichmentModel` events in `notificationEventHandlers.ts`.

### Server-Side Changes:
* Enabled asynchronous task execution in `TerariumApplication.java` by configuring a thread pool for async tasks.
* Introduced a new `NotificationProperties` class in `KnowledgeController.java` to handle additional properties in notifications.
* Removed unused imports and variables, and updated method parameters for clarity and consistency. [[1]](diffhunk://#diff-d6534a7efa7aa2b6c241914d73eb4ecbc163b3da1d509c87c1d828f749a6999fR71-R87) [[2]](diffhunk://#diff-86712ae74e79cfa17607573213ad7f0dd36863a3ef9eac66ef20cdb0c4fecfe7L615-L616)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
```diff
@@ -0,0 +1,228 @@
...
+		// Create the notification group
+		final NotificationGroupInstance<Properties> notificationInterface = new NotificationGroupInstance<>(
+			clientEventService,
+			notificationService,
+			ClientEventType.ENRICHMENT_MODEL,
+			projectId,
+			new Properties(modelId)
+		);
+
+		ExecutorService executor = Executors.newFixedThreadPool(POOL_SIZE);
```
kbirk, line 77: Put the pool on the class. You want a single persistent pool.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
```diff
@@ -0,0 +1,228 @@
...
+				document,
+				model,
+				projectId
+			);
+
+			final TaskRequest taskRequest = new CompoundTask(enrichAmrRequest, modelCardRequest);
+			final TaskResponse taskResponse = taskService.runTask(TaskService.TaskMode.SYNC, taskRequest);
+
+			// Check if the task was un-successful
+			if (!taskResponse.getStatus().equals(TaskStatus.SUCCESS)) {
```
kbirk, line 136: When the TaskService runs them synchronously, they will throw an exception if they fail. So rather than check, use a try catch

kbirk (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
YohannParis: Isn't what's above on L49 ?
https://github.com/DARPA-ASKEM/terarium/pull/5583/files/30f21d6d69df3f8dac13ea8dcb739fb285d08a0a#diff-c904c511e871ff90a3ae8a64d0596ab5cc8bcd9c9125fc9e36d6494ca0aff2c3R49-R50

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
```diff
@@ -0,0 +1,231 @@
...
+					}
+					TaskUtilities.performDKGSearchAndSetGrounding(dkgService, vertex);
+				});
+				enrichedModel.setVerticies(vertices);
+			} else {
+				final List<State> states = enrichedModel.getStates();
+				states.forEach(state -> {
+					if (state == null) {
+						state = new State();
+					}
```
mwdchang, line 170: I don't think this if-section does anything, what are we trying to do?
If state is null, `new State()` does not change the element in the array/list ... 

```
import java.util.List;
import java.util.ArrayList;

public class JavaRef {

    public static void main(String args[]) {
       ArrayList list = new ArrayList();
       list.add(1);
       list.add(null);
       list.add(2);
       list.forEach(elem -> {
            if (elem == null) {
                System.out.println("reassign null");
                elem = Long.valueOf(99);
            }
       });
       System.out.println(list);
    }
}
```

Running this the `null` is unchanged

```
reassign null
[1, null, 2]
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
```diff
@@ -0,0 +1,231 @@
...
+				.orElseThrow(() -> {
+					final String errorString = String.format("Enriched Model %s not found", modelId);
+					log.warn(errorString);
+					return new IOException(errorString);
+				});
+
+			notificationInterface.sendMessage("Updating model grounding...");
+
+			// Update State Grounding
+			if (enrichedModel.isRegnet()) {
```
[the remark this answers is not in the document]
YohannParis: Good point, removed

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/EnrichmentService.java
YohannParis: Good point, I should not have copy/paste code. Fixed.

hmiServerDev.sh
```diff
@@ -167,7 +164,7 @@ case ${COMMAND} in
       if [ ${ENVIRONMENT} == "local" ]; then
         start_local
       elif [ ${ENVIRONMENT} == "staging" ]; then
-        start_staging
+        start_local
```
dvince2, line 167: is there a point any more of having 'staging' as an option at all?

dvince2 (approved):

PR #5600, opened 2024-11-22
exporting project: add simulation to project-assets, cleanup

mwdchang:
### Summary
Project-export is broken, partially because simulations are not extracted into the archive. Moreover they are not a part of project-asset that the exporter iterates over.

This PR adds project-asset entries and does a bit of cleanup of redundant setters

### Testing
- Create project
- Create workflow with model => model-config => simulate, run a simulate, plot some variables
- Export the project into an archive
- Check archive contains the simulation result files (eg result.csv result-summary.csv ... etc), it is just a zip stream, so use whatever extractor is in macos or `unzip -l <archive>` in CLI
- Rename the project
- Import the project, wait 5-10 seconds
- Go into the imported project, verify that workflow is there, we can fetch simulation results and the charts are visible


mwdchang:
CC @liunelson

YohannParis (approved):
Looks good to me

PR #5611, opened 2024-11-22
Ensemble Calibrate fix up stop button

Tom-Szendrey:
# Description
- Add error messages to ensemble calibrate if we run into a failure.
- Update cancelId to include all sim ids
- if we stop the poller we do not need to also check for errors. This should be `else if` not `if`


mwdchang (approved):
Did test, code looks okay

blanchco (approved):

PR #5612, opened 2024-11-22
turning off dual-controller edges per request

mwdchang:
### Summary
A pass at refreshing model-diagrams per feedback from nelson/jamie/pascale

#### Non-stratified
We "remove" the transition boxes to make it look like states are connecting to states, some font styling adjustments

image

#### Stratified model-diagram
Turn off one of the controller edges

Before:

image

After:

image


packages/client/hmi-client/src/model-representation/mira/mira.ts
```diff
@@ -523,14 +523,15 @@ export const convertToIGraph = (
...
 
-			t.controllers.forEach((controllerName) => {
-				graph.edges.push({
-					source: t.name,
-					target: controllerName,
-					points: [],
-					data: { isController: true }
-				});
-			});
+			// Turn off dual-controller edges as per request - Nov 2024
```
Tom-Szendrey, line 526: I suspect ideally this is determined by a flag/toggle rather than just commenting it out

Tom-Szendrey (approved):

blanchco (approved):

packages/client/hmi-client/src/model-representation/mira/mira.ts
mwdchang: I don't think we need this kind of flexibility, however we are still playing around - thus the comment instead of total removal.

PR #5613, opened 2024-11-22
show a max of 4 charts in intervention node

shawnyama:
# Description

* show a max of 4 charts in intervention node


mwdchang (approved):

Tom-Szendrey (approved):
thisll help with those insanely long nodes but i suspect we will want to have another look for all nodes

PR #5620, opened 2024-11-25
remove feature config where applicable

blanchco:
# Description

* Remove the feature-config object where applicable for `tera-asset`
* There are a number of components that do actually use featureConfig still so I have left those alone for now.  It might be worth it to make the prop simply a boolean rather than this object for those components though.


shawnyama (approved):

asylves1 (approved):

YohannParis (approved):
Amazing ⭐

Issue #5473, opened 2024-11-12
[BUG]: Updating the dataset names going into Transform dataset does not update the comment with the mapping.


pascaleproulx:
**Describe the bug**
Don't know which dataset is which if I don't first save the simulation output for reuse. Updating the dataset names does not update the comment with the mapping.

**To Reproduce**
1. After running two different simulate (but NOT saving for reuse), add a transform dataset to the canvas and plug the output of each dataset into it. It will show two pills at the top both named dataset-1 and nothing mapped in the comment. 
2. Go back to the simulate and save for reuse (give different names). These names are reflected in the pills but the comment is not updated. If I disconnect and reconnect after renaming then it works. 

**Expected behavior**
The comment should indicate which dataset is mapped to each dataframe.

**Screenshots**
![Image](https://github.com/user-attachments/assets/49872bdf-aa20-4283-9f04-92640de90216)

Tom-Szendrey:
A simulation run that has not been saved will have a name similar to: `Forecast run f1d4c215-f7e5-49f1-b098-72542f2ea942`

In the first scenario would we like to have `#d1 = Forecast run f1d4c215-f7e5-49f1-b098-72542f2ea942`?

In the second scenario where a user then renames this dataset we would then like to update it to `#d1 = updated_name_here`?

Tom-Szendrey:
Note to complete this task once above questions are answered:
Ensure the following gets updated correctly: https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue#L161-L170

Update logic on how we are setting first code cell here: https://github.com/DARPA-ASKEM/terarium/blob/main/packages/client/hmi-client/src/components/llm/tera-jupyter-chat.vue#L442 
Depending on above answers we may want to watch defaultPreview and reset notebookItems.value[0]

PR #5624, opened 2024-11-25
Data transform update names

Tom-Szendrey:
# Description
The context for the user (aka d1 is mapped to [dataset name]) should not be within the first cell.
More context can be found here: https://github.com/DARPA-ASKEM/terarium/issues/5473

Screenshot of solution:

image

image


shawnyama (approved):

packages/client/hmi-client/src/components/dataset/tera-dataset-jupyter-panel.vue
```diff
@@ -85,23 +85,28 @@
...
-			@new-dataset-saved="onNewDatasetSaved"
-			@download-response="onDownloadResponse"
-			@update-language="(lang) => emit('update-language', lang)"
-			@update-selected-outputs="(outputs) => emit('update-selected-outputs', outputs)"
-			:notebook-session="props.notebookSession"
-		/>
+		<div>
+			<div v-for="(data, indx) in dataContextDescription" :key="indx">
+				<input class="context-description" :value="data" disabled type="text" />
+			</div>
```
mwdchang, line 91: Any particular reason why these are disabled `<input>` elements? Are we trying for some specific effect?

Tom-Szendrey: no deeper meaning aside from `disabled` background colour and grab any rules we have on this that I may have missed

mwdchang: Maybe an unordered-list is better at conveying what the code is trying to do. The background colour is like `var(--surface-100)` or something similar to that.

YohannParis: Are we really shortening `index` with `indx` to save one character?

Tom-Szendrey: No we are shortening it to idx to save two characters :P

mwdchang (approved):

PR #5625, opened 2024-11-25
refactor rename asset

blanchco:
# Description

* Refactor rename asset for the tera-asset component

### DIFF
--- a/packages/client/hmi-client/src/components/asset/tera-asset.vue
+++ b/packages/client/hmi-client/src/components/asset/tera-asset.vue
@@ -14,10 +14,13 @@
 		</div>
 		<!--For naming asset such as model or code file-->
 		<div class="row">
-			<slot name="name-input" />
-			<h4 v-if="!isNamingAsset" :class="{ shrink: shrinkHeader }">
-				{{ name }}
-			</h4>
+			<tera-toggleable-input
+				v-if="[ProjectPages.OVERVIEW, AssetType.Dataset, AssetType.Model].includes(pageType)"
+				:model-value="name"
+				tag="h4"
+				@update:model-value="onRename"
+			/>
+			<h4 v-else>{{ name }}</h4>
 			<slot v-if="!overline" name="edit-buttons" />
 		</div>
 		<!--put model contributors here too-->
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -166,6 +171,23 @@ function onScroll(event: Event) {
...
+		case ProjectPages.OVERVIEW:
+			if (!activeProject.value) return;
+			updateProject({ ...activeProject.value, name: newName }).then(() => {
+				useProjects().refresh();
+			});
+			break;
+		case AssetType.Dataset:
+		case AssetType.Model:
+		default:
+			emit('rename', newName);
```
blanchco, line 186: model and dataset require saving an entire object since we can modify many parts within those asset pages, so im emitting the rename here.

shawnyama (approved):
hell yeee

packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -14,10 +14,13 @@
...
-				{{ name }}
-			</h4>
+			<tera-toggleable-input
+				v-if="[ProjectPages.OVERVIEW, AssetType.Dataset, AssetType.Model].includes(pageType)"
+				:model-value="name"
+				tag="h4"
+				@update:model-value="onRename"
+			/>
+			<h4 v-else>{{ name }}</h4>
 			<slot v-if="!overline" name="edit-buttons" />
```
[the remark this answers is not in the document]
blanchco: will do

packages/client/hmi-client/src/components/asset/tera-asset.vue
```diff
@@ -166,6 +162,23 @@ function onScroll(event: Event) {
 	chosenItem.value = closestItem;
 }
 
+function onRename(newName: string) {
+	if (!newName) return;
```
YohannParis, line 166: I found myself renaming asset to an empty string because I pressed the return key to fast... 
```suggestion
	if (!newName || isEmtpy(newName)) return;
```

packages/client/hmi-client/src/components/asset/tera-asset.vue
YohannParis: That make sense, in that case I would prefer all pageType to emit a `rename` event, and let each page handle how this is done. So in this case, the project page should handle the renaming, not tera-asset. The goal being to simplify tera-asset and remove as many condition in it.

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -227,12 +205,10 @@ async function updateDatasetContent() {
...
 }
 
-async function updateDatasetName() {
-	if (transientDataset.value && !isEmpty(newName.value)) {
-		transientDataset.value.name = newName.value;
-		await updateDatasetContent();
-	}
-	isRenaming.value = false;
+function updateDatasetName(name: string) {
+	if (!transientDataset.value || isEmpty(name)) return;
```
YohannParis, line 209: If tera-asset check if the new name is empty, we don't need to do it here.

```suggestion
	if (!transientDataset.value) return;
```

YohannParis (approved):
I like that how cleaner this looks

packages/client/hmi-client/src/components/asset/tera-asset.vue
blanchco: done, I've moved that snippet to the project overview

PR #5626, opened 2024-11-25
Added small multipes chart support for ensemble calibratoin

jryu01:
# Description

- Added chart setting UI to toggle between small multiples chart for each model vs single ensemble chart
- Added support for rendering small multiples charts for a ensemble variable 
- Added support for ai generated annotation for each chart. 

Screenshot 2024-11-25 at 4 42 16 PM
Screenshot 2024-11-25 at 4 42 29 PM

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -205,13 +206,16 @@
 						</AccordionTab>
 						<template v-if="!isRunInProgress">
 							<AccordionTab header="Ensemble variables over time">
-								<template v-for="setting of selectedEnsembleVariableSettings" :key="setting.id">
+								<div ref="outputPanel"></div>
```
Tom-Szendrey, line 209: You have removed the reference to this (ln 534) 
can this div be removed or is there something I am missing?

Tom-Szendrey (approved):

packages/client/hmi-client/src/services/model-configurations.ts
```diff
@@ -283,3 +283,9 @@ export function getModelInitials(modelConfiguration, source, amrInitials) {
 		};
 	});
 }
+
+// Get the model configuration name for the given model configuration id
+export function getModelConfigName(modelConfigs: ModelConfiguration[], configId: string) {
+	const modelConfig = modelConfigs.find((d) => d.id === configId);
+	return modelConfig?.name ?? '';
+}
```
mwdchang, line 291: We should have a chat about how things are wired together, with the composables and utilities and such. For example you are calling this function inside a loop to get translation names, it might be better if we can cache `{ modelConfigId => name }` somewhere, then the process wouldn't need to be `O(n^2)`

mwdchang (approved):
Looks fine. Didn't take a very close look, but I think it would be helpful if you can do a walkover on the state of things and how different parts are wired together, since the landscape had changed quite a bit since we initially discussed about chart-settings and such.

Some food for thoughts
- Do we know explicitly that we are rendering the chart(s) as a small multiple? I think we should consider adjusting the styling, eg make lines thinner, fewer ticks ...etc
- Likewise may need to tweak when "dates" are on the x-axis instead of just integers

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
jryu01: I'll make the change on a separate branch that I'm currently working on

PR #5642, opened 2024-11-26
provisional option for Vegalite render to image

mwdchang:
### Summary
Provide an option to render as a PNG image (props.interactive = false), this allows us to explicitly invoke `finalize()` after the rendering is done and free up excess memory held on by Vegalite.

Note: Memory clean up of raw data are at the discretion of the parent components and chart itself does not have sufficient knowledge when the data can be released.

Usage example:
```
       <vega-chart
         v-for="setting of selectedVariableSettings"
         :key="setting.id"
         expandable
         are-embed-actions-visible
         :visualization-spec="variableCharts[setting.id]"
         :interactive="false"
       />
```

### Testing
Latch onto staging `yarn staging` and use Gauldie's project as an example:
```
 http://localhost:8080/projects/b1f12eb1-d5ad-46a8-becf-94e98508c833/workflow/9c635377-bd57-4b10-936f-ed639fc2bf86
```

Open up the memory-panel and observe memory usage after all nodes have rendered. You may need to manually click garbage collect. 

image

#### Before
Total memory footprint is around 450MB
image

#### After
Total memory footprint is around 150MB
image


packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -13,7 +13,11 @@
 		</div>
 	</Dialog>
 	<div class="vega-chart-container">
-		<div ref="vegaContainer" />
+		<div v-if="!interactive">
+			<img :src="imageDataURL" alt="chart" />
+		</div>
+		<div v-else ref="vegaContainer" />
```
YohannParis, line 19:
```suggestion
		<img v-if="!interactive" :src="imageDataURL" alt="chart" />
		<div v-else ref="vegaContainer" />
```

YohannParis (approved):

Tom-Szendrey (approved):
LGTM did not test

jryu01 (approved):

PR #5646, opened 2024-11-26
improve a bit search and filter bar

YohannParis:
Tidy the HTML and CSS code to be less clunky and more readable.


packages/client/hmi-client/src/page/Home.vue
```diff
@@ -50,51 +50,39 @@
...
+								@update:modelValue="onToggle"
+								placeholder="Add or remove columns"
+								class="p-inputtext-sm"
+							/>
+							<SelectButton :model-value="view" @change="selectChange" :options="viewOptions" option-value="value">
+								<template #option="slotProps">
+									<span class="p-button-label">{{ slotProps.option.value }}</span>
+								</template>
+							</SelectButton>
+							<aside>
```
mwdchang, line 77: What are the improvements here? 
A one or two liner description would be nice, or conversely don't include people that are not aware of the motivation(s) of this change in the review :)

PR #5647, opened 2024-11-26
Sensitivity analysis template

blanchco:
# Description

* A v1 for a new scenario template - Sensitivity analysis
* I made a base `tera-scenario-template` component that all the scenario templates can use since they all have a similar look/feel.

https://github.com/user-attachments/assets/a43da635-25e3-4fda-a3e9-6c3f803f37da


packages/client/hmi-client/src/components/workflow/scenario-templates/tera-scenario-template.vue
blanchco, line 1: Base class that all scenario template use...there are static variables such as the header and examples that are consistent throughout each scenario design.  This will also allow for consistent styling for every template form we make

packages/client/hmi-client/src/components/widgets/tera-modal.vue
```diff
@@ -112,7 +112,7 @@ main > section {
 }
 
 .content {
-	max-height: 80vh;
+	max-height: 70vh;
```
mwdchang, line 115: Were you the one who made this 80vh previously? Wondering if the reduction will impact someone else.

packages/client/hmi-client/src/components/workflow/scenario-templates/base-scenario.ts
```diff
@@ -1,9 +1,21 @@
...
+}
 export abstract class BaseScenario {
-	public static templateId: string;
+	public static templateId: string = '';
+
+	public static templateName: string = '';
+
+	public static header: ScenarioHeader = {
+		title: ''
+	};
```
mwdchang, line 16: I am not entire sure what you are going for here, the inheritance and static declaration seem to in opposition to each other.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/base-scenario.ts
blanchco: I've moved these out of the class and into the component, which should be easier.
Ex.
```
const header: ScenarioHeader = Object.freeze({
	title: 'Situational awareness template',
	question: "What's likely to happen next?",
	description:
		'Calibrates the model to historical data to obtain the best estimate of parameters for the present, then forecasts into the near future.',
	examples: [
		'Anticipate the arrival of a new variants.',
		'Evaluate the potential impact of growing vaccine hesitancy and declining NPIs.'
	]
});
```

mwdchang (approved):

packages/client/hmi-client/src/components/widgets/tera-modal.vue
blanchco: Yes I made it from 65 -> 80.  This was causing some visual issues when the component got too big (pushing the footer button down past the modal space).  70 seems to be a sweet spot

PR #5650, opened 2024-11-27
fix model config initial error state

blanchco:
# Description

* It looks like initials were forced to be evaluated as numbers, I've lifted that restriction in this PR
* Initials will only be "empty" when there is no value
* initial now updated on blur, which should reduce some api calls


YohannParis (approved):
LGTM

mwdchang (approved):

PR #5651, opened 2024-11-27
feat: knn project search ui

shawnyama:
# Description

* Assets and descriptions related to what was searched are now shown when searching for projects
* I also made the table fill the page like the cards along with making its header sticky
* Added a paginator for the table

https://github.com/user-attachments/assets/942e4842-fea4-42ff-b511-02c38effa52a


packages/client/hmi-client/src/components/home/tera-project-table.vue
```diff
@@ -1,23 +1,51 @@
...
-					<a class="project-title-link" @click.stop="emit('open-project', data.id)">
-						{{ data.name }}
-					</a>
+					<a @click.stop="emit('open-project', data.id)">{{ data.name }}</a>
+					<ul>
+						<li
+							v-for="(asset, index) in data.projectAssets.slice(0, data.showMore ? data.projectAssets.length : 3)"
+							class="flex align-center gap-2"
+							:key="index"
+						>
```
mwdchang, line 31: I dunno if the projectAssets list returned are dynamic or not, but you may want to consider key'ing with assetId instead of index as a safer option, since index does not change even if the array-size or array-elements have changed.

packages/client/hmi-client/src/components/home/tera-project-table.vue
```diff
@@ -92,18 +140,46 @@ function formatStatTooltip(stat, displayName) {
...
 }
+
+watch(
+	() => props.projects,
+	() => {
+		// Reset the page to 0 when a new search is performed
+		if (pageState.page !== 0) {
+			const firstPageButton = projectTableRef.value.$el.querySelector('.p-paginator-first');
+			firstPageButton?.dispatchEvent(new MouseEvent('click'));
+		}
```
mwdchang, line 177: This feels awkward, if a new search is performed wouldn't the entire table be refreshed because we have new search results?

packages/client/hmi-client/src/components/home/tera-project-table.vue
```diff
@@ -92,18 +140,46 @@ function formatStatTooltip(stat, displayName) {
...
+	// Just fetch the asset data we are seeing in the current page
+	projectsWithKnnMatches.value.slice(first, first + rows).forEach(async (project) => {
+		// If assets were fetched before from when we were on that page don't redo it
+		if (!isEmpty(project.projectAssets) && prevSearchQuery === searchQuery) return;
+
+		const projectWithAssets = (await ProjectService.get(project.id)) as ProjectWithKnnSnippet | null;
+		if (!projectWithAssets) return;
+
+		project.projectAssets = projectWithAssets.projectAssets.filter(
+			(asset) => asset.assetName.toLowerCase().includes(searchQuery) && asset.assetType !== AssetType.Simulation // Simulations don't have names
```
mwdchang, line 159: I don't know what you are trying to do here, but we just recently added NotebookSession as a projectAsset and it also doesn't have a name by default.

shawnyama: I am filtering the assets to show below the project title

packages/client/hmi-client/src/components/home/tera-project-table.vue
mwdchang: Oh? then that might be indicative that DataTable has an unexpected reactivity problem/behaviour, and there may be deeper problems that we have not uncovered yet. 

I wonder if using the UUID key trick is a more graceful solution? or whether it will janky? The idea is to force the table to refresh when you have a new search result because of the key-check. But I am not sure if the UI will flicker like mad if we do this.
```
< DataTable :key="tableKey">
</Datatable>


const tableKey = ref(uuid())

function doSearch() {
   ...blahblah...
   tableKey.value = uuid();
} 
```

shawnyama: that works

YohannParis (approved):
Great job 🎉

PR #5662, opened 2024-11-27
5472 task after enriching a model should show the descriptions

### LINKED ISSUE #5472: [TASK]: After enriching a model, should show the descriptions.
**Describe the task**
After enriching a model, should show the descriptions.

Try uploading this unenriched model and the paper.
Create a new workflow and add the model to it.
Open the model and click the enrich metadata button at the top. Wait for it to finish.
=> There is now nice description but they are all hidden by default. We should expand/show all descriptions where there is text after an enrichment.

**Screenshots**
I expanded the first two parameters descriptions manually. They were all collapsed after the enrichment.
Image

mloppie:
# Description

* Automatically expands the descriptions of model parts after enrichnment.
* Refactor to simplify the model part components.

Resolves #5472


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -111,22 +111,27 @@ import { CalendarDateType } from '@/types/common';
 
 const props = defineProps<{
 	item: ModelPartItem;
+	description?: string;
+	name?: string;
+	unitExpression?: string;
+	templateId?: string;
+	id?: string;
 	featureConfig: FeatureConfig;
```
mwdchang, line 119: Maybe I am missing something here, beforehand it is `props.item.description`, and now it is `props.description`. Why does it matter if it is nested under props or nested under props.item?

YohannParis: Better reactivity. I don't even think we need the `item` anymore.

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -152,8 +157,15 @@ watch(
 	{ immediate: true }
 );
 
-const showDescription = ref(false);
-if (props.item.description) showDescription.value = true;
+watch(
+	() => props.item.description,
```
YohannParis, line 161: Should it be `props.description`?

packages/client/hmi-client/src/components/model/model-parts/tera-model-part.vue
```diff
@@ -120,6 +120,11 @@
 						>
 							<tera-model-part-entry
 								:item="child"
```
YohannParis, line 122: Do we need the `item` here?

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/model/stockflow/tera-stockflow-parts.vue
shawnyama, line 1: I think the stockflow and regent parts can be deleted right @mwdchang @YohannParis

YohannParis: Yes they can.

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
mwdchang: In terms of reactivity, I think it might be a incorrect setup in Vue and we can address that. I don't know what to make of whether we need "item" or not, I did not take an in-depth read.

So we have this (or something similar to this):
```
<tera-model-part-entry
  v-else
  :is-time-part="!!isTimePart"
  :item="base"
  :description="base.description"
  :name="base.name"
  :id="base.id"
  :templateId="base.templateId"
  :unitExpression="base.unitExpression"
  :feature-config="featureConfig"
  @update-item="$emit('update-item', { id: base.id, ...$event })"
/>
```

The issue is that we have:
- `props.item.descriptions` and `props.description`
- `props.item.name` and `props.name`
- ...etc

That is both redundant and error-prone. I think if we need sub-items to be reactive, either
```
const description = ref(props.item.description)
```

or

```
const itemRef = ref(props.item)
```

or, as Yohann said if we don't need to governing "item" and just pass all the fields in as first-level props.

packages/client/hmi-client/src/components/model/model-parts/tera-model-part.vue
mloppie: Nope

mwdchang (approved):

Issue #5575, opened 2024-11-20
[TASK]: Remove subset and regrid datasets from the menu


pascaleproulx:
**Describe the task**
Remove subset and regrid datasets from the menu

PR #5669, opened 2024-11-28
remove data-grid and subset-dataset from workflow

mwdchang:
### Summary
Remove regrid and subset-dataset operators from workflow. This remove the registration only, the actual operators are still in-place and will be removed at a later date after next hackathon/eval.

Closes https://github.com/DARPA-ASKEM/terarium/issues/5575


YohannParis (approved):
LGTM

blanchco (approved):

PR #5674, opened 2024-11-28
Decode Project Overview properly before sending for embeddings

kbirk:


YohannParis (approved):
Missing the model.description, which is using the same base64 as project overview

dvince2 (approved):

YohannParis (approved):

PR #5681, opened 2024-11-28
Calibrate ensemble weights charts

jryu01:
# Description

- Added calibrate ensemble weight distribution charts
- Added a checkbox UIs to toggle on/off the loss chart and weight charts

Screenshot 2024-11-27 at 6 54 22 PM
Screenshot 2024-11-27 at 6 54 37 PM
Screenshot 2024-11-27 at 6 54 51 PM


packages/client/hmi-client/src/components/widgets/VegaChart.vue
```diff
@@ -246,6 +246,7 @@ watch(
 		} else {
 			// console.log('render interactive');
 			if (!vegaContainer.value) return;
+			vegaVisualization.value?.finalize(); // dispose previous visualization before creating a new one
```
mwdchang, line 249: Are there any noticeable impact from calling "finalize()" in this case?

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -541,6 +543,53 @@ export function useCharts(
...
+			const modelConfigIds = extractModelConfigIdsInOrder(chartData.value?.pyciemssMap ?? {});
+
+			const data = chartData.value?.result.filter((d) => d.timepoint_id === 0) ?? [];
+			const labelBefore = 'Before calibration';
+			const labelAfter = 'After calibration';
+
+			const colors = CATEGORICAL_SCHEME.slice(1); // exclude the first color which is for ensemble variable
+
+			modelConfigIds.forEach((configId, index) => {
+				const modelConfigName = getModelConfigName(modelConfigs, configId ?? '');
```
mwdchang, line 564: Are you doing `configId ?? ''` because typescript is complaining? It doesn't seem possible for configId to be null/undefined at this stage, and if they are null there are much greater problems. Or am I wrong here and you can actually get null configs?

packages/client/hmi-client/src/components/widgets/VegaChart.vue
jryu01: I wasn't 100% sure so I just did a test. 

https://github.com/user-attachments/assets/025d3fc7-f130-4c61-9045-47c19279aba9

It seems every time chart re renders, it adds up the memory usage and it doesn't free the memory up.
If you call `finalize()` It seems to free up the memory after some time (I think it's when GC kicks in). 

It also adds up the memory whenever the chart component is mounted (e.g. if you open/close drilldown repeatedly, it adds up quite a bit of memory usage) so I think we also need to explicitly clean up on `unmount` as well..

packages/client/hmi-client/src/composables/useCharts.ts
jryu01: Yeah configId is always string and we don't need `?? ''`. Probably it was a copy/paste mistake. Typescript compiler was fine.

packages/client/hmi-client/src/components/widgets/VegaChart.vue
jryu01: Calling `.finalize()`  before rendering new chart spec and calling it on component unmount seems to solve the issue.

mwdchang (approved):

Tom-Szendrey (approved):

PR #5708, opened 2024-11-29
setup compare simulations

shawnyama:
# Description

* setup compare simulations operator (node, drilldown, operation, inputs, outputs)


packages/client/hmi-client/src/components/workflow/ops/compare-simulations/compare-simulations-operation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const CompareSimulationsOperation: Operation = {
+	name: WorkflowOperationTypes.COMPARE_SIMULATIONS,
+	displayName: 'Compare simulations',
+	description: 'Compare simulations',
+	documentationUrl: '',
+	inputs: [
+		{ type: 'datasetId', label: 'Dataset' },
+		{ type: 'datasetId', label: 'Dataset' }
+	],
```
mwdchang, line 14: If we are taking in datasetIds, and "only" datasetIds, I think we are better off just calling this operator CompareDatasets ?? It would be a bit less confusing?? 

Or does this operator only work for dataset derived from a simulation?

YohannParis: Yes, let's just call it "Compare Datasets" for now until Jamie/Pascale says otherwise.

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-node.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+	<section>
+		<tera-operator-placeholder :node="node">Attach datasets/simulation outputs to compare</tera-operator-placeholder>
```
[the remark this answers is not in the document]
shawnyama: the default one is the same one

blanchco (approved):

PR #5705, opened 2024-11-29
Sympy 2 amr

mwdchang:
### Summary
`latex_to_sympy` => `latex_to_amr`  and update logic. It is expected to consume an list of LaTeX equations in the form of a JSON-array string, eg

```
[
  "\\frac{d S(t)}{d t} = - b_q S(t) I(t)",
  "\\frac{d E(t)}{d t} = b_q S(t) I(t) - r_X E(t)",
  "\\frac{d I(t)}{d t} = r_X E(t) - g_p_q I(t)",
  "\\frac{d R(t)}{d t} = g_p_q I(t)"
]
```

and outputs, if possible, a corresponding AMR JSON-string

### Testing
With this SEIR model equation

```
[
  "\\frac{d S(t)}{d t} = - b_q S(t) I(t)",
  "\\frac{d E(t)}{d t} = b_q S(t) I(t) - r_X E(t)",
  "\\frac{d I(t)}{d t} = r_X E(t) - g_p_q I(t)",
  "\\frac{d R(t)}{d t} = g_p_q I(t)"
]
```

Goto and run `http://localhost:3000/swagger-ui/index.html#/mira-controller/latexToAMR` 
- Verify that a response is returned with a JSON string of the AMR

Check the AMR is compatible, go to `https://app.staging.terarium.ai/amr-petri-test`, and paste in the response - you need to clean up the escape double-quotes first. Terarium should be able to parse and render the model

image

PR #5721, opened 2024-12-03
add extractionService for equations-to-model to use MIRA

mwdchang:
### Summary
Option extractionService = mira | skema

see: https://github.com/DARPA-ASKEM/terarium/pull/5705


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -260,29 +261,54 @@ public ResponseEntity<UUID> equationsToModel(
...
-		// Get an AMR from Skema Unified Service
+		// Get an AMR from Skema Unified Service or MIRA
 		final Model responseAMR;
-		try {
-			responseAMR = skemaUnifiedProxy.consolidatedEquationsToAMR(skemaRequest).getBody();
-			if (responseAMR == null) {
-				log.warn("Skema Unified Service did not return a valid AMR based on the provided equations");
-				throw new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, messages.get("skema.bad-equations"));
+		String extractionService = "mira";
+		if (req.get("extractionService") != null) {
```
dvince2, line 267: It feels really incorrect to have this be strongly typed on the front end in `knowledge.ts` and then have to deal with parsing json for values on the back end. This feels like something that should be generated from the back end and used from Types.ts

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -260,29 +261,54 @@ public ResponseEntity<UUID> equationsToModel(
...
+				taskReq.setUserId(currentUserService.get().getId());
+				final TaskResponse taskResp = taskService.runTaskSync(taskReq);
+				final JsonNode taskResponseJSON = mapper.readValue(taskResp.getOutput(), JsonNode.class);
+				final String amrString = taskResponseJSON.get("response").asText();
+				ObjectNode objNode = (ObjectNode) mapper.readTree(amrString);
+
+				final JsonNode testNode = mapper.readValue(amrString, JsonNode.class);
+				responseAMR = mapper.convertValue(testNode, Model.class);
+			} catch (Exception e) {
+				e.printStackTrace();
```
dvince2, line 287: If we want a print out here it should use `log.error("some message", e);`

dvince2 (approved):
Consider turning the request into a typed object?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
YohannParis: I agree with your comment, but this is a temporary solution for the hackathon, until we prove that MIRA is good enough and get rid of SKEMA.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
blanchco: done!

PR #5731, opened 2024-12-03
Update QoI chart data in optimize

Tom-Szendrey:
# Description
Previously our success criteria chart in optimize (the QoI chart) was based solely on the ciemss team's risk result json
This is an issue as they are always providing the same result irrespective of our optimize criteria form.

This PR moves this data creation on our side where we can react to things such as "All time points" vs "Last time points".

# Note:
- This is in draft as I have currently not adapted it to multiple criteria forms. This should be a quick fix tomorrow morning

# Pictures:
Please note here I am asking for the highest value of S for any time point. With S starting at 1 we can see that it correctly picks this. 
This is only a meaningful picture in that previously this chart could not be displayed. It is not necessarily a good use case for this feature but more so a very easy to see example
image

image


YohannParis (approved):
Great job 🎉

PR #5735, opened 2024-12-04
fix model-diagram layout option global bounds

mwdchang:
### Summary
The ortho edge routing's global bound was misconfigured, also added a manual edge-points if algorithm fails.

Before: 
Note D and E ends in transition which is incorrect

image

After:
Now D and E routes back onto themselves (this is location change within the same state)

image

### Test
Use the following, paste the content into http://localhost:8080/amr-petri-test or upload it as a model and view the diagram
[pp-bad-amr-render.json](https://github.com/user-attachments/files/18004396/pp-bad-amr-render.json)


Tom-Szendrey:
I do not see a difference in before vs after for E
I do however for D

Tom-Szendrey (approved):

PR #5741, opened 2024-12-04
Sensitivity analysis template update

blanchco:
# Description

* Creating sensitivity analysis template


mwdchang:
This generally looks  good, but there are problems with the design itself. There are some confusing points and several things that will not work.

- The naming of newly created model-configuration in the scenario bootstrapping process will become confusing after a few rounds
- This scheme will not really work for different types of distributions (eg more than just constant and uniform), but this is a bit low in terms of priority
- It seems to switch every thing to uniform, so it means I cannot set initial conditions for `S0 = 2000 => S0 = 3000`
- This will not work if we select a calibrated model-configuration as the base, because it doesn't have editable parameters, the actual data lives in the inferred_parameter.dll file

mwdchang (approved):
See comment, code looks okay but there are issues with the design itself and UX

YohannParis (approved):

PR #5768, opened 2024-12-05
Dataset statistics

YohannParis:
# Description

* Add a pandas stat task from a Dataset
* Updated `DatasetColumn` with a new property `stats` with proper object representation
* Add a new Service `DatasetStatistics` to handle the task and update the dataset accordingly
* Call the new service in two places, on upload of Dataset from the front-end and when adding a new file


packages/client/hmi-client/src/utils/number.ts
```diff
@@ -73,8 +73,8 @@ export function nistToNumber(numStr: string): number {
  * @param {string} num - The number string to display.
  * @returns {string} The number in either exponential form or NIST form.
  */
-export function displayNumber(num: string): string {
-	const number = fixPrecisionError(parseFloat(num));
+export function displayNumber(num: string | number): string {
+	const number = fixPrecisionError(parseFloat(num.toString()));
```
mloppie, line 77: Is there a reason you check for `NaN` in `numberToNist` but not here?

mloppie (approved):

packages/gollm/tasks/dataset_statistics.py
```diff
@@ -0,0 +1,100 @@
...
+            'total_columns': len(df.columns)
+        })
+
+        taskrunner.log("Statistics from input created")
+        taskrunner.write_output_dict_with_timeout(response)
+
+    except Exception as e:
+        sys.stderr.write(f"Error: {str(e)}\n")
+        sys.stderr.flush()
+        exit_code = 1
```
mwdchang, line 93: Let add a traceBack in the catch so we have a debugging trace:
```
import traceback

... snipped ...

sys.stderr.write(traceback.format_exc())
```

YohannParis: Great suggestion, learned something cool today!

dvince2 (approved):

mwdchang (approved):

PR #5777, opened 2024-12-06
Model comparisons MIRA

YohannParis:
# Description

Implementation of the [new MIRA comparison of model concepts](https://github.com/gyorilab/mira/blob/main/notebooks/hackathon_2024.12/comparison.ipynb)

* Adding controller to fetch the MIRA comparison
* Created the SYNC task in `mira-taskrunner`
* Add the call to the endpoint to the `services/concept.ts`
* Display the result in the `model-comparison` operator
* Display the graph images properly

https://github.com/user-attachments/assets/745bab7a-e1d6-4bc8-b9dd-8ad3e3d7c497


containers/scripts/docker-compose-taskrunner.yml
```diff
@@ -44,6 +44,7 @@ services:
 #      TERARIUM_MQ_PASSWORD: "terarium123"
 #      TERARIUM_MQ_USERNAME: "terarium"
 #      TERARIUM_TASKRUNNER_REQUEST_TYPE: "mira"
+#      MIRA_REST_URL: "http://mira-epi-dkg-lb-dc1e19b273dedaa2.elb.us-east-1.amazonaws.com/"
```
mwdchang, line 47: Is this being used? I can't seem to find a reference to it in the rest of your PR

packages/client/hmi-client/src/components/widgets/tera-csv-table.vue
```diff
@@ -0,0 +1,84 @@
...
+	dynamicTyping: true,
+	skipEmptyLines: true,
+	complete: onComplete,
+	error: onError
+};
+
+// Load CSV when component is mounted
+onMounted(() => {
+	try {
+		Papa.parse(props.csvText, config);
```
mwdchang, line 73: We already have a handful of csv parsers in house
- D3 has capabilities to parse csv and tsf
- We built an in-house `util/dsv.ts#dsvParse` to be a bit more intelligent for parsing matrices in csv/tsf format and numerical values

Since we are a bit short on time - I think we should flag this as to be consolidated. We don't need another CSV parsing library.

containers/scripts/docker-compose-taskrunner.yml
YohannParis: Yes, this is needed by MIRA ontological client.

packages/client/hmi-client/src/components/widgets/tera-csv-table.vue
YohannParis: Nice, sorry, `dsv` was not what I was looking for in the codebase. I will take a look and see if I can refactor.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/CompareModelsConceptsResponseHandler.java
```diff
@@ -0,0 +1,40 @@
...
+		List<String> amrs;
+	}
+
+	@Data
+	public static class Properties {
+
+		List<UUID> modelIds;
+		UUID workflowId;
+		UUID nodeId;
+	}
```
mwdchang, line 33: The abstractions level here feels a little strange: why are the input attributes of a taskrunner-request part of the output/response handler?  But we can sort these out later, just want to make a note.

mwdchang (approved):
Didn't test - some small comments but code generally looks fine.

packages/client/hmi-client/src/components/widgets/tera-csv-table.vue
YohannParis: I will not solve this, as DSV is hard coded for numbers only, and the object returned is not really useful to be displayed to the user.
I could add more feature to DSV to handle that use case, but this is out of scope of this PR...

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/CompareModelsConceptsResponseHandler.java
YohannParis: It's the opposite actually.
Because the task response can be consumed in another place, we can provide extra information as part of the request to be provided in the output.
In this case, this allow the notiifcationGroup to provide a link to the right drilldown to open.

dvince2 (approved):

PR #5780, opened 2024-12-06
Display shortDescription

YohannParis:
# Description

On search results we want to always display the description of an asset.

### DIFF
--- a/packages/client/hmi-client/src/components/home/tera-project-table.vue
+++ b/packages/client/hmi-client/src/components/home/tera-project-table.vue
@@ -69,8 +69,11 @@
 					:asset="asset"
 					@click="emit('open-asset', data.id, asset.assetId, asset.assetType)"
 				/>
-				<p class="font-semibold">{{ capitalize(asset.embeddingType) }}</p>
-				<tera-show-more-text :text="asset.embeddingContent" :lines="3" />
+				<tera-show-more-text :text="asset.assetShortDescription" :lines="3" />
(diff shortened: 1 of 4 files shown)


shawnyama (approved):

packages/client/hmi-client/src/components/home/tera-project-table.vue
```diff
@@ -69,8 +69,11 @@
 					:asset="asset"
 					@click="emit('open-asset', data.id, asset.assetId, asset.assetType)"
 				/>
-				<p class="font-semibold">{{ capitalize(asset.embeddingType) }}</p>
-				<tera-show-more-text :text="asset.embeddingContent" :lines="3" />
+				<tera-show-more-text :text="asset.assetShortDescription" :lines="3" />
+				<template v-if="visibleEmbeddingResult(asset.embeddingType)">
```
shawnyama, line 73: I think you can treat the `TerariumAssetEmbeddingType` as an object so you can do something like:
```
Object.values(TerariumAssetEmbeddingType).includes(asset.embeddingType)
```

blanchco (approved):

packages/client/hmi-client/src/components/home/tera-project-table.vue
YohannParis: Faire, but I wanted to be explicit, as in the near future we want to display hit information even if it's not the name, description.

PR #5783, opened 2024-12-09
utility to export svg into html image

mwdchang:
### Summary
Adds utility functions to convert SVG into image. 

This is not in use for now as where we want to apply SVG=>Image transform (model-diagram), it requires CSS style adjustments and tweaks to the governing logic that are probably better off as a separate PR.

### Testing
You add the this to the end of `renderGraph()` function in tera-model-diagram.vue to test, it appends the image(s) onto the document body. The image(s) are not 100% faithful because of the aforementioned style changes that are not carried over when converting to an image.

```
     const imageTest = await svgToImage(graphElement.value?.querySelector('svg') as any);
     document.body.appendChild(imageTest);
```

image


blanchco (approved):
okay!

shawnyama (approved):
cool

PR #5787, opened 2024-12-09
create notebook session when cloning

blanchco:
# Description

* When duplicating a data transform the notebook clone call would not actually create a new notebook session but just send back a clone of the session, nothing was done with this.  Now we are creating the new notebook session and sending it back

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
@@ -259,6 +259,11 @@ public class NotebookSessionController {
 				projectId,
 				permission
 			);
+
+			final Optional<Project> project = projectService.getProject(projectId);
+
+			projectAssetService.createProjectAsset(project.get(), AssetType.NOTEBOOK_SESSION, newNotebookSession, permission);
+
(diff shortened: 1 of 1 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -255,7 +254,12 @@ ResponseEntity<NotebookSession> cloneNotebookSession(
 			if (session.isEmpty()) {
 				return ResponseEntity.notFound().build();
 			}
-			return ResponseEntity.status(HttpStatus.OK).body(session.get().clone());
+			final NotebookSession newNotebookSession = sessionService.createAsset(
+				session.get().clone(),
+				projectId,
+				permission
+			);
+			return ResponseEntity.status(HttpStatus.OK).body(newNotebookSession);
```
mwdchang, line 262: laksdjflksdajflksadjf

blanchco: You're right, this is wasn't being adding to the project asset table when cloning. I've added some additional lines to to this and it is now being slotted into the project asst table

mwdchang (approved):

PR #5796, opened 2024-12-10
Update NotebookSessionController.java

YohannParis:
Fix how we get an asset

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
@@ -252,7 +252,7 @@ public class NotebookSessionController {
 		);
 		final NotebookSession session = sessionService
 			.getAsset(id, permission)
-			.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("asset.not-found")));
+			.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("notebook-session.not-found")));
 
 		NotebookSession newNotebookSession;
 		try {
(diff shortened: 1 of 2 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -260,9 +261,11 @@ ResponseEntity<NotebookSession> cloneNotebookSession(
 				permission
 			);
 
-			final Optional<Project> project = projectService.getProject(projectId);
+			final Project project = projectService
+				.getProject(projectId)
+				.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("asset.not-found")));
```
mwdchang, line 266: This will get caught by the general catch and it has a different error message, so you need to chain them together, or you need additional catch-clauses.

blanchco (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
```diff
@@ -250,29 +250,26 @@ ResponseEntity<NotebookSession> cloneNotebookSession(
...
+			newNotebookSession = sessionService.createAsset(session.clone(), projectId, permission);
 		} catch (final Exception e) {
 			final String error = "Unable to clone notebook session";
 			log.error(error, e);
 			throw new ResponseStatusException(org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR, error);
 		}
+
+		final Project project = projectService
+			.getProject(projectId)
+			.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("asset.not-found")));
```
mwdchang, line 268: So I dunno how serious you are about handling exceptions the "right" way. We are fetching two types of resources here: project and notebook-session, and they are both throwing the generic "asset.not-found".

Which means if we do encounter this exception, it isn't obvious where the error is on the front-end, on the backend, because we repackage the error with no chaining, it is also not obvious where things have awry. So overall this isn't very useful.

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/NotebookSessionController.java
YohannParis: I agree, but because this is a rare occurence error. It's not too important which of the two is NOT_FOUND.

Nonetheless, you are correct and this is now fixed.

PR #5800, opened 2024-12-10
value of information scenario template

blanchco:
# Description

* New value of information scenario template
* A user can select x number of interventions and y number of parameters with a low and high value.
* For each parameter a new model config with be created with that value as a uniform distribution. For example selecting beta and gamma with min and max value will create 2 new model configurations
* These new configurations will each be attached to a simulate node with each intervention node.
* some minor improvements to reduce repeated code between templates

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
@@ -18,7 +18,7 @@ export const displayParameter = (parameters: ParameterSemantic[], parameterName:
 	return `${parameterName}  [${value}]`;
 };
 
-export const formatUniformDistribution = (parameter: ParameterSemantic) => {
+export const switchToUniformDistribution = (parameter: ParameterSemantic) => {
 	if (parameter.distribution.type !== DistributionType.Uniform) {
 		let minimum = 0;
 		let maximum = 1;
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
@@ -1,0 +1,34 @@
+import { DistributionType } from '@/services/distribution';
+import { ParameterSemantic } from '@/types/Types';
+import { calculateUncertaintyRange } from '@/utils/math';
+
+export const displayParameter = (parameters: ParameterSemantic[], parameterName: string) => {
+	let value = '';
+	const parameter = parameters.find((p) => p.referenceId === parameterName);
+	switch (parameter?.distribution.type) {
+		case DistributionType.Constant:
+			value = `${parameter.distribution.parameters.value}`;
+			break;
+		case DistributionType.Uniform:
+			value = `${parameter.distribution.parameters.minimum} - ${parameter.distribution.parameters.maximum}`;
+			break;
+		default:
+			return '';
+	}
+	return `${parameterName}  [${value}]`;
+};
+
+export const formatUniformDistribution = (parameter: ParameterSemantic) => {
+	if (parameter.distribution.type !== DistributionType.Uniform) {
+		let minimum = 0;
+		let maximum = 1;
+		if (parameter.distribution.type === DistributionType.Constant) {
+			// +10% and -10% of the constant value
+			const { min, max } = calculateUncertaintyRange(parameter.distribution.parameters.value, 10);
+			minimum = min;
+			maximum = max;
+		}
+		parameter.distribution.type = DistributionType.Uniform;
+		parameter.distribution.parameters = { minimum, maximum };
+	}
+};
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
@@ -61,7 +61,7 @@ export class ValueOfInformationScenario extends BaseScenario {
 
 	interventionSpecs: { id: string }[];
 
-	simulateSpecs: { ids: string[] };
+	simulateSpec: { ids: string[] };
 
 	parameters: (ParameterSemantic | null)[];
 
@@ -74,7 +74,7 @@ export class ValueOfInformationScenario extends BaseScenario {
 		this.modelConfigSpec = {
 			id: ''
 		};
-		this.simulateSpecs = {
+		this.simulateSpec = {
 			ids: []
 		};
 		this.interventionSpecs = [{ id: '' }];
@@ -85,7 +85,7 @@ export class ValueOfInformationScenario extends BaseScenario {
 		this.modelSpec.id = id;
 		this.modelConfigSpec.id = '';
 		this.interventionSpecs = [{ id: '' }];
-		this.simulateSpecs.ids = [];
+		this.simulateSpec.ids = [];
 		this.parameters = [null];
 	}
 
@@ -94,7 +94,7 @@ export class ValueOfInformationScenario extends BaseScenario {
 	}
 
 	setSimulateSpec(ids: string[]) {
-		this.simulateSpecs.ids = ids;
+		this.simulateSpec.ids = ids;
 	}
 
 	addInterventionSpec() {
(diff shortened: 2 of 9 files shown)


packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+		case DistributionType.Uniform:
+			value = `${parameter.distribution.parameters.minimum} - ${parameter.distribution.parameters.maximum}`;
+			break;
+		default:
+			return '';
+	}
+	return `${parameterName}  [${value}]`;
+};
+
+export const formatUniformDistribution = (parameter: ParameterSemantic) => {
```
mwdchang, line 21: The function name is somewhat misleading here. You are not formatting a uniform-distribution, you are coercing non-uniform into a uniform format. A better choice of naming would be:
- convertToUniformDistrivution
- switchToUniformDistribution 
- ...etc

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+	if (parameter.distribution.type !== DistributionType.Uniform) {
+		let minimum = 0;
+		let maximum = 1;
+		if (parameter.distribution.type === DistributionType.Constant) {
+			// +10% and -10% of the constant value
+			const { min, max } = calculateUncertaintyRange(parameter.distribution.parameters.value, 10);
+			minimum = min;
+			maximum = max;
```
mwdchang, line 29: Note any percentage of 0 is still 0, you may want to default to [0, 1] if that is the case.

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -0,0 +1,317 @@
...
+
+	public static templateName = 'Value of information';
+
+	modelSpec: { id: string };
+
+	modelConfigSpec: { id: string };
+
+	interventionSpecs: { id: string }[];
+
+	simulateSpec: { ids: string[] };
```
mwdchang, line 65: I didn't notice this before, any particular why the two different ways of plurals `interventionSpecs: { id: string }[];` and `simulateSpec: { ids: string[] };` ? 

Eg why have both `{ ids: string[] }` and `{ id: string }[]` ?

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -0,0 +1,317 @@
...
+		this.parameters = [null];
+	}
+
+	setModelConfigSpec(id: string) {
+		this.modelConfigSpec.id = id;
+	}
+
+	setCalibrateSpec(ids: string[]) {
+		this.simulateSpec.ids = ids;
+	}
```
mwdchang, line 99: The way you build the scenario, it has no calibrate? either typo or you used the wrong operator

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -0,0 +1,317 @@
...
+			if (!parameter) return null;
+			const clonedModelConfig = _.cloneDeep(modelConfig);
+
+			const modelConfigNode = wf.addNode(
+				ModelConfigOp,
+				{ x: 0, y: 0 },
+				{
+					size: OperatorNodeSize.medium
+				}
+			);
```
mwdchang, line 209: Food for thoughts, you have done several scenarios now - if you feel like these addNode/addEdge is a good pattern and also workable for the foreseeable future scenarios, you may want to consider cleaning things up a bit as there are a lot of repeated things (the xy-coodinate and the operator size for example) and have wrapper functions. I know you can definitely do it for addNode, not really sure about addEdge since it might be port-position dependent. 

Anyway food for thought in the future and not needed for this PR, I will leave this with you.

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -0,0 +1,317 @@
...
+					value: [newModelConfig.id],
+					state: _.omit(modelConfigNode.state, ['transientModelConfig'])
+				}
+			});
+
+			return modelConfigNode;
+		});
+
+		// Wait for all modelConfigPromises to resolve and filter non null responses
+		const modelConfigNodes = (await Promise.all(modelConfigPromises)).filter((node) => !!node);
```
mwdchang, line 240: what is with the filter at the end? what does it do?

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -0,0 +1,317 @@
...
+					size: OperatorNodeSize.medium
+				}
+			);
+
+			wf.addEdge(modelNode.id, modelNode.outputs[0].id, modelConfigNode.id, modelConfigNode.inputs[0].id, [
+				{ x: 0, y: 0 },
+				{ x: 0, y: 0 }
+			]);
+
+			setParameterDistribution(clonedModelConfig, parameter.referenceId, parameter.distribution);
```
mwdchang, line 216: I don't understand how converting a parameter to uniform has to do with "representing the extremes of uncertainty" of some parameter as you had outlined in your description above. 

A uniform distribution has no "extreme" regions by definition. Maybe you can elaborate some more on the goal of this scenario to clarify.

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
blanchco: Ya I have already made a ticket for this todo eventually: https://github.com/DARPA-ASKEM/terarium/issues/5699

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
blanchco: This was a type, I've renamed it to simulate

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
blanchco: Gotcha, I'm now using `switchToUniformDistribution`

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
blanchco: Now I've added a check to see if min and max are 0, it will set max to 1

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
blanchco: Since there's an early return in the `flatMap` function when it can't find a parameter.  The return of the promise can be a `WorkflowNode<any>` or `null`.  The filter will filter out all the promises that return `null`.

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
blanchco: For the `simulateSpecs` I just wanted to hold a list of states/observables that the user wants to see there which can represent by an array of ids.  For the interventionSpecs there were some talks about potentially renaming/making new intervention policies within the template, so I wanted to keep a list of objects just in case we needed to keep more than just an `id` when/if that happens

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
mwdchang: So I am not sure why you need to use `flatMap`, the structure is only a single level no?

But that aside, if you know that result cannot be null, just declare or coerce the data as such or sprinkle in `@ts-ignore`. We are not obligated to satisfy Typescript if it makes non-sensical warnings. Using additional control structure not only make reading more confusing, it also introduce additional performance cost with no clear benefits.

mwdchang (approved):

YohannParis (approved):

shawnyama (approved):
lgtm

PR #5802, opened 2024-12-10
Output Settings Primary Color Picker

asylves1:
# Description

image

Design change:
- enable colour picking of primary colour for charts using output settings

Comparison & Error chart will be handled in another PR

Resolves #(issue)


packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -187,7 +187,7 @@ export function useCharts(
 	};
 
 	// Create options for forecast charts based on chart settings and model configuration
-	const createForecastChartOptions = (setting: ChartSetting) => {
+	const createForecastChartOptions = (setting: ChartSetting, color: string | null = null) => {
```
jryu01, line 190: Looks like you don't need an additional argument, `color` since you can access the color via setting object here. e.g. `setting.primaryColor`

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -130,10 +130,10 @@ export function useCharts(
 		setting: ChartSettingEnsembleVariable,
 		modelConfigId = '',
 		multiVariable = false,
-		showBaseLines = false
+		showBaseLines = false,
+		color: string | null = null
```
jryu01, line 134: Same here, `color` argument is not needed since you can access from `setting.primaryColor`

packages/client/hmi-client/src/composables/useChartSettings.ts
```diff
@@ -101,6 +101,24 @@ export function useChartSettings(
 		});
 	};
 
+	const updateChartPrimaryColor = (settings: ChartSetting, color: string) => {
+		const index = chartSettings.value.findIndex(
```
jryu01, line 105: You should be able to find the chart setting by it's id. e.g. `settings.id === chart.id`

packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -69,7 +73,7 @@ const props = defineProps<{
 	generateAnnotation?: (setting: ChartSetting, query: string) => Promise<ChartAnnotation | null>;
 }>();
 
-const emit = defineEmits(['close', 'update-settings-scale', 'delete-annotation', 'create-annotation']);
+const emit = defineEmits(['close', 'update-settings-scale', 'delete-annotation', 'create-annotation', 'change-color']);
```
jryu01, line 76:
```suggestion
const emit = defineEmits(['close', 'update-settings-scale', 'delete-annotation', 'create-annotation', 'update-settings-color']);
```
Maybe we should rename it so that it's consistent with other existing event name.

shawnyama (approved):

packages/client/hmi-client/src/components/widgets/tera-chart-settings-item.vue
```diff
@@ -10,14 +10,17 @@
...
-defineProps({
+const props = defineProps({
 	settings: {
 		type: Object,
 		default: () => ({})
 	}
 });
 defineEmits(['open', 'remove']);
+
+const primaryColor = computed(() => (props.settings.primaryColor ? `4px solid ${props.settings.primaryColor}` : ''));
```
mwdchang, line 23: Better naming here, border-style is not a colour.

mwdchang (approved):

PR #5804, opened 2024-12-10
Make Dockerfile more readable

YohannParis:
Use higher-level `apt` instead of lower-level. This was recommended by @bigglesandginger 

source: https://manpages.debian.org/jessie/apt/apt.8.en.html?ref=itsfoss.com

>The [apt commands](https://manpages.debian.org/jessie/apt/apt.8.en.html?ref=itsfoss.com) have been introduced to solve this problem. apt consists of some of the most widely used features from apt-get and apt-cache, leaving aside obscure and seldom-used features. It can also manage [apt.conf](https://linux.die.net/man/5/apt.conf?ref=itsfoss.com) file.

> With apt, you don’t have to fiddle your way from apt-get commands to apt-cache. apt is more structured and provides the user with the necessary options needed to manage packages.


mwdchang:
@YohannParis did changing from apt-get to apt solve an actual issue? cause github is flagging this as bad practice

```
DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead
```

YohannParis:
Based on @mwdchang recommendation, revert change.
Took advantage of this PR to add long name options for readability.

mwdchang (approved):

PR #5806, opened 2024-12-11
Improve compare simulation charts

shawnyama:
# Description

- use the forecast chart to plot the simulate comparisons
- add output settings
- clean up code for calculating differences between simulations

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
@@ -190,7 +190,7 @@ const commonHeaderNames = ref<string[]>([]);
 const timepointHeaderName = ref<string | null>(null);
 
 const plotOptions = [
-	{ label: 'Raw data', value: PlotValue.VALUE },
+	{ label: 'Raw values', value: PlotValue.VALUE },
 	{ label: 'Percent change', value: PlotValue.PERCENTAGE },
 	{ label: 'Difference', value: PlotValue.DIFFERENCE }
 ];
--- a/packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
@@ -108,7 +108,7 @@
 			>
 				<template #content>
 					<tera-drilldown-section class="px-2">
-						<label>What values do you want to plot?</label>
+						<label>How do you want to plot the values?</label>
 						<div v-for="option in plotOptions" class="flex align-items-center" :key="option.value">
 							<RadioButton
 								v-model="knobs.selectedPlotValue"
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
```diff
@@ -262,60 +293,99 @@ function findDuplicates(strings: string[]): string[] {
 }
 
 async function createCharts() {
+	// FIXME: Temporary check, useChartSettings updates the state directly but the knobs are not updated
+	// If this isn't synced the chartSettings will be reverted to the previous state
+	if (!isEqual(props.node.state.chartSettings, knobs.value.chartSettings)) {
+		knobs.value.chartSettings = props.node.state.chartSettings;
+	}
+
```
shawnyama, line 301: The `useChartSetttings` is set up to `update-state` directly but not with the `knobs`. So this is a workaround so the knob watcher doesn't remove the update from the `update-state`. 

I am starting to think that for the operators in general we should just run `update-state` when the `knobs` watcher is triggered. We are using both ways in most operators. I think I recall there was some infinite loop cases when watching for certain knobs...(idk). But I think it'll be worth cleaning this up if we can

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
```diff
@@ -209,9 +239,10 @@ const updateCriteria = (card: Partial<CriteriaOfInterestCard>, index: number) =>
 
 const knobs = ref<BasicKnobs>({
 	criteriaOfInterestCards: [],
-	selectedPlotValue: PlotValue.PERCENTAGE,
+	selectedPlotType: PlotValue.PERCENTAGE,
 	selectedCompareOption: CompareValue.IMPACT,
-	selectedDataset: null
+	selectedDataset: null,
+	chartSettings: null
```
jryu01, line 245: why chartSettings needs to be inside knobs? In my understanding knobs are for input settings and it should be separated from output settings.

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
```diff
@@ -182,17 +205,24 @@ const isATESelected = ref(false);
 
 const compareCharts = ref<any[]>([]);
 
+const { activeChartSettings, chartSettings, selectedVariableSettings, removeChartSettings, updateChartSettings } =
+	useChartSettings(props, emit);
+
+const selectedCharts = computed(() => {
+	const selectedChartIds = selectedVariableSettings.value.map((setting) => setting.selectedVariables[0]);
+	return compareCharts.value.filter((chart) => selectedChartIds.includes(chart.title.text));
```
jryu01, line 213: Is there specific reason we are maintaining another list of charts, `compareCharts`? I think we should be able to just use existing `useVariableCharts` in `useCharts.ts` with `selectedVariableSettings`

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-drilldown.vue
```diff
@@ -262,60 +293,99 @@ function findDuplicates(strings: string[]): string[] {
...
+						timepoint: parseFloat(timepoints[index])
+					});
+				} else {
+					data[index][name] = value;
+				}
 			});
 		});
-
-		compareCharts.value.push(createDatasetCompareChart(values, headerName));
+		compareCharts.value.push(
```
jryu01, line 369: It looks like we can just re use `useVariableCharts` in `useCharts.ts` instead of creating a chart manually here. You may need to modify some logic inside `useVariableCharts` on how it creates the chart options to provide correct variable names though.

shawnyama:
will address @jryu01 comments in another pr

blanchco (approved):

jryu01 (approved):

PR #5810, opened 2024-12-11
5663 look into simulate operator for polling spinner

### LINKED ISSUE #5663: Look into Simulate Operator for polling + spinner
Dan has raised that we should add the spinner to the simulate node and that we should update to the same polling we are doing elsewhere.

mloppie:
# Description

* Fixes the simulate node spinner by switching to polling from notifications. 
* Also fixes the simulate drilldown spinner, previously it would not appear on the first click of Run.
* Elevates the spinner and blank graphics to the drilldown section component
* Remove spinner and blank in various places to reuse the one in drilldown section
* Changes the poller default threshold and interval to require fewer params in most cases

Resolves #5663

### DIFF
--- a/packages/client/hmi-client/src/api/api.ts
+++ b/packages/client/hmi-client/src/api/api.ts
@@ -108,9 +108,9 @@ export interface PollerResult<T> {
 }
 
 export class Poller<T> {
-	pollingInterval = 2000;
+	pollingInterval = 3000;
 
-	pollingThreshold = 10;
+	pollingThreshold = 350;
 
 	keepGoing = false;
 
--- a/packages/client/hmi-client/src/api/api.ts
+++ b/packages/client/hmi-client/src/api/api.ts
@@ -110,7 +110,7 @@ export interface PollerResult<T> {
 export class Poller<T> {
 	pollingInterval = 3000;
 
-	pollingThreshold = 40;
+	pollingThreshold = 150;
 
 	keepGoing = false;
 
--- a/packages/client/hmi-client/src/api/api.ts
+++ b/packages/client/hmi-client/src/api/api.ts
@@ -110,7 +110,7 @@ export interface PollerResult<T> {
 export class Poller<T> {
 	pollingInterval = 3000;
 
-	pollingThreshold = 350;
+	pollingThreshold = 40;
 
 	keepGoing = false;
 
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -225,6 +225,7 @@
 		<template #preview>
 			<tera-drilldown-section
 				:is-loading="isLoading"
+				:show-slot-while-loading="true"
 				:loading-progress="props.node.state.currentProgress"
 				:is-blank="!showOutputSection"
 				:blank-message="'Click \'Run\' to begin calibrating'"
@@ -251,7 +252,7 @@
 					class="p-3"
 					:summary-id="node.state.summaryId"
 				/>
-				<Accordion :active-index="lossActiveIndex" class="px-2" v-if="!isLoading">
+				<Accordion :active-index="lossActiveIndex" class="px-2">
 					<AccordionTab header="Loss">
 						<!-- Loss chart -->
 						<div ref="lossChartContainer">
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -371,6 +371,7 @@
 						:active-settings="activeChartSettings"
 						:generate-annotation="generateAnnotation"
 						@delete-annotation="deleteAnnotation"
+						@update-settings-color="onColorChange"
 						@close="activeChartSettings = null"
 					/>
 				</template>
@@ -749,6 +750,7 @@ const {
 	comparisonChartsSettingsSelection,
 	removeChartSettings,
 	updateChartSettings,
+	updateChartPrimaryColor,
 	addComparisonChartSettings
 } = useChartSettings(props, emit);
 
@@ -1008,6 +1010,10 @@ const getConfiguredModelConfig = async () => {
 	if (configuredModelId) {
 		configuredModelConfig.value = await getModelConfigurationById(configuredModelId);
 	}
+};
+
+const onColorChange = (color: string) => {
+	if (activeChartSettings.value) updateChartPrimaryColor(activeChartSettings.value, color);
 };
 
 onMounted(async () => {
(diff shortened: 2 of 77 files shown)


packages/client/hmi-client/src/api/api.ts
```diff
@@ -108,9 +108,9 @@ export interface PollerResult<T> {
 }
 
 export class Poller<T> {
-	pollingInterval = 2000;
+	pollingInterval = 3000;
 
-	pollingThreshold = 10;
+	pollingThreshold = 350;
```
mwdchang, line 113: These are not really sensible defaults (17+ minutes) for a poller. For me it is more natural to actively acknowledge that there is a long running process (calibrate, optimize, funman, ensemble-calibrate), then it is to active acknowledge that there is a short-running process (simulate).

Also note right now simulate will poll for 17+ minutes before it fails over, we should terminate simulate if it exceeds say ... 2 minutes.

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -247,13 +281,13 @@ watch(
...
-		state.inProgressForecastId = '';
-		state.forecastId = id;
-		emit('update-state', state);
+		const response = await pollResult(id);
+		if (response.state === PollerState.Done) {
+			const state = _.cloneDeep(props.node.state);
+			state.inProgressForecastId = '';
+			state.forecastId = id;
+			emit('update-state', state);
+		}
```
mwdchang, line 290: I think there may be some misunderstanding on what the task is asking, let me try to elaborate here.

As briefly mentioned before, the push-event had been broken for a significant amount of time, so we wanted to switch over to do polling since we know it works and is consistent across other operators. For the simulate operator, it would either be using polling or push-event, but not both (as it is doing currently) - as you would be kind of double-counting or double-executing in some sense with both mechanisms running simultaneously.

Moreover, there are two items we are polling for:
- A base simulation, that is a simulation without an intervention policy, and
- A simulation with an intervention policy
These are the "baseForecastId", "forecastId" ... fields in the operator's state. 
So I think this is only half working, but may appear to work because:
- you may not have exercised the intervention use case, and/or
- we had addressed a couple of problems with push-events in the meantime

To have a simulate with intervention, you need to create an intervention-policy operator and pipe it into the simulate-operator. Feel free to reach out if more clarification is needed here.

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -223,7 +223,12 @@
 
 		<!-- Output section -->
 		<template #preview>
-			<tera-drilldown-section v-if="showOutputSection">
+			<tera-drilldown-section
+				:is-loading="isLoading"
+				:loading-progress="props.node.state.currentProgress"
+				:is-blank="!showOutputSection"
+				:blank-message="'Click \'Run\' to begin calibrating'"
+			>
```
mwdchang, line 231: I think calibrate and ensemble-calibrate are a bit different, there is a loading spinner, but there is also the rendering of loss-values that gets rendered while the spinner is spinning. 

I am not sure if this covers this particular use case.

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -192,69 +189,42 @@ const isChartsEmpty = computed(
...
-		const simId = simulationNotificationData.simulationId;
+async function processPolling(id, propName) {
+	const response = await pollResult(id);
+	if (response.state === PollerState.Done) {
 		const state = _.cloneDeep(props.node.state);
-		state.errorMessage = { name: '', value: '', traceback: '' };
-		state.inProgressBaseForecastId = '';
-		if (event.data.state === ProgressState.Complete) state.baseForecastId = simId;
+		state[propName] = id;
+		state[`inProcess${propName}`] = ''; // fix after call
```
Tom-Szendrey, line 217: Reminder that this is the issue of capitalization:
Eg: `forecastId` provided and 
inProgress**F**orecastId needed

Tom-Szendrey:
Just confirming @mloppie on your branch while running ensemble calibrate I do see the loss values:

https://github.com/user-attachments/assets/f4b589e1-a95d-4918-9a95-6c36fd301dd3

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -100,8 +100,6 @@ const showSpinner = computed<boolean>(
 const poller = new Poller();
 const pollResult = async (runId: string) => {
 	poller
-		.setInterval(5000)
-		.setThreshold(350)
```
mwdchang, line 104: @Tom-Szendrey we should take this opportunity to adjust intervals and thresholds based on latest knowledge, if we expect the process to take a significant time to finish, specify nothing will give you a 2 minute window.
- ensemble-calibrate
- optimize

Tom-Szendrey: I believe calibrate should be included in that list as well

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -192,69 +189,42 @@ const isChartsEmpty = computed(
...
-		if (event.data.state === ProgressState.Complete) await processResult(simId);
+const poller = new Poller();
+const pollResult = async (runId: string) => {
+	poller.setPollAction(async () => pollAction(runId));
+	const pollerResults = await poller.start();
+	const state = _.cloneDeep(props.node.state);
+	state.errorMessage = { name: '', value: '', traceback: '' };
+
+	if (pollerResults.state === PollerState.Cancelled) {
+		// state.inProgressForecastId = '';
```
Tom-Szendrey, line 200: Confusing to leave these in here as commented out

Tom-Szendrey (approved):
Aside from my current comments this looks good
Thank you for this

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
mloppie: No reason, that's a typo

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -171,8 +171,6 @@ const variableCharts = useVariableCharts(selectedVariableSettings, groundTruth);
 const poller = new Poller();
 const pollResult = async (runId: string) => {
 	poller
-		.setInterval(3000)
-		.setThreshold(350)
```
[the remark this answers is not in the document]
mloppie: Changed to 150 after our offline discussion

PR #5846, opened 2024-12-16
Sympy code pipeline

mwdchang:
### Summary
This PR introduce an alternative pipeline for handling LaTeX => AMR.
- A new gollm task that takes latex to sympy (python code)
- A new mira task that takes sympy python code and produce AMR

Note this does not replace the current pipeline that uses sympy `parse_latex`, once this has been more thoroughly vetted we will slot this in.

### Testing
This is hooked into `http://localhost:3000/mira/latex-to-amr`


dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -269,47 +271,71 @@ public ResponseEntity<JsonNode> generateModelLatex(@RequestBody final JsonNode m
...
 	)
 	public ResponseEntity<JsonNode> latexToAMR(@RequestBody final String latex) {
-		// create request:
-		final TaskRequest req = new TaskRequest();
-		req.setType(TaskType.MIRA);
+		////////////////////////////////////////////////////////////////////////////////
+		// 1. Convert latex string to python sympy code string
+		//
+		// Note this is a gollm string => string task
+		////////////////////////////////////////////////////////////////////////////////
```
YohannParis, line 278: Because it has been decided to split the request between MIRA and GoLLM instead of just one task-runner. This should be in the Knowledge Controller, not in the MIRA one.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/mira/MiraController.java
```diff
@@ -269,47 +271,71 @@ public ResponseEntity<JsonNode> generateModelLatex(@RequestBody final JsonNode m
...
+		final TaskRequest latexToSympyRequest = new TaskRequest();
```
YohannParis, line 279: Why not use the `CompoundTask` for that purpose?

mwdchang: It looks like `CompoundTask` runs tasks independently, it doesn't really have the notion of chaining tasks together (eg what is the handoff between taskA and taskB) or some kind of computation graph.

dgauldie (approved):

PR #5848, opened 2024-12-16
simplify collab/sync

mwdchang:
### Summary
We had added various heuristics onto the workflow canvas that:
- attempt to give a smoother UX
- get around slow/chunky db updates
- asymmetric protocols (rest-api going in, push-events coming out)

These heuristics ended up making things more complicated and at times, can corrupt the workflow data as they can compete with each other. As a few of the things had been partially addressed (db slowness), we will rip out and adjust several heuristics to both simplify the logic and try to ensure better data integrity.

Cleaned up a bit of logging on the server side, it is a bit too excessive.

### Testing
Same workflow across two different tabs should mirror each other. There may still be some quirkiness with syncs, please raise them, however I will resolve them separately - the main point of this PR is to get us back to a more deterministic state and then tackle any problem that arises.


Tom-Szendrey (approved):
Thanks

blanchco (approved):
nice

shawnyama (approved):

dvince2 (approved):

PR #5849, opened 2024-12-16
new sensitivity line chart

blanchco:
# Description

* Added a new sensitivity line chart
* 7 colour quantiles
* There was some _hacking_ to make this work with the chart settings.  Each chart setting has its own parameters and timepoint but for now the controls use the first setting's parameters and timepoint and updates every chart setting with the selected parameters and timepoint.
* using the forecast chart to plot here, but some conditional logic was needed if there was bins.  It _might_ be worth it to make this it's own chart
* Polishing is needed, but this is wanted before the demo

https://github.com/user-attachments/assets/104f6590-825e-4f78-ba9f-cb5d476e8d5b

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
@@ -13,11 +13,10 @@
 		/>
 		<template v-if="type === ChartSettingType.SENSITIVITY && sensitivityOptions">
 			<!--FIXME: It might be better to move these inside the panel so that they can be controlled at an individual chart settings level -->
-
 			<label>Select parameter(s) of interest</label>
 			<MultiSelect
 				:disabled="_.isEmpty(selectedOptions)"
-				:placeholder="_.isEmpty(selectedOptions) ? 'Select variable(s) to display above' : 'Select parameters'"
+				placeholder="Select parameters"
 				:model-value="sensitivityOptions.selectedInputOptions"
 				:options="sensitivityOptions.inputOptions"
 				@change="
--- a/packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
@@ -21,7 +21,7 @@
 				:model-value="sensitivityOptions.selectedInputOptions"
 				:options="sensitivityOptions.inputOptions"
 				@change="
-					$emit('input-selection-change', {
+					$emit('sensitivity-selection-change', {
 						selectedInputVariables: $event.value,
 						timepoint: sensitivityOptions.timepoint
 					})
@@ -33,7 +33,7 @@
 				:disabled="_.isEmpty(selectedOptions)"
 				:model-value="sensitivityOptions.timepoint"
 				@update:model-value="
-					$emit('input-selection-change', {
+					$emit('sensitivity-selection-change', {
 						selectedInputVariables: sensitivityOptions.selectedInputOptions,
 						timepoint: $event
 					})
@@ -104,7 +104,7 @@ const emits = defineEmits([
 	'remove',
 	'selection-change',
 	'toggle-ensemble-variable-setting-option',
-	'input-selection-change'
+	'sensitivity-selection-change'
 ]);
 
 // Settings of the same type that we want to interact with.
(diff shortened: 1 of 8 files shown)


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -10,7 +10,7 @@ import type { FunmanBox, FunmanConstraintsResponse } from './models/funman-servi
 const VEGALITE_SCHEMA = 'https://vega.github.io/schema/vega-lite/v5.json';
 
 export const CATEGORICAL_SCHEME = ['#1B8073', '#6495E8', '#8F69B9', '#D67DBF', '#E18547', '#D2C446', '#84594D'];
-
+export const SENSITIVITY_COLOUR_SCHEME = ['#4575B4', '#91BFDB', '#E0F3F8', '#FFFFBF', '#FEE090', '#FC8D59', '#D73027'];
```
mwdchang, line 13: Would be good to have nature of the colour scheme either commented or in the name itself, eg sequential, diverging, categorical, ordinal ... etc. Since most folks cannot read hex-values very well.

packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
```diff
@@ -11,6 +11,35 @@
 			:variables="selectOptions"
 			@configuration-change="$emit('selection-change', $event.selectedVariable, type)"
 		/>
+		<template v-if="type === ChartSettingType.SENSITIVITY && sensitivityOptions">
+			<!--FIXME: It might be better to move these inside the panel so that they can be controlled at an individual chart settings level -->
+
+			<label>Select parameter(s) of interest</label>
+			<MultiSelect
+				:disabled="_.isEmpty(selectedOptions)"
+				:placeholder="_.isEmpty(selectedOptions) ? 'Select variable(s) to display above' : 'Select parameters'"
```
mwdchang, line 20: The placeholder message is confusing, I'd take out the "select variables to display above part". 

If you need instructional text for the users, put them outside.

packages/client/hmi-client/src/services/charts.ts
blanchco: I've added a comment here to address this (I've also learned what a diverging colour scheme is 😄 )

packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
blanchco: Ive removed this, this was here before I added the label

packages/client/hmi-client/src/services/chart-settings.ts
```diff
@@ -115,6 +119,34 @@ export function updateChartSettingsBySelectedVariables(
...
+			found.selectedInputVariables = options.selectedInputVariables ?? [];
+			found.timepoint = options.timepoint ?? 0;
+		}
+		return (
+			found ??
+			createNewChartSetting(variable, ChartSettingType.SENSITIVITY, [variable], {
+				selectedInputVariables: options.selectedInputVariables ?? [],
+				timepoint: options.timepoint ?? 0
+			})
+		);
```
YohannParis, line 143:
```suggestion
		if (found) {
			found.selectedInputVariables = options.selectedInputVariables ?? [];
			found.timepoint = options.timepoint ?? 0;
			return found;
		} else {
			return createNewChartSetting(variable, ChartSettingType.SENSITIVITY, [variable], {
				selectedInputVariables: options.selectedInputVariables ?? [],
				timepoint: options.timepoint ?? 0
			});
		}
```

YohannParis (approved):

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -652,10 +654,89 @@ export function useCharts(
...
+		// Extract the selected variable values and ensure they are numeric
+		const samples = records.map((record) => {
+			const value = record[selectedVariable];
+			const id = record.sample_id;
+			if (typeof value !== 'number') {
+				throw new Error(`The selected variable "${selectedVariable}" must map to numeric values.`);
+			}
+			if (typeof id !== 'number') {
+				throw new Error('The "sample_id" field must be a numeric value.');
+			}
```
mwdchang, line 675: Is checking for sample_id to be numeric an actual thing we care about? eg when would it go wrong realistically? and can we save some computation and just skip the check?

blanchco: Nah there probably wouldn't be an issue, Ill remove it

mwdchang (approved):

jryu01:
Code looks good to me

PR #5851, opened 2024-12-16
use gollm=>mira equation to amr pipeline

mwdchang:
# Summary
Replace `latex => sympy.parse_latex => mira` pipeline with `latex => gollm => mira` pipeline.
This is the 2nd part to: https://github.com/DARPA-ASKEM/terarium/pull/5846

This PR also clean up the mira-endpoint.
`/mira/latex-to-amr` => `/knowledge/equations-to-model-debug`

### Testing
POST `/knowledge/equations-to-model` should return a model-id, which in turn should reflect the equations.

```
{
  "equations": [
  "\\frac{d S(t)}{d t} = - b_q S(t) I(t)",
  "\\frac{d E(t)}{d t} = b_q S(t) I(t) - r_X E(t)",
  "\\frac{d I(t)}{d t} = r_X E(t) - g_p_q I(t)",
  "\\frac{d R(t)}{d t} = g_p_q I(t)"
   ]
}
```

### DIFF
--- a/packages/client/hmi-client/src/temp/Equations.vue
+++ b/packages/client/hmi-client/src/temp/Equations.vue
@@ -64,7 +64,7 @@ const latex2amr = async () => {
 		resultAmr.value.innerHTML = 'processing...';
 	}
 
-	const resp = await API.post('/mira/latex-to-amr', equations);
+	const resp = await API.post('/knowledge/equations-to-model-debug', equations);
 	const respData = resp.data;
 
 	if (resultCode.value) {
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
Lovely!

Tom-Szendrey (approved):
why the debug in `equations-to-model-debug` ?

mwdchang:
> why the debug in `equations-to-model-debug` ?

@Tom-Szendrey I want to flag it as not meant to be used aside from debugging. The current pipeline is not very transparent, it has 3-4 junctions (llm-enrichment, llm-code-gen, code-eval to sympy, ode-to-amr) where things can go wrong but not really set up to report errors at each stage. Because it is also coupled with SKEMA stuff and model enrichment, and is used as a one-shot model creation process in the UI, it is just not ideal to refactor it until we agree on a more sane process. So for now we have a debugging analog.

PR #5852, opened 2024-12-16
Minor poll updates

Tom-Szendrey:
# Description
This addresses a few minor things
- [X] Simulation polling updates did not include the `update-state` required when polling status is cancelled.
- [X] If simulation fails grab the error message
- [X] Simulation provides a list of Ids to `pyciemss-cancel-button`. We should include this check for our disable logic.
- [X] Add `isCancelling` flag to prevent user from double clicking cancel.

### DIFF
--- a/packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
+++ b/packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
@@ -3,7 +3,7 @@
 </template>
 
 <script setup lang="ts">
-import { computed } from 'vue';
+import { computed, ref } from 'vue';
 import Button from 'primevue/button';
 import { cancelCiemssJob } from '@/services/models/simulation-service';
 import { logger } from '@/utils/logger';
@@ -12,10 +12,14 @@ const props = defineProps<{
 	simulationRunId?: string | string[];
 }>();
 
-const disabled = computed(() => props.simulationRunId === '' || props.simulationRunId?.length === 0);
+const isCancelling = ref(false);
+const disabled = computed(
+	() => props.simulationRunId === '' || props.simulationRunId?.length === 0 || isCancelling.value
+);
 
 const cancelSimulation = async () => {
 	if (!props.simulationRunId) return;
+	isCancelling.value = true;
 	const cancelIds = Array.isArray(props.simulationRunId) ? props.simulationRunId : [props.simulationRunId];
 	cancelIds.forEach(async (id) => {
 		if (!id) return;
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
```diff
@@ -12,10 +12,14 @@ const props = defineProps<{
 	simulationRunId?: string | string[];
 }>();
 
-const disabled = computed(() => props.simulationRunId === '');
+const isCancelling = ref(false);
+const disabled = computed(
+	() => props.simulationRunId === '' || props.simulationRunId?.length === 0 || isCancelling.value
```
YohannParis, line 17: We've been using Lodash `isEmpty` extensively in the application.

```suggestion
	() => isEmtpy(props.simulationRunId) || isCancelling.value
```

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
```diff
@@ -12,10 +12,14 @@ const props = defineProps<{
...
+);
 
 const cancelSimulation = async () => {
 	if (!props.simulationRunId) return;
+	isCancelling.value = true;
```
YohannParis, line 22: When are you setting it back to `false`?

YohannParis (approved):

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
mloppie: typo in `isEmpty`

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
mloppie: Nice catch. I feel like it should be at the start of `pollAction`?

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
Tom-Szendrey: Very good point thank you

packages/client/hmi-client/src/components/pyciemss/tera-pyciemss-cancel-button.vue
Tom-Szendrey: Thank you totally missed this

mwdchang (approved):

PR #5853, opened 2024-12-16
Fix: Calibrate always showing 1%

Tom-Szendrey:
# Description
We are manually sorting the progress coming and then taking the first each time.
Resulting in the first progress % being the only % we see the entire run:

image


YohannParis (approved):
That indeed is quite funny bug

mloppie (approved):

PR #5863, opened 2024-12-17
Add no cache property to tasks

kbirk:
DC Edit:
- do not cache latex-to-sympy task, since it is LLM derived and we have difficulty validating whether it is correct or not
- fix taskrequest wrapper


YohannParis (commented):
Looks good, but I would prefer to reverse the logic for ease of understanding.
`isNoCache` -> `isCache`
defautl to `false` -> `true`.

mwdchang:
> Looks good, but I would prefer to reverse the logic for ease of understanding. `isNoCache` -> `isCache` defautl to `false` -> `true`.

I can change and test it, the gollm task latex-to-sympy we do not want to cache, since there is a definite "correct" result and we don't want the incorrect ones to be cached.

YohannParis (approved):
Should be good to merge, now that the boolean are easier to read.

PR #5867, opened 2024-12-17
sensitivity improvements

blanchco:
# Description
Some improvements for sensitivity charts
* Can now plot observables in sensitivity line chart (was missing the translationMap)
* Default to max timepoint for sensitivity charts
* Using the tooltipFormatter for the label values
* Move chart expression functions from component -> `chart.ts`

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
@@ -302,14 +302,16 @@
 							:sensitivity-options="{
 								inputOptions: Object.keys(pyciemssMap).filter((c) => ['parameter'].includes(modelPartTypesMap[c])),
 								selectedInputOptions: selectedSensitivityChartSettings[0]?.selectedInputVariables ?? [],
-								timepoint: selectedSensitivityChartSettings[0]?.timepoint ?? 0
+								timepoint:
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
```diff
@@ -302,14 +302,16 @@
...
+									selectedSensitivityChartSettings[0]?.timepoint ?? _.last(runResults[selectedRunId])?.timepoint_id
 							}"
 							@selection-change="
 								(e) =>
 									updateSensitivityChartSettings({
 										selectedVariables: e,
 										selectedInputVariables: selectedSensitivityChartSettings[0]?.selectedInputVariables ?? [],
-										timepoint: selectedSensitivityChartSettings[0]?.timepoint ?? 0
+										timepoint:
+											selectedSensitivityChartSettings[0]?.timepoint ?? _.last(runResults[selectedRunId])?.timepoint_id
```
mwdchang, line 314: For instances where you have `abc?.def` consider using non-null assertions or coercion if you know intrinsically they cannot be null or undefined - it produces more efficient code when you compile things to JS. It seems like this is a case where "_.last" should return a result.

blanchco: I've changed this to a computed value `lastTimepoint`.  If there are no run results (i.e. haven't run a simulation yet) it would be undefined but Im now doing those  additional checks in the computed value

mwdchang (approved):

PR #5876, opened 2024-12-19
scenario template improvements

blanchco:
# Description

Address some bugs/improvements raised and attached the tickets:
* outputs now show the id on the label rather than the name to identify them uniquely
* parameters in some templates are sorted alphabetically and can be searched


mwdchang (approved):
Looks okay

asylves1 (approved):

PR #5879, opened 2024-12-19
Calibrate ensemble minor UI tweaks for user

Tom-Szendrey:
# Description
- [x] Calibrate ensemble has a few conditions as to why the run button can be disabled.
This can lead to user confusion if theyre not familiar with the operator.
We should include tooltips similar to how we have done in the optimize-drilldown to help guide the user.

- [x] could the node just have 2 connections for model configurations on creation? Right now it just shows one and doesn't show the 2nd until the 1st is added. That doesn't really suggest that the 2nd is required

# Screenshots:
New node:
image

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
@@ -64,7 +64,9 @@ const runResults = ref<RunResults>({});
 const csvAsset = shallowRef<CsvAsset | undefined>(undefined);
 
 const areInputsFilled = computed(
-	() => props.node.inputs[0].value && props.node.inputs[1].value && props.node.inputs[2].value
+	() =>
+		props.node.inputs[0].value &&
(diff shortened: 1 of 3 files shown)


mloppie (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -63,7 +63,9 @@ const emit = defineEmits(['open-drilldown', 'update-state', 'append-output', 'ap
 const runResults = ref<RunResults>({});
 const csvAsset = shallowRef<CsvAsset | undefined>(undefined);
 
-const areInputsFilled = computed(() => props.node.inputs[0].value && props.node.inputs[1].value);
+const areInputsFilled = computed(
+	() => props.node.inputs[0].value && props.node.inputs[1].value && props.node.inputs[2].value
+);
```
mwdchang, line 68: Note: This technically would not work, if somehow I ended up with
1. model-config
2. nothing after disconnection
3. dataset
4. model-config

Tom-Szendrey: fair point I have updated to filter on modelConfigId's that have values assigned

PR #5894, opened 2024-12-19
don't enable resize-observer in model-diagram if in preview mode

mwdchang:
### Summary
Resize observer causes double rendering pass, often times with no visible changes to the rendering. 
This disables the resize observer in preview mode to speed up loading time in the workflow canvas.

### Testing
If you put a console.log into tera-model-diagram#renderGraph() you can track how many times render happens.


Tom-Szendrey (approved):
Nice catch - did not test

YohannParis (approved):

PR #5931, opened 2024-12-23
Jamie templates style tweaks

jamiewaese-uncharted:
misc style tweaks for the templates dialog
Monosnap Terarium 2024-12-23 17-18-38
Monosnap Terarium 2024-12-23 17-19-44
stuff like that


mloppie (approved):

blanchco (approved):

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
```diff
@@ -433,9 +460,9 @@ const modelConfiguration = ref<ModelConfiguration | null>(null);
 const model = ref<Model | null>(null);
 
 const lastTimepoint = computed<number>(() => {
-	if (isEmpty(runResults.value) || !selectedRunId.value) return 0;
+	if (!runResults.value || !selectedRunId.value) return 0;
```
blanchco, line 463: Not sure why we are changing this back here

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
```diff
@@ -471,13 +498,17 @@ const datasetId = computed(() => {
...
+	return selectedOutput ? selectedOutput.label : '';
+});
+
 const llmThoughts = ref<any[]>([]);
 const llmQuery = ref('');
 
 // input params
 const timespan = ref<TimeSpan>(props.node.state.currentTimespan);
 const numSamples = ref<number>(props.node.state.numSamples);
-const solverStepSize = ref<number>(props.node.state.solverStepSize);
```
blanchco, line 480: Are we getting rid of the solver step size in simulate?  If so we should get rid of it in the html template as well

blanchco (changes requested):

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-ciemss-drilldown.vue
Tom-Szendrey: We should not get rid of `solverStepSize` in simulate it should be editable in a `tera-input-number` which can be disabled depending on the selected `solver method`

blanchco: I've restore the step size, and I think it should be good now

blanchco (approved):

PR #5937, opened 2024-12-24
Jamie vegalite chart tooltip interactions tweaks

jamiewaese-uncharted:
Added this line that follows your mouse, with labels
line with value labels over chart on hover


mloppie (approved):

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -657,6 +659,275 @@ export function createForecastChart(
...
+					value: 0
+				}
+			}
+		};
+		layerSpec.layer.push(pointLayer);
+
+		// Add labels for each point for tooltip.
+		// This is the base layer with a white stroke around it to make the text readable
+		const labelLayerBase = {
+			mark: {
```
mwdchang, line 833: @jamiewaese-uncharted is this meant to replace the existing tooltip layer that is already in place?? I don't think we want to have multiple tooltip-layers acting all at once. Can you elaborate on what these changes are meant to do?

image

jamiewaese-uncharted: yes, this replaces the original tooltip

mwdchang (approved):

PR #5945, opened 2025-01-02
clean up old notebook css

shawnyama:
# Description

* clearing up an old issue, this `toolbar` class used to be wrapped around every `tera-notebook-jupyter-input` but now there is no need for it


mwdchang (approved):
Didn't test, looks okay

blanchco (approved):

PR #5946, opened 2025-01-02
change time unit with respect to the amr

blanchco:
# Description

* address the attached ticket for time with respect to amr
* removed the name, description, and concept from showing on the front end
* if the unit is "day" when creating a model, this will be changed to "date" so that the dropdown option is selected


packages/client/hmi-client/src/components/model/model-parts/tera-time.vue
shawnyama, line 1: This component isn't used anymore. Look at `tera-petrinet-parts` and `createTimeList()` in `/model-representation/service.ts`

shawnyama (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-time.vue
blanchco: In that case I'll remove that component!

dvince2 (approved):

PR #5949, opened 2025-01-03
Add guard against unnecessary workflow updates

mwdchang:
### Summary
We are queuing up workflow-update actions regardless of whether there are progress changes or not, this is wasteful. This PR adds guards to only update if progress actually changed.

### Testing
Run any of Calibrate, Optimize, Ensemble-Calibrate and check the network tab, we should only send updates if progress changes.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
@@ -108,7 +108,7 @@ const pollResult = async (runId: string) => {
 				if (checkpointData) {
 					const state = _.cloneDeep(props.node.state);
 					const newProgress = +((100 * checkpointData.progress) / checkpointData.totalPossibleIterations).toFixed(2);
-					if (newProgress > state.currentProgress) {
+					if (newProgress !== state.currentProgress) {
 						state.currentProgress = newProgress;
 						emit('update-state', state);
 					}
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -103,13 +103,15 @@ const pollResult = async (runId: string) => {
 		.setPollAction(async () => pollAction(runId))
 		.setProgressAction((data: Simulation) => {
 			if (runId === props.node.state.inProgressOptimizeId && data.updates.length > 0) {
+				// TODO: Need to check if "first" is actually the latest update
 				const checkpointData = _.first(data.updates)?.data as CiemssOptimizeStatusUpdate;
```
mwdchang, line 107: @Tom-Szendrey can you check that `first` here actually reflects the latest intermediate result? Recall that our data storage and fetch via hibernate is not reliable in terms of ordering, so typically we'd want to sort them manually .. somewhere...

mloppie (approved):

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
Tom-Szendrey: This was working as is but as you mentioned that the storage and fetch is not 100% reliable I have added the manual sort anyways.
I used `last` to keep consistent with our calibrate node's poller

Tom-Szendrey (approved):

PR #5951, opened 2025-01-03
Fix transitions & observable tables

blanchco:
# Description

Address changes that @liunelson wants for the transitions table:
* removed units from the transitions table
* use `id` instead of the template name for transitions
* show expressions for transitions & observables


mwdchang:
If what Nelson wanted  are not covered by the other bullet points, please elaborate so folks have a better sense of what to look out for.

mloppie (approved):

mwdchang (approved):
Looks okay!

PR #5967, opened 2025-01-06
add new intervention policy in templates

blanchco:
# Description

We can add a new intervention policy within a template:
* open an intervention dropdown
* click create policy button
* policies will not be saved to the project until the create button for the template is pressed AND it is acutally used within the template
* these will be set as blank intervention policies and the user can add the them as needed within the workflow

Added for templates: 
* Value of Information
* Horizontal Scanning
* Decision Making

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/InterventionController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/InterventionController.java
@@ -23,7 +23,9 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.server.ResponseStatusException;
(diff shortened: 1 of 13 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/InterventionController.java
```diff
@@ -102,23 +104,30 @@ public ResponseEntity<InterventionPolicy> getIntervention(
 	)
 	public ResponseEntity<InterventionPolicy> createIntervention(
 		@RequestBody final InterventionPolicy item,
+		@RequestParam(name = "skip-check", required = false) final boolean skipCheck,
```
dvince2, line 107: if this isn't required, do we need to set a `default value`?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/InterventionController.java
blanchco: Good point, I've added a default

Issue #5841, opened 2024-12-13
[FEAT]: provide good default names for simulation output datasets


pascaleproulx:
When using templates, we expect to have to do as little as possible to get to a meaningful result at the end. One thing that is not so smooth is the compare simulation operator: the datasets from the various simulation have abstract names that are not meaningful so you don't know what is what there. 

Request: Provide default names to the simulations (I think we do something similar in stratify?). The name should probably be the name of the intervention and 'baseline' or 'no intervention' when there is no intervention. 

![Image](https://github.com/user-attachments/assets/f21e7050-51c9-4c3a-b493-b3c6c8331648)

YohannParis:
@liunelson could you provide some good default names.

PR #5969, opened 2025-01-07
ranking interventions first pass and better simulation output naming

shawnyama:
# Description

- `modelConfigurationId` and `interventionPolicyId` are stored in the `metadata.simulationAttributes` of datasets resulting from simulations (we will need access to these assets when we rank interventions)
- Selecting model config and variable can be done in the criteria of interest cards
- Named the dataset outputs based on the intervention policy name based on [these instructions
](https://github.com/DARPA-ASKEM/terarium/issues/5841)


YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/SimulationController.java
```diff
@@ -345,7 +345,9 @@ public ResponseEntity<Dataset> createFromSimulationResult(
 		@PathVariable("id") final UUID id,
 		@PathVariable("project-id") final UUID projectId,
 		@RequestParam("dataset-name") final String datasetName,
-		@RequestParam("add-to-project") final Boolean addToProject
+		@RequestParam("add-to-project") final Boolean addToProject,
+		@RequestParam("model-configuration-id") final Optional<UUID> modelConfigurationId,
```
dvince2, line 349: I don't think `Optional` is right here, you can just use `UUID`. If you want to make these params optional it would be done like:
```
	@RequestParam(
			value = "model-configuration-id",
			required = false
		) final String modelConfigurationId,
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/SimulationService.java
```diff
@@ -93,6 +93,8 @@ public Dataset createDatasetFromSimulation(
 		final String datasetName,
 		final UUID projectId,
 		final boolean addToProject,
+		final Optional<UUID> modelConfigurationId,
+		final Optional<UUID> interventionPolicyId,
```
mwdchang, line 97: Similar to Derek's comment, you don't really need Optionals, just 
```
		final UUID modelConfigurationId,
		final UUID interventionPolicyId,
```
you can pass in nulls, Optionals are typically when you programmatically fetch something and you are not sure if you will have a result or not.

mwdchang (approved):

PR #5979, opened 2025-01-07
Transcribe workflow-service to hmi-server - Part 1

mwdchang:
### Summary 
Workflow updates and syncs can be quirky because we update the workflow object in a wholesale manner. This PR changes the paradigm by telling the server what we want to do to the workflow, rather than doing something to the workflow and asking the server to try to save it.

To not make the PR several thousands lines long, these are left in-place for now and will be done later. With exception of node-status the others are rarely used.
- update node-state
- update port-output
- update note-status
- branching

Added
- POST /workflow/:id/select-output/:node-id/:output-id
- POST /workflow/:id/append-output/:node-id
- POST /workflow/:id/node
- POST /workflow/:id/remove-nodes
- POST /workflow/:id/edge
- POST  /workflow/:id/remove-edges

There are analogous to the functions in client service/workflow.ts

### Testing 
Normal workflow operations should work as before

### Addendum 
services/workflow itself is used for template-scenarios, we haven't exactly worked out how that will change, so presently there are duplicated logic in the front-end and back-end


dvince2 (approved):
Didn't test, but nothing seems obviously wrong!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/InputPort.java
```diff
@@ -0,0 +1,23 @@
...
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import java.io.Serializable;
+import java.util.List;
+import java.util.UUID;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.experimental.Accessors;
+
+@Data
+@Accessors(chain = true)
```
Tom-Szendrey, line 13: Not urgent at all of course can be done in later PRs as this is big enough as is
If we add TSModel to these we can remove some of the typing done in `src/types/workflow.ts`

mwdchang: I am purposely NOT generating the workflow types because they are still not syntactically compatible between Java and Typescript.

Tom-Szendrey:
When updating a workflow I see that the updated workflow has no ID:

https://github.com/user-attachments/assets/260cfbce-b5e3-499c-bccf-a7701b391ec4

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
```diff
@@ -97,6 +95,11 @@ watch(
 		const modelId = inputs.find((input) => input.type === 'modelId')?.value?.[0];
 		const state = cloneDeep(props.node.state);
 
+		const modelInputs = inputs.filter((input) => input.type === 'modelId');
+		if (modelInputs[0].status === WorkflowPortStatus.CONNECTED) {
+			isModelInputConnected.value = true;
+		}
```
shawnyama, line 101: Is this the only reason the `immediate: true` is at the end of the watcher? Maybe this should be in an `onMounted`?

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -821,6 +823,105 @@ export const getWorkflow = async (id: string, projectId?: string) => {
...
+	return response?.data ?? null;
+};
+
+export const addEdge = async (id: string, edge: WorkflowEdge) => {
+	console.log('>> workflowService.addEdge', edge.source, edge.target);
+	const response = await API.post(`/workflows/${id}/edge`, edge);
+	return response?.data ?? null;
+};
+
+export const removeNodes = async (id: string, nodes: string[]) => {
```
shawnyama, line 913: I think it'd be more clear to call these nodeIds
```suggestion
export const removeNodes = async (id: string, nodeIds: string[]) => {
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
```diff
@@ -202,8 +208,498 @@ public Optional<Workflow> updateAsset(
 		return result;
 	}
 
+	@Observed(name = "function_profile")
+	private void cascadeInvalidStatus(final WorkflowNode sourceNode, final Map<UUID, List<WorkflowNode>> nodeCache) {
+		List<WorkflowNode> downstreamNodes = nodeCache.get(sourceNode.getId());
+		if (downstreamNodes == null) return;
+
+		for (final WorkflowNode node : downstreamNodes) {
+			System.out.println(">> " + node.getId());
```
YohannParis, line 217:
```suggestion
```

packages/client/hmi-client/src/services/workflow.ts
YohannParis, line 1: Could we remove all those console.log?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
```diff
@@ -264,4 +273,278 @@ public ResponseEntity<ResponseDeleted> deleteWorkflow(
...
+				org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR,
+				messages.get("workflow.update.create-output")
+			);
+		}
+
+		broadCastWorkflowChange(updated.get(), projectId);
+		return updated.map(ResponseEntity::ok).orElseGet(() -> ResponseEntity.notFound().build());
+	}
+
+	@PostMapping("/{id}/node")
```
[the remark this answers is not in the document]
mwdchang: I like to avoid using verbs on the path when creating API endpoints, as it is accessing some type of hierarchical resources, and `add-node` is not a resource, plus the PUT or POST already has a add/modify connotation.  

But as we can see a lot of times there are special cases, here with deleting nodes/edges:
- DELETE /workflow/:id/node/:node-id would be the idea approach, but not efficient in bulk
- DELETE /workflow/:id/node/id1,id2,id3 was discouraged
- DELETE /workflow/:id/node, with a body works, but is also unconventional as DELETE can, but is not expected to have a body part

So in the end it all kind of get mixed up

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
mwdchang: This needs to be watched because the connection can change. 

The reason of immediate is to compensate for what was previously a semi-computed isModelInputConnected variable which, as far as I know, wasn't hooked up correctly. Model-config, likewise has the same kind of scheme and I made the same corresponding change.

packages/client/hmi-client/src/services/workflow.ts
mwdchang: At the moment they are useful for me to debug, some of these are automatically triggered (eg finishing simulations, or in case of model and dataset operator defaulting an output). Seeing how they interleave give some clues as to why things may go awry.


I will remove them in my part-2 PR.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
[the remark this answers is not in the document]
YohannParis: Thank you for the detailled answer.

packages/client/hmi-client/src/services/workflow.ts
mwdchang: done

PR #5981, opened 2025-01-08
modify readers to have admins, evict cache when permissions changed

bigglesandginger:
# Description

Should add Admin Group users to reader contributors list (as opposed to the group itself).
Should evict the cached reader's list when permissions for a project change.

This is untested as currently I am failing to connect to S3 (I assume all are failing, but incase you can test without...)

Resolves #(issue)


dvince2 (approved):

Tom-Szendrey:
Just testing this locally now
I will only check a part of Dan's matrix

Tom-Szendrey (approved):
THANK YOU

PR #5982, opened 2025-01-08
add model from literature scenario

blanchco:
# Description

* Adding the "Reproduce model from literature" scenario 
* The user can select `n` amount of documents and set a goal. 
* The resulting workflow will be:
Document --> Model from Equations --> Model config -->  Simulate
Document --> Model from Equations --> Model config -->  Simulate

AND every Model from equations node will be attached to a single compare Model node


blanchco:
Not able to run locally at the moment so I coded this blind 😄 .  Waiting to be able to run locally to make this ready for review

blanchco:
Got it to start working, looking good!

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/model-from-literature/model-from-literature-scenario.ts
```diff
@@ -0,0 +1,182 @@
...
+			// Add nodes for each document spec
+			const documentNode = wf.addNode(
+				DocumentOp,
+				{ x: 0, y: 0 },
+				{
+					size: OperatorNodeSize.medium
+				}
+			);
+
+			const modelFromEquationsNode = wf.addNode(
```
mwdchang, line 97: As part of my changes to move the workflow-services into the server side, I _am_ changing these `addNode`, `addEdge` ... etc functions. So we should sync up on what you need for the templates because I will probably override the logic you are depending on.

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/model-from-literature/model-from-literature-scenario.ts
blanchco: As per our discussion sibling functions will be made in the meantime, and we will further discuss next week

PR #5983, opened 2025-01-08
Minor Fixes - Simulate, Calibrate

Tom-Szendrey:
# Description
There were a few small ugly things that were found and could use a fix
- Calibrate `update-state` was called as the `knobs watcher` would trigger from the `initialize()` 
- Simulate was missing a `update-state` call on the `endTime`
- Simulate was missing an import for Primevue's SelectButton leading to spam warnings


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -1048,7 +1048,8 @@ onMounted(async () => {
 
 watch(
 	() => knobs.value,
-	async () => {
+	async (newValue, oldValue) => {
```
YohannParis, line 1051: the async is not necessary if their is no await.
```suggestion
(newValue, oldValue) => {
```

YohannParis (approved):

mwdchang (approved):

PR #5989, opened 2025-01-09
Jamie add dragndrop upload file zone to resource panel

jamiewaese-uncharted:
BEFORE
In order to upload a file, you had to click the "Upload" button in the resource panel then drag files into the drop zone in the "Upload resources" modal. It's a pretty typical UX flow and it's fine. 

A few months ago Yohann suggested that it would be cool if we could also drag files directly onto the resource panel. This is a step in that direction. 

This also addresses a separate issue:  When we have two or three uploads in the cue, and then we add one more, the upload status bar gets displayed on all the files, not just the one that's uploading. It's just messy.
image

AFTER
This feature enables files to be dropped directly into the resources panel. They get passed through to the upload file modal and get uploaded as if you had dropped it into the drop zone in the modal.

Also, I fixed the status bar issue so that we only show one.

drop files on resource panel to upload


YohannParis (approved):
I'm impressed

mwdchang (approved):

PR #6020, opened 2025-01-11
5908 uncertainty intervals on compare datasets operator

jryu01:
# Description

- Prepare chart data in a way similar to what we do for other operators. This allows quantile (uncertainty interval) charts to be added and annotations to work properly.
- Leverages pyciemss map for better display names for the variables (see the drop down list for selecting variables)
- Added uncertainty interval charts mode
- Some refactors and clean ups

**Uncertainty interval**
Screenshot 2025-01-10 at 8 07 10 PM
**Variable names**
Screenshot 2025-01-10 at 8 07 27 PM
**Comparing non simulation datasets**
(For now, with the assumption of same column names and the existence of `timepoint_id` column)
Screenshot 2025-01-10 at 8 07 49 PM
**Annotation fixed**
Screenshot 2025-01-13 at 11 33 59 AM

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/compare-datasets/compare-datasets-utils.ts
```diff
@@ -0,0 +1,166 @@
...
+	summaryResults: DataArray[];
+	datasetResults: DataArray[];
+}
+
+export function isSimulationData(dataset: Dataset) {
+	// assume if the dataset has these two files, it's a simulation dataset
+	return (
+		dataset.fileNames?.find((name) => name === 'result.csv') &&
+		dataset.fileNames?.find((name) => name === 'result_summary.csv')
+	);
```
mwdchang, line 19: I think this is the part where @shawnyama had embedded some telltale metadata if the dataset came from a simulation. I am not totally sure if it is better here, but you may want to if it is worthwhile to sync up so we have one idiomatic way of checking provenance.

shawnyama: Actually datasets coming from a simulation always had this `dataset.metadata.simulationId` attribute. I added `dataset.metadata.simulationAttributes` which hold a `modelConfigurationId` and a `interventionPolicyId` but I think its more clear to check for `dataset.metadata.simulationId`

shawnyama (approved):

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/compare-datasets-utils.ts
jryu01: I updated to check `dataset.metadata.simulationId` instead. Thanks!

PR #6028, opened 2025-01-13
[FEAT]: Copy model interventions and configs with model #5739

### LINKED ISSUE #5739: [FEAT]: When sharing a model from one project to another, want the model in the second project to have the same configurations
We have people collaborating on a problem involving multiple models. Each person creates and configures one of the models and they want to pool the configured models (model + model configuration) into a single shared project. Currently, we can only share models between projects but the shared model loses all the configurations (except the default one). Downloading the AMR doesn't help because the configurations are not stored in the AMR.

dvince2:
# Description

Manually checking for a model to be copied, then check for intervention policies or configurations associated with that model to clone. The manual check has to be done because currently we walk the json of the `model` in a one-direction way, but models don't contain references to their configs/interventions... thats the other way around. This was a quicker fix for a deadline. Better fix would be to have models include references outward and then this problem goes away.

Resolves #5739

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/ProjectController.java
@@ -866,7 +866,7 @@ public class ProjectController {
 			if (owningProjectId != null) {
 				// if the asset is already under another project, we need to clone it and its
 				// dependencies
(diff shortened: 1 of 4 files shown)


mwdchang (approved):
Didn't test, seems okay.

A side question - are we exporting/importing calibrated model-configurations correctly? Because it has a proxy-reference to a simulation-id which is subsequently used to fetch the "calibration result" from S3.

YohannParis (approved):
LGTM

PR #6038, opened 2025-01-13
fix rendering edges that are created through quick-menu

mwdchang:
### Summary
Force edges to re-evaluate when adding connecting operators through the quick-menu.  This is a new problem due to splitting apart the updates APIs, and some certain actions no longer trigger edge evaluations because things are done piece-wise.

### Testing
Add a model to workflow
- Connect to model-config
- From the quick-menu in model-config (the "+" button), create a connection to calibrate and calibrante-ensemble

Both the calibrate/calibrate-ensemble node and connecting edge should be shown.


YohannParis (approved):

Tom-Szendrey (approved):
fair enough thank you!!

PR #6046, opened 2025-01-13
clean up compiler warnings

mwdchang:
Clean up compiler warnings


YohannParis (approved):
Thank you, it was on my to-do list for a year now.

PR #6052, opened 2025-01-14
fix compare dataset charts from template

blanchco:
# Description

* A recent change to the chart selection in compare dataset caused the template default selection to break since we were selecting different variables, this is now fixed with this changed
* removed now unused functions


YohannParis (approved):
Nice clean up

shawnyama (approved):

PR #6054, opened 2025-01-14
tweak the template builder for horizon scanning and value of information

blanchco:
# Description

* Tweak the template building for horizon scanning and value of information templates based on feedback

Horizon Scanning:
* Two parameters of interest were chosen - mu, beta
* Currently, the (incorrect) configurations are: (1) low mu, (2) high mu, (3) low beta, (4) high beta
* The correct configurations should be the cartesian product of the extrema: (1) low mu, low beta, (2) low mu, high beta, (3) high mu, low beta, (4) high mu, high beta
* If there are N parameters of interest, then there should be 2^N such configurations, not just 2 * N.

Value of Information:
* Currently, every source of uncertainty (a parameter and a (min, max) pair) is mapped to their own configuration
* All the sources of uncertainty should be included into a single configuration, not one configuration per source.


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
```diff
@@ -19,6 +19,10 @@ export const displayParameter = (parameters: ParameterSemantic[], parameterName:
 	return `${parameterName}  [${value}]`;
 };
 
+export function cartesianProduct<T>(arrays: T[][]): T[][] {
+	return arrays.reduce((acc, curr) => acc.flatMap((a) => curr.map((b) => [...a, b])), [[]] as T[][]);
+}
```
mwdchang, line 24: This is a bit hard to understand, please provide an example as a part of the comment.

blanchco: comment + example added

mwdchang (approved):

PR #6055, opened 2025-01-14
Simulate - Save Solver Step Size with output

Tom-Szendrey:
# Description
- Chart settings should not be stored in output's state as per a quick conversation in the design channel.
- make sure that simulate follows the same set up as the rest of the TA3


blanchco (approved):

Tom-Szendrey:
Draft as Simulate is not providing entire state to output. 
Calibrate, optimize, ensembles all do provide entire state.

Before merging this we should determine if we want to omit the chart settings from output and have all of these operators have the same solution

YohannParis (approved):
That's an interesting fix... I'm not sure that will scale well. But evaluation time it is!

PR #6056, opened 2025-01-14
Workflow sync fix part 3

mwdchang:
### Summary
This PR address a couple of sync issues

1. In part-1 https://github.com/DARPA-ASKEM/terarium/pull/5979 we changed appendOutput to also include a state, mostly this was because the state changes were done in part-2 https://github.com/DARPA-ASKEM/terarium/pull/6019 and it acted as a kind of polyfill. Basically in part-2 I forgot to remove it. Here we take this out but left adding a state as an option as it is probably a better way to handle new outputs and we can migrate slowly.

2. We introduced a bug with ☝️ in workflowservice.java checking nulls where we need to check for json-nulls on top of it.

3. To actually address the problem @Tom-Szendrey  found. There are scenarios where update-state event are emitted back-to-back-to-back programmatically. In order to ensure state variable dependencies are sane, we need to update the local copy immediately rather than debouncing the update to the DB.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -299,7 +299,7 @@ const saveNodeStateHandler = debounce(async () => {
 	const updatedWorkflow = await workflowService.updateState(wf.value.getId(), nodeStateMap);
 	nodeStateMap.clear();
 	wf.value.update(updatedWorkflow, false);
-}, 300);
+}, 250);
 
 const saveWorkflowHandler = () => {
 	saveWorkflowDebounced();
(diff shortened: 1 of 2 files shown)


dvince2 (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -299,7 +299,7 @@ const saveNodeStateHandler = debounce(async () => {
 	const updatedWorkflow = await workflowService.updateState(wf.value.getId(), nodeStateMap);
 	nodeStateMap.clear();
 	wf.value.update(updatedWorkflow, false);
-}, 300);
+}, 250);
```
Tom-Szendrey, line 302: Whats the thought process behind changing this? 
asking out of curiosity more than concern

Tom-Szendrey (approved):
Thank you - tested and fixed issue I was seeing

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
Tom-Szendrey: Makes sense, thank you

YohannParis (approved):
LGTM

PR #6058, opened 2025-01-14
filter rows and columns in matrix

shawnyama:
# Description

* Filter rows and columns
* Changed keys to be the row/col number instead of the index
* Fixed matrix scrolling

https://github.com/user-attachments/assets/f23309d6-a3e7-4e0c-9377-20801f740be9

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
@@ -32,6 +32,12 @@
 			label="Paste"
 			class="ml-auto"
 			severity="secondary"
+			v-tooltip.left="{
+				value:
+					filteredRowNames.length !== matrix.length || filteredColumnNames.length !== matrix[0].length
+						? `Hidden rows/columns won't be ignored.`
+						: ''
+			}"
 			:disabled="clipboardText === ''"
 		/>
 	</div>
--- a/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
@@ -1,62 +1,83 @@
 <template>
-	<div class="matrix-toolbar">
-		<Dropdown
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
```diff
@@ -301,6 +324,16 @@ onUnmounted(() => {
 	window.clearInterval(timerId);
 });
 
+// This just prepares the filters when the matrix is assigned for the first time
+watch(
+	() => matrix.value,
```
YohannParis, line 329: No need to change it, but you can do just like this to watch a computed or ref.

```suggestion
	matrix,
```

YohannParis (approved):
Nice clean up 🎉

packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
```diff
@@ -1,62 +1,85 @@
...
+			/>
+		</div>
+		<div v-if="matrix.length > 1">
+			<label>Rows</label>
+			<MultiSelect
+				v-model="filteredRowNames"
+				:options="matrix.map((r) => r[0].rowCriteria)"
+				:max-selected-labels="2"
+				filter
+			/>
```
mwdchang, line 19: This task ain't that easy ... 

Filter a matrix breaks copy-and-paste function.

For example, I stratify S such that I have:
- S_foo = 123
- S_bar = 456

If I filter to only show `S_bar` and paste in "999", S_foo is the one that is updated even though S_bar is the one that showing in the matrix. 

Also pasting seems to update the matrix but breaks the UI when the matrix-modal is dismissed, though I think this might be a separate issue - I will file a separate ticket and link here.

Before we start addressing this, check with Pascale and design folks to see if there are any additional requirements when pasting a grid of numbers while filtering is active.

shawnyama: @pascaleproulx said that accommodating this would be a refinement. For now I can add a little warning message when some rows/cols are unselected to let the user know that paste may paste into rows/column you have hidden

Issue #6031, opened 2025-01-13
[BUG]: "Compare datasets" operator does not seem to accept more 3-4 input ports


liunelson:
**Describe the bug**

The "Compare datasets" operator is supposed to be able to accept an arbitrary number of input dataset. However, when connect a 3rd or 4th dataset, (1) the connection does not stick reliably and (2) no new input appears to accept the next input.

https://app.staging.terarium.ai/projects/33fdb387-8e64-49d8-b766-34d6d39ded74/workflow/7d44149b-5604-4ee2-a314-2f45b29b0ef4?configHash=957d2a25&state=8bef5ab0e01c4cf08dce89736a683a82&session_state=70c2f048-6e52-4118-a1ec-c13ca45af672&code=773b22b7-d62e-4d73-9270-31e1741861a6.70c2f048-6e52-4118-a1ec-c13ca45af672.316f53e1-2b9f-439b-a300-2a5d6726355f

![Image](https://github.com/user-attachments/assets/493f3e46-8771-42b9-aac6-10a19df093da)

YohannParis:
I can replicate this issue by not being able to link more that *2* dataset.
https://app.staging.terarium.ai/projects/ec1c9313-26f9-4d35-8e60-ecb491eafaa0/workflow/15770927-8ac7-4169-9a67-1cdfa71a8e4c

PR #6061, opened 2025-01-14
hook up append-input event in workflow to server workflow update

mwdchang:
Closes https://github.com/DARPA-ASKEM/terarium/issues/6031

### Summary
`append-input` event wasn't wired up correctly. It only changes the client's version of the workflow, so the server's version does not have any of the newly created ports. This PR should address that.

Note: Things are a bit messy currently, as there is the workflow-wrapper, that functionally overlaps with the server logic. While the workflow-wrapper is more or less deprecated with respect to terarium-workflow.vue, it is used in the scenariot-templates. I marked a couple of areas where we should start tidying up things but these are out of scope here.


YohannParis (approved):
LGTM

PR #6071, opened 2025-01-15
Fetch simulation result from `getRunResultCSV` to leverage cached results

jryu01:
# Description

For simulation data result in compare datasets operator, using `getRunResultCSV` allows to use already cached results from the simulation operators which are inputs to this node. 

Resolves #(issue)


Tom-Szendrey (commented):
`getDatasetResultCSV` uses a cache as well?
Oh my mistake I see, because the inputs are simulations we want to use the simulation cache

Tom-Szendrey (approved):

YohannParis (approved):

PR #6080, opened 2025-01-15
Redirect to login screen on all 401 responses

bigglesandginger:
# Description

When an API or client-event call returned a 401, we checked the local login token to see if it was "valid" cancelling a 401.

isUserLoggedIn only checks the clinet's token, not reliable nor correct, but as it seems to be required to be called to make the login function even exist... who writes this sort of... but as the token is untrustworthy, we will force the login process

Resolves #(issue)


YohannParis:
The error comes from the fact that yes, if the `isUserLoggedIn` is `true`, there is no `login` method available.

from `node_modules/oidc-spa/src/oidc.ts`
```typescript
export declare namespace Oidc {
    export type Common = {
        params: {
            issuerUri: string;
            clientId: string;
        };
    };

    export type NotLoggedIn = Common & {
        isUserLoggedIn: false;
        login: (params: {
            doesCurrentHrefRequiresAuth: boolean;
            extraQueryParams?: Record<string, string>;
        }) => Promise<never>;
        initializationError: OidcInitializationError | undefined;
    };

    export type LoggedIn<DecodedIdToken extends Record<string, unknown> = Record<string, unknown>> =
        Common & {
            isUserLoggedIn: true;
            renewTokens(): Promise<void>;
            getTokens: () => Tokens<DecodedIdToken>;
            subscribeToTokensChange: (onTokenChange: () => void) => { unsubscribe: () => void };
            logout: (
                params:
                    | { redirectTo: "home" | "current page" }
                    | { redirectTo: "specific url"; url: string }
            ) => Promise<never>;
            subscribeToAutoLogoutCountdown: (
                tickCallback: (params: { secondsLeft: number | undefined }) => void
            ) => { unsubscribeFromAutoLogoutCountdown: () => void };

            /**
             *
             * backFromLoginPages: The user has returned from the identity server's login pages. The session was established using query parameters provided by the identity server.
             * sessionStorageRestoration: The user just reloaded the page. An OIDC token stored in the session storage was used to restore the session.
             * silentSignin: Silent Single Sign-On (SSO) was achieved by creating an iframe to the identity server in the background. HttpOnly cookies were utilized to restore the session without redirecting the user to the login pages.
             */
            loginScenario: "backFromLoginPages" | "sessionStorageRestoration" | "silentSignin";
        };
```

YohannParis (approved):
That an interesting approach !

PR #6082, opened 2025-01-15
5965 task compare datasets node thumbnail should be one of the charts

### LINKED ISSUE #5960: [DESIGN]: nodes can become unwieldy with many graphs.
Zach has brought up that nodes can become unwieldy with many graphs.
His suggestion is that we allow users to select which graphs from the drilldown should be displayed on the node itself.
Another possible solution is to cap each node to one graph - this would be quite quick

### LINKED ISSUE #5963: [TASK]: Chart settings panel should include option to display that chart on node thumbnail
CURRENT
There's no way for users to control what charts to show on the node preview thumbnails.

PROPOSED
The chart settings panel should have a checkbox for users to choose whether they want to display that chart on the node thumbnail or not. It can be put in its own section at the bottom of the panel. 

The default setting should be whatever we currently show by default in a node. This is so users can remove charts from the node if they don't want them, or add others that they may prefer.

Image

### LINKED ISSUE #5965: [TASK]: Compare datasets node thumbnail should be one of the charts
CURRENT
The compare dataset node has a preview thumbnail with an illustration of two different plants. It does not show any useful charts however.

Even if charts have been created, we don't show them in the node preview.
Image
Image

PROPOSED
Once charts have been created, we should show them in the preview thumbnail instead of the illustration.

The chart settings panel part of this may be covered in #5963 
But the node may need adjusting to actually display the selected charts.

mloppie:
# Description

* Updates the compare dataset node to show charts
* Also compare dataset shows a spinner while running, and a placeholder. 
* Adds option to hide a chart from the node in the output settings.
* New NodeCharts abstraction to contain the logic.
* Refactor nodes to use the new abstraction and gain the chart hiding feature.

Screenshot 2025-01-15 at 2 24 41 PM

Screenshot 2025-01-15 at 3 35 11 PM

Resolves #5960 
Resolves #5963 
Resolves #5965


packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -93,6 +98,7 @@ const emit = defineEmits(['close', 'update-settings', 'delete-annotation', 'crea
 
 // Log scale
 const useLog = computed(() => props.activeSettings?.scale === 'log');
+const isHiddenInNode = computed(() => !!props.activeSettings?.hideInNode);
```
YohannParis, line 101: Why not casting it as boolean here to avoid the casting in the prop `:model-value` line 18?
```suggestion
const isHiddenInNode = computed<boolean>(() => !!props.activeSettings?.hideInNode);
```

packages/client/hmi-client/src/components/workflow/ops/node-charts.vue
YohannParis, line 1: we pre-pend all the component with `tera-`.

packages/client/hmi-client/src/composables/useChartSettings.ts
```diff
@@ -22,7 +22,7 @@ import { WorkflowNode } from '@/types/workflow';
  */
 export function useChartSettings(
 	props: { node: WorkflowNode<{ chartSettings: ChartSetting[] | null }> },
-	emit: (event: 'update-state', ...args: any[]) => void
+	emit: (...args: any[]) => void
```
jryu01, line 25: Why do we remove `event: 'update-state'`? I think this component only uses 'update-state' event.

packages/client/hmi-client/src/components/workflow/ops/node-charts.vue
```diff
@@ -0,0 +1,42 @@
...
+import { computed } from 'vue';
+import VegaChart from '@/components/widgets/VegaChart.vue';
+import TeraProgressSpinner from '@/components/widgets/tera-progress-spinner.vue';
+import TeraOperatorPlaceholder from '@/components/operator/tera-operator-placeholder.vue';
+import { WorkflowNode } from '@/types/workflow';
+
+const props = defineProps<{
+	node: WorkflowNode<any>;
+	preparedCharts: any;
+	chartSettings: any;
```
jryu01, line 31:
```suggestion
	chartSettings:  ChartSetting[];
```

packages/client/hmi-client/src/components/workflow/ops/node-charts.vue
```diff
@@ -0,0 +1,42 @@
...
+<script setup lang="ts">
+import { computed } from 'vue';
+import VegaChart from '@/components/widgets/VegaChart.vue';
+import TeraProgressSpinner from '@/components/widgets/tera-progress-spinner.vue';
+import TeraOperatorPlaceholder from '@/components/operator/tera-operator-placeholder.vue';
+import { WorkflowNode } from '@/types/workflow';
+
+const props = defineProps<{
+	node: WorkflowNode<any>;
+	preparedCharts: any;
```
jryu01, line 30: Can we give more specific type other than `any`?

packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
mloppie: The boolean casting is also there for `useLog` and `isSmallMultiples` so I was continuing the pattern; but I like this better I'll change the others too

packages/client/hmi-client/src/composables/useChartSettings.ts
mloppie: Because the compare datasets node doesn't emit `update-state` and it was failing lint

packages/client/hmi-client/src/components/workflow/ops/node-charts.vue
mloppie: I originally had `Ref<ChartData | null, ChartData | null> | Record<string, VisualizationSpec[]>` but that was failing lint

mloppie: There's an issue I was having where something like `useComparisonCharts` is typed as `VisualizationSpec` even though it's an array. If I change the type in `useCharts` it breaks all the charts that don't use this component, and so it seemed out of scope for this week

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-node.vue
```diff
@@ -1,30 +1,106 @@
 <template>
 	<section>
-		<tera-operator-placeholder :node="node">Attach datasets/simulation outputs to compare</tera-operator-placeholder>
-		<Button v-if="hasAtLeastTwoValues" label="Open" @click="emit('open-drilldown')" severity="secondary" outlined />
+		<NodeCharts
```
mwdchang, line 3: Convention wise, we tend to use kebab-cases for our components. So `<node-charts...` instead of `<NodeCharts`. Also I know that @YohannParis liked to have a "tera-" prefix in front of our own components though I am not exactly sure what is the sentiment at this moment.

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,42 @@
...
+
+const props = defineProps<{
+	node: WorkflowNode<any>;
+	preparedCharts: any;
+	chartSettings: any;
+	chartSettingKey?: string;
+	isLoading?: boolean;
+	placeholder?: string;
+	processing?: string;
+	interactive?: boolean;
```
mwdchang, line 36: We probably do not need the interactive flag, as the charts in the graph nodes are generally too small for interactive use, so they will be false all the time.

jryu01 (approved):

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,40 @@
...
+	<tera-progress-spinner v-if="isLoading" :font-size="2" is-centered style="height: 100%">
+		<div v-if="processing">{{ processing }}</div>
+		<div v-else>Processing...</div>
+	</tera-progress-spinner>
+	<div v-else-if="visibleChartSettings.length">
+		<div v-for="setting of visibleChartSettings" :key="chartSettingKey || setting.id">
+			<vega-chart
+				:expandable="expandable"
+				:are-embed-actions-visible="areEmbedActionsVisible"
+				:visualization-spec="preparedCharts[chartSettingKey || setting.id]"
```
mwdchang, line 11: Sorry, to clarify by earlier comment.  `interactive` prop is an opt-out mechanism - if you don't explicitly set to false the charts will be interactive. You can verify this by looking at simulate charts which has tooltip interactions if interactive mode is on.

image

For the nodes pascale had asked the interactions to be turned off (to save memory, as we would render to png/svg and dispose the vegalite context).

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-node.vue
```diff
@@ -18,7 +18,6 @@
 					:key="setting.id"
 					:visualization-spec="variableCharts[setting.id]"
 					:are-embed-actions-visible="false"
-					:interactive="false"
```
Tom-Szendrey, line 21: I believe we want the nodes to be uninteractive

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+	<tera-progress-spinner v-if="isLoading" :font-size="2" is-centered style="height: 100%">
```
Tom-Szendrey, line 2: If I am understanding correctly this spinner is for the node running not the charts loading data
It is unintuitive to me to have this spinner hidden in this component
I prefer these spinners to be owned by the node. Nodes tend to be quite simple I dont think saving 3 lines per node is worth it here 

I am open to having my mind changed

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,41 @@
...
+<script setup lang="ts">
+import { computed } from 'vue';
+import VegaChart from '@/components/widgets/VegaChart.vue';
+import TeraProgressSpinner from '@/components/widgets/tera-progress-spinner.vue';
+import TeraOperatorPlaceholder from '@/components/operator/tera-operator-placeholder.vue';
+import { WorkflowNode } from '@/types/workflow';
+
+const props = defineProps<{
+	node: WorkflowNode<any>;
+	preparedCharts: any;
```
Tom-Szendrey, line 30: We should have the type for this rather than any

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,41 @@
...
+	chartSettings: any;
```
Tom-Szendrey, line 31: This should be type` ChartSetting[] | null` I believe rather than any

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -217,12 +217,11 @@
...
-										v-for="(spec, index) of ensembleVariableCharts[setting.id]"
-										:key="index"
-										expandable
-										:are-embed-actions-visible="true"
-										:visualization-spec="spec"
+									<tera-node-charts
+										:node="node"
+										:prepared-charts="ensembleVariableCharts"
+										:chart-settings="selectedEnsembleVariableSettings"
+										:expandable="true"
```
Tom-Szendrey, line 224:
```suggestion
										expandable
```

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
```diff
@@ -6,12 +6,7 @@
 			:visualization-spec="lossChartSpec"
 		/>
 		<div v-if="outputData">
-			<vega-chart
-				v-for="(chart, index) of ensembleVariableCharts"
-				:key="index"
-				:interactive="false"
-				:visualization-spec="chart"
-			/>
+			<vega-chart v-for="(chart, index) of ensembleVariableCharts" :key="index" :visualization-spec="chart" />
```
Tom-Szendrey, line 9: If we are creating a component specifically to handle the charts that nodes deal with I think that this node should follow the others and use it as well

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
jryu01: yeah I think the spinner needs to be on the node not inside here. While it's loading, we won't even render the charts.  And without the spinner, I wonder if we need this component at all.

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
```diff
@@ -0,0 +1,41 @@
...
+		<div v-for="setting of visibleChartSettings" :key="chartSettingKey || setting.id">
+			<vega-chart
+				:expandable="expandable"
+				:are-embed-actions-visible="areEmbedActionsVisible"
+				:visualization-spec="preparedCharts[chartSettingKey || setting.id]"
+				:interactive="false"
+			/>
+		</div>
+	</div>
+	<tera-operator-placeholder v-else-if="placeholder" :node="node">
```
jryu01, line 16: I think we just need one place holder at the node level not for each individual chart.

mloppie: I've updated the `node-charts` to be `tera-node-preview` and it can now handle groups of charts and no longer needs to be instatiated multiple times in a node

packages/client/hmi-client/src/components/workflow/ops/tera-node-preview.vue
```diff
@@ -0,0 +1,60 @@
...
+import VegaChart from '@/components/widgets/VegaChart.vue';
+import { ChartSetting } from '@/types/common';
+import TeraProgressSpinner from '@/components/widgets/tera-progress-spinner.vue';
+import TeraOperatorPlaceholder from '@/components/operator/tera-operator-placeholder.vue';
+import { WorkflowNode } from '@/types/workflow';
+
+const props = defineProps<{
+	node: WorkflowNode<any>;
+	preparedCharts: any;
+	chartSettings: ChartSetting[] | null;
```
Tom-Szendrey, line 45: I may be mistaken but I believe this is now
`ChartSetting[][] | ChartSettings[] | null`

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -217,12 +217,11 @@
 						<template v-if="!isRunInProgress">
 							<AccordionTab v-if="selectedEnsembleVariableSettings.length > 0" header="Ensemble variables over time">
 								<div class="flex flex-row" v-for="setting of selectedEnsembleVariableSettings" :key="setting.id">
-									<vega-chart
-										v-for="(spec, index) of ensembleVariableCharts[setting.id]"
-										:key="index"
-										expandable
-										:are-embed-actions-visible="true"
-										:visualization-spec="spec"
+									<tera-node-charts
```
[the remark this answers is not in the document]
mloppie: It shouldn't; reverted

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-node-ciemss.vue
mloppie: There's no chartSettings being used here to key the chart spec like in the other cases. I'm not sure `vega-chart` needs a wrapper that doesn't add any convenience

packages/client/hmi-client/src/components/workflow/ops/tera-node-charts.vue
mloppie: We don't unfortunately, as discussed above: https://github.com/DARPA-ASKEM/terarium/pull/6082#discussion_r1917724004

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/compare-datasets-utils.ts
```diff
@@ -169,3 +177,161 @@ export function buildChartData(
...
+	baselineDatasetIndex,
+	selectedPlotType
+) {
+	chartData.value = buildChartData(
+		datasets.value,
+		datasetResults.value,
+		baselineDatasetIndex.value,
+		selectedPlotType.value
+	);
+}
```
shawnyama, line 269: I think these params should be typed. I am not getting any charts appear for myself, it seems like I am passing datasets that are currently undefined

mloppie: I didn't change any code in the drilldown, just moved it to the util.

PR #6086, opened 2025-01-15
Jamie adjust situational awareness template

jamiewaese-uncharted:
BEFORE
Some templates had a bit of a confusing workflow due to the order of the form fields. For instance, Situational awareness had the dataset field above the configuration field. 

Also, some templates had a field for selecting interventions, but it wasn't disabled before a model is selected.

Also, the radio button list wasn't beautiful.
image

Also, there were a bunch of titles that weren't sentence case

AFTER
image


YohannParis (approved):

blanchco (approved):
looks good

blanchco (changes requested):
It looks like some of the fields are based off model configs rather than models, let me change these

blanchco (approved):

PR #6088, opened 2025-01-16
revive workflow annotations

mwdchang:
### Summary
This PR sets up the plumbing and support for adding text-annotations onto a workflow canvas. A new Workflow field

```
private Map<UUID, WorkflowAnnotation> annotations;
```

is added as a new json-blob column in the database for storing annotations. This PR provides the basic UI to do CRUD-operations on workflow-annotations, however Jamie will be refining the UI in his pass after this PR has been merged.

Note: Annotation are not versioned or tombstoned. It means that if you delete an annotation while another person is looking at the same workflow, there is a chance the delete will get reverted. If it is important we can add the version support, otherwise I am more inclined to just keep it simple.

https://github.com/user-attachments/assets/4aad8b4f-40c2-4462-b26d-0d36eed7a9e0

### Testing
The context-menu has a new "Add annotation" option. You can add, remove, update, and drag the annotations on the canvas.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.UUID;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.experimental.Accessors;
+
+@Data
+@Accessors(chain = true)
+public class WorkflowAnnotation implements Serializable {
+
+	private UUID id;
```
Tom-Szendrey, line 16: We may want to store width and height

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -916,6 +925,18 @@ export const addEdge = async (id: string, edge: WorkflowEdge) => {
 	return response.data ?? null;
 };
 
+export const addOrUpdateAnnotation = async (id: string, annotation: WorkflowAnnotation) => {
+	console.log('>> workflowService.addOrUpdateAnnotation');
```
Tom-Szendrey, line 929: Assuming this is intentional for a short time but figured I would point out console.log

packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -916,6 +925,18 @@ export const addEdge = async (id: string, edge: WorkflowEdge) => {
...
+	const response = await API.post(`/workflows/${id}/annotation`, annotation);
+	return response.data ?? null;
+};
+
+export const removeAnnotation = async (id: string, annotationId: string) => {
+	console.log('>> workflowService.removeAnnotation');
```
Tom-Szendrey, line 935: If intention feel free to just resolve this comment I do not mind past you being aware

Tom-Szendrey (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
```diff
@@ -663,6 +664,21 @@ public void updateNodeStatus(final Workflow workflow, final Map<UUID, String> st
 		}
 	}
 
+	public void addOrUpdateAnnotation(final Workflow workflow, final WorkflowAnnotation annotation) {
+		if (workflow.getAnnotations() == null || workflow.getAnnotations().isEmpty()) {
+			workflow.setAnnotations(new HashMap<UUID, WorkflowAnnotation>());
+		}
+		if (annotation.getId() == null) {
```
dvince2, line 671: Just to confirm that hibernate isn't doing this automagically, and this won't clash with that?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
mwdchang: I am thinking the UI should determine this,

Tom-Szendrey: Fine by me, simple is good

packages/client/hmi-client/src/services/workflow.ts
mwdchang: Noted - they will be here for maybe another week. 

But please do not fall into the trap of pointing out console.log or System.out.println, there are much more important things to pay attention on.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
```diff
@@ -0,0 +1,22 @@
...
+	private String type;
```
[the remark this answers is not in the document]
mwdchang: Both "type" and "textSize" are placeholders for Jamie.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
YohannParis: Then why the `textSize` if UI should determine design?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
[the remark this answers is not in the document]
Tom-Szendrey: Im not sure how either will be used 
Is it expected that this class will get updated with the UI tweaks coming in? 
It will be good to solidify what the class is and needs relatively soon so workflows that contain annotations do not run into compatibility issues

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/WorkflowService.java
mwdchang: Correct. WorkflowAnnotation is packaged into Workflow object. The unique-id is here is to make CRUD operations easier to manage, since deletion is possible using a pure-index approach is not reliable.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
mwdchang: Jamie wants to have small/medium/large text sizes as determined by the modeller.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/workflow/WorkflowAnnotation.java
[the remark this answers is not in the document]
mwdchang: Probably something like this


image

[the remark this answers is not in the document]
Tom-Szendrey: As long as you and Jamie are on the same page as each other I am not arguing just mentioning as these fields were unintuitive to me

PR #6089, opened 2025-01-16
Rest workflow zoom

mwdchang:
### Summary
Pascale was asking for a reset workflow canvas feature.

This is a somewhat working reset zoom for the workflow. We need to "exposed" a setZoom function in tera-infinite-canvas in order to make this happen.

Note the math is not precise, because we don't have easy access to the actual dimensions of the nodes which varies, only their XY-coords, so the restoration fitting is not exact, but probably good enough.

https://github.com/user-attachments/assets/f00d12ca-db9f-4ff9-bf4c-8cf16a01a2b8

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -829,7 +829,6 @@ function relinkEdges(node: WorkflowNode<any> | null) {
 }
 
 function resetZoom() {
-	console.log('resetting zoom!!!');
 	let scaleFactor = 1;
 	let translateX = 1;
 	let translateY = 1;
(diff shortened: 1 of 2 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -18,6 +18,15 @@
...
 				<tera-toggleable-input :model-value="wf.getName()" @update:model-value="updateWorkflowName" tag="h4" />
 				<div class="button-group">
+					<Button
+						id="add-component-btn"
+						icon="pi pi-expand"
+						label="Reset Zoom"
+						@click="resetZoom"
+						size="small"
+						class="white-space-nowrap"
+					/>
```
YohannParis, line 28: The button should not be green, as it's not a primary button.
```suggestion
					<Button
						id="add-component-btn"
						icon="pi pi-expand"
						label="Reset Zoom"
						@click="resetZoom"
						size="small"
						class="white-space-nowrap"
						severity="secondary"
					/>
```

YohannParis (approved):
Neat!

PR #6090, opened 2025-01-16
value of information layout

mwdchang:
### Summary
value of information layout, based on latest tweaks from https://github.com/DARPA-ASKEM/terarium/pull/6054

Layout-wise, this is identical to that of of `Decision making` template.

image


packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -329,7 +329,37 @@ export class ValueOfInformationScenario extends BaseScenario {
 		await Promise.all(interventionPromises);
 
 		// 4. Run layout
-		wf.runDagreLayout();
+		// The schematic for decision-making is as follows
```
blanchco, line 332:
```suggestion
		// The schematic for value-of-information is as follows
```

blanchco (approved):

YohannParis (approved):

PR #6092, opened 2025-01-16
horizon scanning layout

mwdchang:
### Summary
Rectangular grid like with model-config rows and intervention-columns (if applicable)

image


blanchco (approved):
nice 👏

PR #6096, opened 2025-01-16
fix compare models not connecting from output menu selection

### LINKED ISSUE #6012: [BUG]: Clicking the '+' from a model and selecting 'compare models' does not link the model to the operator.
**Describe the bug**
Clicking the '+' from a model and selecting 'compare models' does not link the model to the operator.

https://github.com/user-attachments/assets/4da3f4af-9036-478d-9dc6-5edfc9fc1ebc

mloppie:
# Description

* When there exists more than one potential match for input ports it now just connects to the first available

Resolves #6012


mwdchang (approved):
I think this may be okay, I can't recall what we were trying to guard against though, @asylves1 do you recall??

YohannParis (approved):

PR #6098, opened 2025-01-16
Try GitHub actions arm64

YohannParis:
Try to see if avoiding emulation helps:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


YohannParis:
It is not ready yet!

PR #6104, opened 2025-01-16
5675 task improve mapping feature

YohannParis:
This pull request introduces a new `ContextMatcher` service to improve the search functionality in the `DKGController`.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ContextMatcher.java
```diff
@@ -0,0 +1,153 @@
...
+
+		int bestDistance = Integer.MAX_VALUE;
+		boolean matched = false;
+		String termLower = term.toLowerCase();
+
+		for (String keyTerm : keyTerms) {
+			int distance = levenshteinDistance(termLower, keyTerm);
+			if (distance < bestDistance) {
+				bestDistance = distance;
+				if (distance <= Math.min(3, keyTerm.length() / 2)) {
```
mwdchang, line 97: 1. Why are you splitting things by underscores? Doesn't that just make things much more complicated? 
2. You need a comment to explain what this `Math.min(3, keyTerm.length() / 2))` is doing

YohannParis: 1. I could remove the split, not necessary.
2. I have no idea, I just took a levensthein distance algo online. If you have a better one, I'm happy to update.

mwdchang: So I read it again and I think I have a better understanding. The distance logic is a bit weird that it juggles a "distance" and "bestDistance", but that aside ...

It is a threshold to what is considers to be a match, the cost of change has to be smaller than the minimum of
- 3, or
- you changed less than half of yourself

So if you have really short strings like 1-3 characters, this heuristic may be a little wonky.

YohannParis: I removed the underscore spliting.
about the math, I tried with just `S` and it returned accurate matches from the list of keys.

YohannParis: But if you are worried that the math is wrong, I am more than happy to change it.

mwdchang (approved):
code looks alright, didn't test

dvince2 (approved):

PR #6112, opened 2025-01-17
Comparison chart settings options for TA3 operator

### LINKED ISSUE #5172: [FEAT]: Comparison charts for TA3 operators
We currently support looking at simulation output one model variables per chart. It works fine except when a model is stratified, the user need to compare the strata members so this feature would support that.

- [x] There should be an additional section in the output settings of the ta3 operators called 'comparison charts'. 

- [x] The multi-select UI widget should support typing to filter the variables list in the dropdown and bulk select all the variables left after filtering. E.g. type 'I_' should filter the list to just the model variables related to the infected states e.g. I_young, I_middle, etc.
- [x] Implement rules such as: 
- a) if there are 5 or less variables selected, create one chart with one line per variable, each in a different colour (with legend)
- b) if there are 6 or more variables selected, create a panel of small multiples (one thumbnail chart per variable), all with the same y axis
- When the user clicks on the gear icon for the new chart, they can then change the default from the rule above (e.g. switch between 'All in one chart' and 'Small multiples', and if 'Small multiples' is selected, opt for unlocking the y axis or show before (grey lines) too.
Image

jryu01:
# Description

- Updated comparison chart settings UI
- Added an option for sharing same Y axis for small multiples
- Added an option for show/hide before and after run result
- Made annotation working

https://github.com/user-attachments/assets/7d5de62c-1a9d-4dd3-b384-b35743f10b0b

**Annotations**

Screenshot 2025-01-16 at 7 42 09 PM
Screenshot 2025-01-16 at 7 42 17 PM

Resolves #5172


jryu01:
There are some layout issues with comparison charts in TA3 operator drilldown components and some improvements still needs to be made. I'll handle that in separate PR.

packages/client/hmi-client/src/types/common.ts
```diff
@@ -201,6 +205,13 @@ export interface ChartSettingSensitivity extends ChartSettingBase, ChartSettingS
 	type: ChartSettingType.SENSITIVITY;
 }
 
+export interface ChartSettingComparison extends ChartSettingBase {
+	type: ChartSettingType.VARIABLE_COMPARISON;
+	smallMultiples?: boolean;
+	shareYAxis?: boolean;
+	showBeforeAfter?: boolean;
```
mwdchang, line 212: Not super clear what `showBeforeAfter` means, can we add a comment here? or is there a better name for this toggle?

jryu01: Just added a comment with the explanation.

mwdchang (approved):
Didn't look at the code closely, but overall looks nice!

YohannParis (approved):

PR #6117, opened 2025-01-17
default timepoint 100 sensitivity analysis scenario

blanchco:
# Description

As per request, default timepoint to 100 in the sensitivity chart settings when creating from the sensitivity analysis scenario template


packages/client/hmi-client/src/components/workflow/scenario-templates/sensitivity-analysis/sensitivity-analysis-scenario.ts
```diff
@@ -175,7 +175,7 @@ export class SensitivityAnalysisScenario extends BaseScenario {
 		simulateChartSettings = updateSensitivityChartSettingOption(simulateChartSettings as ChartSettingSensitivity[], {
 			selectedVariables: this.simulateSpec.ids,
 			selectedInputVariables: this.parameters.map((parameter) => parameter!.referenceId),
-			timepoint: 0
+			timepoint: 100
```
mwdchang, line 178: Do we need some additional guard when you actually apply this? 
We should always have a timepoint=0, but we may not have a timepoint=100 because someone only runs simulate for 50 steps.

blanchco: Fair, I'm looking at it a different way and I made a commit to maybe fix a race condition since this should be happening in the first place.  Locally it always selects the latest timepoint but it doesn't work on staging.  It looks like multiple `'update-state'` requests are being sent after the simulation is finished.  I'm only calling it once now

shawnyama (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/sensitivity-analysis/sensitivity-analysis-scenario.ts
mwdchang: `update-state` may indicate a sync issue, let touch base and discuss, as there may be other problems

YohannParis (approved):

PR #6122, opened 2025-01-17
Jamie interactive legend for charts

jamiewaese-uncharted:
Added interactive legend feature to forecast charts.
interactive legends - sample

It's not perfect. Ideally the unselected lines in the sample layer would fade out too. And the points and value labels in the tooltip.
Monosnap Terarium 2025-01-17 10-57-08

Also ideally we could add this feature to the other charts as well, especially the Compare Datasets quantile charts.

Unfortunately this is all over my head.


jryu01:
FYI: I'm working on this.

jryu01:
https://github.com/user-attachments/assets/77dad077-51b5-41a2-a487-784c7f0f546a

I've made changes to include followings

- Added legend selection for quantile charts
- Hide tooltips for line for selected legends

Tom-Szendrey (approved):

PR #6125, opened 2025-01-17
Fix how documentation url is provided

YohannParis:
# Description

* Small fix to allow for the freshest documentation URLs.
* Instead of being part of the workflow state, now they are linked to the operation types.
* Should help documentation update.
* I did not remove the documentationUrl from the operation type for previously created workflow.


packages/client/hmi-client/src/components/workflow/util.ts
```diff
@@ -120,3 +120,53 @@ export const getGraphDataFromDatasetCSV = (
...
+			return 'https://documentation.terarium.ai/config-and-intervention/optimize-intervention-policy/';
+		case WorkflowOperationTypes.REGRIDDING:
+			return 'https://darpa-askem.github.io/askem-beaker/contexts_climate_data_utility.html';
+		case WorkflowOperationTypes.SIMULATE_CIEMSS:
+			return 'https://documentation.terarium.ai/simulation/simulate-model/';
+		case WorkflowOperationTypes.SIMULATE_ENSEMBLE_CIEMSS:
+			return 'https://documentation.terarium.ai/simulation/simulate-ensemble/';
+		case WorkflowOperationTypes.STRATIFY_MIRA:
+			return 'https://documentation.terarium.ai/modeling/stratify-model/';
+		case WorkflowOperationTypes.SUBSET_DATA:
```
mwdchang, line 164: I think it is better if the documentation urls remain with the operators, such that it is self contained. I think it will be easier to maintain as the information is in one place/directory and not spilled over to some other component.

Something like
```
getDocumentationUrl(node) {
   return regstry.getOperator(node.operatorType).documentationURL
}
```

mloppie (approved):

packages/client/hmi-client/src/components/workflow/util.ts
YohannParis: The reason I did not do that, was because:
* that information `documentationURL` is only being used by the method `getDocumentUrl()` and being called only in the node header.
* I thought it will be heavy to load the full registry within the `utils.ts`
* I did not thought it would be a good place to put that funtion within `tera-workflow.vue`
* Another solution was to put the registry in a composables for ease of access... but that's an overkill just to display urls

But I will just try to implement the `getDocumentationUrl()` within `tera-workflow.vue` and use some system to pass it on to the drilldown.

YohannParis:
This is not worth my time!

PR #6134, opened 2025-01-20
sync filtered rows and cols with new matrix format

shawnyama:
# Description

* sync filtered rows and cols when matrix format changes
* `isReadOnly` prop shouldn't restrict showing values and showing Evaluate expressions toggle

image


YohannParis (approved):
LGTM

mwdchang (approved):
thanks

PR #6136, opened 2025-01-20
Patch model LaTeX generator to account for case where a state variable has no rate law

liunelson:
# Description

* Jamie reported an error with the [Fang2020 model](https://github.com/user-attachments/files/18481276/Fang2020.1.json)
* Issue is caused by the state variable `Total_population` having no associated rate law, leading to an expression `d X / d t = 0` which the model latex generator does not account for.
* Fix is applied by adding an attribute check (if no `atoms` attribute, don't try to substitute `Y` with `Y(t)`)


mwdchang (approved):
ok!

YohannParis (approved):

PR #6138, opened 2025-01-20
Get all project bit faster.

YohannParis:
When fetching ALL the projects, we do not need the `overviewContent` at all.
The overview is fetched when it need to be displayed.


mwdchang (approved):
Sure

PR #6142, opened 2025-01-20
Add dragndrop images to create model from equations

blanchco:
# Description

* make equations from image actually work.  We don't need to attach it to the document, Im not sure why that logic was there so I've removed it.
* Jamie has made some changes here to drag n drop an image to scan for equations
* Some python tweaks so that the job runs properly

https://github.com/user-attachments/assets/9eac28ed-0a71-4563-9401-badaeb8fb589


dvince2 (approved):
Didn't test - back end looks good.

PR #6144, opened 2025-01-20
have tera-input accept new values pasted in stratified matrix

shawnyama:
# Description

* The input fields expect a string and we used to be passing numbers to it when pasting into the matrix
* Passing strings now

https://github.com/user-attachments/assets/61a45ba3-ba95-4b39-9ee0-3568add24d84

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
@@ -171,34 +171,27 @@ let timerId = -1;
 const clipboardText = ref('');
 const updateByMatrixBulk = (matrixToUpdate: MiraMatrix, text: string) => {
 	const parseResult = dsvParse(text);
-	const updateList: { id: string; value: number }[] = [];
+	const updateList: { id: string; value: string }[] = [];
 
 	matrixToUpdate.forEach((row) => {
 		row.forEach(async (matrixEntry) => {
-			// If we have label information, use them as they may be more accurate, otherwise use indices
-			if (parseResult.hasColLabels && parseResult.hasRowLabels) {
-				const match = parseResult.entries.find(
-					(entry) => entry.rowLabel === matrixEntry.rowCriteria && entry.colLabel === matrixEntry.colCriteria
-				);
-				if (match) {
-					updateList.push({
-						id: matrixEntry.content.id,
-						value: match.value
-					});
-				}
-			} else {
-				const match = parseResult.entries.find(
-					(entry) => entry.rowIdx === matrixEntry.row && entry.colIdx === matrixEntry.col
-				);
-				if (match) {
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix.vue
```diff
@@ -171,31 +171,21 @@ let timerId = -1;
...
+				// If we have label information, use them as they may be more accurate, otherwise use indices
+				parseResult.hasColLabels && parseResult.hasRowLabels
+					? entry.rowLabel === matrixEntry.rowCriteria && entry.colLabel === matrixEntry.colCriteria
+					: entry.rowIdx === matrixEntry.row && entry.colIdx === matrixEntry.col
+			);
+			if (match) {
+				updateList.push({
+					id: matrixEntry.content.id,
+					value: match.value.toString()
+				});
```
mwdchang, line 188: How confident are we that this "toString()" fix will not break some other areas? Because while initials are strings (expressions), for parameters we are more or less assuming that they are numbers., and this might cause problems for downstream operators. 

Have we tested pasting values for parameters and see it working with TA3 and Funman operators?

shawnyama: parameters work. I'll try some other ta3 ops though it should be similar bc they all use tera-model

mwdchang: We are definitely saving them as "strings" which is technically not quite right.  


Printing out what are are saving/getting for model-config
```
        "parameterSemanticList": [
             ... snipped...
            {
                "id": "953739a0-8883-40ab-bc95-c0e6f82d502a",
                "createdOn": "2025-01-20T21:39:54.568+00:00",
                "updatedOn": "2025-01-20T21:57:53.148+00:00",
                "referenceId": "alpha_def",
                "distribution": {
                    "type": "Constant",
                    "parameters": {
                        "value": "0.34"
                    }
                },
                "default": false
            },
```

It might work for TA3 because we explicitly added a shim to turn string-like values into floats, but it doesn't exactly work 100% of the time 😓  See 

https://github.com/DARPA-ASKEM/pyciemss-service/blob/599cc5f420868f344405d3e330bd0e1cec29b7b5/service/utils/tds.py#L147

Probably `funman` should be tested. 

But there are a lot of other things unaccounted for that might go wrong if we just change types at this point. 
It may be best to keep initials strings and parameters numbers. And we ca double check if we can convert everything to strings after the eval.

shawnyama: ok I decided to convert to string when its a rate or initial matrix

blanchco (approved):
okay 👌

PR #6153, opened 2025-01-21
Observable Chart missing

asylves1:
# Description
image

Bug:
- Observable chart in the validation configuration drill down was not appearing

Testing:
- create a model configuration with a observable
- connect node to a validation configuration node
- add constraint using observable and run
- observable chart & setting should appear

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
@@ -203,20 +203,26 @@
 								/>
 								<span class="ml-4" v-else> To view parameter charts select some in the Output settings. </span>
 							</AccordionTab>
-							<AccordionTab v-if="!isEmpty(calibratedConfigObservables)" header="Observables">
+							<AccordionTab
+								v-if="!isEmpty(calibratedConfigObservables) || !isEmpty(observableCharts)"
+								header="Observables"
+							>
 								<tera-model-part
+									v-if="!isEmpty(calibratedConfigObservables)"
 									class="pl-4"
 									:part-type="PartType.OBSERVABLE"
 									:items="calibratedConfigObservables"
 									:feature-config="{ isPreview: true }"
 								/>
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
```diff
@@ -301,6 +312,27 @@
 							@remove="removeChartSetting"
 						/>
 						<hr />
+						<h5>Observable</h5>
+						<tera-chart-control
+							class="w-full"
+							:chart-config="{
+								selectedRun: 'fixme',
```
YohannParis, line 319: Could you please add a comment about this `fixme`.

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
mwdchang: The "fixme" looks like a copy-n-paste, it is sort of a legacy construct from when we were supporting multiple number of runs per output. The fixmes were there when we were transitioning out of the multi-run pattern - if there is time we should remove it.

mwdchang: Move to different task

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
```diff
@@ -920,6 +971,17 @@ function updateSelectedParameters(event) {
 	updateParameterCharts(); // Rerender when we remove/add charts since they are all contained in the same visualization
 }
 
+function updateSelectedObservables(event) {
```
jryu01, line 974: I think you can just use `updateChartSettings` method in `useChartSettings.ts` instead.

PR #6158, opened 2025-01-21
Filter out unused variables from the label expression

jryu01:
# Description

Filter out unused variables from the label expression. 
This should prevent to have very long label expression which causes `Maximum call stack size exceeded ` error from the chart when there's a lot of variables in the operator. 

Resolves #(issue)


mwdchang (approved):
Great thanks!

PR #6163, opened 2025-01-21
Calibrate ensemble chart to allow variables not mapped

Tom-Szendrey:
# Description
- _unrelated_ datasetColumn options should be restricted to current file 
- Update getAsConfiguredModel to be a more user friendly interface
- Add mapping to `getSelectedOutputEnsembleMapping` to also include not selected mapping
- Rename `getSelectedOutputEnsembleMapping` to be more accurate
- By default do not show ensemble charts in drilldown.

Looking into:
- Remove _other_ chart all together?
- Ground truth not being shown in small multiples

# Video:
https://github.com/user-attachments/assets/95e8f02a-e049-4334-aa49-8310261b8152

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
@@ -102,8 +102,7 @@ export function getChartEnsembleMapping(
 
 	// For every State Variable that has not been mapped in the ensembleMapping
 	// We will fill in here so the user can still see these if they want
-	const allStates = Object.keys(stateToModelConfigMap);
-	allStates.forEach((state) => {
+	Object.keys(stateToModelConfigMap).forEach((state) => {
 		const modelConfigurationsMap = {};
 		stateToModelConfigMap[state].forEach((id) => {
 			modelConfigurationsMap[id] = state;
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
@@ -13,6 +13,7 @@ import { EnsembleModelConfigs, ModelConfiguration } from '@/types/Types';
 import { WorkflowNode } from '@/types/workflow';
 import { getActiveOutput } from '@/components/workflow/util';
 import { CalibrateMap, setupModelInput } from '@/services/calibrate-workflow';
+import { getAsConfiguredModel } from '@/services/model-configurations';
 import {
 	CalibrateEnsembleCiemssOperationState,
 	CalibrateEnsembleMappingRow,
@@ -87,6 +88,7 @@ export function formatCalibrateModelConfigurations(
 
 export function getSelectedOutputEnsembleMapping(
 	node: WorkflowNode<CalibrateEnsembleCiemssOperationState>,
+	variableChartOptionsObject: any,
 	hasTimestampCol = true
 ) {
 	const wfOutputState = getActiveOutput(node)?.state;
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
```diff
@@ -97,6 +99,21 @@ export function getSelectedOutputEnsembleMapping(
 			datasetMapping: wfOutputState?.timestampColName ?? '',
 			modelConfigurationMappings: {}
 		});
+
+	// For every State Variable that has not been mapped in the ensembleMapping
+	// We will fill in here so the user can still see these if they want
+	const allStates = Object.keys(stateToModelConfigMap);
+	allStates.forEach((state) => {
```
YohannParis, line 106: Why the need to create a variables?
```suggestion
	Object.keys(stateToModelConfigMap).forEach((state) => {
```

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
```diff
@@ -803,7 +803,7 @@ onMounted(async () => {
 	const modelConfigurationId = props.node.inputs[0].value?.[0];
 	if (!modelConfigurationId) return;
 	const modelConfiguration = await getModelConfigurationById(modelConfigurationId);
-	configuredInputModel = await getAsConfiguredModel(modelConfiguration);
+	configuredInputModel = await getAsConfiguredModel(modelConfiguration.id as string);
```
YohannParis, line 806: No need to update, but I prefer using the proper type when available, as if we change it to something else, those typing bug will show up right away in the ts linter.

```suggestion
	configuredInputModel = await getAsConfiguredModel(modelConfiguration.id as ModelConfiguration["id"]);
```

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/calibrate-ensemble-util.ts
Tom-Szendrey: No explicit need
I have been told in the past it makes it easier to read/follow to break chunks like that up so I did

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
Tom-Szendrey: ModelConfiguration["id"] is type `string | undefined`
I see no issue using this and updating `getAsConfiguredModel` to the undefined scenario 
Instead here I am not doing this check as we know that it is defined at this point (or there are more serious issues)

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -456,15 +457,23 @@ const isRunInProgress = computed(() => Boolean(inProgressCalibrationId.value ||
...
 const lossChartSpec = ref();
 const lossValues = ref<{ [key: string]: number }[]>([]);
 const LOSS_CHART_DATA_SOURCE = 'lossData';
 // Model:
+const modelConfigIds = computed(() =>
+	props.node.inputs.filter((input) => input.type === 'modelConfigId' && input.value).map((input) => input.value?.[0])
+);
 const listModelLabels = ref<string[]>([]);
 const allModelConfigurations = ref<ModelConfiguration[]>([]);
+const stateToModelConfigMap = ref<{ [key: string]: string[] }>({});
```
mwdchang, line 476: nitpick: this probably doesn't need to be a ref, since you seemingly write once and never change it again?

Tom-Szendrey (commented):
Please do not hit merge for me until Jaehwan has had a chance to review

jryu01 (approved):

jryu01:
While testing, I found some issues


https://github.com/user-attachments/assets/0ef318ef-0dbd-44e0-b17b-a5bf2e8fac1d

1. Multiple variables of same name are selected.
2. When small multiples mode is off, charts is empty.

PR #6164, opened 2025-01-21
watch for `unitExpression` and `name` changes in `tera-model-part-entry`

shawnyama:
# Description

* new values from model enrichment need to be watched for the input fields to update

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -128,6 +128,8 @@ const emit = defineEmits(['update-item']);
 const nameText = ref(props.name);
 const unitExpression = ref(props.unitExpression);
 const descriptionText = ref(props.description);
+const showDescription = ref(!!descriptionText.value);
+
 const query = ref('');
 const results = ref<DKG[]>([]);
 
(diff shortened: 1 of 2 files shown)


mloppie (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -162,15 +164,27 @@ watch(
...
+		unitExpression.value = newUnitExpression;
+	}
+);
+
+watch(
+	() => props.name,
+	(newName) => {
+		nameText.value = newName;
+	}
+);
```
mwdchang, line 179: Would `toRef` work? it may be a bit more succinct over using a watcher then you don't really have this idea of having 2 copies of the same thing.

```
const unitExpression = toRef(props. unitExpression)
...
```

shawnyama: never mind, it works for enrichment but you can't modify it in the input field anymore. the toRef always points to the prop and isn't editable

mwdchang: Oh I see - I thought we only need to watch for changes and do not need to edit

mwdchang (approved):

PR #6172, opened 2025-01-22
Bug - Calibrate ground truth set incorrectly

asylves1:
# Description

image

Bug: 
- X-axis ground truth not being set correctly

Testing:
- Add a model and dataset 
- create a model config
- create a calibrate from the model config & dataset
-  ground truth points should be in the correct x-axis locations

Resolves #(issue)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -52,6 +52,7 @@
 									:placeholder="mappingDropdownPlaceholder"
 									v-model="knobs.timestampColName"
 									:options="datasetColumns?.map((ele) => ele.name)"
+									@change="updateTimeline()"
```
Tom-Szendrey, line 55: Any reason to not call the existing `updateState` and update it to include the `timestampColName`?

Tom-Szendrey (approved):

shawnyama (approved):

PR #6177, opened 2025-01-22
compare dataset select first dataset by default

blanchco:
# Description

* Select the first dataset by default in the compare datasets operator


shawnyama:
My branch shall take over

PR #6181, opened 2025-01-22
do not generate equations if num states + num observables > 100

mwdchang:
### Summary
In discussion with jaehwan, pascale and tom on perf and stabilization, we decided to not run genreate_model_latex task if there are too many expressions that need to be parsed in sympy, which is the bottleneck. We proxy this idea by restricting the number of equations (number of states + number of observables). 

The 250 is just a somewhat reasonable max that a modeller might go through.

image

### Testing
If you don't have a large model handy, lower the 250 threshold to 2.

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
@@ -9,17 +9,22 @@
 		@update-model-from-equation="updateModelFromEquations"
 	>
 		<template #math-editor>
-			<tera-math-editor
-				v-for="(eq, index) in equations"
-				:key="index"
-				:index="index"
-				:is-editable="isEditable"
-				:is-editing-eq="isEditing"
-				:latex-equation="eq"
-				@equation-updated="setNewEquation"
-				@delete="deleteEquation"
-				ref="equationsRef"
-			/>
+			<div v-if="exceedEquationsThreshold == false">
+				<tera-math-editor
+					v-for="(eq, index) in equations"
+					:key="index"
+					:index="index"
+					:is-editable="isEditable"
+					:is-editing-eq="isEditing"
+					:latex-equation="eq"
+					@equation-updated="setNewEquation"
+					@delete="deleteEquation"
+					ref="equationsRef"
+				/>
+			</div>
+			<div v-else>
+				<p>Number of equations exceed LaTex generator threshold.</p>
+			</div>
 		</template>
 	</tera-equation-container>
 </template>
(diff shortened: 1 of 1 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
```diff
@@ -9,17 +9,22 @@
...
+					:is-editing-eq="isEditing"
+					:latex-equation="eq"
+					@equation-updated="setNewEquation"
+					@delete="deleteEquation"
+					ref="equationsRef"
+				/>
+			</div>
+			<div v-else>
+				<p>Number of equations exceed LaTex generator threshold.</p>
+			</div>
```
YohannParis, line 27: No need for superflous elements

```suggestion
			<template v-if="exceedEquationsThreshold == false">
				<tera-math-editor
					v-for="(eq, index) in equations"
					:key="index"
					:index="index"
					:is-editable="isEditable"
					:is-editing-eq="isEditing"
					:latex-equation="eq"
					@equation-updated="setNewEquation"
					@delete="deleteEquation"
					ref="equationsRef"
				/>
			</template>
			<p v-else>Number of equations exceed LaTex generator threshold.</p>
```

packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
```diff
@@ -89,6 +95,19 @@ watch(
...
 	async (newSemantics, oldSemantics) => {
 		if (isEqual(newSemantics, oldSemantics)) return;
+
+		// If there are too many states, don't bother generating the equations - Jan 2025
+		let eqLength = 0;
+		eqLength += props.model.model.states.length;
+		if (props.model.semantics?.ode.observables) {
+			eqLength += props.model.semantics.ode.observables.length;
+		}
+		if (eqLength > 250) {
```
Tom-Szendrey, line 105: Very big nitpick but while I see it:
I prefer calling this 250 a var for readability and incase of re-use

```suggestion
    const equationThreshold = 250;
		if (eqLength > equationThreshold) {
```

YohannParis (approved):
LGTM

packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
mwdchang: Linter doesn't like this as it thinks the `v-if` does not exist

packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
mwdchang: Done, and i also lowered the threshold to 100.

PR #6187, opened 2025-01-22
horizon scanning tweaks

blanchco:
# Description

* Some name changes for the horizon scanning created model configurations
* Added a baseline configuration for the horizon scanning template (selected config w/o interventions)
* Simulation results name now like `Configuration (intervention)` so that this can be carried into downstream nodes

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/scenario-templates/horizon-scanning/horizon-scanning-scenario.ts
+++ b/packages/client/hmi-client/src/components/workflow/scenario-templates/horizon-scanning/horizon-scanning-scenario.ts
@@ -188,6 +188,34 @@ export class HorizonScanningScenario extends BaseScenario {
 			}
 		});
 
+		const modelConfig = await getModelConfigurationById(this.modelConfigSpec.id);
+
+		const baseModelConfigNode = wf.addNode(
+			ModelConfigOp,
+			{ x: 0, y: 0 },
+			{
+				size: OperatorNodeSize.medium
+			}
+		);
+
+		const baseSimulateNode = wf.addNode(
+			SimulateCiemssOp,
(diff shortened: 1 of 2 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/horizon-scanning/horizon-scanning-scenario.ts
```diff
@@ -231,8 +290,8 @@ export class HorizonScanningScenario extends BaseScenario {
 
 		// Generate Cartesian product of parameter extrema
 		const parameterExtrema = this.parameters.map((parameter) => [
-			{ id: parameter!.id, value: parameter!.low },
-			{ id: parameter!.id, value: parameter!.high }
+			{ id: parameter!.id, type: 'Low', value: parameter!.low },
+			{ id: parameter!.id, type: 'High', value: parameter!.high }
```
mwdchang, line 294: Where does this new field type` being used?

mwdchang: Ah okay I see, would it be more appropriate to call it "label" instead?

mwdchang (approved):
Looks okay, small comment about naming.

packages/client/hmi-client/src/components/workflow/scenario-templates/horizon-scanning/horizon-scanning-scenario.ts
blanchco: Ya that would work.  I've changed this and have done it for the value of information template as well

PR #6189, opened 2025-01-22
Value of information template changes

blanchco:
# Description

Number of changes to the value of information template:

- Create multiple configurations like in the Horizon Scanning for each end of the selected parameters also doing the cross product like in Horizon scanning
- Interventions are required for this template
- Added a baseline scenario like for horizon scanning (selected configuration w/o interventions)
- Select Rank criteria by default for the compare datasets node


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/value-of-information/value-of-information-scenario.ts
```diff
@@ -327,40 +482,58 @@ export class ValueOfInformationScenario extends BaseScenario {
...
+				}
+			});
+		});
+
+		// Data comparison
+		if (interventionNodes.length > 0) {
+			compareDatasetNode.x = modelNode.x + (2 + interventionNodes.length) * nodeGapHorizontal;
+		} else {
+			compareDatasetNode.x = modelNode.x + 3 * nodeGapHorizontal;
+		}
```
mwdchang, line 536: In the template you seem to indicate that intervention configuration is no longer optional, if that is the case we can simplify this data-comparison section.

blanchco: Good catch, I've simplified this now

mwdchang (approved):

PR #6190, opened 2025-01-22
Get stats on get Dataset if uploaded by Beaker

YohannParis:
Temporary solution.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/DatasetController.java
```diff
@@ -147,19 +147,39 @@ public ResponseEntity<Dataset> getDataset(
...
+					datasetStatistics.add(dataset, datasetUrl);
+				} catch (final Exception e) {
+					log.error("Error calculating statistics for dataset {}", dataset.getId(), e);
+				}
+
+				// Update and fetch updated dataset
+				datasetService.updateAsset(dataset, projectId, Schema.Permission.WRITE);
+				dataset = datasetService
+					.getAsset(id, permission)
+					.orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("dataset.not-found")));
```
dvince2, line 179: Because this is just for stats generation (which is sketchy at best here.....), do we want to throw an exception here and fail the request?

YohannParis: Good point, updated to be non-breaking

blanchco (approved):

dgauldie (approved):

PR #6199, opened 2025-01-23
switch to target-cascade instead of source-cascade

mwdchang:
### Summary
We are incorrectly cascading in this scenario when either B or C is unlinked/removed, because cascade starts "source node" when B is removed C is invalidated, and when C is removed B is invalidated.

```
      /==== B
 A ==   
      \==== C
```

This PR switches to use "target node"

### Testing
On a workflow, connect multiple model-configs to a model operator, remove or unlink one of the model-config operators, the other model-configs should remain "valid".


Tom-Szendrey (approved):
Nice

blanchco (approved):

PR #6209, opened 2025-01-23
Add description to "other values" in the model config modals

Tom-Szendrey:
# Description
- add the model's parameter's description to the config's "other values" modal.
- Do the same with model's initial value's description 

# Screenshots:

image

image

image

image


packages/client/hmi-client/src/components/model/petrinet/tera-parameter-other-value-modal.vue
```diff
@@ -160,12 +163,14 @@ const props = defineProps<{
 }>();
 
 const otherValueList = ref(props.otherValueList);
+console.log(otherValueList);
```
YohannParis, line 166:
```suggestion
```

YohannParis (approved):
LGTMDNT

PR #6214, opened 2025-01-23
Scenario template: option to auto run simulations

blanchco:
# Description

* Option to choose simulate settings and auto run simulations with the scenario templates (decision making, horizon scanning, value of information, and sensitivity analysis)
* Those simulation nodes whose intervention policies are blank will not be run as these will be invalid when making a request

https://github.com/user-attachments/assets/3ad5240a-c649-4abd-a723-27d948442f7e

Some things I've noticed and good for followup PRs: 
* The edges seem to freakout a little especially when there's many simulations running at once, refreshing the workflow will re-connect them in the proper places
* Use the `tera-timestep-calendar` component in the settings so that the user can use the date.  This isn't trivial to add unfortunately
* For the template like horizon scanning and value of information, I'm finding when there's so many simulations running at once sometimes the simulation will not attach to the node and it will have to be re-run
* Green header spam! For the completed simulations especially for the larger templates


YohannParis (approved):
Why not run _Compare Datasets_ as well?

mwdchang:
> Why not run _Compare Datasets_ as well?

@YohannParis There is an ordering here, compare-dataset would require simulations to finish first. Since we don't have a centralized orchestration system, this may end up with a lot of bespoke logic all over the place.

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
```diff
@@ -136,6 +144,84 @@ export const convertToCalibrateMap = (rows: CalibrateEnsembleMappingRow[]): Map<
...
+
+// find all forecast nodes and run simulations + update state
+export const runSimulations = async (
+	wf: workflowService.WorkflowWrapper,
+	options: ReturnType<typeof createDefaultForecastSettings>,
+	interventionPolicies: InterventionPolicy[] = []
+) => {
+	const forecastNodes = wf.getNodes().filter((node) => node.operationType === SimulateCiemssOp.name);
+	await Promise.all(
+		forecastNodes.map(async (node) => {
```
mwdchang, line 197: I am thinking you _might_ want to do this iteratively, rather than going for broke and do `Promise.all`, the reason being to not suddenly overwhelm the servers/services.

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
```diff
@@ -193,34 +193,36 @@ export const runSimulations = async (
...
+			? makeForecastRequest(node, modelConfigId, options, policy.id)
+			: Promise.resolve('');
 
-			const simulationPromise = makeForecastRequest(node, modelConfigId, options);
+		const simulationPromise = makeForecastRequest(node, modelConfigId, options);
 
-			const [baseSimulationId, simulationId] = await Promise.all([baseSimulationPromise, simulationPromise]);
+		/* eslint-disable no-await-in-loop */
+		const [baseSimulationId, simulationId] = await Promise.all([baseSimulationPromise, simulationPromise]);
+		/* eslint-enable no-await-in-loop */
```
YohannParis, line 215: You can remove those disable comment by removing the `await`.
```ts
		const [baseSimulationId, simulationId] = Promise
				.all([baseSimulationPromise, simulationPromise])
				.then(() => {
						wf.updateNode(node, { .... });
				});
```

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
blanchco: Done, I had to add a couple eslint enables/disables cause of some linting "errors" around this though

packages/client/hmi-client/src/components/workflow/scenario-templates/scenario-template-utils.ts
YohannParis: Why is it marked as resolved, are we fine with await in for loops?
If yes, then edit the eslint config.

mwdchang: Was it me? I closed the other thread about not sending a massive amount of request all at once, which is not related to this thing??

But anyway, my opinion is that we need to treat both es-linter and ts-linter as guidelines and not some sacred creed. Which means we should make allowances based on the problem context. I have no problems with a few sprinkling of eslint-ignore or ts-ignore, as long as the readers understands why that is - some times it beats blindly following linter rules and you end up with head scratching source code.

blanchco: @YohannParis I just tried this out and it doesn't seem to work as expected.  I need to `await` for all the promises to resolve for the simulations since they're all added to the payload to create a workflow.  Shall I remove the await in loops from the eslint file instead??

mwdchang: Let's leave this as is. It isn't always appropriate to have awaits inside loops, but I think it is justified here.

PR #6221, opened 2025-01-24
ensure timestamp column name is present when we run calibrate, clean …

mwdchang:
### Summary
This is an additional fix on top of https://github.com/DARPA-ASKEM/terarium/pull/6172 to have the ground-truth data setup correctly. The problem seems to be that, if you never change the timestamp mapping, this `timestampColName` is not set correctly

### Testing
In calibrate, do not touch the timestamp mapping, but set up other variable mappings.
Run calibrate, when it is finished the variable chart should have the ground truth data correctly displayed and not all bunched together.


YohannParis (approved):

Tom-Szendrey (approved):
thanks

PR #6222, opened 2025-01-24
add iteration config for kraken

mwdchang:
### Summary
Add an iteration option for running profiles, this determines how many times a task should be repeated by an instance.

So the config format is: `<profile>:<instances>:<iterations>`

e.g.
```
python kraken.py browser:2:2 forecaster:1:2
```


YohannParis (approved):
LGTMDNT

PR #6239, opened 2025-01-24
6114 bug validate configuration manual test

### LINKED ISSUE #6114: [BUG]: "Validate configuration" manual test
**Describe the bug**

The test mostly passes, except these issues:
1. Progress is stuck at 83.33% when re-run the test at Step 6
2. Trying again, no issue if the number of steps is reduced to `7` or tolerance to `0.2` (reduced compute)
3. The x-axis of the charts should be defined by the "Start time" and "End time" in the input settings; currently, the x-axis only go as far as the data points.
4. We should probably stop using the `>=` and `<=` constraints; they should be replaced with a bounding box constraint to avoid the +/-inf y values; for example, "The state variable(s) S, R should be [between] 180 and 450 from Day 30 to Day 34.

https://app.staging.terarium.ai/projects/7af5ffef-009d-406a-9270-98429df73dcb/workflow/0de19473-6e7d-4973-af78-e82e00d8f96f

mloppie:
# Description

* Added a button to cancel running validate configuration tasks as discussed. 
Screenshot 2025-01-24 at 12 52 58 PM

Resolves #6114


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/funman/FunmanController.java
```diff
@@ -141,4 +143,32 @@ public ResponseEntity<Simulation> createValidationRequest(
 
 		return ResponseEntity.ok(newSimulation);
 	}
+
+	@PutMapping("/{task-id}")
```
mwdchang, line 147: Would be more appropriate and consistent to have this route as a DELETE instead of a PUT.

```
@DeleteMapping("/{task-id}")
```

mwdchang (approved):

packages/client/hmi-client/src/services/models/funman-service.ts
```diff
@@ -60,6 +60,17 @@ export async function makeQueries(body: FunmanPostQueriesRequest, modelId: strin
 	}
 }
 
+export async function cancelQueries(taskId: string) {
+	try {
+		const resp = await API.delete(`/funman/queries/${taskId}`);
+		const output = resp.data;
+		return output;
```
YohannParis, line 67: This is how we've been doing in most service calls
```suggestion
		const { data } = await API.delete(`/funman/queries/${taskId}`);
		return data;
```

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-drilldown.vue
```diff
@@ -649,6 +651,18 @@ async function run() {
 	emit('update-state', state);
 }
 
+async function stop() {
+	const state = cloneDeep(props.node.state);
+	const response = await cancelQueries(state.inProgressId);
+	if (response) {
+		console.log(response);
+	}
```
YohannParis, line 659:
```suggestion
	await cancelQueries(state.inProgressId);
```

YohannParis (approved):

PR #6240, opened 2025-01-24
new way of scoring intervention rank

shawnyama:
# Description

* New method for calculating the score for intervention Ranking results
image


packages/client/hmi-client/src/components/workflow/ops/compare-datasets/compare-datasets-utils.ts
```diff
@@ -269,28 +270,42 @@ export function generateRankingCharts(
...
+	// For each criteria
+	allRankedCriteriaValues.forEach((criteriaValues, index) => {
+		const rankMutliplier = node.state.criteriaOfInterestCards[index].rank === RankOption.MINIMUM ? -1 : 1;
+
+		// Calculate mean and stdev for this criteria
+		const meanValue = criteriaValues.reduce((acc, val) => acc + val.score, 0) / criteriaValues.length;
+		let stdevValue = Math.sqrt(
+			criteriaValues.reduce((acc, val) => acc + (val.score - meanValue) ** 2, 0) / (criteriaValues.length - 1)
+		);
+		if (stdevValue === 0) stdevValue = 1;
```
mwdchang, line 282: We have some of these common metric calculations like stddev under `utils/stats.ts` - can you see if we can reuse them rather than making separate copies?

shawnyama: thanks

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/compare-datasets-utils.ts
```diff
@@ -269,28 +271,36 @@ export function generateRankingCharts(
...
-		}))
-		// Sort from lowest to highest value
-		.sort((a, b) => a.score - b.score)
-		// Instead of the values, we want to rank by score
-		.map((value, index) => ({ ...value, score: index + 1 }));
-
-	rankingResultsChart.value = createRankingInterventionsChart(rankingResultsScores, interventionNameColorMap);
+			configName: ''
+		})) // Sort from highest to lowest value
+		.sort((a, b) => a.score + b.score);
```
mwdchang, line 301: Your comment `Sort from highest to lowest value` does not match up with your code ... which adds the two scores together, which in general does not yield highest to lowest.

mwdchang (approved):

blanchco (approved):

PR #6243, opened 2025-01-24
fix dupe model creation on intervention policy creation

blanchco:
# Description

* Fixes an issue where a new model would be added to the resources page when an intervention policy was created
* This seems to be a caveat of the cloning process.  Looking at the backend if we add an asset to the project when it already exists it will clone it and it's dependent assets hence why the model was cloned.  Since we are already adding the intervention policy in the backend when it is created we do not need to add it in the front end and this cloning is prevented

***EDIT***
* After a chat with Dan, going to keep things consistent here and I've removed the adding of the asset to the project in the backend.  I will be adding that functionality to the templates themselves for consistencies sake

### DIFF
--- a/packages/client/hmi-client/src/services/save-asset.ts
+++ b/packages/client/hmi-client/src/services/save-asset.ts
@@ -58,10 +58,7 @@ export async function saveAs(
 		return;
 	}
 
-	// this is already done in the backend for intervention policies
-	if (assetType !== AssetType.InterventionPolicy) {
-		await useProjects().addAsset(assetType, response.id, projectId);
-	}
+	await useProjects().addAsset(assetType, response.id, projectId);
 
 	// After saving notify the user and do any necessary actions
 	logger.info(`${response.name} saved successfully in project ${useProjects().activeProject.value?.name}.`);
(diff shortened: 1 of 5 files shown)


shawnyama (approved):

packages/client/hmi-client/src/services/save-asset.ts
```diff
@@ -57,7 +57,11 @@ export async function saveAs(
 		logger.error(`Asset can't be saved since target project doesn't exist.`);
 		return;
 	}
-	await useProjects().addAsset(assetType, response.id, projectId);
+
+	// this is already done in the backend for intervention policies
+	if (assetType !== AssetType.InterventionPolicy) {
+		await useProjects().addAsset(assetType, response.id, projectId);
+	}
```
mwdchang, line 64: Talked to Cole, it seemed like we are doing an additional step for intervention-policy for no particular reason, and that is causing the duplication problem. I think rather than making a special-case here to fix another special-case which may not be needed in the first place, we should address the redundant special-case instead.

blanchco: Okay things should be consistent now, I have removed adding the asset to the project on that backend and have added the functionality to the template code for now

mwdchang (approved):

PR #6253, opened 2025-01-27
Correct timing on when we save code to state

Tom-Szendrey:
# Description
Previously we were calling `update-state` (within saveCodeToState) only after our `append-output` (within handleModelPreview) 
This changes the timing such that we call `update-state` with our executed code whenever we get the response
And we call `append-output` whenever we get our preview 

# Video
https://github.com/user-attachments/assets/12d63b23-ed42-459f-9a78-4bb1df7b3a8f


YohannParis (approved):
LGTM

blanchco (approved):

mwdchang (approved):
sure

PR #6255, opened 2025-01-27
fix intervention policy scenario chart selection

blanchco:
# Description

* fixes an issue in some scenarios where intervention policy chart select would just be blank.
Fixes:
* do not add intervention policies chart selection if the policy is blank
* added chart selection support for horzion scanning and value of information templates


YohannParis (approved):
LGTM

shawnyama (approved):

PR #6257, opened 2025-01-27
6228 bug matrix popup from a stratified model appears cutoff

jamiewaese-uncharted:
# Not a perfect solution but better than before

This is not a z-index issue. The problem is the tooltip is too close to the top of the section that it's in, so the top goes out of bounds. We would have to add a bunch of top padding to the section to see it, but that would be weird.
image

## My imperfect solution
Instead of positioning the tooltip above the trigger, I centered it over the trigger. This moves it down a fair amount. It may still get cut off if it's large, but this is a decent temporary solution.
matrix hover

A better solution would be to move the tooltip to the top level layer, like app.vue, and then trigger it there. But that feels a bit drastic just as we're about to begin the eval. 

I also made the text smaller in two places:
image
image

Oh, and I changed the field in the Stratify operator for naming the strata groups from a regular input to a Textarea that grows if the content is larger than one line.
image
image

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix-preview.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/model-configurations/tera-stratified-matrix-preview.vue
@@ -25,7 +25,7 @@
 				</template>
 			</div>
 		</section>
-		<p>Click to open</p>
+		<p class="ml-auto">Click to open</p>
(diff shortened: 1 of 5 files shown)


YohannParis (approved):
LGTM, we could fix that tooltip properly after the eval.

mwdchang (approved):

PR #6260, opened 2025-01-27
User warning message for validate config

asylves1:
# Description
Screenshot 2025-01-28 at 9 02 56 AM
Screenshot 2025-01-27 at 5 39 32 PM

Design Change:
- display message if funman process % has been stuck on a process for over 5mins

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
+++ b/packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
@@ -13,8 +13,7 @@
 			<tera-progress-spinner v-if="isLoading" :font-size="2" is-centered>
 				<span v-if="!loadingMessage">Processing... </span>
 				<span v-if="loadingProgress">{{ loadingProgress }}% </span>
-				<span v-if="loadingProgress && loadingMessage">{{ loadingMessage }}</span>
-				<span v-if="loadingProgress === 0 && loadingMessage">{{ loadingMessage }}</span>
+				<span v-if="displayMessage">{{ loadingMessage }}</span>
 			</tera-progress-spinner>
 			<div v-else-if="isBlank" class="empty-state">
 				<Vue3Lottie :animationData="EmptySeed" :height="150" :width="150" loop autoplay />
@@ -33,7 +32,7 @@ import { Vue3Lottie } from 'vue3-lottie';
 import EmptySeed from '@/assets/images/lottie-empty-seed.json';
 import TeraProgressSpinner from '../widgets/tera-progress-spinner.vue';
 
-defineProps<{
+const props = defineProps<{
 	isLoading?: boolean;
 	showSlotWhileLoading?: boolean;
 	isBlank?: boolean;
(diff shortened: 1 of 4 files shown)


packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-node.vue
```diff
@@ -87,15 +93,30 @@ const getStatus = async (runId: string) => {
 	return pollerResults;
 };
 
+function startTimer() {
```
YohannParis, line 96: This should be within `funman-operation.ts` not in the node.

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-node.vue
```diff
@@ -87,15 +93,30 @@ const getStatus = async (runId: string) => {
...
+	clearTimeout(timer.value);
+	timer.value = setTimeout(
+		() => {
+			const state = _.cloneDeep(props.node.state);
+			state.message = "Process is stuck in Funman, click 'Stop' to cancel";
```
YohannParis, line 101: The `click stop to cancel` doesn't work on the node as a message.
I would prefer the whole system to not be a bunch of props to pass a message around, but instead a boolean prop that mentionned that it is stuck. Then each the drilldown preview and the node can each convey that information the best for them.

packages/client/hmi-client/src/components/drilldown/tera-drilldown-section.vue
```diff
@@ -14,6 +14,7 @@
 				<span v-if="!loadingMessage">Processing... </span>
 				<span v-if="loadingProgress">{{ loadingProgress }}% </span>
 				<span v-if="loadingProgress && loadingMessage">{{ loadingMessage }}</span>
+				<span v-if="loadingProgress === 0 && loadingMessage">{{ loadingMessage }}</span>
```
YohannParis, line 17: I prefer using a computed here than to have two identical lines of HTML

```vue	
<span v-if="displayMessage">{{ loadingMessage }}</span>
	...
const displayMessage = computed(() => (loadingProgress || loadingProgress === 0) && loadingMessage );
```

YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/funman/tera-funman-node.vue
```diff
@@ -107,4 +127,20 @@ watch(
...
+		} else {
+			message.value = '';
+			const state = _.cloneDeep(props.node.state);
+			state.isRequestUnresponsive = false;
+			emit('update-state', state);
+			clearTimeout(timer.value);
+		}
+	},
+	{ immediate: true, deep: true }
+);
```
mwdchang, line 145: This feels more complicated than it needs to be. I think we should get by with just the poller since it pretty much provide the same information as the additional watcher that is being introduced here.

Also if you are using a setTimeout you should dismiss it in beforeUnmount or onUnmount hook.

PR #6270, opened 2025-01-28
attempt to minimize port update collisions

mwdchang:
### Summary
TLDR
- Add more update control to update nodes.inputs and nodes.outputs
- Add POST /update-position endpoint
- Tweak relinkEdges

The auto-run-simulate option in the scenario template builder introduce a situation where we can get lots multiple simulations that finish within milliseconds of each other, and they all publish to the same destination (the compare-model/compare-dataset operators). This situation, in turn, can create problem updating the ports, we have been updating node.inputs and node.outputs as a whole rather than individual elements, similar problem when we update the workflow as a whole instead of individuals.

We also extend the previous sync changes
- Move branching off the debouncer, the save debouncer is now removed
- Move update positions into its own api-end point to just update positions

### Testing
- Create a scenario template with horizon-scanning or value-of-information, with multiple configs-parameters and interventions. Ensure that auto-run-simulation is on. Create the workflow, when all the simulations are finished the "compare-dataset" operator should have the proper inputs (only one blank slot) and the edges properly connected.

- With either an older or new workflow, move the nodes arounds, refresh, and check the positions are where you left them


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
```diff
@@ -327,6 +328,51 @@ public ResponseEntity<Workflow> selectOutput(
...
+			currentUserService.get().getId(),
+			projectId
+		);
+
+		final Optional<Workflow> workflow = workflowService.getAsset(id, permission);
+		final Optional<Workflow> updated;
+
+		try {
+			workflowService.updatePositions(workflow.get(), payload);
+			updated = workflowService.updateAsset(workflow.get(), projectId, permission);
```
YohannParis, line 363: I do not see a `workflow.isEmpty()` test here before accessing the Optional Workflow.

YohannParis (approved):

blanchco (approved):
Tested quite and few times and it looks good!

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/dataservice/WorkflowController.java
mwdchang: That is a good point - we haven't been doing this kind of safety checks in WorkflowController. 

I will do a separate cleanup pass for the class to keep this PR more closely contained with the original intent.

PR #6277, opened 2025-01-28
Display concept and context fields

YohannParis:
# Description

Screenshot 2025-01-28 at 10 59 18

https://github.com/user-attachments/assets/b4708188-b990-428b-827d-798a483651a6

# Changes
* Added `modifiers` to the `Grounding.java` to make sure we can do better concept comparison with stratification.
* Created a new `tera-concept.vue` to handle editing a `Grounding` object.
* Update `tera-model-parts.vue` to leverage the `computed()` + `v-model` pattern instead of `@change` + `watch` pattern.


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -125,71 +90,63 @@ const props = defineProps<{
...
-const unitExpression = ref(props.unitExpression);
-const descriptionText = ref(props.description);
-const showDescription = ref(!!descriptionText.value);
-
-const query = ref('');
-const results = ref<DKG[]>([]);
+const nameText = computed({
+	get: () => props.name,
+	set: (newName) => emit('update-item', { key: 'name', value: newName })
+});
```
shawnyama, line 96: does this pattern work with enriching as well as editing?

shawnyama (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
YohannParis: Yes, it's the same patern, enriching is done on the back-end, not on the front-end.

mloppie (approved):

dvince2 (approved):

PR #6281, opened 2025-01-29
Added 'Normalize' option in the output settings of Simulate operator

jryu01:
# Description

In comparison chart setting, added an option to show the chart with normalized stratified model data

https://github.com/orgs/DARPA-ASKEM/projects/3/views/8?pane=issue&itemId=94761557&issue=DARPA-ASKEM%7Cterarium%7C6173

Displaying the equations used to calculate the normalized charts (non-editable in this pass) is not yet implemented with this PR. 

Screenshot 2025-01-28 at 8 19 50 PM
Screenshot 2025-01-28 at 8 20 11 PM

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/utils/math.ts
+++ b/packages/client/hmi-client/src/utils/math.ts
@@ -283,3 +283,54 @@ export function calculateUncertaintyRange(value: number, percentage: number): { 
 		max
 	};
 }
+
+/**
+ * Sums the corresponding elements of multiple arrays of numbers.
+ *
+ * @param arrays - An array of arrays of numbers.
+ * @returns A new array where each element is the sum of the corresponding elements of the input arrays.
+ * @throws Will throw an error if the input arrays are not of the same length.
+ */
+export function sumArrays(...arrays: number[][]): number[] {
+	if (arrays.length === 0) {
+		throw new Error('At least one array is required');
+	}
+
+	const length = arrays[0].length;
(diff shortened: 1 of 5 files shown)


packages/client/hmi-client/src/utils/math.ts
```diff
@@ -283,3 +283,54 @@ export function calculateUncertaintyRange(value: number, percentage: number): {
...
+	if (array1.length !== array2.length) {
+		throw new Error('Arrays must be of the same length');
+	}
+
+	const result = new Array(array1.length);
+	for (let i = 0; i < array1.length; i++) {
+		if (array2[i] === 0) {
+			throw new Error('Division by zero');
+		}
+		result[i] = array1[i] / array2[i];
```
mwdchang, line 333: Do we care about divisions by zero? Or is that dealt with somewhere else?

jryu01: hm.. I'm not handling it but we should I think. What should we expect in the case of division by zero? It's an edge case but it could happen when values of all variables are zero. I guess we would just return 0 in that case?

jryu01: I updated to return `NaN` on division by zero, and return 0 when calculating percentage with NaN

mwdchang (approved):
Didn't test, code looks okay

Tom-Szendrey (approved):

PR #6289, opened 2025-01-29
set time point to end time in sensitivity analysis template

blanchco:
# Description

* Now that we have the simulation settings in the scenario template, we can set it as the end time point indicated by the template


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/scenario-templates/sensitivity-analysis/sensitivity-analysis-scenario.ts
```diff
@@ -194,7 +194,7 @@ export class SensitivityAnalysisScenario extends BaseScenario {
 		simulateChartSettings = updateSensitivityChartSettingOption(simulateChartSettings as ChartSettingSensitivity[], {
 			selectedVariables: this.simulateSpec.ids,
 			selectedInputVariables: this.parameters.map((parameter) => parameter!.referenceId),
-			timepoint: 0
+			timepoint: this.simulateSpec.endTime
```
mwdchang, line 197: What if the workflow did not come from scenario-template?

PR #6290, opened 2025-01-29
add a base simulation for each config in scenario templates

blanchco:
# Description

* Add a base simulation node for each generated  model configuration in the horizon scanning and value of information templates

Now looks like this 😄 
Screenshot 2025-01-29 at 12 57 49 PM

Testing: 
* Create a horizon scanning and/or a value of information workflow
* see that there are base simulation nodes (no interventions) for each generated configuration


mwdchang (approved):
Yikes, maybe we should make the edges lighter, e.g. inversely proportional with respect to the size of the graph 🤔

PR #6295, opened 2025-01-29
Temporary fix for performance issue. Disabling embeddings for now.

### LINKED ISSUE #6294: Disable embedding generation

dvince2:
# Description

* Temporary fix for performance issue. Disabling embeddings for now.

Resolves #6294


mwdchang (approved):
I thought there would be more? For example when/where does the embedding for project.overviewContent gets turned into a vector?

YohannParis (approved):
love it

PR #6299, opened 2025-01-29
Ta3 operators to include rk4 as a solver option

Tom-Szendrey:
# Description
- Update states to have method type to ciemss method's enum 
- Update enum to include rk4
- Ensure when rk4 or euler are selected user can write in `step_size`

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
@@ -1,3 +1,4 @@
+import { CiemssMethodOptions } from '@/services/models/simulation-service';
 import { ChartSetting } from '@/types/common';
 import type { TimeSpan } from '@/types/Types';
 import { Operation, WorkflowOperationTypes, BaseState } from '@/types/workflow';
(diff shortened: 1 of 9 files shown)


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
```diff
@@ -13,7 +14,7 @@ export interface SimulateCiemssOperationState extends BaseState {
 	currentTimespan: TimeSpan;
 	numSamples: number;
 	solverStepSize: number;
-	method: string;
+	method: CiemssMethodOptions;
```
YohannParis, line 17: Gorgeous!

YohannParis (approved):
Great work

packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/simulate-ciemss-operation.ts
mwdchang: Should be fine, but just a reminder that `state` fields are intentionally kept very simple, because they needs to be able to be serialize into a json-string.

mwdchang (approved):

PR #6301, opened 2025-01-29
Remove the unused asset elasticsearch stuff

kbirk:
This removes the individual asset indices and all the related logic that was associated with it.

With the addition of the projects index, with it storing embeddings of each asset under the project, there is no purpose in having individual indices for each asset. If we aren't doing a KNN search, we can just use SQL queries.

The app no longer actually read from the indices. It only wrote to them.


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -706,19 +705,22 @@ public Model runAlignAMR(
...
-						final TerariumAssetEmbeddings embeddings = embeddingService.generateEmbeddings(cardText);
 
-						modelService.uploadEmbeddings(modelId, embeddings, hasWritePermission);
-						notificationInterface.sendMessage("Embeddings created");
-					} catch (final Exception e) {
-						log.warn("Unable to generate embedding vectors for model");
-					}
-				}
-			}
+			// TODO: do you want to update the model asset embeddings in the project index?
```
kbirk, line 709: These weren't used anymore, but if you want the model embeddings updated after extraction in the project index, here is where to put it.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/data/TerariumAssetService.java
```diff
@@ -48,9 +48,7 @@
 @Data
 @RequiredArgsConstructor
 @Slf4j
-public abstract class TerariumAssetServiceWithoutSearch<
-	T extends TerariumAsset, R extends PSCrudSoftDeleteRepository<T, UUID>
->
+public abstract class TerariumAssetService<T extends TerariumAsset, R extends PSCrudSoftDeleteRepository<T, UUID>>
```
kbirk, line 51: Such a nicer name...

YohannParis (approved):
Thank you very much for the proper work!

dgauldie (approved):

PR #6311, opened 2025-01-30
6292 concept search when enriching datasets

### LINKED ISSUE #6292: concept search when enriching datasets

dvince2:
# Description

* Doing a bit of refactoring to handle the fact that `id` was duplicated between groundings and a dataset columns UUID.
* Dataset columns now hook into the same code path for model enrichment
* fixed a bug where if there are no terms to search for errors were thrown

Resolves #6292


YohannParis (commented):
I don't think the term `conceptId` is the best fit, as it does not represent or link to a concept, this is the mathematical representation within an ODE. Maybe `mathId` or `symbolicId` would be better. 
@mwdchang could you weight on this please?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/dataset/DatasetColumn.java
```diff
@@ -95,6 +99,16 @@ public void updateMetadata(final JsonNode metadata) {
 		}
 	}
 
+	@TSIgnore
+	public String getConceptReference() {
+		return conceptId;
```
YohannParis, line 104: Shouldn't this be the `name` of the column?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -1086,6 +1087,26 @@ public ResponseEntity<TaskResponse> createEnrichDatasetTask(
...
+		// Update Grounding
+		final Optional<Dataset> newDataset = datasetService.getAsset(datasetId, permission);
+		if (newDataset.isEmpty()) {
+			log.warn(String.format("Dataset %s not found", datasetId));
+			throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("dataset.not-found"));
+		}
+
+		if (newDataset.get().getColumns() != null && !newDataset.get().getColumns().isEmpty()) {
+			final List<DatasetColumn> columns = newDataset.get().getColumns();
+			TaskUtilities.performDKGSearchAndSetGrounding(dkgService, columns);
```
dgauldie, line 1099: do we still do this? I thought we only search the curated concepts

dgauldie (approved):

PR #6315, opened 2025-01-30
Jamie fix confusing create model from equation workflow

jamiewaese-uncharted:
# Fix UX issue with Create model from equations
image

image

# fixed
image
image
image
image
Monosnap Monosnap 2025-01-30 12-44-35


YohannParis (approved):
Such a nice UX improvment!

PR #6320, opened 2025-01-30
5999 bug enrich metadata in model without a document does not add descriptions of the variables and parameters even if these are produced for the model card

dgauldie:
# Description

Resolves #(issue)


YohannParis (approved):
I still recommend to remove those lines: https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/TerariumApplication.java#L26-L42

mloppie (approved):

packages/gollm/common/prompts/dataset_enrichment.py
```diff
@@ -1,36 +1,38 @@
...
-
-sourceData, considerationsForUsingTheData, and additionalInformation should not be filled in for this task.
-
-Use the following dataset statistics as a reference:
-
----START DATASET---
-{dataset}
----END DATASET---
+DO_NOT_HALLUCINATE = """
+Do not hallucinate. Do not make up information. Only use the information provided in the dataset and the document.
```
dvince2, line 34: 🤣

dvince2 (approved):

packages/gollm/common/prompts/dataset_enrichment.py
YohannParis: LOL.
Could we also add the three laws of robotics.

YohannParis (approved):

PR #6322, opened 2025-01-30
Display normalize equations on the chart setting panel

### LINKED ISSUE #6173: [FEAT]: Add 'Normalize' option in the output settings of Simulate and Compare datasets
Without normalization, comparison of different demographics might be misleading e.g. if trying to understand if a particular demographic segment is more likely to die than other, without normalization, you might only be able to see that the most populous segment is more affected.

- [x] In the Simulate Output settings, after adding a '**Comparison chart**' and opening its own settings, you should see an option to **Normalize** the variables in the comparison chart (a check box). 
- [x] For transparency, display the **equations** used to calculate the normalized charts (non-editable in this pass)
- [x] The units of the charts should be in % of demographic. The normalize function is independent of wether the charts are small multiples or in one chart
- [x] The algorithm to compute the normalized data should be based on looking at the variables names. This feature is intended to work specifically with stratified models. If the user toggles it for other kind of models it may not make sense but we can live with that for now.

Image

jryu01:
# Description
Screenshot 2025-01-30 at 3 49 45 PM

Resolves #6173

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
@@ -191,11 +191,11 @@ const cancelGenerateAnnotation = () => {
 
 <style scoped>
 .chart-settings-panel-anchor {
-	position: fixed;
-	top: 7.5rem;
-	right: var(--gap-8);
+	position: absolute;
+	top: 0;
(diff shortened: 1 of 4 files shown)


mwdchang (approved):
Seems alright, didn't test.

mloppie (approved):

PR #6325, opened 2025-01-31
more selective node/edge positions update

mwdchang:
### Summary
Workflow node sizing jiggles around when workflow is first loaded onto the canvas, so we are updating positions that get overwritten later when the node finishes loading itself. This PR makes the update more selective by tracking nodes/edges that instead of always updating everything.


mwdchang:
> LGTM. There are a few places that refer to `newNodePosition` that I am not sure if we want to add them to this node set or not it is hard for me to tell at a glance

@Tom-Szendrey No I don't think that is needed unless someone changed something. newNodePosition is kind of an ephemeral thing to track where to place the node on the canvas, when the node is instantiated it will emits its position and that is tracked.

YohannParis (approved):

Issue #6067, opened 2025-01-15
Add ATE & WIS metric to 'Compare dataset' operator


pascaleproulx:
![Image](https://github.com/user-attachments/assets/74c6bcfd-9183-4e1b-b6f6-83004d9dbc0c)

The table should be similar to the one mentioned in #6091 . Like in that issue, let's move the checkbox to the output settings ('Show impact metric table') with only one option 'Average treatment effect (ATE)'.
![Image](https://github.com/user-attachments/assets/70cda40b-8c01-4545-93fe-eca665babbbd)

liunelson:
Example workflow demonstrating these two metrics:
https://app.staging.terarium.ai/projects/7af5ffef-009d-406a-9270-98429df73dcb/workflow/c5496254-1328-41f3-a414-1b8b73ac697b

![Image](https://github.com/user-attachments/assets/a48b2db2-5cb9-4b84-af8a-a8189bbaf768)

## Average Treatment Effect (ATE)

ATE measures how much impact a given intervention policy has relative to some baseline scenario. Large ATE means large impact.

```python
def average_treatment_effect(df_baseline: pd.DataFrame, df_treatment: pd.DataFrame, outcome: str) -> tuple[pd.DataFrame, pd.DataFrame]:
    
    num_samples_0 = len(df_baseline['sample_id'].unique())
    x0 = df_baseline[['timepoint_unknown', outcome]].groupby(['timepoint_unknown'])
    x0_mean = x0.mean()
    x0_mean_err = x0.std() / np.sqrt(num_samples_0)
    
    num_samples_1 = len(df_treatment['sample_id'].unique())
    x1 = df_treatment[['timepoint_unknown', outcome]].groupby(['timepoint_unknown'])
    x1_mean = x1.mean()
    x1_mean_err = x1.std() / np.sqrt(num_samples_1)

    ate = (x1_mean - x0_mean).reset_index()
    ate_err = (np.sqrt(x0_mean_err ** 2.0 + x1_mean_err ** 2.0)).reset_index()

    return (ate, ate_err)
```

Plot ATE and its 68% uncertainty interval

![Image](https://github.com/user-attachments/assets/35816048-7c5c-44c0-8515-e83859286ab8)

```python
import matplotlib.pyplot as plt

fig, axes = plt.subplots(1, 2, figsize = (12, 6))
fig.suptitle = 'Average Treatment Effect'

for outcome in ('S_state', 'I_state', 'R_state'):
[...]

PR #6330, opened 2025-01-31
wis table

shawnyama:
# Description

* WIS table calculated by following instructions found [here](https://github.com/DARPA-ASKEM/terarium/issues/6067)
* Moved calculation focused functions to `stats.ts`
 
Screenshot 2025-01-31 at 1 44 24 PM


packages/client/hmi-client/src/utils/stats.ts
```diff
@@ -61,3 +62,126 @@ export function quantile(numberList: number[], q: number) {
...
+) {
+	if (!leftQuantile) {
+		leftQuantile = variableObservations[alpha / 2];
+	}
+	if (!rightQuantile) {
+		rightQuantile = variableObservations[1 - alpha / 2];
+	}
+
+	if (checkConsistency && leftQuantile > rightQuantile) {
+		logger.error('Left quantile must be smaller than right quantile.');
```
mwdchang, line 110: What are you trying to do with by showing this error message?

shawnyama: This happens sometimes so just want to make people aware the quantiles taken from the dataset were bad. aka their dataset isn't setup for this

packages/client/hmi-client/src/utils/stats.ts
```diff
@@ -61,3 +62,126 @@ export function quantile(numberList: number[], q: number) {
...
+// specified or the quantiles need to be specified via q_left and q_right.
+// This is a dumbed down version of the original function as it's just handling the case where the left and right quantiles are not specified, etc.
+export function getIntervalScore(
+	groundTruthObservations: Record<number, number>,
+	variableObservations: Record<number, number>,
+	alpha: number,
+	percent: boolean,
+	checkConsistency: boolean,
+	leftQuantile: number | null = null,
+	rightQuantile: number | null = null
```
mwdchang, line 100: I do not recommend chaining optional parameters like this, it seems to suggest that both `leftQuantile` and `rightQuantile` are optional.  But if I only want to provide a `rightQuantile` I have to specify a leftQuantile (albeit sending in a null). 

From a modelling perspective, I think it is better to put optional fields into an object like:
```
{
  leftQuantile?: number,
  rightQuantile?: number,
 .... other stuff
}
```

mwdchang (approved):
A few small comment, I don't know the WIS calculation so I didn't verify correctness.

packages/client/hmi-client/src/utils/stats.ts
mwdchang: Then maybe we should expand the message saying that? because to me this is kind of vague and I am not sure if people knows what to do to solve this error.

blanchco (approved):

PR #6331, opened 2025-01-31
Jamie add filter to dropdown in intervention cards

jamiewaese-uncharted:
## BEFORE
It can be hard to find the parameter you want in the intervention card dropdowns if there are a lot of them.

Pascale sent me this today:
image

## AFTER
I added filter and autoFilterFocus to the dropdowns.
added filter to parameter dropdown

I also added a conditional to the filter so we only show it if there are more than five items in the menu. It felt weird to show a filter on a menu with few items.
image
image


YohannParis (approved):
LGTM

shawnyama (approved):

PR #6332, opened 2025-01-31
added modifiers to grounding. getCuratedGrounding does not overwrite …

dgauldie:
…modifiers

# Description

Resolves #(issue)


YohannParis (approved):
Nice catch!

mwdchang (approved):
I am not sure what we are planning to do with the modifier column, I suspect it will be unused because AMR's grounding modifiers are just a part of the model json-blob and is unrelated to these changes.

PR #6333, opened 2025-01-31
Compare scenario chart share intervention colouring with ranking charts

### LINKED ISSUE #6274: [TASK]: Use same colour in 'compare scenarios' than in 'Ranking of interventions'
**Describe the task**
In ranking of interventions, we use the colour to differentiate interventions. But when looking at the same data in the compare scenario option, we use different colours (and not enough of them to be all different).

The user can use the interactive legend to see which configuration correspond to which line in the compare scenario option.

Let's be consistent and only use colours to differentiate the interventions and the baseline in 'compare scenarios' too.

**Screenshots**

Image

Image

mloppie:
# Description

* Pulls the intervention color palette creation out to a util to be shared. 
* Compare scenario charts now uses color to represent interventions
* Coloring is consistent across ranking and comparison charts

Screenshot 2025-01-31 at 12 10 40 PM
Screenshot 2025-01-31 at 12 10 31 PM
Screenshot 2025-01-31 at 12 10 26 PM
Screenshot 2025-01-31 at 12 10 00 PM

Resolves #6274


packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -540,22 +540,42 @@ export function useCharts(
...
 	) => {
 		const compareDatasetCharts = computed(() => {
 			const charts: Record<string, VisualizationSpec> = {};
 			if (!isChartReadyToBuild.value) return charts;
 
-			// Make baseline black
-			const colorScheme = cloneDeep(CATEGORICAL_SCHEME);
-			colorScheme.splice(baselineIndex.value, 0, 'black');
+			const interventionNameColorMap: Record<string, string> = {};
+			const interventionNameScoreMap: Record<string, string> = {};
```
jryu01, line 553: I don't see why these variables are needed to be declared out side of `setInterventionColorAndScoreMaps` function and let function modifies them. Can we move them inside the function and let `setInterventionColorAndScoreMaps` return them?

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -540,22 +540,42 @@ export function useCharts(
...
+			setInterventionColorAndScoreMaps(
+				datasets,
+				modelConfigurations,
+				interventionPolicies,
+				interventionNameColorMap,
+				interventionNameScoreMap
+			);
+
+			const variableColorMap = datasets.value.map(({ name }) => {
+				const interventionName = name?.match(/\(([^)]+)\)/);
```
jryu01, line 563: Is it reliable to get intervention name from dataset name? What if dataset name doesn't contain intervention name?

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -540,22 +540,42 @@ export function useCharts(
 	const useCompareDatasetCharts = (
 		chartSettings: ComputedRef<ChartSetting[]>,
 		selectedPlotType: ComputedRef<PlotValue>,
-		baselineIndex: ComputedRef<number>
+		// baselineIndex: ComputedRef<number>,
```
jryu01, line 543: remove

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -540,22 +540,42 @@ export function useCharts(
...
+		datasets: Ref<Dataset[]>,
+		modelConfigurations,
+		interventionPolicies
```
jryu01, line 546: Please give it a proper type.

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1230,3 +1250,38 @@ export function useCharts(
 		useEnsembleErrorCharts
 	};
 }
+
+export function setInterventionColorAndScoreMaps(
+	datasets,
+	modelConfigurations,
+	interventionPolicies,
+	interventionNameColorMap,
+	interventionNameScoresMap
```
jryu01, line 1259: Can we type these parameters please?

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1230,3 +1250,38 @@ export function useCharts(
...
+		);
+
+		const policyName = policy?.name ?? 'no policy';
+
+		if (!modelConfiguration?.name) {
+			return;
+		}
+
+		if (!interventionNameColorMap[policyName]) {
+			interventionNameScoresMap[policyName] = [];
```
jryu01, line 1278: The value of  `interventionNameScoresMap` isn't used inside this function. Why do we need this inside of this function? Can we make this function only work with the color mapping if it doesn't do anything with the scores?

packages/client/hmi-client/src/composables/useCharts.ts
mloppie: But I'm matching them to the datasets for the order; how should I get the name from the policy while mapping the datasets?

shawnyama (approved):

packages/client/hmi-client/src/composables/useCharts.ts
jryu01: It looks like each dataset has intervention id in its metadata? and the policy name is used to create the `interventionNameColorMap`
So, for each dataset, 
```
const policy = interventionPolicies.value.find(
			({ id }) => id === dataset.metadata.simulationAttributes?.interventionPolicyId
		);
const policyName = policy?.name;
```
I think dataset.name is not always guaranteed to have intervention name in it.

mloppie: But that won't necessarily match the name in the chart and so doesn't key correctly

jryu01 (approved):

PR #6357, opened 2025-02-03
moving update-state in simulate.

Tom-Szendrey:
# Problem:
Previously we would `update-state` that the inProgressIds are all finished prior to starting our process result. 
Our process result would take a bit of time to `createLLMSummary` and `createDatasetFromSimulationResult` before `append-output` 
During the time that we have updated the state we are no long showing spinner. 
During the time we are generating this llm summary and dataset we do not yet have output to show.
This lead to an awkward blip showing the previous run result or (no run result)

# Solution:
We will now update our state as we are appending our output to ensure inProgressIds are marked as finished at the same time that we also have our new dataset to display


packages/client/hmi-client/src/components/workflow/ops/simulate-ciemss/tera-simulate-node-ciemss.vue
```diff
@@ -160,14 +166,21 @@ Provide a summary in 100 words or less.
...
+		state.errorMessage = {
+			name: 'Failed to create dataset',
+			value: '',
+			traceback: `Failed to create dataset from simulation result: ${runId}`
+		};
+		emit('update-state', state);
 		return;
 	}
 
+	// Note: this will also update state
```
mwdchang, line 178: I do not understand your comment here, this will update the output-port's state. But you are **not** updating the node's state if that is the intent here. Can you elaborate/explain a bit?

Tom-Szendrey: This comment is for the reader to understand that there is not a missing emit('update-state' ...)

I have updated to the form thanks to your explaination

```
	emit(
		'append-output',
		{
			output-state
		},
		state
```

mwdchang (approved):

PR #6364, opened 2025-02-03
Testing building MIRA with linux ARM

YohannParis:
Changes to GitHub Actions workflow:

`.github/workflows/publish-mira-taskrunner.yml`Added `workflow_dispatch`: Allow manual triggering of the workflow, and updated the ARM64 job to use `ubuntu-24.04-arm` instead of `ubuntu-24.04`.


mwdchang (approved):

.github/workflows/publish-mira-taskrunner.yml
```diff
@@ -6,6 +6,7 @@ name: Build and Publish mira-taskrunner
 
 # yamllint disable-line rule:truthy
 on:
+  workflow_dispatch:
```
dgauldie, line 9: what does this do?

YohannParis:
closed because done by #6483

PR #6366, opened 2025-02-03
sensitivity score chart improvements

blanchco:
# Description

* Show top 20 parameters in sensitivity score charts
* Score charts are now reactive to timepoint changes


YohannParis (approved):

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -1323,21 +1323,52 @@ export function createSimulateSensitivityScatter(samplingLayer: SensitivityChart
...
+		.slice(0, parametersShownCount);
+
+	let title = '';
+	if (foramttedData.length === parametersShownCount) {
+		title = `Top ${parametersShownCount} most sensitive parameters displayed.`;
+	}
+	const notShownCount = data.size - foramttedData.length;
+	if (title && notShownCount > 0) title += ` ${notShownCount} parameters not shown.`;
+
+	const globalFont = 'Figtree';
```
mwdchang, line 1340: @jryu01 : This is a bit silly, that we repeated declare a globalFont at a local scope, maybe we can mark this as cleanup todo?

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1128,13 +1128,13 @@ export function useCharts(
...
-		let rankingScores: Map<string, Map<string, number>> = new Map();
+		const rankingScores = ref<Map<string, Map<string, number>>>(new Map());
+		const timepoint = ref(0);
 
 		const fetchSensitivityData = async () => {
 			// pick the first setting's timepoint for now
-			const timepoint = chartSettings.value[0].timepoint;
 			const { result } = chartData.value as ChartData;
-			const sliceData = result.filter((d: any) => d.timepoint_id === timepoint) as any[];
+			const sliceData = result.filter((d: any) => d.timepoint_id === timepoint.value) as any[];
```
mwdchang, line 1137: Some of these `any` stuff should be able to be typed out? eg result is of type `DataArray`

blanchco: I removed some `any`s since typing was already implicitly implied here

mwdchang (approved):

packages/client/hmi-client/src/services/charts.ts
jryu01: yeah I think we should

PR #6368, opened 2025-02-03
Ts/132 optimize update to obj func

Tom-Szendrey:
# Description
- Clean up optimize types to better depict what we have rather than lists of single elements ect.
- Allow optimize to have an intervention that contains `new value and start time` that will contain a different obj function for each of `new value` and `start time`

Lots of testing to be done as this is a big and breaking change.


dvince2 (approved):
looks good; didn't test

YohannParis (approved):

PR #6397, opened 2025-02-04
Stratify model with model config

mwdchang:
### Summary
Allow stratification via a model-configuration to carry forward the values of the config, as oppose to the values of the original model.

### Testing
Start with an SIR model
- Create from it a new model-configuration with a different starting S0 value, say 123456
- Feed this model-config into stratify operator, and run a stratification that includes the S-state by say 3-groups
- Feed the stratify output into another model-config operator, verify that the S0 is 123456 and not the S0 value from the origin model

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-node.vue
@@ -34,11 +34,12 @@ watch(
 	async () => {
 		const input = props.node.inputs[0];
 		// Create a default if we dont have an output yet:
-		if (input && !props.node.outputs[0].value) {
+		// if (input && !props.node.outputs[0].value) {
+		if (input && input.value) {
 			let baseModelId = '';
 
 			if (input.type === 'modelId') {
-				baseModelId = input.value?.[0];
+				baseModelId = input.value[0];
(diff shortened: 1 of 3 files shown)


packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-node.vue
```diff
@@ -27,31 +27,56 @@ const props = defineProps<{
...
+				{
+					type: StratifyMiraOperation.outputs[0].type,
+					label: modelName ?? 'Default Model',
+					value: baseModelId,
+					state: {
+						strataGroup: cloneDeep(props.node.state.strataGroup),
+						strataCodeHistory: cloneDeep(props.node.state.strataCodeHistory)
+					}
+				},
+				state
```
YohannParis, line 78: Why do we send the state twice is some fashion?

Tom-Szendrey: We will send output's state within that inner json and the last section here (line 78) will update the general state

YohannParis: Thank you both for the explaination

shawnyama (approved):

YohannParis (approved):

PR #6403, opened 2025-02-04
input delay

asylves1:
# Description
image

Issue:
- On large Models, updating Parameters/Initials can take a long time to display input

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -65,6 +65,7 @@
 
 <script setup lang="ts">
 import { computed, ref } from 'vue';
+import { debounce } from 'lodash';
 import Button from 'primevue/button';
 import Dropdown from 'primevue/dropdown';
 import TeraConcept from '@/components/widgets/tera-concept.vue';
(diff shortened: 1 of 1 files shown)


shawnyama (approved):

blanchco (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -90,25 +91,29 @@ const props = defineProps<{
...
+
 const nameText = computed({
 	get: () => props.name,
-	set: (newName) => emit('update-item', { key: 'name', value: newName })
+	set: (newName) => debouncer('name', newName)
 });
 const unitExpression = computed({
 	get: () => props.unitExpression,
-	set: (newUnitExpression) => emit('update-item', { key: 'unitExpression', value: newUnitExpression })
+	set: (newUnitExpression) => debouncer('unitExpression', newUnitExpression)
```
mwdchang, line 104: I don't think this works, because the later debouncer will replace the earlier debouncers, so would that mean you just cancel your prior updates?

e.g `unitExpression` will override earlier `grounding`

asylves1: I didn't see that during my testing.

YohannParis (approved):

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
mwdchang: We chatted. This is not an ideal solution but suffice for the time being. We can merge and do a follow up to make this mechanic more robust.

PR #6416, opened 2025-02-05
Use getFileName to filter in compare dataset 

Tom-Szendrey:
# Description
1) Move the getFileName from the calibrate util to a the dataset util
2) utilize this getFileName to filter out columns that do not originate from the main source.

Here we see that we are displaying columns from `result_summary.csv` and `risk.json` which ideally we should not.
Adding this filter will remove these additional columns.

image


shawnyama (approved):
cool, I'll keep note of this

PR #6425, opened 2025-02-05
Layout worker

mwdchang:
### Summary
Move model-diagram layout to web worker threads

### Testing
Drilldowns with model-diagrams should be non-blocking, you don't need to wait for diagram to finish to use the drilldown.


YohannParis (approved):

blanchco (approved):
Cool! Tested and works well

PR #6431, opened 2025-02-05
Layout speedup

mwdchang:
### Summary
Use cheaper astar routing if graph size is large to speed up layout process


packages/client/hmi-client/src/services/graph.ts
```diff
@@ -299,12 +336,15 @@ export const runDagreLayout = <V, E>(graphData: IGraph<V, E>, lr: boolean = true
 		});
 
 		// manual route controller edges
-		rerouteEdges(
-			graphData.nodes,
-			graphData.edges.filter((edge: IEdge<any>) => edge.data?.isController === true)
-		);
+		const controllerEdges = graphData.edges.filter((edge: IEdge<any>) => edge.data?.isController === true);
+		if (controllerEdges.length < 8) {
```
blanchco, line 340: Curious as to why we landed chose 8 to run the fast algo!

asylves1 (approved):

packages/client/hmi-client/src/services/graph.ts
mwdchang: That seems to be roughly the number when the models get complicated.

YohannParis (approved):

PR #6432, opened 2025-02-05
Update model_comparison.py

YohannParis:
Trying to make each mira concept comparison method non blocking, and return null if they have issues.

Reasons is that if any of those bugs, which happens more often than necessary, we still can move forward, we could display a better message to the user.


mwdchang (approved):
Seems to work, a couple of comments:
- This is merging back to `main`, do you mean the release branch instead
- The errors are visible in the mira-taskrunner but are not sent back to the hmi-server, not sure if this is the intent

PR #6435, opened 2025-02-05
optimize how model-configuration communicates update events

mwdchang:
### Summary
A more efficient way to update both output and state at the same time in model-config operator. 
This should reduce the flickering due to updates happening back-to-back

Where we used to have this sequence of events (you can see it in the dev-console)
```
append-output
update-position
update-state
update-position
```

This is now
```
append-output
update-position
```

### Testing
Connect model => model-config, the model-config operator node should be relatively stable.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
@@ -96,16 +96,20 @@ watch(
 		if (configs[0]?.id) {
 			const config = configs[0];
 			state.transientModelConfig = config;
-			emit('update-state', state);
+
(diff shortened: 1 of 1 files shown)


blanchco (approved):
tested, looks good

Tom-Szendrey (approved):
Nice clean up!

YohannParis (approved):

YohannParis:
Should all operator be updated with this new way of doing `append-output`?

mwdchang:
> Should all operator be updated with this new way of doing `append-output`?

@YohannParis I think this is going to be case-by-case, but already asked @Tom-Szendrey to look at and to aduit the TA3 operators as that is where we originally saw these kind of workflow update-pattern.

PR #6441, opened 2025-02-05
no constants in ranking charts

blanchco:
# Description
This is not needed for the eval so I am branching off main

* will no longer calculate ranking score for constants

Resolves #(issue)


mwdchang (approved):
Sure

jryu01 (approved):
LGTM

PR #6443, opened 2025-02-05
[TASK]: Grounding extra concept to be updated #6436

### LINKED ISSUE #6436: Grounding extra concept to be updated
We need to remove the `context` property and only use the `modifiers` in Grounding.java.
https://github.com/DARPA-ASKEM/terarium/blob/main/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/Grounding.java#L36-L46

We do not need to care about the loss of data in the DB because Jackson will not parse it.

dvince2:
# Description

* Using the `modifiers` instead of `context`

Resolves #6436


packages/client/hmi-client/src/components/widgets/tera-concept.vue
```diff
@@ -55,7 +55,7 @@ function saveConcepts() {
 		}
 
 		if (!isEmpty(contextConcepts)) {
```
YohannParis, line 57: Variable using the word `context` shall be updated to.

```suggestion
		if (!isEmpty(modifiersConcepts)) {
```

dvince2: thanks good catch!

dgauldie:
I'm confused now. I thought the modifiers were just used by MIRA to indicate stratification.

YohannParis:
> > I'm confused now. I thought the modifiers were just used by MIRA to indicate stratification.
> 
> I'm going to defer to @YohannParis for this question

Actually, this was a mistake on our parts.

`context` [do not exist in the AMR Schema](https://github.com/DARPA-ASKEM/Model-Representations/blob/main/petrinet/petrinet_schema.json#L198-L210)

Therefore they will be gone. This was after a debugging session with @bgyori into why some MIRA concept comparison method failed on the application.

YohannParis (approved):

PR #6444, opened 2025-02-05
Fix dataset mapping in compare datasets

shawnyama:
# Description

* mapping shows expecting variable name format (not pyciemss)
* generate mapping options in `onMounted` instead of in html
* higher compatibility for mapping different variable names
* this is half the battle of solving [this](https://github.com/orgs/DARPA-ASKEM/projects/3/views/2?filterQuery=sprint%3A%40current+-status%3A%22Done%22+assignee%3A%40me&pane=issue&itemId=96464115&issue=DARPA-ASKEM%7Cterarium%7C6348), regular dataset assets aren't supported in compare datasets so I'll see what I can do in a second pass

image


packages/client/hmi-client/src/components/model/petrinet/tera-initial-table.vue
```diff
@@ -110,6 +120,12 @@ const initialList = computed<
 		.filter(({ baseInitial }) => baseInitial.toLowerCase().includes(filterText.value.toLowerCase()));
 });
 
+const hasEmptyExpressions = computed(() => ({ baseInitial }) => {
+	const semanticsForThisInitial = props.modelConfiguration.initialSemanticList.filter((s) =>
+		s.target.startsWith(`${baseInitial}_`)
+	);
+	return semanticsForThisInitial.some((s) => !s.expression);
+});
```
mwdchang, line 128: What do these petrinet component changes (tera-parameter-entry, tera-initial-entry, ... etc) have to do with compare-dataset? Are these fixing something else?

mwdchang: false alarm, branch is out of sync, I updated it

PR #6447, opened 2025-02-05
Update LaTeX-to-SymPy to support no-derivative equation substitution

liunelson:
# Description

* Paper authors often write their ODEs with quantities that they separately define
* Example, `N = S(t) + I(t) + R(t)` is to be substituted into the equation `\frac{d S(t)}{d t} = -beta * S(t) * I(t) \frac{1}{N}`
* This uses SymPy to handle the substitution

Resolves #(issue)


YohannParis (approved):

dgauldie:
we can't just merge this into the release branch, can you fix this PR so that it can be merged properly

dgauldie (changes requested):
can you fix this so that we can merge into the release branch

YohannParis:
> can you fix this so that we can merge into the release branch

I did fix it, I don't think @liunelson would know how to revert some commits like this.

dgauldie (approved):

PR #6449, opened 2025-02-05
Download equations

YohannParis:
Requested by users for the eval

https://github.com/user-attachments/assets/401cef0d-5152-4b6d-88b4-9f11ddced623


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
```diff
@@ -311,6 +326,31 @@ const notIncludedEquations = computed(() =>
...
+		})
+		.catch(() => {
+			btnCopyLabel.value = 'Failed to copy equations to clipboard';
+		})
+		.finally(() => {
+			setTimeout(() => {
+				btnCopyLabel.value = 'Copy to Clipboard';
+			}, 2000);
+		});
+};
```
mwdchang, line 351: We got a `utils/clipboard`, it doesn't have this feature I think, but you should copy the function there to keep things in one place.

YohannParis: I did it, not without some huff and puff!

mwdchang (approved):

PR #6461, opened 2025-02-06
add observables from the model and not model-configuration

YohannParis:
# Description

* We used to add the _Observables_ from the model-configuration and not the model, or only the ones from the model have Groundings.
* Do some renaming and typing to match that the options for mapping from the model are either `State` or `Observable`.
* Only map concept to the minimum of model variables or dataset fields.


mloppie (approved):

Tom-Szendrey (approved):
Thanks

Tom-Szendrey (approved):

dvince2 (approved):

PR #6477, opened 2025-02-06
Operator initialization tweaks

mwdchang:
### Summary
Small tweaks to model and dataset operators, so nodes are initialized with a single append-output instead append-output + update-state events.

### Testing
Model and dataset operators should work as normal


Tom-Szendrey (approved):
love it, thank you

YohannParis (approved):

PR #6482, opened 2025-02-06
Stop browser back navigation on overscroll

YohannParis:
# Description

When scrolling to the right or left to fat (overscroll) on the notebook editor (or any other). The browser will navigate to the page before, or after.

This is to improve UX.


packages/client/hmi-client/src/assets/css/reset.css
```diff
@@ -2,6 +2,7 @@
 html,
 body {
 	overflow-x: hidden;
+	overscroll-behavior-x: none; /* Stop browser back navigation on overscroll */
```
mwdchang, line 5: Wondering if this is a bit extreme - I for one tend to use the two finger scroll to go forward and back quite a bit cause it is convenient and I don't use a mouse.

jryu01: me too. Does it prevent me from going back with two finger swipe gesture?

YohannParis: I agree, I use that feature a lot.

YohannParis:
This is not a bug but a feature of the browser, we cannot fix it.

PR #6488, opened 2025-02-07
Comparison chart colour selection

asylves1:
# Description
image

Design Change:
 - Enable Colour Selection for Comparison Chart

Testing:
- Create a Simulate
- Add a Comparison Chart
- click on the setting cog for the chart in the side panel
- change colour for one of the variable
- 
Resolves #(issue)


packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -468,7 +468,20 @@ export function useCharts(
 					statLayerVariables.push(`${chartData.value?.pyciemssMap[v]}_mean:pre`);
 				}
 			});
-			colorScheme.push(...CATEGORICAL_SCHEME);
+			if (!setting.variableColors || setting.variableColors === undefined) {
+				setting.variableColors = {};
```
jryu01, line 472: Hm.. I think we should avoid modifying setting object here as it makes things more difficult to track what's happening in where. We should only read from setting in this file and update should be handled in `useChartSetting.ts`

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -468,7 +468,20 @@ export function useCharts(
...
+			variables.forEach((variable, index) => {
+				if (setting.variableColors?.[variable] && !isEmpty(setting.variableColors[variable])) {
+					colorScheme.push(setting.variableColors[variable]);
+				} else {
+					variableColors[variable] = CATEGORICAL_SCHEME[index % CATEGORICAL_SCHEME.length];
+					colorScheme.push(variableColors[variable]);
+				}
+			});
+			Object.assign(setting.variableColors, variableColors);
+			setting.colorScheme = [...colorScheme];
```
jryu01, line 484: Yeah I would like to avoid modifying setting object directly here, is there other work around?

packages/client/hmi-client/src/components/widgets/tera-chart-settings.vue
```diff
@@ -82,24 +82,18 @@
 		</template>
 		<template v-if="type === ChartSettingType.VARIABLE_COMPARISON">
 			<!-- TODO: Move this part to it's own component, tera-chart-settings-item-comparison or inside tera-char-settings-item -->
```
jryu01, line 84: Thanks for refactoring!. We can remove this comment now I think

packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -114,7 +130,27 @@ const props = defineProps<{
...
-const emit = defineEmits(['close', 'update-settings', 'delete-annotation', 'create-annotation']);
+type Items = { name: string; id: string; primaryColor: string };
+
+const variables = computed(() => {
+	const items: Items[] = [];
+	if (!props.activeSettings) return items;
+	const activeSettings = cloneDeep(props.activeSettings as ChartSettingComparison | null);
+	if (activeSettings?.type !== ChartSettingType.VARIABLE_COMPARISON) return [];
+	activeSettings.selectedVariables.forEach((value) => {
+		const item = { name: value, id: activeSettings.id, primaryColor: activeSettings.variableColors?.[value] ?? '' };
```
jryu01, line 141: Instead of modifying setting (`setting.variableColors = {};`), line 472 in `useChart.ts`, what if we just fall back to init `varaibleColors` with default CATEGORICAL_SCHEME values here if `variableColors` is undefined yet?

packages/client/hmi-client/src/composables/useCharts.ts
jryu01: See https://github.com/DARPA-ASKEM/terarium/pull/6488/files#r1946744670

packages/client/hmi-client/src/components/widgets/tera-chart-settings-panel.vue
```diff
@@ -186,10 +222,25 @@ const createAnnotationDebounced = _.debounce(createAnnotation, 100);
...
+	const activeSettings = cloneDeep(props.activeSettings as ChartSettingComparison | null);
+	if (activeSettings?.type === ChartSettingType.VARIABLE_COMPARISON && activeSettings.variableColors) {
+		activeSettings.variableColors[name] = event.target?.value;
+		emit('comparison-selection-change', {
+			id: activeSettings.id,
+			selectedVariables: activeSettings.selectedVariables,
+			variableColors: activeSettings.variableColors
+		});
+	}
+};
```
jryu01, line 236: Nit pick. Can we relocate this method and also `const variables` under `// === Settings for comparison method ===` section (line 166 - 181) for better readability? I just put all comparison chart related things there. It would be nice to have some consistency

jryu01 (approved):

jryu01:
Note: https://github.com/DARPA-ASKEM/terarium/pull/6547 needs to be reviewed and merged to this branch.

blanchco (approved):

PR #6494, opened 2025-02-07
heatmap scatter switch for sensitivity-charts

blanchco:
# Description

* Heatmaps for sensitivity charts
* ability to switch between sensitivity charts and heatmaps

https://github.com/user-attachments/assets/4015595a-5027-40c1-9b17-bb79b181119f

### DIFF
--- a/packages/client/hmi-client/src/services/charts.ts
+++ b/packages/client/hmi-client/src/services/charts.ts
@@ -66,6 +66,10 @@ export interface ForecastChartOptions extends BaseChartOptions {
 	yExtent?: [number, number];
 }
 
+export interface SensitivityChartOptions extends ForecastChartOptions {
+	isHeatmap?: boolean;
+}
+
 export interface ForecastChartLayer {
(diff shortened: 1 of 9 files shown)


jryu01 (approved):
Didn't test but code looks good!

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -66,6 +66,10 @@ export interface ForecastChartOptions extends BaseChartOptions {
 	yExtent?: [number, number];
 }
 
+export interface SensitivityChartOptions extends ForecastChartOptions {
+	isHeatmap?: boolean;
```
mwdchang, line 70: I see `isHeatmap: boolean" and some notion of sensitivity-chart-type `scatter | heatpmap` . Is there a particular reason we want to have two different ideas (boolean | enum) to represent the same thing? or am I missing something here?

blanchco: For consistency I'll change this to use an enum!

mwdchang (approved):

PR #6550, opened 2025-02-10
Jryu/fix error and histogram charts warnings

jryu01:
# Description

Screenshot 2025-02-06 at 4 58 57 PM

Fix above warnings from the histogram and error charts (that usually happens in calibrate drilldown)

- Fetch all data at once to prevent chart from rendering with partial data
- Calculate and provide x extent of the error charts explicitly to avoid warnings thrown with dataset with length 1. 

Resolves #(issue)


blanchco (approved):
nice!

PR #6551, opened 2025-02-10
Release/v2.31.0

dgauldie:
# Description

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/layout/controller.ts
+++ b/packages/client/hmi-client/src/layout/controller.ts
@@ -11,18 +11,23 @@ export default class LayoutController {
 
 	constructor() {
 		this.worker = new Promise((resolve, reject) => {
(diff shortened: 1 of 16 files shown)


packages/client/hmi-client/src/layout/controller.ts
```diff
@@ -11,18 +11,23 @@ export default class LayoutController {
...
+						resolve(worker);
+					} else {
+						console.error('Pyodide Init Error');
+						reject();
+					}
+				};
+			} catch (e) {
+				console.error('Worker Creation Error:', e);
+				reject(e);
+			}
```
mwdchang, line 30: FYI we moved this file into `src/web-workers/layout` to put all the web workers together. So we will need to adjust this after the fact.

PR #6552, opened 2025-02-10
fix calibration chart edge case

blanchco:
# Description

* Fixes an edge case in the parameter distribution charts where if the parameter was a constant the displayed data would be confusing to the user as to what number is shown on the chart.
* To fix this I have removed binning when the parameter is a constant so that it will only show a bar at the constant value as seen below:

BEFORE:
Screenshot 2025-02-10 at 2 14 42 PM

AFTER:
Screenshot 2025-02-10 at 2 10 35 PM

Resolves #(issue)


YohannParis (approved):
Nice work on the typing.

packages/client/hmi-client/src/services/charts.ts
```diff
@@ -7,6 +7,7 @@ import type { ChartAnnotation, FunmanInterval } from '@/types/Types';
 import { CalendarDateType, SensitivityChartType } from '@/types/common';
 import { countDigits, fixPrecisionError } from '@/utils/number';
 import { format } from 'd3';
+import { BinParams } from 'vega-lite/build/src/bin';
```
mwdchang, line 10: Slightly suspicious of the path here, please keep an eye out when we are doing CI-building that it gets picked up correctly.

mwdchang (approved):

packages/client/hmi-client/src/services/charts.ts
blanchco: I'll keep my eyes open 👀

PR #6554, opened 2025-02-10
filter parameter selection in stratify and simulate drilldown

blanchco:
# Description

* Can now search for parameters in stratify and sensitivity chart paramarter dropdowns


YohannParis (approved):
Like a boss!

PR #6555, opened 2025-02-10
Each input should have it's own debounce

asylves1:
# Description

image

Testing
- Upload a larger model and open it
- Make sure the diagram is in view
- Try changing the an Initial Name & description quickly
- One input change maybe overwritten by the debounce

Resolves #(issue)


mwdchang:
I will open a separate PR for this change.

PR #6556, opened 2025-02-10
fix for copying project name

asylves1:
# Description
image

Issue: 
- When copying a project on the home, the default input isn't correct

Testing:
- try copying a project on the home page
- default name should look like 
  - `Copy of <Name>`

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
@@ -57,6 +57,7 @@ import { RoutePath, useCurrentRoute } from '@/router/index';
 import { RouteName } from '@/router/routes';
 import { useProjects } from '@/composables/project';
 import { useProjectMenu } from '@/composables/project-menu';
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
```diff
@@ -96,6 +96,9 @@ watch(
 	async () => {
 		if (isCopyDialogVisible.value) {
 			project.value = await useProjects().activeProject.value;
+			if (project.value === null) {
+				project.value = menuProject.value;
+			}
 			const name = project.value.name ?? null;
 			projectName.value = name ? `Copy of ${name}` : 'Enter new project name';
 		}
```
YohannParis, line 104: We have `getActiveProjectName()` available.

```suggestion
		if (isCopyDialogVisible.value) {
			const name = await useProjects().getActiveProjectName();
			projectName.value = isEmpty(name) ? 'Enter new project name' : `Copy of ${name}`;
		}
```

YohannParis (approved):

PR #6559, opened 2025-02-11
just testing threads, do not merge

mwdchang:
Just testing thread tracking and handling.

Experiment on hold until further discussions on priorities - Feb 11.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TaskManager.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TaskManager.java
@@ -1,0 +1,42 @@
+package software.uncharted.terarium.hmiserver.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.Future;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.stereotype.Service;
+
+@Service
+@Slf4j
+public class TaskManager {
+
+	private ThreadPoolTaskExecutor executor;
+	private Map<UUID, Future<?>> taskTable = new HashMap();
+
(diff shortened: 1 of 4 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/TaskManager.java
```diff
@@ -0,0 +1,42 @@
...
+		executor.setMaxPoolSize(2);
+		executor.setQueueCapacity(500);
+		executor.setThreadNamePrefix("GithubLookup-");
+		executor.initialize();
+	}
+
+	public void submitTask(final UUID id, Runnable r) {
+		final Future futureTask = executor.submit(r);
+		taskTable.put(id, futureTask);
+	}
```
YohannParis, line 30: Should only the `Runnable r` be passed as an arguement and the `id` be returned by the method?

mwdchang: TLDR: No, but yes if we change the interface to be less generic

Not with accordance to my proposal on orchestration - you need to pass in an Id. The reason is kind of the chicken-and-egg type.

My proposal is to track the metadata against the a database table, and the thread itself is responsible for updating any intermediate progress as well as the final result. If the task is short lived, by the time the id is returned the thread is done and it has no where to update to. So the table entry needs to exist before we can run the thread - kind of like a registration process. 

We could change the interface such that that instead passing in a `Runnable` we pass in a `TerariumRunnable` wich some kind of TerariumRunnable.register(...) method. But the basic idea is the same, the table entry needs to exist before we start the thread.

YohannParis: That makes a lot of sense, thank you.

YohannParis:
I will close the PR to keep things clean. But the branch will be kept and move over the new organization.

PR #6560, opened 2025-02-11
Sort model equations alphabetically and wrap equations with more than 5 terms

liunelson:
# Description

* Users requested the model equations to be sorted alphabetically such that they are easier to review and compare with literature
* Ditto for the observables
* Wrapped around equations that are too long
* SymPy does not allow me to maintain a given order of terms in a multiplication or addition (so not alphabetical term order within each equation)

Screenshot 2025-02-10 at 11 42 39 PM

### DIFF
--- a/packages/mira/tasks/generate_model_latex.py
+++ b/packages/mira/tasks/generate_model_latex.py
@@ -44,40 +44,69 @@ def main():
             time = "t"
         t = sympy.Symbol(time)
 
-        # Construct Sympy equations
-        odesys = []
+        # Add "(t)" to all the state variables as time-dependent functions
         for var, terms in odeterms.items():
+            for i, term in enumerate(terms):
+                if hasattr(term, 'atoms'):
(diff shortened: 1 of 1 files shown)


YohannParis:
After talking with @pascaleproulx and @dgauldie, this is a nice to have feature.
But the best solution would be to have:
- a button on the top of the equations field to allow you to sort them alphabetically.
     Do Observables be listed at the end, or even better, should the non time derived equations be listed after the one derived?
- a button to wrap/unwrap equations within the field

liunelson:
Based on what David said, we could merge this PR and then consider some toggles afterwards.

I don't think a toggle is necessary since the original sort order is essentially arbitrary (as generated by the MIRA AMR-to-MMT algo) and an really long, unwrapped equation is always hard to read.

Yep, the observable expressions are listed at the bottom (sorted alphabetically amongst themselves). No wrap-around for them though since SymPy doesn't allow that easily.

PR #6567, opened 2025-02-11
stratify use input temporal context in output

blanchco:
# Description

* A request for when we attach a model configuration node to a stratify operator we want that model configuration's temporal context to be carried over to the output's temporal context.  Since this temporal context is not apart of the AMR, we need to get the temporal context of the input model configuration, and update the output default configuration of the stratify model.

So...
1. Input Model config -> stratify/edit
2. creates new model and a default configuration
3. get first configuration
4. update first configuration with input model config's temporal context


YohannParis (approved):

packages/client/hmi-client/src/components/workflow/ops/stratify-mira/tera-stratify-node.vue
```diff
@@ -61,6 +66,16 @@ watch(
 			if (!baseModelId) return;
 
 			const model = await getModel(baseModelId);
+			// we want to update the new model's configuration with the old one's temporal context
+			if (modelConfiguration?.temporalContext) {
+				getModelConfigurationsForModel(baseModelId).then((modelConfigurations) => {
```
mwdchang, line 71: This seems weird, I don't understand why this is a separate block in stead of under the `else if (input.type === 'modelConfigId')` block.

Also I don't think this is helping matters because we need to think about the _output_ and not the _input_ 

Let't may be have chat to clear things up

blanchco: After our chat, closing this PR in favour of https://github.com/DARPA-ASKEM/terarium/pull/6572

blanchco:
closing in favour for https://github.com/DARPA-ASKEM/terarium/pull/6572

PR #6572, opened 2025-02-11
stratify use input temporal context in output

blanchco:
# Description

* as per a suggestion by @mwdchang it might be better to send in an optional model config id and try to populate the temporal data in the default configuration in the backend

* A request for when we attach a model configuration node to a stratify operator we want that model configuration's temporal context to be carried over to the output's temporal context.  Since this temporal context is not apart of the AMR, we need to get the temporal context of the input model configuration and use the temporal context in the created default configuration

* Added an optional model config id to the create model and create new from old model endpoints to take metadata from and populate

### DIFF
--- a/packages/client/hmi-client/src/services/model.ts
+++ b/packages/client/hmi-client/src/services/model.ts
@@ -10,17 +10,25 @@ import { fileToJson } from '@/utils/file';
 import { Ref } from 'vue';
 import { DateOptions } from './charts';
 
-export async function createModel(model: Model): Promise<Model | null> {
+export async function createModel(
+	model: Model,
+	modelConfigurationId?: ModelConfiguration['id']
+): Promise<Model | null> {
 	delete model.id;
-	const response = await API.post(`/models`, model);
+	const response = await API.post(`/models`, { model, modelConfigurationId });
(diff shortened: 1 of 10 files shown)


packages/client/hmi-client/src/services/model.ts
```diff
@@ -10,17 +10,25 @@ import { fileToJson } from '@/utils/file';
...
 import { DateOptions } from './charts';
 
-export async function createModel(model: Model): Promise<Model | null> {
+export async function createModel(
+	model: Model,
+	modelConfigurationId?: ModelConfiguration['id']
+): Promise<Model | null> {
 	delete model.id;
-	const response = await API.post(`/models`, model);
+	const response = await API.post(`/models`, { model, modelConfigurationId });
```
mwdchang, line 18: It might make more coherent sense to send `modelConfigurationId` as a query parameter than a part of the request body. eg

```
POST /models?modelConfigurationId=xyzxyz
```

It is a little weird to mix, but I think it conveys a better idea of what the route is about, that is, posting a new model constrained by an optional model-configuration.

mwdchang (approved):

YohannParis (approved):

PR #6577, opened 2025-02-12
5964 task add date formatting to forecast chart tooltip if relevant

### LINKED ISSUE #5964: [TASK]: Add date formatting to forecast chart tooltip if relevant
The tooltip in the Forecast Chart (the line charts in simulate) has been replaced with a pretty line that follows your mouse and draws value labels next to each line.
Image

Is good. However when a reference date has been set, the tooltip label only shows the raw time step number, not a date. Somehow we're formatting this number correctly when we draw the X axis. I couldn't figure out how to apply that number formatting to the time step label.

Goal: Add date formatting to the tooltip label (if relevant).

ALSO

Pascale noticed that if you add an annotation using the AI that includes a date in the prompt (e.g., "Draw a line at January 6") the AI doesn't know what time step that is. Mentioning this here because it's possible the solution to this task isn't just front-end number formatting for the label, but actually adding the formatted dates to the data somehow, and using that to fix both these issues.

mloppie:
# Description

* reuses the axis date format expression 
* transforms and adds a column to the statisticsLayer data when possible
* conditionally references the transformed date data when available

Resolves #5964 
Screenshot 2025-02-12 at 11 27 07 AM


jryu01:
LGTM

jryu01 (approved):

YohannParis (approved):

Issue #6409, opened 2025-02-04
[BUG]: Model inputs delay cleanup


asylves1:
**Describe the bug**
There should be a better why to handle the input delay then using debounce. The previous debounce can be overwritten by the newest debounce counting on the size of the debounce delay. This change was originally done as a quick fix.

**To Reproduce**
1. Upload a larger model and open it
2. Make sure the diagram is in view
3. Try changing the an Initial Name & description quickly
4. One input change maybe overwritten by the debounce

**Expected behavior**
Both input changes should appear

**Screenshots**
![Image](https://github.com/user-attachments/assets/4b65a9f9-b70d-40af-9aae-2a3325862613)

**Additional context**
This was a rushed change for the eval.
The change is in the file `tera-model-part-entry.vue`

PR #6578, opened 2025-02-12
Use different debouncer for each field so they do not override each other

mwdchang:
Closes: https://github.com/DARPA-ASKEM/terarium/issues/6409

### Summary
Have separate debouncers for each field so the queued up emit-events from field-A are not overwritten by event from fieldB.

### Testing
In model page, try to change variable's description, unit, name, ... etc. Should all work as expected and not lose previous edits.


jryu01 (approved):

YohannParis (approved):

YohannParis:
I still beleive that the debouncing should be done on the code receive the emit, not the emit itself.

PR #6581, opened 2025-02-13
Operator input/output label reactive

YohannParis:


packages/client/hmi-client/src/services/workflow.ts
```diff
@@ -649,14 +644,20 @@ export function getOutputLabel(outputs: WorkflowOutput<any>[], id: string) {
...
+
+		// default use single output type
+	} else {
+		assetId = selectedOutput.value?.[0];
 	}
 
-	// default use single output type
-	return useProjects().getAssetName(selectedOutput.value?.[0]) || selectedOutput.label;
+	const label = useProjects().getAssetName(assetId).value;
+	return computed(() => (!isEmpty(label) ? label : (selectedOutput.label ?? '')));
```
mwdchang, line 660: There seems to be a couple of extra layers of unnecessary computed going. on here, what is the problem here? The PR description is blank so I am not even sure what to test for ??

YohannParis:
The goal was to solve the issue that input/output are not reactive to change. Maybe this old PR is not worth reopening anymore.

PR #6584, opened 2025-02-14
show rank charts in compare datasets node

blanchco:
# Description

* Show ranking charts in the compare datasets node

Took the opportunity to do _some_ refactoring here:
* state save on unmount instead of every knob change
* added types where needed
* moving the building the ranking chart to the `useCharts.ts` as done with the impact charts for consistency
* `Run` no longer needed on ranking charts, they will calculate automatically when changes are made the the criteria cards


packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-node.vue
```diff
@@ -1,12 +1,34 @@
...
+			v-if="node.state.selectedCompareOption !== CompareValue.RANK"
 			:node="node"
 			:prepared-charts="comparisonCharts"
 			:is-loading="isFetchingDatasets"
 			:chart-settings="selectedVariableSettings"
 			:placeholder="'Attach datasets/simulation outputs to compare'"
 		/>
+		<template v-else>
+			<tera-progress-spinner v-if="isFetchingDatasets" :font-size="2" is-centered style="height: 100%">
+				<div>Processing...</div>
```
YohannParis, line 13: Is the `<div>` necessary.

YohannParis (approved):

mwdchang (approved):
Didn't really test, code seems okay

packages/client/hmi-client/src/components/workflow/ops/compare-datasets/tera-compare-datasets-node.vue
blanchco: It is not.  I have removed this div!

PR #6587, opened 2025-02-14
Add save for reuse to calibrate ensemble

Tom-Szendrey:
# Description
- Add save for reuse to calibrate ensemble.

image

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
@@ -196,12 +196,23 @@
 		<section :tabName="DrilldownTabs.Notebook">
 			<h4>Notebook</h4>
 		</section>
+		<!-- Output Section -->
 		<template #preview>
 			<tera-drilldown-section
 				:is-loading="isRunInProgress"
 				:show-slot-while-loading="true"
 				:loading-progress="node.state.currentProgress"
 			>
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
```diff
@@ -459,11 +472,19 @@ const isRunInProgress = computed(() => Boolean(inProgressCalibrationId.value ||
...
+const csvAsset = shallowRef<CsvAsset | undefined>(undefined);
 const datasetColumnNames = computed(
 	() =>
 		dataset.value?.columns?.filter((col) => col.fileName === currentDatasetFileName.value).map((col) => col.name) ??
 		([] as string[])
 );
+const outputDatasetId = computed(() => {
+	if (!selectedOutputId.value) return '';
+	const output = props.node.outputs.find((o) => o.id === selectedOutputId.value);
+	return output?.value?.[0] ?? '';
```
mwdchang, line 485: Can use null assertion here instead of option chaining, the latter is expensive after transpile, the former just disappears

```
return output!.value?.[0] ?? '';
```

mwdchang (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ensemble-ciemss/tera-calibrate-ensemble-ciemss-drilldown.vue
Tom-Szendrey: Thanks good catch!
I have updated this

PR #6597, opened 2025-02-18
show dynamic intervention description in drilldown

blanchco:
# Description

* Dynamic interventions are now displayed in the description
* Fixed an issue on the node where the console would produce errors if no interventions were defined on the node
* Added a better way of iterating over the interventions.  I am using a `set` to determine what interventions are defined in the drilldown and iterating over that in the description panel


packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
```diff
@@ -292,6 +295,20 @@ interface BasicKnobs {
 	transientInterventionPolicy: InterventionPolicy;
 }
 
+const displayedInterventions = computed(() => {
+	// we want a set of the appliedTo interventions in the drilldown
+	const selectedInterventionSet: Set<string> = new Set();
```
mwdchang, line 300: How is this better than the `Object.keys(...)` implementation? I am confused to what this is trying to do.

mwdchang: Chatted, we should have additional documentation and comments here, because it is using one list eg `[a, b, c, d, e]` to iterate over another list `[1, 2, 3]` which is a bit of an unusual pattern.

blanchco: Hopefully made it clearer with the latest commit. 

We want to display all selected intervention descriptions in the side panel, however we only chart `static` interventions at this moment. So this is a superset of `preparedCharts`.  If the variable in the superset exists inside prepared charts, we display the chart as well as the description, otherwise we just display the description.

mwdchang (approved):

PR #6608, opened 2025-02-20
Refactor: Simplify notifications in model config and intervention policy

mloppie:
# Description

* less code duplication/more reuse.
* common actions moved to composable.

This is the first of several PRs on notifications


packages/client/hmi-client/src/composables/useClientEvent.ts
```diff
@@ -14,3 +16,13 @@ export function useClientEvent(
 		await Promise.all(eventTypes.map((type) => unsubscribe(type, handlerFn)));
 	});
 }
+
+// accepts a string[] to update with in progress task ids
+export function createClientEventHandler(taskIds: string[]) {
```
mwdchang, line 21: I am not sure if we are over-thinking or trying to over-refactor here. But I think the whole idea is to be able to have customized event handlers based on context. So probably at the least we should rename this `createClientEventHandler` to something more appropriate?

Also I don't know what the bottom assignment:
```
		if ([TaskStatus.Success, TaskStatus.Cancelled, TaskStatus.Failed].includes(event.data.status)) {
			taskIds = taskIds.filter((id) => id !== event.data.id);
		}
```
is trying to do? You are changing the functional params `taskIds` which will not work since you are assigning a new reference.

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
```diff
@@ -39,19 +38,12 @@ const props = defineProps<{
...
-
-useClientEvent(ClientEventType.TaskGollmConfigureModelFromDocument, configModelEventHandler);
-useClientEvent(ClientEventType.TaskGollmConfigureModelFromDataset, configModelEventHandler);
+useClientEvent(
+	[ClientEventType.TaskGollmConfigureModelFromDocument, ClientEventType.TaskGollmConfigureModelFromDataset],
+	createClientEventHandler(props.node.state.modelConfigTaskIds)
+);
 
-const isLoading = computed(() => modelConfigTaskIds.value.length > 0);
+const isLoading = computed(() => props.node.state.modelConfigTaskIds.length > 0);
```
YohannParis, line 46: How is this passing the TS linter without optional chaining? How `props.node.state.modelConfigTaskIds` not `null` or `undefined` when the component is mounted? The `lenght` property should not work.

packages/client/hmi-client/src/composables/useClientEvent.ts
YohannParis, line 21: what is the advantage to pass a list `taskIds` here instead of a single `taskId`?

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
```diff
@@ -58,44 +58,36 @@
 import { computed, ref, watch } from 'vue';
 import { WorkflowNode, WorkflowPortStatus } from '@/types/workflow';
 import Button from 'primevue/button';
-import _, { cloneDeep, groupBy } from 'lodash';
+import _ from 'lodash';
```
YohannParis, line 61: I prefer we do not use the `_` and just select which method to use.

packages/client/hmi-client/src/composables/useClientEvent.ts
mloppie: I'm trying to break up what I have in progress. This is a pattern that was already used multiple times, and that I am using again in model-operation-document in another branch so felt strongly that it needs to be shared. `useClientEvent` takes a function as a param and model comparison supplies it's own. Making a generic fit all solution event handler factory is out of scope.

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-config-node.vue
mloppie: It's set to empty array in the state init

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
mloppie: sure, I just didn't like that it was doing both. Happy to switch the other way

packages/client/hmi-client/src/components/navbar/tera-notification-panel.vue
```diff
@@ -88,22 +97,21 @@ const getTitleText = (item: NotificationItem) => `${item.typeDisplayName} from`;
...
-const isQueued = (item: NotificationItem) => item.status === ProgressState.Queued;
+const showProgress = (item: NotificationItem) =>
+	item.status === ProgressState.Running ||
+	item.status === ProgressState.Cancelling ||
+	item.status === ProgressState.Queued;
 const isFailed = (item: NotificationItem) =>
 	item.status === ProgressState.Failed || item.status === ProgressState.Error;
-const isRunning = (item: NotificationItem) => item.status === ProgressState.Running;
-const isCancelling = (item: NotificationItem) => item.status === ProgressState.Cancelling;
-const isCancelled = (item: NotificationItem) => item.status === ProgressState.Cancelled;
```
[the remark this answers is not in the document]
mloppie: Fair enough. I don't like having multiple names for the same state/value. It makes the code harder to search and trace and learn, but don't feel strongly enough about it to defend if you want it reverted

packages/client/hmi-client/src/composables/useClientEvent.ts
jryu01: I agree with Dan that we should at least rename the function to a specific name for what it does.   `createClientEventHandler` implies it's a generic handler creation function but it looks like what you are actually doing is for creating a specific handler for some of the `TaskResponse` type events.

mloppie: Sure thing, and I don't know the context for all the different event types besides what I've been working on. Renamed.

packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-node.vue
```diff
@@ -58,34 +58,26 @@
...
 import TeraProgressSpinner from '@/components/widgets/tera-progress-spinner.vue';
 import { InterventionPolicyState } from './intervention-policy-operation';
 
 const emit = defineEmits(['open-drilldown', 'update-state']);
 const props = defineProps<{
 	node: WorkflowNode<InterventionPolicyState>;
 }>();
+useClientEvent(
+	[ClientEventType.TaskGollmInterventionsFromDocument, ClientEventType.TaskGollmInterventionsFromDataset],
+	createInProgressClientEventHandler(props.node.state, 'taskIds')
```
mwdchang, line 77: You may want to consider passing in `props.node`, rather than `props.node.state` there are some cases where user actions or upstream changes that will cause node.state to be reassigned/refreshed, then this function would be pointing to a dangling address. We can probably remediate this state situation to do smarter assignments/merges, but I think that will take a bit of thinking and work ... 


So:

```
createInProgressClientEventHandler( node: WorkflowNode<any>, taskIdKey: string) { ... }
```

instead of

```
createInProgressClientEventHandler(state, taskIdKey: string) { ... }
```

mloppie: Good to know, thanks

mwdchang (approved):

PR #6612, opened 2025-02-20
Update simulate ensemble to use new charts.

Tom-Szendrey:
# Description:
- Update simulate ensemble to use new charts. (update state, update which components we are utilizing to create chart data and chart specs)
- update simulate ensemble to use new node preview.
- remove now unused and depreciated charts.
- remove now unused and depreciated `getRunResultsCiemss`
- Rename `showSpinner` to `isLoading` to try to keep consistency between nodes.

# Testing
- Should be able to run a simulate ensemble with the same process as before. You will need a new node as the state has been updated.
- Should see a default chart for every ensemble variable mapped on your first run (or any run while the output settings are blank)
- Should be able to utilize any chart features you are used to using in any of the other drilldowns.

# follow up pr 
1) remove the `other` label in the chart as there is not groundTruthLayer or whatever this is supposed to be the label for,
2) Update from `uuid/state_var` to `state_var`


packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/simulate-ensemble-ciemss-operation.ts
```diff
@@ -28,7 +29,7 @@ export const normalValues = Object.freeze({
 });
 
 export interface SimulateEnsembleCiemssOperationState extends BaseState {
-	chartConfigs: string[][];
+	chartSettings: ChartSetting[] | null;
```
mwdchang, line 32: @jryu01 what is the difference between "chartSettings" being null versus being an empty array?

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
```diff
@@ -1,53 +1,82 @@
...
 const selectedRunId = ref<string>('');
+const areInputsFilled = computed(() => props.node.inputs.filter((ele) => ele.value).length > 1);
+// -------------- Charts && chart settings ----------------
+const chartSize = ref<{ width: number; height: number }>({ width: 180, height: 120 });
+const allModelConfigurations = ref<ModelConfiguration[]>([]);
+const preparedChartInputs = usePreparedChartInputs(props, runResults, runResultsSummary, pyciemssMap);
+const stateToModelConfigMap = ref<{ [key: string]: string[] }>({});
+const selectedOutputMapping = computed(() => getChartEnsembleMapping(props.node, stateToModelConfigMap.value));
+
+const { useEnsembleVariableCharts } = useCharts(
```
mwdchang, line 57: I don't see any charts in the ensemble workflow node after running ensemble-simulate a couple of times. Can you describe what is happening here?

I configured this setup in the drilldown:

image

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/simulate-ensemble-util.ts
```diff
@@ -30,3 +42,62 @@ export function formatSimulateModelConfigurations(
...
+
+export function getChartEnsembleMapping(
+	node: WorkflowNode<SimulateEnsembleCiemssOperationState>,
+	stateToModelConfigMap: { [key: string]: string[] }
+) {
+	const wfOutputState = getActiveOutput(node)?.state;
+	const mapping = _.clone(wfOutputState?.mapping ?? []);
+
+	// For every State Variable that has not been mapped in the ensembleMapping
+	// We will fill in here so the user can still see these if they want
```
mwdchang, line 54: The comment is a bit confusing - what are you trying to say?

Tom-Szendrey: This is trying to say that we will add individual model options for the user to select on top of the ensemble mapping that was provided for the run.

The following picture the user has only mapped `Sus` but the user should still be allowed to select charts for `Sus`, `S`, `I` and `R` 
image

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
Tom-Szendrey: There is an issue around updating the `hideInNode` value for some charts. This setting can be seen if you click on the settings cog beside `Sus` or `I` in your output settings.

I believe @jryu01 is looking into the issue around updating this 

By default this value is set to true to not flood the node with charts the user has not explicitly asked to be shown.

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/simulate-ensemble-ciemss-operation.ts
jryu01: When it's null, we know the chartSettings hasn't been initialized at all (e.g. Right after the simulate/calibrate run). When it's empty, it tells that it might have been initialized already but there's no settings item currently. The settings can be empty when user deselects all existing charts.

mwdchang: Do we have a real need to distinguish between the two different states or can they be  consolidated? 
One thing I would like to reduce is this type of disjunctive types because it feels like we are doing a lot of unnecessary null-checks.

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
Tom-Szendrey: `I think hideInNode not working properly in calibrate ensemble is a separate issue.` 
I have tracked this down and [corrected this here]([7e78f51](https://github.com/DARPA-ASKEM/terarium/pull/6612/commits/7e78f51a8699fdb8d19e72136cc01f614baa74ad)) 

`I'm not sure if tera-node-preview handles that.`
I will check and correct as needed thank you

jryu01: > I have tracked this down and corrected this [here](https://github.com/DARPA-ASKEM/terarium/pull/6612/commits/7e78f51a8699fdb8d19e72136cc01f614baa74ad)

Nice thanks!

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/simulate-ensemble-ciemss-operation.ts
jryu01: If we are fine with the behaviour we set the default charts on every run whenever there are no charts (even though they are cleared by user),  I think we can stick to `[]`

mwdchang: alright, anyway, just some food for thought

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/tera-simulate-ensemble-node-ciemss.vue
Tom-Szendrey: Should be able to correctly update the `hideInNode` value now.
It should correctly show chart in node (when this setting is false)
image

packages/client/hmi-client/src/components/workflow/ops/simulate-ensemble-ciemss/simulate-ensemble-ciemss-operation.ts
Tom-Szendrey: I would be happy to update this thought process such that we do not have `chartSettings[] | null` in all of our operator's states. I suspect that treating empty as a good enough reason to populate it should be fine from a user's perspective.
I think this should be done in a new PR as this one is already covering quite a lot of ground

mwdchang (approved):
Per discussion, there are a few small cleanups for both ensemble operators to be resolved in a differentPR.

YohannParis (approved):

Issue #6445, opened 2025-02-05
[FEAT]: The transition table in the Model page should not list the AMR inputs/outputs, should be subject/outcome/controller(s)


liunelson:
**Is your feature request related to a problem? Please describe.**

The users are seeing subject/outcome/controller(s) when editing the model in MIRA. However, they see these inputs/outputs = terminology they have not seen before. Properly accounting of the controller(s) is important for correct stratification.

**Describe the solution you'd like**

Compute the subject/outcome/controller(s) for the transitions and show those instead.

<img width="1393" alt="Image" src="https://github.com/user-attachments/assets/e2f13dc1-cb48-4985-a025-10aa8db7326e" />

PR #6613, opened 2025-02-21
transition table: add controllers, remove concepts

mwdchang:
Closes: https://github.com/DARPA-ASKEM/terarium/issues/6445

### Summary
- Remove concept search/entries as transition concepts do not exist
- Split **input/output** => **subject/output/controllers** per Nelson request
- Start to split apart "createPartsList(...)", this function seemingly treats state/transition/parameter as if they are the same object, which they are not. It makes it difficult to modify/tweak things. 
- Minor clean up in tera-model-part-entry for unused props

### Testing
Where models are displayed:
- Input/Output becomes Subject/Outcome/Controllers in transition table
- No more concepts in the transition table

image

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -7,9 +7,10 @@
 				<tera-input-text v-else placeholder="Add a name" v-model="nameText" />
 			</span>
 			<span class="unit" :class="{ time: isTimePart }">
-				<template v-if="input || output">
-					<span><label>Input:</label> {{ input }}</span>
-					<span><label>Output:</label> {{ output }}</span>
+				<template v-if="subject || outcome">
+					<span><label>Subject:</label> {{ subject }}</span>
+					<span><label>Outcome:</label> {{ outcome }}</span>
+					<span v-if="controllers"><label>Controllers:</label> {{ controllers }}</span>
 				</template>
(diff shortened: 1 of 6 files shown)


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -7,9 +7,10 @@
 				<tera-input-text v-else placeholder="Add a name" v-model="nameText" />
 			</span>
 			<span class="unit" :class="{ time: isTimePart }">
-				<template v-if="input || output">
-					<span><label>Input:</label> {{ input }}</span>
-					<span><label>Output:</label> {{ output }}</span>
+				<template v-if="subject || outcome">
+					<span><label>Subject:</label> {{ subject }}</span>
```
Tom-Szendrey, line 11: Do we need to have {{ subject ?? "" }} to handle the case where we have an outcome therefore the above if passes, but we do not have a subject?

Same goes for the reverse scenario with having a subject but not having a controller

Tom-Szendrey (approved):
Thank you!!

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
mwdchang: If you meant the formatting I can take a run at improving it. 
If you meant the predicate check that is fine as is.

mwdchang: There
image

PR #6615, opened 2025-02-21
model sanity check: check for closed system

mwdchang:
### Summary
Add a check to see if the system is closed (preserves constant population). This is not an error but something to be called out for, as it can lead to overflow/underflow in simulations. 

This is potentially part of the suite of checks to be deployed into the operators.

image

### Testing
Can test with this model 
[no-mass.json](https://github.com/user-attachments/files/18907661/no-mass.json)

and pasting the content into http://localhost:8080/amr-petri-test


blanchco (approved):

jryu01 (approved):

Tom-Szendrey (approved):
nice and clean i like it

PR #6643, opened 2025-02-24
Pyciemss better error messages

Tom-Szendrey:
# Description
- Update our error component to check for `Error:` and split the last error message out of the trace to help the reader see the most relevant error message first.
- Still keep full traceback if the user wants to view this in an accordion tab that will be defaulted to be closed.
- While Im at it remove `error-message-background` and replace with otherwise used `surface-error`

If we cannot find `Error:` for some reason we will default to open the full trace and leave the last error tab closed

# Screenshots
Here is an example where the pyciemss team is helping with the underflow message:

image

Here is another example for a negative number of samples:
image


YohannParis (approved):
hell yeah!

PR #6645, opened 2025-02-24
support additional sensitivity timeslice options

blanchco:
# Description

* Support additional timeslice options for sensitivity analysis
peak value (all sample max values and analyze the selected state variable value)
peak timepoint (all sample max values and analyze the timepoint)

https://github.com/user-attachments/assets/4bdc8532-0743-43d5-be64-0dbe87bd7ab3


YohannParis (approved):
LGTM

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1026,13 +1027,65 @@ export function useCharts(
...
+	 *          - For MAX/ARGMAX: Records where each variable reaches its maximum value per sample
+	 */
+	function getSlicedData(
+		records: Record<string, any>[],
+		options: { method: SensitivityMethod; timepoint?: number; selectedVariables: string[] }
+	): Map<string, Record<string, any>[]> {
+		const slicedRecords: Map<string, Record<string, any>[]> = new Map();
+		// the records of interest we want are at the indicated timepoint
+		if (options.method === SensitivityMethod.TIMEPOINT) {
+			options.selectedVariables?.forEach((selectedVariable) => {
```
mwdchang, line 1053: According to your function definition, `options.selectedVariables` cannot be undefined.

Also there is something a bit fishy here. you are iterating over the selectedVariables and put them into a map with some data, but all the data are the same (you are filtering by timepoint and never touched the selectedVariables).  So either the logic is incorrect or you are doing a lot of extra work.

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1026,13 +1027,65 @@ export function useCharts(
...
+				const data = records.filter((d) => d.timepoint_id === options.timepoint);
+				if (!slicedRecords.has(selectedVariable)) {
+					slicedRecords.set(selectedVariable, data);
+				}
+			});
+
+			// the records of interest we want are the max value of each sample
+		} else if (options.method === SensitivityMethod.MAX || options.method === SensitivityMethod.ARGMAX) {
+			options.selectedVariables?.forEach((selectedVariable) => {
```
mwdchang, line 1062: Same as other comments, you are not allowing `selectedVariables` to become undefined.

I don't know how exactly you are initializing `records` but if they are casted as `Record<string, any>` you might need to coerce them to numbers to get a true numerical inequality comparison.

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1026,13 +1027,65 @@ export function useCharts(
...
+				const variableMax = records.reduce<Map<number, Record<string, any>>>((acc, record) => {
+					const id = record.sample_id;
+					// loop over all selected variables and get the max value
+					const value = record[selectedVariable];
+					if (!acc.has(id) || value > acc.get(id)?.[selectedVariable]) {
```
mwdchang, line 1067: More efficient if you do 

```
acc.get(id)!.[selectedVariable])
```

Since it is pretty much guaranteed, and unnecessary option chaining produce inefficient code inside loops.

packages/client/hmi-client/src/composables/useCharts.ts
```diff
@@ -1113,27 +1166,31 @@ export function useCharts(
...
-					if (translationMap && translationMap[key]) {
-						const newKey = translationMap[key];
-						r[newKey] = obj[key];
-					} else {
-						r[key] = obj[key];
-					}
+			const dataTranslated = slicedData
+				.entries()
+				.next()
+				.value?.[1].map((obj) => {
```
mwdchang, line 1182: What is up with this `entries().next().value?.[1]` stuff? 
Don't we want to apply to all variables?

packages/client/hmi-client/src/types/common.ts
```diff
@@ -215,13 +215,20 @@ export interface ChartSettingSensitivityOptions {
...
 export enum SensitivityChartType {
 	SCATTER = 'scatter',
 	HEATMAP = 'heatmap'
 }
 
+export enum SensitivityMethod {
+	TIMEPOINT = 'value at timepoint',
+	MAX = 'peak value',
+	ARGMAX = 'peak timepoint'
+}
```
mwdchang, line 230: This enum feels a bit weird, in that the labels are meaningful but the enums themselves are a bit ambiguous.
Why not just:

```
export enum SensitivityMethod {
	TIMEPOINT = 'timepoint',
	PEAK_VALUE = 'peak value',
	PEAK_TIMEPOINT = 'peak timepoint'
}

```

packages/client/hmi-client/src/composables/useCharts.ts
blanchco: done!

packages/client/hmi-client/src/composables/useCharts.ts
blanchco: done and done!

packages/client/hmi-client/src/composables/useCharts.ts
blanchco: My intent was to perform this slicing operation once so that the follow-up functions...

a) Binning into color quantiles (createSensitivityBins)
b) Computing the Pyodide ranking score (getRankingScores)

...could directly use the filtered records without having to repeat the mapping themselves.

I agree that for `value at timepoint` method, each state variable would end up with the same set of records. However, for `peak value` and `peak timepoint` methods, each state variable has a distinct set of records to bin and score. I wanted to make sure that the follow-up functions handle data in a uniform way without doing the mapping themselves.

packages/client/hmi-client/src/composables/useCharts.ts
blanchco: I changed this in my latest commit to make it a `little` more readable but it does the same thing.

We only need to translate for the scatter/heatmap charts and we only have to do this for one of the selected state records since the values will always be the same.

packages/client/hmi-client/src/composables/useCharts.ts
mwdchang: okay - then does it make sense to take it out of the loop so we are not running the same operations over and over?

```
                         const data = records.filter((d) => d.timepoint_id === options.timepoint);
			options.selectedVariables.forEach((selectedVariable) => {
				if (!slicedRecords.has(selectedVariable)) {
					slicedRecords.set(selectedVariable, data);
				}
			});
```

liunelson (approved):
I didn't check the code but the recorded behaviour looks right to me.

mwdchang (approved):

PR #6648, opened 2025-02-24
4101 document node extraction status notification enhancements

### LINKED ISSUE #4101: [PAPER CUT]: No indication on the Resource tree that there is a task running for a resource
- Scenario: Model Extraction
- Step: 2.2
- Link to the project: https://app.staging.terarium.ai/projects/f9b1e1ca-2042-4948-9c4d-54c7f6827a04/model/1c597226-eea1-41bf-af49-54ed8f00fa65

**Describe the issue**
The task steps tell me to wait for document extraction to be completed, and yes the bell icon will inform me that a task was completed (or running) but without arbitrarily clicking on the bell to see if there is a long running task currently being performed, there is no way to understand that the document is being processed and not "ready to use"

mloppie:
# Description

* Adds processing status of pdf extraction to document node and drill down

Screenshot 2025-02-24 at 6 09 03 PM
Screenshot 2025-02-24 at 6 08 58 PM


packages/client/hmi-client/src/composables/useClientEvent.ts
```diff
@@ -17,12 +24,23 @@ export function useClientEvent(
...
-		if (!taskIds?.includes(event.data?.id)) return;
+		if (!taskIds?.includes(event.data?.id) || !event.data) return;
 		if ([TaskStatus.Success, TaskStatus.Cancelled, TaskStatus.Failed].includes(event.data.status)) {
-			state[taskIdKey] = taskIds.filter((id) => id !== event.data.id);
+			state[taskIdsKey] = taskIds.filter((id) => id !== event.data.id);
+		}
+	};
+}
+
+export function createTaskProgressClientEventHandler(state, progressKey: string) {
```
mwdchang, line 37: There are some inconsistencies between `createTaskListClientEventHandler` and `createTaskProgressClientEventHandler`, even though both have a "state" function parameter, you seem to be passing in "props.node" in some cases and "props.node.state" in other cases. 

I am guessing they should be unified here?  and per my musing the last time, it may be a bit more resilient to use "props.node" instead of "props.node.state" when invoking these functions, as props.node should always guarantee to point to the correct memory address.

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
```diff
@@ -6,6 +6,10 @@
 	>
 		<tera-columnar-panel>
 			<tera-drilldown-section :is-loading="isFetchingPDF">
+				<div v-if="props.node.state?.taskProgress">
+					Processing...
+					<ProgressBar :value="props.node?.state?.taskProgress * 100" />
+				</div>
```
mwdchang, line 12: The title of the PR talks about situational awareness on the resource tree/panel. The content of the PR is focusing on showing progress on the workflow. These are (at least so far) two unrelated things, can you clarify the intent?  Or are we missing some commits?

packages/client/hmi-client/src/composables/useClientEvent.ts
mloppie: I had been avoiding that because I feel like it violates the interface segregation clean coding principle by sharing more than is needed ( https://dev.to/moh_moh701/c-clean-code-solid-principles-51ed ) but the more familiar I get with the notifications and how I see the work expanding I am more inclined to agree with you

packages/client/hmi-client/src/components/workflow/ops/document/tera-document-operation-drilldown.vue
mloppie: In discussions there are several related notification tasks that I'm working on simultaneously, and I'm breaking it up into small PRs to make it easier to review. The pr took one of the task titles, I have renamed

packages/client/hmi-client/src/composables/useClientEvent.ts
mwdchang: "props.node.state" is not really an interface, one that can be reasoned with in a systematic manner outside of their defined scopes. So I don't think the principles applies here.

YohannParis (commented):
I do not see an indicator for the document in the ressource panel?

mloppie:
Ok I will take my resource pane work to another branch. I think it should be done later and for all asset types rather than just documents anyway

YohannParis (approved):

PR #6658, opened 2025-02-25
model check warnings/errors - part 1

mwdchang:
### Summary
- Add type (model, state, transition) and entity identifier to error/warning messages so they can be traced back to the model-parts.
- Added provisional errors/warnings prompts in tera-model, will need refinements

Note the error messages will need to be tweaked quite a bit than they are currently. An error is not precisely confined to a particular section, but rather it is about the consistency between 2 or more sections in the model operator. For example a transition is invalid if it does not have corresponding entries in the state section of the model, but both sections are technically valid if we look at them individually, so we need to do some iterations in terms of best messaging and placements of these errors/warnings.

image

### Testing
You can test with a valid but semantically incorrect model, for example
- [no-mass.json](https://github.com/user-attachments/files/18973343/no-mass.json)
- [dc-sir-bad-initials.json](https://github.com/user-attachments/files/18973462/dc-sir-bad-initials.json)


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -1,5 +1,12 @@
 <template>
 	<section class="flex flex-column">
+		<div v-if="modelErrors.length > 0">
+			<ul>
+				<li v-for="err of modelErrors" :key="err.id">
+					<span :class="err.severity">{{ err.content }}</span>
+				</li>
+			</ul>
+		</div>
```
YohannParis, line 9: We are constitently use `isEmpty` to do those conditional test.
```suggestion
			<ul v-if="!isEmtpy(modelErrors)">
				<li v-for="err of modelErrors" :key="err.id">
					<span :class="err.severity">{{ err.content }}</span>
				</li>
			</ul>
```

YohannParis (approved):

blanchco (approved):
Nice!

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
mwdchang: done

PR #6659, opened 2025-02-25
clean up redundant prop collapsed-items

mwdchang:
### Summary
Simplify props for tera-model-part component, we are seemingly passing in redundant data structures

image

### Testing
Find a stratified model, the operations `add xyz to all children` should work, eg update all state units for a stratified state V.


YohannParis (approved):

mwdchang:
> Fine by me but I have minimal context so i may be missing something

@Tom-Szendrey there are two props in the component, "items" and "collapsedItems", collapsedItems exists inside items, so it is redundant and confusing.

Issue #6651, opened 2025-02-25
[FEAT]: Expose "logging_step_size" as "Number of Timepoints" in the simulation operator UI


liunelson:
**Is your feature request related to a problem? Please describe.**

It is currently not possible to change the spacing between timepoints (i.e. time step size). This produces simulation results of insufficient temporal resolution when the number of timepoints is relatively small. 

For example, a simulation is required to run for 10 days and thus has only 10 timepoints.

**Describe the solution you'd like**

The PyCIEMSS interfaces have the argument `logging_step_size` which defaults to `1.0`, meaning "Sample the solution at 1 timepoints per day". If the user wants to higher temporal resolution, they would simply reduce this value (`0.1` means 10 timepoints per day).

We should expose this argument as an additional numeric field beside "Start time" and "End time", possibly as "Number of Timepoints" (`logging_step_size = (EndTime - StartTime) / NumberOfTimepoints` where `NumberOfTimepoints >= 1`).

PR #6665, opened 2025-02-26
Add numberOfTimepoints for each ta3 operator 

Tom-Szendrey:
# Related and Required:
https://github.com/DARPA-ASKEM/pyciemss-service/pull/139

# Description
- Add logging step size to our `SimulationRequest` and `EnsembleSimulationCiemssRequest`
- Add `numberOfTimepoints` to each ta3 operator for their respective simulation runs. This will be used to set `logging_step_size`
- Add `isNumberOfTimepointsManual` to each ta3 operator to denote whether we should calculate `numberOfTimepoints` for the user or let them manually type it.
- Minor clean and fix to calibrate's state management
- Remove `timeSpan` from simulation's state and replace with `endTime` as the `startTime` is always fixed to `0` like the other operators.

# Example why we want this:
If I want to run a simulation for 10 days and want to receive data twice a day there is currently no way of doing this.
Now I will be able to mention that I want end time to be 10 and number of timepoints to be 20.
This will also help when a user has an intervention triggering on a half day for example day 2.5
More information can be found [here](https://github.com/DARPA-ASKEM/terarium/issues/6651)

# Sample video:

https://github.com/user-attachments/assets/ae5259d9-41b4-4ef0-ad03-5b69e3fca292


packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/calibrate-operation.ts
```diff
@@ -69,9 +71,11 @@ export const CalibrationOperationCiemss: Operation = {
 			errorMessage: { name: '', value: '', traceback: '' },
 			numIterations: 100,
 			numSamples: 100,
-			endTime: 100,
+			endTime: 90,
 			stepSize: 1,
-			learningRate: 0.03
+			learningRate: 0.03,
+			numberOfTimepoints: 90,
```
mloppie, line 77: Is there a case where the number of time points and the end time would be different?

mloppie (approved):

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-node-ciemss.vue
```diff
@@ -257,6 +257,7 @@ watch(
 				start: 0,
 				end: state.endTime
 			},
+			loggingStepSize: state.endTime / state.numberOfTimepoints,
```
mwdchang, line 260: These logginStepSize seem to be different than that mentioned in the issue this PR is linked to, which is:
```
(EndTime - StartTime) / NumberOfTimepoints
```

Can we double check what it is we weant?

mwdchang: Also can `loggingStepSize` be fractional? If so it might mess up some of our dataset parsing which assumes that these are whole numbers 1, 2, 3 ... etc

YohannParis (approved):

PR #6667, opened 2025-02-26
modify-run-task-sync-behaviour

blanchco:
# Description

* Modifying the behaviour of the runTaskSync function such that we now have to explicitly handle the task failures and how we want to handle them rather than generically throwing a RuntimeException.  This can either mean bubbling up the error to the front end or simply just logging the error in the backend.

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
@@ -43,6 +43,7 @@ import software.uncharted.terarium.hmiserver.models.dataservice.simulation.Progr
 import software.uncharted.terarium.hmiserver.models.task.TaskRequest;
 import software.uncharted.terarium.hmiserver.models.task.TaskRequest.TaskType;
 import software.uncharted.terarium.hmiserver.models.task.TaskResponse;
+import software.uncharted.terarium.hmiserver.models.task.TaskStatus;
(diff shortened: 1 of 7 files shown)


blanchco:
I couldn't do the services one by one since we are getting rid of the RuntimeExceptions on a failed task which maybe lead to something unexpected...so here it all is 😄

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -520,6 +524,9 @@ public Future<TextExtraction> extractTextFromPDF(
 
 		return executor.submit(() -> {
 			final TaskResponse resp = taskService.runTaskSync(req);
+			if (resp.getStatus() != TaskStatus.SUCCESS) {
+				throw new RuntimeException("Text extraction failed" + resp.getStderr());
+			}
```
mwdchang, line 529: nitpick, your messages are like `"...failed" + stderr()`, add a space between concatenations `..."failed " + stderr()`

mwdchang (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
blanchco: done in latest commit!

PR #6671, opened 2025-02-26
6618 feat extracting model configuration as latex

### LINKED ISSUE #6618: [FEAT]:  Extracting model-configuration as LaTeX

YohannParis:
Generate a LaTeX table of model configuration.

Screenshot 2025-02-26 at 15 00 50

The result LaTeX should be visible in an [online LaTeX viewer](https://texviewer.herokuapp.com)

Screenshot 2025-02-26 at 15 11 52

Resolves #6618


liunelson (approved):
Did not check code, just the LaTeX output

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/ModelConfigurationLatexTable.java
```diff
@@ -0,0 +1,162 @@
...
+				String description = "";
+				String source = escapeLatex(param.getSource());
+
+				// Get the values from the Parameter
+				final ModelParameter parameter = model
+					.getParameters()
+					.stream()
+					.filter(s -> s.getId().equals(param.getReferenceId()))
+					.findFirst()
+					.orElse(null);
```
mwdchang, line 94: Consider caching states and parameters so this won't become a perf problem with larger models.

Something like
```
final Map<UUID, State> map = new HashMap<>();
for (final State state: model.getStates()) {
  map.put(state.getId(), state);
}

...

for (InitialSemantic initial : modelConfiguration.getInitialSemanticList()) {
  ...
  state = map.get(initial.getTarget());
 ...
}

```

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/ModelConfigurationLatexTable.java
YohannParis: Fixed! Thanks @mwdchang

PR #6672, opened 2025-02-26
fix enrichment button not being asset specific

mloppie:
# Description

* fix enrichment button not being asset specific 
* improve button styling with enrich status
* code cleanup
* add spinners to dataset and model nodes when enrichment is in progress

Screenshot 2025-02-26 at 2 14 05 PM
Screenshot 2025-02-26 at 2 13 44 PM


YohannParis (approved):

packages/client/hmi-client/src/composables/useClientEvent.ts
```diff
@@ -50,3 +50,15 @@ export function createTaskProgressClientEventHandler(node: WorkflowNode<Document
...
 	};
 }
+
+export function createEnrichClientEventHandler(taskId: Ref, assetId: string | null, emit) {
+	return async (event: ClientEvent<TaskResponse>) => {
+		if (taskId.value !== event.data?.id) return;
+		if (assetId !== event.data.additionalProperties.datasetId && assetId !== event.data.additionalProperties.documentId)
+			return;
+		if ([TaskStatus.Success, TaskStatus.Cancelled, TaskStatus.Failed].includes(event.data.status)) {
+			taskId.value = event.data.status === TaskStatus.Failed ? TaskStatus.Failed : '';
```
mwdchang, line 60: I don't understand, in the upper portion of this function it seems like we are treating `taskId` as if it is a unique identifier, and here we are reassigning `taskId` to the status of the job itself. It seems like there is some double-encoding going on and I don't think it is clear to the readers what to expect, especially if we are changing a Ref which can unintended consequences if it is being used by the caller.

mloppie: I can explain, the issue is I'm working on a few things related to notifications and writing some "while I'm in there" parts but not doing a great job keeping them segregated for smaller PRs. The taskId is indeed unique, but it's specific value does not matter outside of the event handler. It is just used as a proxy in the component to the status of in progress or failed or whatever. In another branch I have renamed the prop in the component and use the OperatorStatus enum. That is in the process of getting cleaned up and will make more sense in the next one

mwdchang: Okay - We (or at least I) do not have foresights beyond what is been presented in the PR, so if there are things being added that are in some kind of transitional phase, would appreciate if you can call them out next time so we know what to look at or not to look at.

Also as far as tracking a particular task is concerned, I am not sure what you have in mind here - it appears to be session-based and would not survive a page refresh if the task in question is still in progress.

mwdchang (approved):

packages/client/hmi-client/src/composables/useClientEvent.ts
mloppie: Sure thing.

The event handler factory uses subscribe, the same as the notification panel, so it will survive a refresh as well as anything else in the app currently. We've spoken about ways to improve that architecture and I'm in favour. For the scope of this PR though, you can now click enrich on a model or dataset and it won't make the buttons disable and spin for all other models and datasets.

Issue #6673, opened 2025-02-26
[SUBTASK]: 2.5. Update Project Controller to use new permissions


dvince2:
Projects have a slightly different format.

PR #6675, opened 2025-02-26
[SUBTASK]: 2. Modify pantera auth to use rebac and replace auth

### LINKED ISSUE #6653: [SUBTASK]: 2. Modify pantera auth to use rebac and replace auth
**Describe the task**
As a stepping stone, move all rebac authentication to the `HasProjectAccessAspect` class and replace calls in all our classes to use correct annotation.

dvince2:
# Description

* Changing all of the explicit calls to our rebac service to instead go through the new annotation.
* `ProjectController` is going to be done seperately as its a bit more involved in issue #6673
* Changed the default param we look for for the project id from `id` (pantera) to `projectId` (terarium standard). Updated the tests to reflect this

Resolves #6653

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
@@ -18,6 +18,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.server.ResponseStatusException;
+import software.uncharted.terarium.hmiserver.annotations.HasProjectAccess;
 import software.uncharted.terarium.hmiserver.controller.SnakeCaseController;
 import software.uncharted.terarium.hmiserver.models.dataservice.AssetType;
 import software.uncharted.terarium.hmiserver.models.dataservice.model.configurations.ModelConfiguration;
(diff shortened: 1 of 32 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
```diff
@@ -72,14 +73,11 @@ private static class SimulationRequestBody<T> {
 
 	@GetMapping("/{id}")
 	@Secured(Roles.USER)
+	@HasProjectAccess
```
YohannParis, line 76: They are no `level` specifiy because by default is READ only?
If so, why not have different annotation? like:
- `@HasProjectAccessRead`
- `@HasProjectAccessWrite`
- `@HasProjectAccessOwner`
- etc.

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/simulationservice/SimulationRequestController.java
dvince2: In theory, this is more cleanly used where `@HasProjectAccess` is set on the entire class, then you just override a handful of endpoints with write access.
However we have a number of endpoints on each controller which have no permission checks so it became easier to just annotate each endpoint. In another pass im going to see whether the endpoints that don't have permissions were oversights or on purpose. There are also a number of endpoints which clearly had the wrong permissions before (we were checking for READ on POST/PUTs, etc).  I changed a couple of those but realized I should upgrade those at the same time as the other changes as a 'permission enhancements' PR

blanchco (approved):

PR #6684, opened 2025-02-27
6681 feat download model configuration as csv

### LINKED ISSUE #6681: [FEAT]:  Download Model Configuration as CSV
Provide two CSV textarea for initial and parameters per model-configuration.

YohannParis:
# Description

Screenshot 2025-02-27 at 13 49 49

Screenshot 2025-02-28 at 10 41 07

Resolves #6681


packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/ModelConfigurationToTables.java
```diff
@@ -13,9 +12,92 @@
...
 
-public class ModelConfigurationLatexTable {
+public class ModelConfigurationToTables {
 
-	public static String generateLatexTable(Model model, ModelConfiguration modelConfiguration) {
+	public static String generateCsv(Model model, ModelConfiguration modelConfiguration) {
+		StringBuilder csv = new StringBuilder();
+
+		// Add the header
+		csv.append("ID,Name,Value,Units,Description,Source\n");
```
mwdchang, line 21: Are we sure we want these columns across both initials and parameter? because it isn't obvious how to tell them apart, for example:

```
S, S, 1.0, null, null, null
beta, beta, 0.02, null, null, null
```

One is initial and one is parameter. Should we have an additional `type` column to differentiate?

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-configuration-item.vue
```diff
@@ -83,6 +100,13 @@ const contextMenuItems = ref([
 			showLatexTable();
 		}
 	},
+	{
+		label: 'CSV table',
+		icon: 'pi pi-table',
+		command() {
+			showCSVTable();
```
mwdchang, line 107: @liunelson what if the model-configuration is one that came from a calibration process that points to a dill file?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/utils/ModelConfigurationToTables.java
YohannParis: Sure, if necessary

Tom-Szendrey (approved):

packages/client/hmi-client/src/components/workflow/ops/model-config/tera-model-configuration-item.vue
liunelson: > @liunelson what if the model-configuration is one that came from a calibration process that points to a dill file?

@YohannParis 
The example above doesn't show it. However, the LaTeX table example did cover a similar case as your calibrated one.

```python
Type, ID, Name,Value,Units,Description,Source
parameter,β,beta,Uniform(min = 0, max = 1),1/(person*day),infection rate, Paris et al. 1999
parameter,γ,gamma,CalibratedNormal(mu = 0.1, sigma = 0.4),person/day,recovey rate,Chang et al. 2025
```

PR #6704, opened 2025-03-03
Add ability to define intervention percentages

blanchco:
# Description

Here is a v1 of adding percentages to interventions.  There are a number of q's in the design as to how we would want to continue this.

I've added a `value_type` field to Intervention which indicates whether the value is a value or percentage.  The calculations for final value of the intervention are in this pyciemss ticket https://github.com/DARPA-ASKEM/pyciemss-service/pull/140

https://github.com/user-attachments/assets/9c0f770d-e6b5-40d1-bb77-9f9f6c349d33

https://github.com/user-attachments/assets/e78af651-e47b-421e-8b6e-a4405c184c6d

CB Edit:
After a discussion with Pascale, Nelson, and Dan the following has been added:
1. Do not allow for intervention percentages on states
2. For calibrated configurations use the `mean` to calculate the percentage off of
3. For optimize we will resolve the value that is used in the intervention cards.  That is if beta = 1 and the percentage is 10%.  We will resolve the intervention to be 0.1 where applicable


blanchco:
This will require https://github.com/DARPA-ASKEM/pyciemss-service/pull/140 to be merged as well

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionValueType.java
```diff
@@ -0,0 +1,11 @@
+package software.uncharted.terarium.hmiserver.models.simulationservice.interventions;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public enum InterventionValueType {
+	@JsonProperty("value")
```
Tom-Szendrey, line 6: Nitpick:
I believe the convention of enums is to be all caps. That should just involve removing this `@JsonProperty`

Tom-Szendrey:
The percent is not being used correctly at the moment by ta3 operations such as simulate
If i provide an intervention to update beta to 10% of its value at timestep 1 we are currently setting beta to value 10 at timestep 1.

image

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
```diff
@@ -847,6 +853,19 @@ const setInterventionPolicyGroups = (interventionPolicy: InterventionPolicy) =>
 	emit('update-state', state);
 };
 
+const getInitialGuessValue = (intervention: StaticIntervention) => {
```
Tom-Szendrey, line 856: This looks like something that would better live in `packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-utils.ts`

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-utils.ts
```diff
@@ -171,7 +172,8 @@ export async function getOptimizedInterventions(optimizeRunId: string) {
 						appliedTo: paramName,
 						type: InterventionSemanticType.Parameter,
 						timestep: newTimestepAsList[0],
-						value: paramValue as number
+						value: paramValue as number,
+						valueType: InterventionValueType.Value
```
Tom-Szendrey, line 176: Here we will need to keep the `InterventionValueType`.
This function is about rebuilding interventions from the optimize call.
This specific section is when the user has asked for a better start time for their intervention.

Here is an example of why this is an issue
If a user has said `give me a better start time for 10% of beta`
We are currently rebuilding that as `new better start time for beta to be value 10`

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/tera-optimize-ciemss-drilldown.vue
blanchco: Done and I've renamed it to `resolveInterventionValue`

packages/client/hmi-client/src/components/workflow/ops/optimize-ciemss/optimize-utils.ts
blanchco: I've make some changes so I _believe_ this should work as expected now.  I am using that function `resolveInterventionValue` so send the actual value of the intervention in the optimize payload.

Using your example if beta is 1 and the user asks `give me a better start time for 10% of beta`,  the value sent to the optimize call resolved to be 0.1.  And what it rebuilt is now `new better start time for beta to be value 0.1`

Tom-Szendrey (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/simulationservice/interventions/InterventionValueType.java
dvince2: seconded!

PR #6714, opened 2025-03-04
Optimize over grouped interventions

Tom-Szendrey:
# Description
- Update that state such that each policy group form can contain a single intervention in form `StaticIntervention | DynamicIntervention`
- Update drilldown, static, and group forms to follow suit 
- WARNING this updates the state for optimize and is not backwards compatible


mloppie (approved):
Nice

blanchco (approved):

PR #6718, opened 2025-03-04
Clean 4 function locations

Tom-Szendrey:
# Description
- remove calibrate's `mergeResults` that was also used in the ta3 operators. This can be replaced with `services/dataset` version of the same thing.
- move `renameFnGenerator` that is used by all ta3 operators. Was in `calibrate-util` now is in `simulation-service`
- move 2 optimize functions from `-operation` to `-util`

**No functional changes**


mwdchang (approved):
Didn't test, looks okay

mloppie (approved):

PR #6722, opened 2025-03-04
rename reset_request/reset_response

mwdchang:
Requires: https://github.com/DARPA-ASKEM/askem-beaker/pull/207

### Summary
Corresponding event renaming changes for beaker PR. Rename reset_request and reset_response to avoid collision and confusion with internal kernerl-level events.

### Testing
There is a caveat that our AspectJ is not correctly configured, to test currently you need to go into "ModelController.java" and remove the `@HasProjectAccess` annotation/aspect from the `GET /{id}` route.

After this stratify operations should work in both wizard and notebook views.


Tom-Szendrey (approved):
Looks fine, Did not test. If you want me to test please slack me

YohannParis (approved):
I can't test locally. Will do on staging

mwdchang:
> I can't test locally. Will do on staging

Note staging will not work until we fix the AspectJ issue, as it is blocking beaker from accessing models. 
See: https://github.com/DARPA-ASKEM/terarium/issues/6723

PR #6724, opened 2025-03-04
fix warnings in drag and drop validation

mloppie:
# Description

* There used to be warnings when uploading documents because the validation was comparing arrays in the wrong direction.
* This PR fixes that

Before:
Screenshot 2025-03-04 at 4 09 47 PM


Tom-Szendrey (approved):
LGTM - didnt test

YohannParis (approved):

PR #6725, opened 2025-03-04
clean up mira_model context references and unused components

mwdchang:
### Summary
We are using "mira_model_edit" instead of "mira_model", this PR cleans up the references and removes components that are no longer in-use.

### Testing
No real functional change, app should work same as before


YohannParis (approved):

Tom-Szendrey (approved):
Thank you

PR #6733, opened 2025-03-05
6729 bug model config csv and latex download should reflect correct values for calibrated model configurations

### LINKED ISSUE #6729: [BUG]: model-config CSV and LaTeX download should reflect correct values for calibrated model-configurations
Follow up from: https://github.com/DARPA-ASKEM/terarium/pull/6684#pullrequestreview-2659204108

The download-csv and download-latex for model-configuration draws from `getParameterSemanticList()` function. This is correct except for the case where the model-configuration is the result of a calibration, here `getParameterSemanticList()` would either be empty or reflect that of the base model and not the config itself.

To get the "correct" (eg sampled from simulate runs) we need to use the values in `getInferredParameterList()` instead of `getParameterSemanticList(...)`.

So it will look something like:

```
if (modelConfig.getSimulationId() == null) {
  // Regular config
  for (val parameter : modelConfig.getParameterSemanticList()) { ... }
} else {
  // Calibrated config
  for (val parameter : modelConfig.getInferredParameterList()) { ... }
}
```

Note while we calculate the "mean" and "standard deviation" for calibrated parameters, we don't actually know what type of distribution they are suppose to be, so the type is labelled as `inferred`.

in tera-calibrate-node-ciemss:

``` 
       inferredParameters.push({
         source: 'calibration',
         type: SemanticType.Inferred,
         referenceId: parameterId,
         default: false,
         distribution: {
           type: 'inferred',
           parameters: {
             mean,
             stddev
           }
         }
       });
```

YohannParis:
Disable the downloading as CSV and LaTeX for calibrated model-configurations.
This is not really an issue as we cannot see them from the model-configuration operator, but we might in the future.

Resolves #6729


mwdchang (approved):
Sure

Tom-Szendrey (approved):

PR #6735, opened 2025-03-05
Fix: Compare Dataset node's charts

Tom-Szendrey:
# Description

We can now properly see charts on the node.
The node will now have correct/matching legends with drilldown

- clean up duplicated spinner as its already in the node-preview
- chartData to be computed rather than set up in initialized. This will keep it up to date as needed and follows more closely with other drilldowns.
- Do not set the value of `datasets` via `initialize` handling a `ref`. This was causing duplication issues when initialize was called back to back before finishing. For more details see below.

Last point explained further:
We were calling `initialize` on input change in the node. If a user were to add multiple inputs before the datasets were pulled we could have duplicate datasets being appended. Code example below:

```
//Call initialize 
datasets.value = [];
datasetPromises = A
//Call initialize again 
datasets.value = []
datasetPromises = B
datasets.value.push(A)
datasets.value.push(B)

```


jryu01 (approved):
Code looks fine

blanchco (approved):

PR #6736, opened 2025-03-05
Simplify model ratelaw

mwdchang:
### Summary
Add a simplify model (AMR) taskrunner task. This is used to simplify complicated rate laws that may result in stratification explosions.

### Test
Kick off a task with `POST http://localhost:3000/mira/simplify` 
[Here](https://github.com/user-attachments/files/19196020/SIR_model_modified.json)
 is a sample request body that is an SIR that has been modified to utilize in order to have a transition that is grouped and can be changed.

[Here](https://github.com/user-attachments/files/19196038/SIR_Stratified_Not_Factored.json) is an SIR that has no groups and therefore will not have a meaningful change


mwdchang:
FYI @Tom-Szendrey 

Changes needed:
- We need to call "check_simplify_rate_laws" instead of "simplify_rate_laws". The interface is a bit different in that you might need to convert AMR to MMT and back to AMR. See [check_simplify_rate_laws](https://github.com/DARPA-ASKEM/mira/blob/6e74747eef77ccca604753aa361a55004213b9be/mira/metamodel/ops.py#L574) for details
- Need to massage the output model of check_simplify_rate_laws to am AMR

YohannParis (approved):
LGTM

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/tasks/TaskServiceTest.java
```diff
@@ -337,6 +337,22 @@ public void testItCanSendAmrToMmtRequest() throws Exception {
...
+	public void testItCanSendSimplifyModelRequest() throws Exception {
+		final ClassPathResource resource = new ClassPathResource("mira/SIR_model_modified.json");
+		final String content = new String(Files.readAllBytes(resource.getFile().toPath()));
+
+		final TaskRequest req = new TaskRequest();
+		req.setType(TaskType.MIRA);
+		req.setScript("mira_task:simplify_model");
+		req.setInput(content.getBytes());
+
+		final TaskResponse resp = taskService.runTaskSync(req);
```
mwdchang, line 351: What are you testing? There is no assertions in this test.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/ClientEventType.java
```diff
@@ -29,6 +29,7 @@ public enum ClientEventType {
 	TASK_GOLLM_DATASET_STATISTICS,
 	TASK_MIRA_AMR_TO_MMT,
 	TASK_MIRA_GENERATE_MODEL_LATEX,
+	TASK_MIRA_SIMPLIFY_MODEL,
```
mwdchang, line 32: This event type `TASK_MIRA_SIMPLIFY_MODEL` is not being used.

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/tasks/TaskServiceTest.java
Tom-Szendrey: That we are able to send the request. I am not certain as to why all of the tests are to exclusively check if we can kick the job off and nothing more but this is the same formatting as many of the tests in this file such as:
```
testItCanSendTextExtractionRequest
testItCanSendLatexToAMRRequest
testItCanSendGenerateModelLatexRequest
testItCanSendAmrToMmtRequest
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/ClientEventType.java
Tom-Szendrey: yes, from my understanding it will be used in the next PR. 
I will remove it from here as it can be added back easily enough

packages/server/src/test/java/software/uncharted/terarium/hmiserver/service/tasks/TaskServiceTest.java
mwdchang: Okay, let's leave it.  But I would like to stay away from adding tests that by large, do not do anything - they will just bloat the CI-build time.

Tom-Szendrey: Very reasonable. I figured it may be important for reasons I do not understand as there are many like it but that makes sense.
I will leave it for now

Tom-Szendrey (approved):

PR #6767, opened 2025-03-07
Fix intervention policy deletion

blanchco:
# Description

* fixes an issue with intervention policies where deleting an intervention card would cause some bugginess.
* we should not be using the array index as a key but instead this uuid

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
@@ -109,25 +109,23 @@
 					Add and configure intervention settings for this policy.
 					<ul class="flex flex-column gap-2">
 						<li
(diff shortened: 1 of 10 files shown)


packages/client/hmi-client/src/components/workflow/ops/intervention-policy/tera-intervention-policy-drilldown.vue
```diff
@@ -113,7 +113,7 @@
 								firstRow,
 								firstRow + MAX_NUMBER_OF_ROWS
 							)"
-							:key="firstRow + index"
+							:key="intervention.id"
```
mwdchang, line 116: Is is the heart of the fix? If so we might want to just generate a better surrogate key. The idea of `firstRow + index` is basically the same thing as the `uuidv4()`. And I think preferably we would do one way or another, and not trying to juggle two different semantics that are both trying to introduce uniqueness to the same set of items.

mwdchang (approved):

dvince2 (approved):

PR #6768, opened 2025-03-07
6673 subtask 25 update project controller to use new permissions

dvince2:
# Description

*Locking down the project controller. This time the `#id` needs to be set specifically as the API expects `projectId` by default


Tom-Szendrey (approved):
LGTM - Note did not test

blanchco (approved):

PR #6775, opened 2025-03-07
ocr taskrunner plumbing

mwdchang:
### Summary
This is the initial setup to have an all-in-one extraction taskrunner for pdf and other documents. It is meant to unify these bespoked pipelines below, in order to have a systematic way to address provenance related matters where all three are referring to the same basis
- equation_extraction
- text_extraction
- table_extraction

The coarse logic is roughly as follow:
```
- Run Docling layout and text extraction to get bounding boxes and their classifications
- For each result.text = text do additional inference
- For each result.formula = text do additional inference
- For each result.table do additional inference
- Collect and trim into final data structure
```

### Note
This is a memory hungry process, pay attention to the output logs, if you see something like this
```
DEBUG:docling.pipeline.base_pipeline:Finished converting page batch time=611780.866
2025-03-11 20:28:50,190 WARN exited: process1 (terminated by SIGKILL; not expected)
2025-03-11 20:28:51,201 INFO spawned: 'process1' with pid 80530
2025-03-11 20:28:52,206 INFO success: process1 entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
```
it may indicate that your docker environment ran out of memory and we are not able to run the extraction. You will need to either stop other containers to free up memory, or allocate more resources to your docker environment.

### Testing
Under `src` run 
```
docker build -t ocr-test --platform linux/arm64 -f packages/ocr_extraction/Dockerfile .
```

Then you can run the docker image as something like:

```
docker run \
    -p 8002:8002 \
    -e TERARIUM_MQ_ADDRESSES="amqp://rabbitmq:5672" \
    -e TERARIUM_MQ_PASSWORD="terarium123" \
    -e TERARIUM_MQ_USERNAME="terarium" \
    -e TERARIUM_TASKRUNNER_REQUEST_TYPE="ocr_extraction" \
    -e OPEN_AI_API_KEY="open ai api key" \
    --network terarium \
    ocr-test
```

Note the rabbitMQ dependency, you either need to run terarium `./hmiServerDdev.sh start staging` or need to adjust the environment values accordingly.

You can test a PDF with this curl command, replace the file part as needed, the result will be a formatted json with text, image and table sections, where each section item will have a "provenance" used for tracking.

```
curl -X POST \
	-H "Content-Type: multipart/form-data" \
	-H 'accept: application/json' \
	--connect-timeout 100 \
	--max-time 1000 \
       	"http://127.0.0.1:8002/predict" \
	-F "file=@/Users/dchang/workspace/askem/pdfs/Measles_small.pdf"
```


Tom-Szendrey:
I am able to run this and test with my own pdf files to see results. 
I suspect there are many more PRs to come so feel free to ignore but we may want to update the `README` if we intend on only running this ocr locally

Tom-Szendrey (approved):
Should be noted when testing I saw output in docker logs but the output of the curl was 
`Empty reply from server`

packages/ocr_extraction/requirements.txt
```diff
@@ -0,0 +1,9 @@
+fastapi
+uvicorn
+pydantic
+python-multipart
+docling==2.25.2
+augraphy
+rapidocr-onnxruntime
+openai
+git+https://github.com/DARPA-ASKEM/TexTeller.git
```
YohannParis, line 9: Could this be in terarium repo?

YohannParis (approved):

packages/ocr_extraction/requirements.txt
mwdchang: Probably but I don't know pip well enough to wrangle that because you have a situation of having a repo inside a repo. Having said that I don't think it is a good idea to subsume library dependencies.

mwdchang:
> I am able to run this and test with my own pdf files to see results. I suspect there are many more PRs to come so feel free to ignore but we may want to update the `README` if we intend on only running this ocr locally

Not just locally. The orchestration bits will be done next round.

PR #6799, opened 2025-03-12
Model from eq verification pt.1

blanchco:
# Description

* Part 1 for the model from equations verification flow.
* Navigate to next and previous equations with buttons
* Improved data structure of equations, state to accommodate designs.  (Separate included and excluded equations lists, uuids for equations to key over)
* Removed places where AssetBlock wasn't used anymore, except for `code-asset`.  It might be worth it to remove code asset related files in another PR since there's a lot yet it is not used anymore

https://github.com/user-attachments/assets/e3a1a8f8-e620-4bb4-8ac6-85835cc55357

TODOs in upcoming PRs
* Keep provenance of PDF pages when cleaning equations
* Scroll to next equation when press next/previous
* Step 2: enrichment


YohannParis (approved):

mwdchang (approved):
I sorta feel like maybe there are better data-structures we can use here that are more aligned with the types of navigation we have in mind - maybe something like a doubly-linked-list to facilitate these types of next/prev interactions, and we don't need to do these "findIndex" every time. 

Maybe we should touch base and discuss before we get too far into the next parts.

PR #6804, opened 2025-03-12
Fix: Correct Optimize Chart Label

Tom-Szendrey:
# Description
Correct the sorting used to determine the average risk so that we have the correctly our value.
Note that the sorting was using the default sort which will sort as though we are looking at strings and not in number order

# Pictures:
New:
image

Old:

image

# Test:
Run `yarn staging` and check out my optimize node:
`/projects/43693d8d-b162-413d-b119-0610f61c9e70/workflow/e8fb32d1-f0b3-4afe-9fa9-aff8ebfe6e3d?operator=512d4ff4-a7d7-4121-acf3-8b7bcc129adf`

[new sort.txt](https://github.com/user-attachments/files/19215221/new.sort.txt)
[Old Sort.txt](https://github.com/user-attachments/files/19215222/Old.Sort.txt)


blanchco (approved):
🤩

YohannParis (approved):

PR #6805, opened 2025-03-12
supporting week-interval for model and simulations

mwdchang:
### Summary
Support "week" intervals for model and simulations. 

https://github.com/user-attachments/assets/def735fb-ea63-496d-b723-8bb1f766f153

### Note 
Our notion for the "start of the week" is whatever date the modeller chose in the model-configuration operator. If the modeller goes backward (using the calendar to set timestep), we assume they will select the same weekday as they had originally selected on the model-configuration operator.

### Testing
- Save a model with timeunit = week
- Create a config and pick a starting date/day
- Run a simulation, verify that the intervals for setting "end time" and the chart's data points are multiples of seven days

### DIFF
--- a/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
+++ b/packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
@@ -29,9 +29,9 @@
 						option-value="value"
 						:options="[
 							{ label: 'Days', value: CalendarDateType.DATE },
+							{ label: 'Weeks', value: CalendarDateType.WEEK },
 							{ label: 'Months', value: CalendarDateType.MONTH },
-							{ label: 'Years', value: CalendarDateType.YEAR },
-							{ label: 'Weeks', value: CalendarDateType.WEEK }
+							{ label: 'Years', value: CalendarDateType.YEAR }
 						]"
 					/>
 					<tera-input-text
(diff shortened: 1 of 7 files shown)


packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
```diff
@@ -30,7 +30,8 @@
 						:options="[
 							{ label: 'Days', value: CalendarDateType.DATE },
 							{ label: 'Months', value: CalendarDateType.MONTH },
-							{ label: 'Years', value: CalendarDateType.YEAR }
+							{ label: 'Years', value: CalendarDateType.YEAR },
+							{ label: 'Weeks', value: CalendarDateType.WEEK }
```
YohannParis, line 34: Does it matters in the dropdown? Will they be organise properly `Days, Weeks, Months, Years`?

YohannParis (approved):
LGTM

blanchco (approved):
📆 Nice

packages/client/hmi-client/src/components/model/model-parts/tera-model-part-entry.vue
mwdchang: I had half a mind of putting the odd one out at the bottom, but sure, I will put the intervals in order

PR #6817, opened 2025-03-14
HITL verificiation review enrichments

blanchco:
# Description

* Review enrichments in the model from equations tab

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-operation.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-operation.ts
@@ -10,12 +10,29 @@ export interface EquationBlock {
 	extractionError?: boolean;
 }
 
+export enum EnrichmentType {
+	DESCRIPTION = 'description',
+	PARAMTER = 'parameter',
+	STATE = 'state',
+	OBSERVABLE = 'observable',
+	TRANSITION = 'transition',
+	CUSTOM = 'custom'
+}
+
+export interface EnrichmentBlock {
+	content: any;
+	path: string[];
+	include: boolean;
+	type: EnrichmentType;
+}
+
 export interface ModelFromEquationsState {
 	includedEquations: AssetBlock<EquationBlock>[];
 	excludedEquations: AssetBlock<EquationBlock>[];
 	text: string;
 	modelFramework: string;
 	modelId: string | null;
+	enrichments: AssetBlock<EnrichmentBlock>[];
 }
 
 export const ModelFromEquationsOperation: Operation = {
@@ -35,7 +52,8 @@ export const ModelFromEquationsOperation: Operation = {
 			excludedEquations: [],
 			text: '',
 			modelFramework: 'petrinet',
-			modelId: null
+			modelId: null,
+			enrichments: []
 		};
 		return init;
 	}
--- a/packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
+++ b/packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
@@ -4,16 +4,20 @@ import { Model } from '@/types/Types';
 import { b64EncodeUnicode } from '@/utils/binary';
 import { createModelMap } from '@/model-representation/service';
 import _ from 'lodash';
+import { formatTitle } from '@/utils/text';
 import { EnrichmentBlock, EnrichmentType } from './model-from-equations-operation';
 
-export function formatTitle(title: string): string {
-	return title
-		.replace(/([A-Z])/g, ' $1')
-		.replace(/^./, (str) => str.toUpperCase())
-		.trim();
+interface ModelEnirchmentResponse {
+	modelCard: any;
+	modelEnrichment: {
+		states: any[];
+		parameters: any[];
+		observables: any[];
+		transitions: any[];
+	};
 }
 
-export const createEnrichmentCards = (enrichments: any) => {
+export const createEnrichmentCards = (enrichments: ModelEnirchmentResponse) => {
 	const cards: AssetBlock<EnrichmentBlock>[] = [];
 
 	function hasNestedObjects(value: any): boolean {
@@ -24,9 +28,7 @@ export const createEnrichmentCards = (enrichments: any) => {
 	}
 
 	function processModelCardSection(content: any, parentPath: string[] = []) {
-		if (typeof content !== 'object' || Array.isArray(content) || content === null) {
-			return;
-		}
+		if (!hasNestedObjects(content)) return;
 
 		Object.entries(content).forEach(([key, value]) => {
 			const currentPath = [...parentPath, key];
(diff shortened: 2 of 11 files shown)


packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+import { createModelMap } from '@/model-representation/service';
+import _ from 'lodash';
+import { EnrichmentBlock, EnrichmentType } from './model-from-equations-operation';
+
+export function formatTitle(title: string): string {
+	return title
+		.replace(/([A-Z])/g, ' $1')
+		.replace(/^./, (str) => str.toUpperCase())
+		.trim();
+}
```
YohannParis, line 14: This would be best in the `/utils/text.ts`.

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+		if (typeof value !== 'object' || Array.isArray(value) || value === null) {
+			return false;
+		}
+		return true;
+	}
+
+	function processModelCardSection(content: any, parentPath: string[] = []) {
+		if (typeof content !== 'object' || Array.isArray(content) || content === null) {
+			return;
+		}
```
YohannParis, line 29:
```suggestion
		if (!hasNestedObjects(content)) return;
```

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	// Start processing from modelCard
+	processModelCardSection(enrichments.modelCard, ['modelCard']);
+	processModelEnrichment(enrichments.modelEnrichment);
+
+	return cards;
+};
+
+function buildHTMLDescriptionFromEnrichments(enrichments: AssetBlock<EnrichmentBlock>[]): string {
```
YohannParis, line 121: We should talk about `<template>` to do this next week.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template#examples

YohannParis (approved):
LGTM

packages/client/hmi-client/src/model-representation/service.ts
```diff
@@ -498,3 +498,38 @@ export function createTimeList(time) {
 		isParent: false
 	}));
 }
+
+export function createModelMap(model: Model): {
+	states: Map<string, State & RegNetVertex>;
```
mwdchang, line 503: We don't need RegNet anymore

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+import { EnrichmentBlock, EnrichmentType } from './model-from-equations-operation';
+
+export function formatTitle(title: string): string {
+	return title
+		.replace(/([A-Z])/g, ' $1')
+		.replace(/^./, (str) => str.toUpperCase())
+		.trim();
+}
+
+export const createEnrichmentCards = (enrichments: any) => {
```
mwdchang, line 16: Do we have typing information for "enrichments"? 

If not it would help to provide an example data structure here as part of the function's comment.

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-operation.ts
```diff
@@ -10,12 +10,29 @@ export interface EquationBlock {
...
+	type: EnrichmentType;
+}
+
 export interface ModelFromEquationsState {
 	includedEquations: AssetBlock<EquationBlock>[];
 	excludedEquations: AssetBlock<EquationBlock>[];
 	text: string;
 	modelFramework: string;
 	modelId: string | null;
+	enrichments: AssetBlock<EnrichmentBlock>[];
```
mwdchang, line 35: What exactly is this used for? 

This is potentially problematic because the the Workflow state structure is meant to carry states, not data. With large data size we will have synchronization problems (for the user interface) and database write issues (at the backend).

blanchco: These are enrichment blocks, we need to keep track of which enrichments we are applying to the model and which ones aren't.  Similar to how the equations are kept in the state

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
blanchco: done

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
blanchco: done

packages/client/hmi-client/src/model-representation/service.ts
blanchco: done

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+		model.metadata.description = description;
+	}
+
+	includedEnrichments.forEach((enrichment) => {
+		const semanticId = enrichment.asset.content.id;
+
+		if (enrichment.asset.type === EnrichmentType.STATE) {
+			const foundState = modelMap.states.get(semanticId);
+			if (foundState) {
+				Object.assign(foundState, _.omit(enrichment.asset.content, 'id'));
```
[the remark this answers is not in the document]
blanchco: I'm now explicitly setting what we get, which is description and units

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
blanchco: Sounds good!

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
blanchco: I've added a rudimentary interface for the type here

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-operation.ts
mwdchang: The difference is that enrichment has a lot more textual information, which will increase the size of the state at a faster rate than that of the equations. Also I probably made the same comments about the equations should not be a part of the state but we did it anyway, keep in mind we have our fair share of technical debts 🤣

mwdchang (approved):
Just a reminder we need to redo State here to hold states or data-references, not the actual data.

PR #6838, opened 2025-03-17
Implement the simply_model UI 

Tom-Szendrey:
# Description
- Add end point in service/model.ts 
- add initial implementation on the model page.

Here when the user opens a model page we will check if this model can be simplified using our mira taskrunner.
If we find that the model can be simplified we will inform the user with an info bar and allow them to save the resulting simplified model as a new model that will be saved to the project

# Testing:
- Add the following model to your project.
- Open the model via the project asset side panel/Model drilldown/Stratify drilldown/Model edit drilldown
- Ensure you see the info banner at the top and click on the green text saying to save the model.
- See an info toaster that the model was saved and check the project's assets to confirm you see the new model saved.

[SIR With Groups simplified.json](https://github.com/user-attachments/files/19296588/SIR.With.Groups.simplified.json)

# Screenshot:
image

### DIFF
--- a/packages/client/hmi-client/src/components/model/tera-model.vue
+++ b/packages/client/hmi-client/src/components/model/tera-model.vue
@@ -297,7 +297,6 @@ defineExpose({ temporaryModel });
 
 <style scoped>
 .save-simplified-button {
-	/* color: blue; */
 	display: contents;
 }
 .btn-group {


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/SimplifyModelResponse.java
Tom-Szendrey, line 1: @dvince2 Is this the correct location for this file?

YohannParis (approved):

PR #6853, opened 2025-03-18
Fixing dataset access for users who only have read access to a project.

dvince2:
# Description

* The legacy permission call here thew an exception and we can't have that. Changing to new API


mwdchang (approved):
Didn't test, code looks okay

YohannParis (approved):

PR #6854, opened 2025-03-18
fix unused model card prompt import

blanchco:
# Description
It looks like the cleanup equation task was erroring out (at least for me locally) due to the model card prompt being removed.

The script was still looking for this unused import and erroring out.  Removing the import will fix this


mwdchang (approved):
sure

YohannParis (approved):

PR #6859, opened 2025-03-19
Compare Dataset Criteria fixes

Tom-Szendrey:
# Description
- Rename TimepointOption  `OVERALL` to `PEAK` as to not be confused with the new `AVERAGE` option. 

Correct logic for the follow:
- PEAK -> Should find each dataset variable's peak independently. 
- LAST -> should not assume each dataset has the same length.
- FIRST -> Just reduce the size of the payload to be only the relevant data.

Add logic for:
- AVERAGE -> this will find the average value for a variable within the each dataset independently. 

# NOTE:
I am not sure if cleaning this up any further is worth it. If you have suggestions or believe it is please let me know.
If you find this hard to follow or read please let me know

# Video:
Here I have set up two SIR simulation runs. The only difference here is that one runs for 20 timesteps and the other for 90.

**Before:**

https://github.com/user-attachments/assets/64f2199c-37cc-4d30-b469-15ce43001658

**After:**

https://github.com/user-attachments/assets/34a1eb5b-85e2-4cce-8204-41733a4074d5


YohannParis (approved):
LTTM

blanchco (approved):

PR #6863, opened 2025-03-19
fix: Sensitivity chart tooltip point colors do not match the corresponding chart line #6935

### LINKED ISSUE #6935: [BUG]: Sensitivity chart tooltip point colors do not match the corresponding chart line
Sensitivity chart tooltip point colors do not match the corresponding chart line

Why are all the circle markers orange in colour? Shouldn't they match the colour of the sample trajectory?

Image

DSheffield:
Update the generation of the sensitivity chart 'tooltip' layer so the point colours match the line colours.  As a bonus, the change also makes the point labels visible and coloured as well.

image

I'm not entirely happy with this solution - feels like piling one more hack on existing ones, but suspect it would take a more significant rework / refactor to fix more cleanly.  Any opinions?

Resolves #6935


packages/client/hmi-client/src/services/charts.ts
```diff
@@ -642,7 +642,7 @@ export function createForecastChart(
 	}
 
 	// Helper function to capture common layer structure
-	const newLayer = (layer: ForecastChartLayer, markType: string) => {
+	const newLayer = (layer: ForecastChartLayer, markType: string, layerName: string = '') => {
```
DSheffield, line 645: Added name props to layers to aid with debugging

blanchco (approved):
nice!

YohannParis (approved):

PR #6864, opened 2025-03-20
denormalize extraction data structure

mwdchang:
### Summary
The first two points here were geared for tabular lookups in DBs so we wouldn't need to resort to json-queries, I don't think we will go these route, but whether these attrs are nested or flattened shouldn't do too much harm.
- Denormalize data structure to make page and source attributes more readily accessible in ExtractionItem
- For similar reason above, flatten provenance
- Introduce type and subType rather than explicit text/picture/table sections
- Java data models
- Added the idea of "rawText" and "text", this is to support secondary processes. For example we have latex-string and cleaned-latex-string

### DIFF
--- a/packages/client/hmi-client/src/types/Types.ts
+++ b/packages/client/hmi-client/src/types/Types.ts
@@ -418,6 +418,14 @@ export interface EvaluationScenarioSummary {
     timestampMillis: number;
 }
 
+export interface Extraction {
+    extractedBy: string;
+    pages: any;
+    body: any;
+    groups: any;
+    extractions: ExtractionItem[];
+}
+
 export interface ExtractionResponse {
     id: string;
     status: string;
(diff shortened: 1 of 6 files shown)


mwdchang:
> I _think_ Extraction will need a UUID so I can reference it from an enrichment

Do you mean to reference `ExtractionItem` @blanchco ? 
Extraction object doesn't really have an id per se because it falls under the purview of the document-asset.

blanchco:
> > I _think_ Extraction will need a UUID so I can reference it from an enrichment
> 
> Do you mean to reference `ExtractionItem` @blanchco ? Extraction object doesn't really have an id per se because it falls under the purview of the document-asset.

I was more thinking about how I can relate it to what we plan to be building out with enrichments down the line.  But if it's under document asset, I can reference the document asset id and get the provenance-enrichment relationship from there.

PR #6867, opened 2025-03-20
Fix: Model Config warnings for stratified parameters

Tom-Szendrey:
# Description
Correct logic around the warning colour for parameters within a stratified model. 
Previously if any of the parameters within a group were set to constant it would warn user that values were missing.

# Solution:
- Move helper function `isParameterInputEmpty` and utilize it for both stratified and non stratified parameters.

# Screenshots:

Before:
image

After
image


mwdchang (approved):
okay!

PR #6869, opened 2025-03-20
Ocr taskrunner task

mwdchang:
### Summary
This PR ass the storage (document_asset.extraction column) and plumbing to call the ocr-taskrunner when the PDF document is uploaded.

Note 1:  `document_asset.extractions` will be phased out.
Note 2: This may consume a lot of processing memory, you may want to test only with a very small document for now

### Testing
Run the `POST knowledge/pdf-extractions` endpoint against an existing document. Note swagger does not work, you'd have to either go through the application or do it via curl, for example: 

```
curl -v -X 'POST' \
  'http://localhost:3000/knowledge/pdf-extractions?document-id=de0ff82b-33a6-4a37-90df-3aadfd888f26&project-id=3d7b0bff-5c55-48b4-bc4c-0dd7aa876967' \
  -H 'accept: */*' \
  -H "${AUTHORIZATION_HEADER}" \
  -d ''
```

### DIFF
--- a/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
+++ b/packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
@@ -53,6 +53,7 @@ public class DocumentAsset extends TerariumAsset {
 	@TSOptional
 	@Type(JsonType.class)
 	@Column(columnDefinition = "json")
+	@Deprecated
 	private Map<String, JsonNode> metadata;
 
 	@TSOptional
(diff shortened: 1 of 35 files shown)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
```diff
@@ -53,6 +54,7 @@ public class DocumentAsset extends TerariumAsset {
 	@TSOptional
 	@Type(JsonType.class)
 	@Column(columnDefinition = "json")
+	@Deprecated
```
YohannParis, line 57: Why not removing them?

blanchco (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
mwdchang: The removal/cleanup changes too many things as to distract away from the task, these deprecated fields are still in heavy rotation - just that we don't use the processing results (as far as I am aware). Also need to do database migration related matters.

dvince2 (approved):

jryu01 (approved):

PR #6871, opened 2025-03-20
Broken project copy on home page

dvince2:
# Description

* Use correct reference to project.
* Set project name

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
+++ b/packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
@@ -74,7 +74,7 @@ const saveCopy = async () => {
 	isCopyDialogVisible.value = false;
 	router.push(RoutePath.Home);
 
-	const copiedProject = await useProjects().clone(project.value.id);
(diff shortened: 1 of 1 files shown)


Tom-Szendrey (approved):

packages/client/hmi-client/src/components/widgets/tera-common-modal-dialogs.vue
```diff
@@ -96,7 +96,8 @@ watch(
 	() => isCopyDialogVisible.value,
 	async () => {
 		if (isCopyDialogVisible.value) {
-			const name = await useProjects().getActiveProjectName();
+			project.value = useProjects().activeProject.value;
```
YohannParis, line 99: What is this line for?

dvince2: There were two issues in this code. The first was where the copy happens we use `project.value` but it was never set (a PR in january took away this line and broke it). The second is on the front page we didn't set the project clicked on and were only looking at the active project.

YohannParis (approved):

PR #6886, opened 2025-03-24
TA3 operators to have correct cancel ids

Tom-Szendrey:
# Description
Update the ta3 operators to have their cancel id logic slightly cleaner.
We can provide a list of ids, so where relevant we will send a list of any ids that pass `Boolean(id)` aka not `""`

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
+++ b/packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
@@ -24,7 +24,7 @@
 							:disabled="isEmpty(node.outputs[0].value)"
 						/>
 						<span class="flex gap-2">
-							<tera-pyciemss-cancel-button class="mr-auto" :simulation-run-id="cancelRunIds" />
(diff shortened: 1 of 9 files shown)


mwdchang (approved):
Sure

packages/client/hmi-client/src/components/workflow/ops/calibrate-ciemss/tera-calibrate-ciemss-drilldown.vue
```diff
@@ -24,7 +24,7 @@
 							:disabled="isEmpty(node.outputs[0].value)"
 						/>
 						<span class="flex gap-2">
-							<tera-pyciemss-cancel-button class="mr-auto" :simulation-run-id="cancelRunId" />
+							<tera-pyciemss-cancel-button class="mr-auto" :simulation-run-id="cancelRunIds" />
```
YohannParis, line 27: You should update this component prop name to imply plurals.
Even better would be to make the prop an array everywherw, which seems to be the case anyway.

Tom-Szendrey: Fair point, done

PR #6887, opened 2025-03-24
ocr extraction improvements

mwdchang:
### Summary
- Clean up Java classes
- Start to add utility methods in Extraction.java for working with extraction data, eg: Format extractions to be used for various enrichment processes
- Add post processing phase for LLM-cleaned latex equations, we use rawText and text to differentiate
- Improve logging

### Testing
Upload a small 1, 2 page PDF. Wait for a few minutes for extraction to finish up (check the document_asset.extraction column).  Check that latex equations are extracted (if applicable) and have both rawText and text fields populated.

For example:
```
      "rawText": "\\frac{dE}{dt} = \\beta SI-(\\mu+\\alpha+\\sigma)E\\qquad\\qquad\\qquad\\qquad\\qquad\\qquad \\qquad\\qquad (2.2)",
      "text": "\\frac{d E(t)}{d t} = \\beta * S(t) * I(t) - (\\mu + \\alpha + \\sigma) * E(t)",

```

```
      "rawText": "\\frac{dS}{dt} = b-\\beta SI-\\mu S,\\hskip 142.26378pt (2.1)",
      "text": "\\frac{d S(t)}{d t} = b - \\beta * S(t) * I(t) - \\mu * S(t)",

```


blanchco (approved):

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extraction/Extraction.java
```diff
@@ -2,29 +2,71 @@
...
 public class Extraction {
 
 	private String extractedBy;
 
 	// Layout infomration, not used, type these out later - Mar 20
-	private JsonNode pages;
-	private JsonNode body;
-	private JsonNode groups;
+	// private JsonNode pages;
+	private Map<String, ExtractionPage> pages;
```
dvince2, line 19: Will these require a migration?

dvince2 (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extraction/Extraction.java
mwdchang: I don't believe so, this is typing out and should be compatible with what had been in the JsonNode.

PR #6889, opened 2025-03-24
remove deprecated document_asset.assets

mwdchang:
### Summary
Remove deprecated column "assets" and associated clean ups, as well as migration script to drop the column from the DB.

Makes the code a little leaner and easier to understand.

### Testing
- Start as normal `./hmiServerDev.sh start staging run` and check in DB column is dropped. 
- App should work normally (eg, view pdf document)


Tom-Szendrey (approved):

dvince2 (approved):
Failing test, but everything looks good

PR #6890, opened 2025-03-24
Update model enrichment flow

blanchco:
# Description

An initial commit to get things rolling for changes to enrichments.

1. Enrichment storage and data structure
This idea here is for the `Model` to have the enrichment list stored in it's metadata.  Now we no longer have to store the enrichments in state which will improve the workflow save/update time.

Enrichments have a structure like:
```
private UUID id; 
private String label; // label of the enrichment
private EnrichmentSource source; // how the enrichment was generated
private EnrichmentTarget target; // where the enrichment will be applied
private JsonNode content; // any related content to the enrichment
private UUID extractionAssetId; // id of the document 
private String[] extractionItemIds; // list of ids for provenance from the document extractions
private boolean included; //whether an enrichment has been applied to the asset
```

Selecting enrichments and then saving the model will apply the enrichments to the model

2. Model Enrichment output changes

The JSON structure of the model enrichment job has changed and now each area for enrichment will have both a content object and a sources[] field to be returned

Testing

1. Upload the
[Measles_1_page.pdf](https://github.com/user-attachments/files/19469083/Measles_1_page.pdf)
pdf document and wait for extractions to finish
2. In a workflow attach the uploaded document to a model from equations node
3. Once a model has been created wait for enrichments to be finish.
4. Observe an enrichments list where you can include the enrichments you want.  To apply them save the model
5. Observe that each enrichment has a extractionItemIds array which can be traced back to the document's extraction items


packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/enrichment/Enrichment.java
```diff
@@ -0,0 +1,19 @@
+package software.uncharted.terarium.hmiserver.models.dataservice.enrichment;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.UUID;
+import lombok.Data;
+import software.uncharted.terarium.hmiserver.annotations.TSModel;
+
+@Data
+@TSModel
+public class Enrichment {
```
mwdchang, line 10: There are a couple of fields that seem to be doing double duties and mean different things in different context, for example:
- label: seem to be both the id, and some table/model-card stuff
- type: seem to denote both where to apply the enrichment (eg, state vs transition) and how the enrichment came about (regular vs custom)

This makes the object hard to reason with, because nowhere are these implicit semantics mentioned. It might be better to  introduce new fields rather than having to deal with double meanings.

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
```diff
@@ -1,177 +1,82 @@
...
 
-export function updateModelWithEnrichments(model: Model, enrichments: AssetBlock<EnrichmentBlock>[]) {
+export function updateModelWithEnrichments(model: Model, enrichments: Enrichment[]) {
 	const modelMap = createModelMap(model);
-	const includedEnrichments = enrichments.filter((enrichment) => enrichment.asset.include);
+	const includedEnrichments = enrichments.filter((enrichment) => enrichment.included);
+	// update enrichments in model enrichments map
+	if (model.metadata && model.metadata.enrichments) {
+		model.metadata.enrichments = enrichments;
+	}
```
mwdchang, line 43: Why do we care if `model.metadata.enrichments` exists or not if you are going to replace it anyway?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/enrichment/Enrichment.java
blanchco: Label, is just used as a label for the enrichment for display purposes.  The only id I am using is the enrichment id!

I have removed type and split it into source and target, hopefully that will make things more clear.
Source -> enum how the enrichment is created (right now just system and custom)
Target -> where the enrichment will be applied to in the asset

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/model-from-equations-utils.ts
blanchco: fair, have remove that check

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/extraction/Extraction.java
```diff
@@ -69,4 +69,18 @@ private static void extractTextToBuffer(
 				}
 			});
 	}
+
+	// Collect simplified extraction items, used when sending for an LLM request to save tokens
+	@JsonIgnore
+	public List<LightweightExtractionItem> getLightweightExtractions() {
```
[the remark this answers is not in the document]
blanchco: Okay I've moved the class to be an inner class and I am filtering out the those extraction that are of type "picture".  I can filter out future type/subTypes if we need!

mwdchang (approved):

Issue #6642, opened 2025-02-24
[FEAT]: paginate metadata table in the Dataset page


liunelson:
**IPlease describe.**
The metadata table in the Dataset page should be paginate (overwhelming when there are more than a few columns in the table)

PR #6897, opened 2025-03-25
Dataset Drilldown - Add paginator to column info section

Tom-Szendrey:
# Description
As described in [this issue](https://github.com/DARPA-ASKEM/terarium/issues/6642) it can be disorienting when a dataset has many many columns.
Here I have added a paginator as suggested

image

### DIFF
--- a/packages/client/hmi-client/src/components/dataset/tera-dataset.vue
+++ b/packages/client/hmi-client/src/components/dataset/tera-dataset.vue
@@ -29,13 +29,7 @@
 			</AccordionTab>
 			<AccordionTab header="Column information">
 				<ul>
-					<li
-						v-for="(column, index) in columnInformation.slice(
-							columnInfoFirstRow,
-							columnInfoFirstRow + MAX_NUMBER_OF_ROWS
-						)"
-						:key="column.name"
-					>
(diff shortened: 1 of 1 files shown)


YohannParis (approved):

packages/client/hmi-client/src/components/dataset/tera-dataset.vue
```diff
@@ -28,11 +28,30 @@
...
 			</AccordionTab>
 			<AccordionTab header="Column information">
-				<tera-column-info
-					v-for="(column, index) in columnInformation"
-					:key="index"
-					:column="column"
-					@update-column="updateColumn(index, $event.key, $event.value)"
+				<ul>
+					<li
+						v-for="(column, index) in columnInformation.slice(
```
DSheffield, line 33: I'm not a big fan of having much JS in the html, so I'd maybe move this to a computed like `displayedColumns` or `columnsOnPage` or something.  But it's a minor point.

Tom-Szendrey: Fair point that is a bit cleaner I will update this

DSheffield (approved):

PR #6899, opened 2025-03-25
6654 subtask 3 create shadow auth system

### LINKED ISSUE #6654: [SUBTASK]: 3. Create shadow auth system
**Describe the task**
Restore the pantera auth system to use postgres in addition to rebac to do a soft migration. Default to postgres and then fallback to rebac

dvince2:
# Description

* Setting up table for permissions
* Setting up new objects for permission
* Project services set up

Resolves #6654


packages/client/hmi-client/src/types/Types.ts
YohannParis, line 1: Do all of those needs to be exposed to the front-end?

YohannParis (approved):

packages/client/hmi-client/src/types/Types.ts
dvince2: Looking at how these are used in Pantera, it appears so... or at least most of them.  A lot of them end up getting used on the admin panel for viewing/editing permissions for users.

PR #6913, opened 2025-03-27
Hookup OCR equations to model from equations

blanchco:
# Description

* Hookup _new_ OCR formulas to model from equations drilldown
* modified `EquationBlock` data structure accordingly to hold a provenance reference, addtional info can be lookedup in a map (i.e bounding boxes, page number, etc...)
* OCR formulas are already edited by AI, so will no longer be cleaning equations on the front end
* removed some unused function related to the `old` equations-to-model endpoint
* removed the old equations-to-model endpoint entirely

### DIFF
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-equation-container.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-equation-container.vue
@@ -76,6 +76,6 @@ main.is-editing {
 	border: 4px solid transparent;
 	border-radius: var(--border-radius);
 	position: relative;
-	top: 0rem;
+	top: 0;
 }
 </style>
--- a/packages/client/hmi-client/src/components/model/petrinet/tera-equation-container.vue
+++ b/packages/client/hmi-client/src/components/model/petrinet/tera-equation-container.vue
@@ -1,22 +1,5 @@
 <template>
 	<main :class="isEditingStyle">
-		<section v-if="isEditable" class="controls">
-			<Button v-if="isEditing" class="p-button-sm p-button-outlined" @click="emit('cancel-edit')" label="Cancel" />
-			<Button
-				v-if="isEditing"
-				class="p-button-sm"
-				:disabled="disableSave"
-				@click="emit('update-model-from-equation')"
-				label="Save"
-			/>
-			<Button
-				v-else
-				class="p-button-sm p-button-outlined"
-				@click="emit('start-editing')"
-				:loading="!!isUpdating"
-				label="Edit"
-			/>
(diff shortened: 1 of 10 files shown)


packages/client/hmi-client/src/components/model/petrinet/tera-equation-container.vue
```diff
@@ -95,6 +76,6 @@ main.is-editing {
 	border: 4px solid transparent;
 	border-radius: var(--border-radius);
 	position: relative;
-	top: v-bind('mathContainerStyle');
+	top: 0rem;
```
YohannParis, line 79:
```suggestion
	top: 0;
```

YohannParis (approved):
I'm loving it!

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
```diff
@@ -178,14 +177,18 @@
...
-												rows="1"
-												placeholder="Add an expression with LaTeX"
-												class="w-full overflow-y-scroll"
-											/>
+											<template v-if="selectedItem === equation.id">
+												<Textarea
+													v-model="equation.asset.text"
+													autoResize
+													rows="1"
+													placeholder="Add an expression with LaTeX"
```
mwdchang, line 185: nitpick: "Add an expression with LaTeX" sounds a little weird, it is kind of like saying "add code snippet with javascript" ... I think "Add LaTeX expression" would be better.

packages/client/hmi-client/src/components/workflow/ops/model-from-equations/tera-model-from-equations-drilldown.vue
```diff
@@ -601,30 +612,25 @@ onMounted(async () => {
...
 						id: uuidv4(),
-						name: 'Equation',
+						name: `Equation ${index}`,
 						asset: {
-							pageNumber: index + 1,
-							text: equation
+							text: eq.text,
+							provenance: {
+								documentId: document.value!.id!,
+								extractionAssetId: eq.id
```
mwdchang, line 628: The "extractionAssetId" would be the docmentId, since the extractions are derived from the document asset.

Since you already have a documentId, maybe you meant to say "extractionItemId" or something??

blanchco: Yes, sorry I meant `extractionItemId`, I have changed this accordingly

PR #6934, opened 2025-03-28
Use document-asset.extraction, markdown/html/text support

mwdchang:
### Summary
- Leverage document-asset.extraction instead of document-asset.extractions. The field is now commented out, the removal and database clean up will be done separately.
- Support for non-pdf types: html, text, markdown
- Expand taskrunner input to accept filename (for inference) and llm type, on top of byte-stream

### Test
You will need to rebuild and start ocr-extraction-taskrunner locally. Once done
- Upload a small markdown document, check DB that it has extraction column filled in
- Upload a small pdf document, check DB that it has extraction column filled in


packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/gollm/GoLLMController.java
```diff
@@ -1211,13 +1213,13 @@ public ResponseEntity<TaskResponse> createDocumentQuestionTask(
...
 		// make sure there are document extractions available
-		if (document.getExtractions() == null || document.getExtractions().isEmpty()) {
+		if (document.getExtraction() == null) {
 			log.warn(String.format("Document %s has no extracted text", documentId));
 			throw new ResponseStatusException(HttpStatus.NOT_FOUND, messages.get("document.extraction.not-done"));
 		}
 
 		try {
-			input.setDocument(objectMapper.writeValueAsString(document.getExtractions()));
+			input.setDocument(objectMapper.writeValueAsString(document.getExtraction().getDocumentText()));
```
blanchco, line 1222: We should use the `getLightweightExtractions` here as well

mwdchang: Please provide a reason for this claim.

blanchco: When we ask a question about the document, we want to receive both an answer and the source of that answer, as this will serve as an additional enrichment.

mwdchang: Sure, but there are no provisions for sources/provenances in the current prompt construct:

https://github.com/DARPA-ASKEM/terarium/blob/dd29ac34acee3f448aba6a3a7bb88804833706ad/packages/gollm/common/prompts/document_question.py

So you need to do some experimentations with prompt engineering, and let me know what format works for you.

packages/ocr_extraction/src/server.py
```diff
@@ -62,7 +62,11 @@ async def process_and_predict(file: UploadFile = File(...), llm_model: str = For
     logging.info("In predict")
     file_bytes = await file.read()
     logging.info(f"File length = {len(file_bytes)}")
-    docstream = DocumentStream(name="test", stream=BytesIO(file_bytes))
+
+    filename = "document.txt"
+    if file.filename is not None:
+        filename = file.filename
```
YohannParis, line 68: Why we need to hardcode the value here?
Shouldn't the taskrunner provide a document already?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -253,22 +253,14 @@ public ResponseEntity<Void> pdfExtractions(
...
-			} catch (InterruptedException | ExecutionException e) {
-				log.error("Error extracting PDF", e);
-				throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, messages.get("document.extraction.failed"));
-			}
-
 			try {
 				newF.get();
 			} catch (Exception e) {
 				e.printStackTrace();
+				log.error("Error extracting PDF", e);
```
YohannParis, line 262:
```suggestion
				log.error("Error extracting Document", e);
```

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
```diff
@@ -253,22 +253,14 @@ public ResponseEntity<Void> pdfExtractions(
...
+				throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, messages.get("document.extraction.failed"));
```
YohannParis, line 263: Need to update this message as well in case it contains the word PDF

packages/server/src/main/java/software/uncharted/terarium/hmiserver/models/dataservice/document/DocumentAsset.java
YohannParis, line 1: Why not removing them instead of commenting them out?

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskUtilities.java
```diff
@@ -90,7 +90,8 @@ public static TaskRequest getEnrichDatasetTaskRequest(
 		input.setLlm(llm);
 		if (document != null) {
 			try {
-				input.setDocument(objectMapper.writeValueAsString(document.getExtractions()));
+				// FIXME: Specialization for dataset enrichment?
+				input.setDocument(objectMapper.writeValueAsString(document.getExtraction().getDocumentText()));
```
YohannParis, line 94: Should we add a method `getText` on the `Document` class to provide `document.getExtraction().getDocumentText()` more efficiently?

YohannParis (approved):

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/TaskUtilities.java
mwdchang: Probably not, not in this PR anyway.
There is already a `DocumentAsset.getText()` ... which is not really being used if I am reading the code right, that field needs to be scrubbed.

packages/server/src/main/java/software/uncharted/terarium/hmiserver/controller/knowledge/KnowledgeController.java
mwdchang: message looks okay.
```
document.extraction.failed = Document extraction failed. Please try again later.
```

PR #6936, opened 2025-03-28
Added core UIs to the pdf viewer

jryu01:
# Description

https://github.com/user-attachments/assets/823a5a9b-9061-4776-ac45-88cb50932736

Resolves #(issue)

### DIFF
--- a/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
+++ b/packages/client/hmi-client/src/components/documents/tera-document-asset.vue
@@ -1,7 +1,7 @@
 <template>
 	<tera-asset :id="assetId" :name="document?.name ?? ''" :is-loading="documentLoading" :is-document="true">
 		<p class="pl-3" v-if="documentLoading">Loading...</p>
-		<tera-pdf-embed v-if="pdfLink" :pdf-link="pdfLink" :title="document?.name || ''" />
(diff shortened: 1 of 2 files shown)


packages/client/hmi-client/index.html
```diff
@@ -13,7 +13,6 @@
 			document.cookie = name + "=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;";
 		});
 	</script>
-	<script src="https://documentservices.adobe.com/view-sdk/viewer.js" async></script>
```
YohannParis, line 16: I will then remove all Adobe API traces from the application now.

YohannParis: Made here #6941

YohannParis (approved):

mwdchang:
I pushed in a couple of commit to have provisional support for highlighting equations and enrichments. The interactions are a little weird because some areas capture the click events and do not seem to bubble up - so we will need to have a round of refinements.

mwdchang (approved):

packages/server/src/test/resources/application-test.properties
```diff
@@ -7,8 +7,6 @@
 adobe_key=nothing
```
dvince2, line 7: can this be deleted too?

jryu01: Done

PR #6941, opened 2025-03-31
Remove Adobe API keys

YohannParis:
Now that we do not use the Adobe PDF JS tool, we can remove the API key.


packages/client/hmi-client/src/components/model/petrinet/tera-model-equation.vue
```diff
@@ -40,8 +40,6 @@ const props = defineProps<{
 	model: Model;
 }>();
 
-const emit = defineEmits(['model-updated']);
```
dvince2, line 43: intentional?

YohannParis: Yes, it was not used and it create issue on Github PRs

dvince2: 👍

PR #6943, opened 2025-03-31
Add Counterfactual Template

Tom-Szendrey:
# Description
- New template
- minor css updates to the template modal to correctly allow overflow.

image

image


packages/client/hmi-client/src/components/workflow/scenario-templates/counterfactual/counterfactual-scenario.ts
```diff
@@ -0,0 +1,273 @@
...
+
+		wf.addEdge(datasetNode.id, datasetNode.outputs[0].id, calibrateNode.id, calibrateNode.inputs[1].id, [
+			POSITION,
+			POSITION
+		]);
+
+		wf.addEdge(interventionNode.id, interventionNode.outputs[0].id, calibrateNode.id, calibrateNode.inputs[2].id, [
+			POSITION,
+			POSITION
+		]);
```
mwdchang, line 230: I find it a bit strange that the intervention operator feeds into the calibrate operator - that seems to void the idea of counter-factual? that is you are kind of forcing your base-case to be your what-if case, then you ask what is the diff between the base-case and the what-if case.

packages/client/hmi-client/src/components/workflow/scenario-templates/counterfactual/tera-counterfactual-template.vue
```diff
@@ -0,0 +1,154 @@
...
+				const modelObs = model.value.semantics?.ode.observables;
+				if (modelObs) {
+					modelObs.forEach((obs) => modelStateAndObsOptions.value.push(obs.id));
+				}
+			}
+		}
+
+		isFetchingModelData.value = false;
+	},
+	{ deep: true }
```
mwdchang, line 142: You either
- do not need a deep watcher because the value being watched is just a string, or
- you might be watching the wrong variable

packages/client/hmi-client/src/components/workflow/scenario-templates/counterfactual/counterfactual-scenario.ts
Tom-Szendrey: I will check in the design channel

packages/client/hmi-client/src/components/workflow/scenario-templates/counterfactual/tera-counterfactual-template.vue
Tom-Szendrey: Good catch!

YohannParis (approved):

PR #6951, opened 2025-04-01
Update final mira endpoint

YohannParis:
- Rename the MIRA service, and have it run both on deploy and local dev docker compose.
- Point `mira-taskrunner` and `beaker` to it.
- Update `hmi-server` to point to it.


containers/scripts/docker-compose-data-services.yml
```diff
@@ -1,5 +1,15 @@
 ---
 services:
+  mira:
+    container_name: mira
+    image: ghcr.io/darpa-askem/mira:latest
+    networks:
+      - terarium
+    environment:
+      MIRA_NEO4J_URL: 'bolt://neo4j:7687'
```
mwdchang, line 9: Does this mean we won't be able to strip out neo4j from or tech stack? Or is this like an embeded neo4j that runs inside mira container?

dvince2 (approved):

containers/scripts/docker-compose-data-services.yml
YohannParis: The latter, it's is own neo4j. So when we are done with the DKG, it's just removing one single image from the stack.

YohannParis:
This is ready to be merged when the base is changed to `main`.

dgauldie (approved):

PR #6952, opened 2025-04-01
Model config enrichments backend

blanchco:
# Description

* Updated configurations from documents prompt to have extractionItemIds to reference provenance from document
* model configuration now has an enrichments field to determine which enrichments the user has applied to the configuration

Note:
I have created a _new_ json output for this process as the configuration from dataset is still using the older one

Before we would get back a list of configurations:
```
{
  "conditions": [
    {
      "modelId": "model-123",
      "name": "Baseline scenario",
      "description": "Configuration representing baseline disease spread",
      "initialSemanticList": [
        {
          "expression": "9900",
          "expressionMathml": "<math><cn>9900</cn></math>",
          "target": "S",
          "source": "Text section 2.1"
        }
      ],
      "inferredParameterList": [],
      "observableSemanticList": [],
      "parameterSemanticList": [
        {
          "distribution": {
            "parameters": {
              "value": 10000,
              "minimum": null,
              "maximum": null
            },
            "type": "Constant"
          },
          "referenceId": "N",
          "source": "Table 1"
        }
      ]
    }
  ]
}
```

After now we are getting a custom object which is parsed down the line into enrichment objects:
```
{
  "conditions": [
    {
      "name": {
        "content": "Baseline scenario",
        "extractionItemIds": ["ext_1"]
      },
      "description": {
        "content": "Configuration representing baseline disease spread",
        "extractionItemIds": ["ext_1", "ext_3"]
      },
      "initialSemanticList": [
        {
          "content": {
            "expression": "9900",
            "expressionMathml": "<math><cn>9900</cn></math>",
            "target": "S"
          },
          "extractionItemIds": ["ext_2", "ext_6"]
        }
      ],
      "parameterSemanticList": [
        {
          "content": {
            "distribution": {
              "parameters": {
                "value": 10000,
                "minimum": null,
                "maximum": null
              },
              "type": "Constant"
            },
            "referenceId": "N"
          },
          "extractionItemIds": ["ext_2"]
        }
      ]
    }
  ]
}
```


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ConfigureModelFromDocumentResponseHandler.java
```diff
@@ -63,38 +66,94 @@ public static class Properties {
...
+			enrichment.setId(UUID.randomUUID());
+			enrichment.setTarget(EnrichmentTarget.STATE);
+			enrichment.setSource(EnrichmentSource.GOLLM);
+			enrichment.setContent(initial.getContent());
+			enrichment.setExtractionAssetId(props.documentId);
+			enrichment.setExtractionItemIds(initial.getExtractionItemIds());
+			enrichment.setIncluded(false);
+			enrichments.add(enrichment);
+		});
+		condition.parameterSemanticList.forEach(parameter -> {
```
mwdchang, line 115: Are we guaranteed to always have `condition.initialSemanticList` and `condition.parameterSemanticLis` available for access? or do we need to do null-checks?

mwdchang (approved):
looks okay, a few comments here and there

packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/tasks/ConfigureModelFromDocumentResponseHandler.java
blanchco: Yes it should always be available, according to the json output that we are defining is it not nullable

dvince2 (approved):

PR #6953, opened 2025-04-01
collapse observables if they are "stratified"

mwdchang:
### Summary
To accommodate "manually stratified" observables in a model, where the observable names are of the format:
```
<observable_name>_<group_1>_<group_2>_<group_3> ...
```

Effectively we strip out all the `<group>` components and make a unique list in the end.

Note actual stratification operator does not produce stratified observables, this is a manual construct and thus the logic is heuristically based, aka our mileage varies.

image

### Testing
You can use 
[CriticalMineralsModel.json](https://github.com/user-attachments/files/19555217/CriticalMineralsModel.json)


Tom-Szendrey (approved):
LGTM

YohannParis (approved):

PR #6954, opened 2025-04-01
Correct loss chart updating in calibrate drilldown

Tom-Szendrey:
# Description
Inserting the new data was not correctly working. I tried to get it to work for a bit but I could not figure it out. As the data for this chart is small and we are already willing to redraw it in the nodes I will do the same in the drilldown.

# Testing
You should not see the loss chart correctly being updated while the calibration is running rather than a relatively straight line
image


mwdchang (approved):
okay!

Issue #6900, opened 2025-03-25
[BUG]: Checkboxes get unchecked


pascaleproulx:
**Describe the bug**
On staging, in Simulate, trying to select states to display in a comparison chart for sidarthe: https://app.staging.terarium.ai/projects/424b5910-8aa1-4f26-8033-f94e0b634881/workflow/25361d73-1681-4835-9bc8-16d4bb1ebcd7?operator=60fa767f-9447-434f-ac72-bb0311319351

https://github.com/user-attachments/assets/383399ac-44f6-4fe2-9e76-6262d4ccfa60

PR #6957, opened 2025-04-01
fix: Checkboxes get unchecked #6900

DSheffield:
Set common state change save interval.  Increase sync interval to 2 sec.

### DIFF
--- a/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
+++ b/packages/client/hmi-client/src/components/workflow/tera-workflow.vue
@@ -237,8 +237,8 @@ import { subscribe, unsubscribe } from '@/services/ClientEventService';
 import { activeProjectId } from '@/composables/activeProject';
 
 const WORKFLOW_LOCAL_STORAGE_UPDATE_INTERVAL = 4000;
-const WORKFLOW_SYNC_INTERVAL = 2000; // how long to wait before applying changes received from server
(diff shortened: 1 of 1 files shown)


packages/client/hmi-client/src/components/workflow/tera-workflow.vue
```diff
@@ -236,7 +236,9 @@ import * as InterventionPolicyOp from '@/components/workflow/ops/intervention-po
 import { subscribe, unsubscribe } from '@/services/ClientEventService';
 import { activeProjectId } from '@/composables/activeProject';
 
-const WORKFLOW_SAVE_INTERVAL = 4000;
+const WORKFLOW_LOCAL_STORAGE_UPDATE_INTERVAL = 4000;
+const WORKFLOW_SYNC_INTERVAL = 2000; // how long to wait before applying changes received from server
```
DSheffield, line 240: I set this to 2 sec rather than 5, because for this bug I think this just needs to be longer than the save interval such that if a user action causes multiple state changes and corresponding events from the server, then the debounce timer will keep getting reset until the update messages finish coming in.

If you think we should still move it up to 5 sec we can change - this is my first go round on this ride so haven't observed the full thing in action much.

mwdchang: I'd probably shorten 2000 => 1500 so the workflow UI is a bit more responsive

mwdchang (approved):

YohannParis (approved):
Lovely

PR #6960, opened 2025-04-02
FE implementation for model configuration enrichments

blanchco:
# Description

* Enrichment selection option within the model configuration item component.
* Can "Save" when there are values missing in the original configuration, this allows us to apply enrichments without needing to create another configuration.
* created a few helper functions in both the front and back end 
* reworked the `tera-asset-block` component so that common functionality doesn't have to be redefined whenever we use enrichments (i.e Next/previous buttons, selection class)

https://github.com/user-attachments/assets/35cf32de-a245-457a-96ec-25c256948847

https://github.com/user-attachments/assets/bafdc464-7705-4dc5-bbd8-074c85f5ace9


YohannParis (approved):

dvince2 (approved):
back end lgtm

PR #6961, opened 2025-04-02
fix backend units tests for knowledge controller

blanchco:
# Description

* There is an issue with the knowledge controller tests after recent changes removing the text attribute for the DocumentAsset.  This should allow the tests to pass again


mwdchang (approved):
Thanks!

Weird this wasn't caught in my Java classes clean up ...

PR #6962, opened 2025-04-02
Simulate ensemble fail gracefully

Tom-Szendrey:
# Description
We were missing the following from simulate ensemble's polling. 
This was causing an issue where if the simulation failed in this way we would keep polling/not reset the inprogressid


YohannParis (approved):

blanchco (approved):
🥇

PR #6967, opened 2025-04-03
OCR Extraction fixes

mwdchang:
### Summary 
Fix possible LLM errors and general cleanup
- We expect a 1-to-1 relationship for input/output for cleaning latex equations, it seems like sometimes LLM might not respect it, so for now we will add a guard
- General python linter clean up

### Testing
Test extraction with 
[qa_SIR.pdf](https://github.com/user-attachments/files/19590738/qa_SIR.pdf)


packages/server/src/main/java/software/uncharted/terarium/hmiserver/service/ExtractionService.java
```diff
@@ -729,8 +731,16 @@ public Future<DocumentAsset> extractPDFAndApplyToDocumentNew(final UUID document
 				if (cleanupResp != null && cleanupResp.getOutput() != null) {
 					int counter = 0;
 					for (JsonNode eq : output.get("response").get("equations")) {
-						formulaItems.get(counter).setText(eq.asText().trim());
-						counter++;
+						// FIXME: It seems like on rare occasions the lengths of the original equations and the cleaned
+						// equations mismatch. May be better to changet the prompt to return a dictionary.
```
YohannParis, line 735:
```suggestion
						// equations mismatch. May be better to change the prompt to return a dictionary.
```

YohannParis (approved):
LGTM

jryu01 (approved):

PR #6970, opened 2025-04-03
disconnect tests/pocs that we probably wont use

mwdchang:
### Summary
Disconnect tests and proof-of-concepts that we do not use to make the build a bit leaner, the reduction is js-chunk size is about 5%.


YohannParis (approved):
Could we delete them instead?

mwdchang:
> Could we delete them instead?
I think they still have some value if we want to debug a module in a stand-alone manner, it is just that we don't need them to be a part of the build any longer as proof of concepts for pascale et al.

PR #6972, opened 2025-04-04
Gollm model query

mwdchang:
I need this here mostly to make change tracking easier - but maybe we can just merge this in as these are all new code and not in use.

This is prototyping right now for model-introspection (asking questions about how the model can be adapted to some specific situation/scenario).


YohannParis (approved):
Looks good to me.

